PR #6, opened 2023-06-06
capitalization fixes to enable building on linux

m-m-adams:
Fixed capitalization in the project configuration as well as all imports. Project now builds successfully on linux


processor286:
I don't know if it's appropriate for me to comment here, and I'll gladly leave if so, but it seems to me your PR is attempting to include your personal /home/ directory into the paths in some places. I doubt this is acceptable.

m-m-adams:
Hmm, you're right the launch files seem to include the full path instead of the local workspace version. They're created automatically by e2 studio so should probably be added to gitignore

litui:
The launch paths can be truncated within e2 Studio itself. When you browse for a path, you can edit the field and remove the portions preceding the workspace. That's how I set up the launch files included in the previous commits.

m-m-adams:
Yes, the path is truncated in the gui option (line 119, org.eclipse.cdt.launch.PROGRAM_NAME). The other lines 97 and 108 org.eclipse.cdt.debug.gdbjtag.core.imageFileName and symbolsFileName don't appear to be editable in the gui and are where the full path to my project dir and formerly Rohan's project dir appear

m-m-adams:
Reverted to just include capitalization fixes, will put slashes in a separate PR

jamiefaye:
I will merge this in a couple of hours after I get a chance to make sure it
builds OK at my end.

– Jamie

PR #11, opened 2023-06-06
ignore cpu specific header file

m-m-adams:
This commit does two things:

1. Reverts the cpu_specific.h file to have HAVE_OLED and HAVE_RTT set to 1, matching the official branch, the comments in uart_all_cpus, and the code guide

2. Adds this file to gitignore so that user configurations do not cause extra changes. Once build scripts are written that could be reverted

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /e2-build-release-oled/
 *.launch
 /toolchain/*
+src/drivers/RZA1/cpu_specific.h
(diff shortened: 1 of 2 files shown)


litui:
I'm not sure why these defaults matter given that I have divided out build targets that set these defines in the preprocessor directives...  e2's config serves as the build scripting for the moment.

m-m-adams:
Not sure - I didn't get RTT output until changing the define but if there's a better spot I have no particular attachment to this change

litui:
Are you working from build-release or build-debug?  I have HAVE_RTT enabled in the preprocessor directives on the debug build. Should be anyway, unless those aren't working for some reason.

litui:
image

m-m-adams:
Not working on my end (now they are stuck at 1 and 1 regardless of profile) but it's possibly my janky E2 setup. Closing this PR since your solution is better and should work

PR #17, opened 2023-06-07
Increase resolution of patch cable strength in the sound editor

bfredl:
This is not nearly done yet but it would make sense to use draft PR:s to indicate what people are exploring/working on.

The Sound engine (and synth XML files) support much higher resolution for patch cable strength internally, than the -50 to +50 steps available in soundeditor when you make modulation mappings.
It is already possible to adjust the strength by editing the XML manually (or by some third party editor which could have implemented this), and the resolution could be increased in the sound editor only, while keeping full backward/forward compat of synth XML files.

to be done:
- [x] ~~rework the conversion code to use floats instead of magic integer constants~~, and allow even higher resolution on demand.
- [x] rework the OLED UI layout (the additional digits might overlap parameter name, need to use smaller font for digits)
- [x] ~~update the UI code for 7seg display~~ this should already be handled by MenuItemDecimal but a 7seg owner verifying this would be good

### DIFF
--- a/src/MenuItemPatchCableStrength.cpp
+++ b/src/MenuItemPatchCableStrength.cpp
@@ -35,6 +35,8 @@
 MenuItemPatchCableStrengthRegular patchCableStrengthMenuRegular;
 MenuItemPatchCableStrengthRange patchCableStrengthMenuRange;
 
+extern bool movingCursor;
+
 #if HAVE_OLED
 void MenuItemPatchCableStrength::renderOLED() {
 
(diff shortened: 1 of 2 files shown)


litui:
Interesting idea of having draft PRs as works-in-progress. I tend to see that as a use for "enhancement" issues rather than PRs, but I'm curious what @jamiefaye thinks.

bfredl:
the benefit of draft PR:s is that you can comment/give suggestions directly on the code diffs and have it tracked as part of the PR discussion even as code is updated/rebased.

bfredl:
I opted to document the meaning of the integer constants instead of changing to float, which seemed a more conservative change and works well enough.

Updated the UI for OLED. I think the 7seg UI should work as it is as this is implemented by the `MenuItemDecimal` parent class and look the same as for (unmodulated) Osc1/osc2 transpose which also has two decimals and max two digits before the decimal point.

jamiefaye:
On 7LED, this fails with a visit to the abort_handler.

PaulFreund:
I didn't look at the code yet but there seem to be different implementations of numeric values for different synth parameters and cable patches. Some of these are not supported as MIDI CC and for some negative values didn't work with any gear I tested. Would it make sense to touch this problem at the same time? Also would it make sense to restrict the range to what MIDI can represent?

bfredl:
@PaulFreund not sure I follow, are you talking about generating midi CC from the mod matrix? Or more like editing synth params over CC?

PaulFreund:
I have an external MIDI controller connected (in this case via DIN) open a parameter like Transpose and hold LEARN/INPUT, then I move the CC knob and the display shows LEARNED. I can now control the parameter with the CC. This does not work with Detune Number (Can't Learn) and it also does not allow negative numbers for some reason. For patch cables (For example mapping transpose to LFO2) I can learn a CC to the cable strength but only in positive direction, so if we want to keep compatibility we might need to limit the range to what MIDI supports

bfredl:
That sounds to me like we should rather improve the midi learn implementation, I e have more CC values  as a modulation source so you can specify a range and offset for CC controlled parameters (this will also apply to patch cables as these can be modulated in turn)

There is no compatibility issue with this PR tho as it only improves the resolution and doesn't change the range, so I'd rather address the MIDI issue in a follow up.

PaulFreund:
Perfect, just wanted to make sure :) Thanks!

bfredl:
@jamiefaye before the 7SEG emulation is in place I'm fumbling around a bit in the blind, but I pushed a change which might (or might not) fix the 7SEG issue. would you mind try again?

jamiefaye:
I will do on Monday, as my Deluge development setup is at the lab space I
rent.

- Jamie

ok-reza:
Would love to see this feature be toggleable in the Community Settings.

bobtwinkles:
> Would love to see this feature be toggleable in the Community Settings.

that should be relatively straightforward to implement. Can you file a bug? Easier to track that way than a comment in a closed PR =)

PR #18, opened 2023-06-07
Build scripts, work-in-progress

litui:
Testing out this draft PR business for myself as a way to track work-in-progress. Not bad, not bad!

This is the repo where I'm doing work that will end up being the build scripts for the Deluge. It is based on the "FBT / Flipper Build Tool" included in the flipperdevices/flipperzero-firmware repository, which is a similar parallel to the Deluge's situation: need for multi-platform (Mac, Windows, Linux) support with automation and an ARM cross-compile target with special considerations.

The build system is based on python/scons with platform-specific scripting in bash, cmd, and powershell as needed for toolchain fetching, etc.


scripts/toolchain/dbtenv.sh
```diff
@@ -0,0 +1,343 @@
...
+    fi
+    dbtenv_check_if_sourced_multiple_times;
+    dbtenv_check_env_vars || return 1;
+    dbtenv_check_download_toolchain || return 1;
+    dbtenv_set_shell_prompt;
+    dbtenv_print_version;
+    PATH="$TOOLCHAIN_ARCH_DIR/python/bin:$PATH";
+    PATH="$TOOLCHAIN_ARCH_DIR/gcc-arm/bin:$PATH";
+    PATH="$TOOLCHAIN_ARCH_DIR/openocd/bin:$PATH";
+    PATH="$TOOLCHAIN_ARCH_DIR/openssl/bin:$PATH";
```
bobtwinkles, line 322: It would be nice to have a switch that turns path hacking off, since binary compatibility on linux is an eldritch hell. On nixos it's easier for me to provide these binaries on the system path than it is to get random downloaded executables working.

litui: I hear you, and a switch might be a nice addition, but also the Flipper Zero folks put a lot into making their tool work well in these scenarios. If you'd like to see how it runs, feel free to check out their source and give the flipperdevices/flipperzero-firmware "fbt" command a run. =)

bobtwinkles:
```
/home/bob_twinkles/Code/fbt/flipperzero-firmware/toolchain/x86_64-linux/python/bin/python3: line 15: /home/bob_twinkles/Code/fbt/flipperzero-firmware/toolchain/x86_64-linux/python/bin/python3.11-bin: cannot execute: required file not found
```
On `nixos` you can't even trust the path to libc (which is what's happening here), so binaries usually need [`patchelf`](https://github.com/NixOS/patchelf) to run. Or better, built from source as a native nix derivation with the right rpath already embedded. If you don't want to support this kind of weirdness I'll happily carry a local patch or spin up a container instead =P

litui: Interesting! I think I'm going to aim to support the "big 3" out of the box, but after the build scripts are working I'm happy to see if we can integrate support for nixos and other unusual platforms.

litui:
Closing this PR since I've been working from a different branch. Will open a new, active PR for the current codebase.

PR #31, opened 2023-06-08
clang-format the whole tree

bobtwinkles:


bobtwinkles:
We should consider the following options, which have potentially larger impacts but might improve readibility
- https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignafteropenbracket this is inconsistent in the existing code
- https://clang.llvm.org/docs/ClangFormatStyleOptions.html#sortincludes this would be nice but might break things
- https://clang.llvm.org/docs/ClangFormatStyleOptions.html#bracewrapping `BreakAfterFunction` in particular.

jamiefaye:
Had build errors on Windows e2studio. I will await the post-Draft era.

bobtwinkles:
I disabled formatting of the NE10 intrinsics, it should build now.

bobtwinkles:
`SortIncludes` 100% breaks the code, maybe on day we'll do a cleanup pass on that and make it work but that's probably best saved for another change.

Psirus:
How did you arrive at these settings for clang-format?

bobtwinkles:
I screwed up my find-foo and didn't format about half the files, should be fixed now.

> How did you arrive at these settings for clang-format?

Attempting to minimize disruptive changes to the core Deluge code (`src/*.cpp`). I'm not 100% happy with the effect it has on `src/drivers` so I plan to revisit that in a bit (hence the draft status).

bobtwinkles:
OK, I think this is about as good as it gets (at least not without starting to patch clang-format or getting really granular with `clang-format off`. I've also added a github action which will check for formatting regressions. An example run can be found [here](https://github.com/bobtwinkles/DelugeFirmware/actions/runs/5217431857/jobs/9417258925).

bobtwinkles:
An example failure run for the action is shown here: https://github.com/bobtwinkles/DelugeFirmware/pull/2

Issue #26, opened 2023-06-08
Random and S&H LFOs, triplet and dotted LFO sync


phfalk:
I always wanted to have some more randomness for modulation that changes more frequently than the "Random" modulation source which only outputs a new random value every time a note is triggered. How I imagine this to work:

  - [x]  Add LFO that just outputs a random value every time it is called. This might be a little too jittery to be useful
    - sort of working but more noisy than musical 
  - [x]  Add another LFO that takes a random value and holds it for some time until a new random value is generated
  - [x] Explore what kinds of smoothing or restrictions in terms of how far subsequent values can be from each other would be useful
  - [x] Now that I understand more about how the LFOs work, next step could be triplet or maybe even dotted LFO sync timings

alex-reid:
I've been looking into dotted notes for sync settings. From what I can gather at a high level, sync setting are stored as an `int` from 0 to 9 with that number being a divisor for the sync length hence why the sync settings are whole steps. My thoughts around this are that maybe implementing a `bool` for dotted notes along with the 0 to 9 sync settings might be the way to go.

Looking at @dctucker's excellent class diagrams, the UI side of things is handled by [MenuItemSyncLevel](https://dctucker.github.io/DelugeFirmware/docs/de/d7d/class_menu_item_sync_level.html) and is fed to the arp, delay, lfo and sidechain. 

I can't currently get the toolchain set up because i'm on a mac, but if I did, I'd probably approach it by trying to get the UI working and then delve into getting it hooked up to the individual things that work off that.

phfalk:
Corresponding MR was merged

manysounds:
Current community build (c95a27c) LFO shape buttons hang the machine.

bobtwinkles:
That was a bug due to the menu item refactor, please try the latest nightly: https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5827496457

The pages site build is broken so it is currently out of date.

bobtwinkles:
Actually that won't have some other fixes to know bugs, latest build is here: https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5830090263

In the future, please prefer opening new bugs to commenting on closed ones :sweat_smile:

PR #32, opened 2023-06-08
Random and S&H LFOs and new LFO (and Delay, Arp, ...) sync types

phfalk:
WIP on #26.

### DIFF
--- a/src/lfo.cpp
+++ b/src/lfo.cpp
@@ -40,6 +40,25 @@ int32_t LFO::render(int numSamples, int waveType, uint32_t phaseIncrement) {
 	case LFO_TYPE_TRIANGLE:
 		value = getTriangle(phase);
 		break;
+
+	case LFO_TYPE_SAH:
+		if (phase == 0) {
+			value = CONG;
+			holdValue = value;
+		}
+		else if (phase + phaseIncrement * numSamples < phase) holdValue = CONG;
+		else value = holdValue;
+		break;
+
+	case LFO_TYPE_RWALK:
+        uint32_t range = 4294967295u / 20;
+		if (phase == 0) {
+			value = holdValue + (range / 2) - CONG % range;
+			holdValue = value;
+		}
+		else if (phase + phaseIncrement * numSamples < phase) holdValue += (range / 2) - CONG % range;
+		else value = holdValue;
+		break;
 	}
 
 	phase += phaseIncrement * numSamples;
--- a/src/lfo.cpp
+++ b/src/lfo.cpp
@@ -51,12 +51,23 @@ int32_t LFO::render(int numSamples, int waveType, uint32_t phaseIncrement) {
 		break;
 
 	case LFO_TYPE_RWALK:
-        uint32_t range = 4294967295u / 20;
+		uint32_t range = 4294967295u / 20;
 		if (phase == 0) {
-			value = holdValue + (range / 2) - CONG % range;
+			value = (range / 2) - CONG % range;
 			holdValue = value;
 		}
-		else if (phase + phaseIncrement * numSamples < phase) holdValue += (range / 2) - CONG % range;
+		else if (phase + phaseIncrement * numSamples < phase)
+			// (holdValue / -16) adds a slight bias to make the new value move
+			// back towards zero modulation the further holdValue has moved away
+			// from zero. This is probably best explained by showing the edge
+			// cases:
+			// holdValue == 8 * range => (holdValue / -16) + (range / 2) == 0 =>
+			// next holdValue <= current holdValue
+			// holdValue == 0 => (holdValue / -16) + (range / 2) == range / 2 =>
+			// equal chance for next holdValue smaller or lager than current
+			// holdValue == -8 * range => (holdValue / -16) + (range / 2) ==
+			// range => next holdValue >= current holdValuie
+			holdValue += (holdValue / -16) + (range / 2) - CONG % range;
 		else value = holdValue;
 		break;
 	}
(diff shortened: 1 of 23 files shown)


m-m-adams:
For a more musical rand I think you want to lower the range - try using the random function with a limit that's within the bounds of the other LFO types

Another possibility would be generating Random Normal distributed values - generate 2 linear ones, cast to float and scale to 0-1, and then get two random normal values y1=sqrt(-2*ln(n1))*cos(2*pi*n2) and y2 = sqrt(-2*ln(n1))*sin(2*pi*n2). Alternatively you can make a 1000 points random distribution lookup table and randomly choose an index in it.

Another possibility would be instead making the noise LFO into a random walk - to do this you would start at 0, generate a random number r between 0 and 10, and move the output r-5 spots. This will create a walk that trends around 0 but wanders up and down at different speeds


As an entirely seperate comment should the LFO types be switched to an enum if we're going to keep adding them? I suspect there will be more to come, I'd personally like to make some one shot decay and attack envelopes in different shapes 
Similarly the if/else code should probably be replaced by a map before it expands too much more

phfalk:
I have now tweaked the random LFO into a random walk, which is quite fun to use.

I have also done a lot of refactoring to add new triplet and dotted sync types and make them work with everything that uses tempo sync. The code is not great because everything is quite entangled, but adding enums for sync type and level adds some level of compile time checking as opposed to using ints and constants everywhere. I am still not completely sure if using enums instead of constants is a good idea over all, though.

The last commit in the chain isn't really necessary, because selecting the new sync types is disabled in the compressor menu, but it prepares the compressor to be used with the new sync types in case that is needed in the future.

src/lfo.cpp
```diff
@@ -40,6 +40,25 @@ int32_t LFO::render(int numSamples, int waveType, uint32_t phaseIncrement) {
...
+			holdValue = value;
+		}
+		else if (phase + phaseIncrement * numSamples < phase) holdValue = CONG;
+		else value = holdValue;
+		break;
+
+	case LFO_TYPE_RWALK:
+        uint32_t range = 4294967295u / 20;
+		if (phase == 0) {
+			value = holdValue + (range / 2) - CONG % range;
```
m-m-adams, line 56: For LFO2 this should reset to a constant instead of using the holdValue

src/lfo.cpp
```diff
@@ -40,6 +40,25 @@ int32_t LFO::render(int numSamples, int waveType, uint32_t phaseIncrement) {
...
+			holdValue = value;
+		}
+		else if (phase + phaseIncrement * numSamples < phase) holdValue += (range / 2) - CONG % range;
```
m-m-adams, line 59: This can drift quite far from the midpoint (I had it mapped to filter and it totally closed it for a long time)

Suggest weighting the walk value so that it tends to come back towards range

m-m-adams (changes requested):
Sounds good overall, some minor issues with the random walk not conforming to the LFO2 philosophy

src/lfo.cpp
phfalk: That makes sense. I am just testing a function that adds more bias for walking back the more the value deviates from the origin.

src/lfo.cpp
phfalk: You mean that LFO should essentially always start at zero modulation? That is a little difficult to do when at the same time LFO1 sync should not return to zero at the start of every sync interval, which we can of course discuss as well. In both cases the LFO phase is reset to 0.

I think these random LFOs will always behave a little different than the original periodic ones. One example is that I made them tick half as fast, so only one value change per sync interval as opposed to two maxima in the periodic LFOs, which means two flips for square. In my opinion, that is not a bad thing as long as it is somewhat intuitive.

Interested to hear your opinion on this.

m-m-adams: Yeah thats a good point, I didn't consider how it would interact with LFO1. I don't have a strong opinion on whether the LFO2 version would ideally reset to zero or to a random value on with every note on but the current free running behaviour is un intuitive with how LFO2 normally works. 

I don't have a good solution since I forgot about LFO1 but I'll comment again if I think of one

phfalk: I tried to replicate what you describe, but I am not sure if I completely understand what you mean. To me, it sounds like LFO2 resets correctly on every trigger. Not to zero, but to a new random value, which would make sense with how it is implemented at the moment. The phase gets reset to zero when a new note triggers and that triggers a new LFO cycle, which means a new value for the random LFOs. That is very similar to the tempo synced behavior in LFO1, but feels a little different than for example "sine" which will always start at zero. To me, it feels somewhere in between free running (in terms of modulation) and re-triggering (in terms of rhythm).

Does that make sense? I might be completely misunderstanding your point, though.

m-m-adams: The S+H mode sounds like you're describing but the random walk doesn't. If you wait until it's audibly at a low point then when you hit a new note on it stays low. I think this is because the phase reset at 0 still uses the previous hold value 

It's not a huge deal and if it can't be changed without affecting LFO1 then I don't think it's a show stopper, the overall effect definitely sounds cool

phfalk: Got it now, thank you. I just pushed some changes that should address that and the biasing.

src/lfo.cpp
jamiefaye: Let me know when this is ready to go into community branch.

soymonitus:
Please @phfalk I beg you, implement the same triplets and dotted options for the Delay sync!

phfalk:
@soymonitus, for some reason I can not reply directly to your comment, so I'm making a new comment. The new sync options for delay and arp are already implemented as part of this MR and the MR has been merged into the community branch. You should be able to try it if you flash a community build.

PR #35, opened 2023-06-09
Fix typo in InstrumentClipView.cpp

eltociear:
lengh -> length


litui:
We don't have a pull request standard defined yet, but I'd argue fixes for single typos in comments is a low-effort PR that we might want to discourage. While PRs are not just for code changes, this type of PR clutters the commit history and rewards "Hacktoberfest" style credit-seeking behaviour (looking at @eltociear's account, it's pretty clear this is the goal).

PR #38, opened 2023-06-09
CODE_OF_CONDUCT, CONTRIBUTING, and FUNDING files

litui:
- Scope expanded slightly as I realized these would also be important documents to include.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,43 @@
+# How to contribute to DelugeFirmware/community
+
+Above all, thank you for your interest in both the Synthstrom Deluge and being a contributor!
+
+Please see also, the [Code of Conduct](CODE_OF_CONDUCT.md) document.
+
+## Bug reports
+
+* Please submit an issue highlighting as much detail and context about the bug as possible
+
+* Provide detail on whether the bug was found on an OLED or 7seg Deluge
+
+* Where the bug exists in build scripts, e2, or other development, building, and debugging processes, please provide logs
+
+* Bugs directly relating to related projects should be submitted to the appropriate issue tracker:
+    * DelugeProbe: [litui/delugeprobe](https://github.com/litui/delugeprobe)
+
+## Pull Requests (PR)
+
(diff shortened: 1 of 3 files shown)


litui:
Good call, @BenRGH.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+Above all, thank you for your interest in both the Synthstrom Deluge and being a contributor!
+
+Please see also, the [Code of Conduct](CODE_OF_CONDUCT.md) document.
+
+## Bug reports
+
+* Please submit an issue highlighting as much detail and context about the bug as possible
+
+* Provide detail on whether the bug was found on an OLED or 7seg Deluge
```
dctucker, line 11: Yes!

Not trying to derail this PR at all, but we could set up issue templates so that folks filing bugs are immediately presented with empty fields to populate. It's been super useful or my day-job, at least, at keeping the description quality high.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
+# How to contribute to DelugeFirmware/community
+
+Above all, thank you for your interest in both the Synthstrom Deluge and being a contributor!
+
+Please see also, the [Code of Conduct](CODE_OF_CONDUCT.md) document.
+
+## Bug reports
+
+* Please submit an issue highlighting as much detail and context about the bug as possible
```
henning, line 9: I would propose to phrase it somehow different. 
What's mainly important to me as developer to fix a bug is reproducibility, not necessarily that someone posts all logs and output and whatever might be but probably is not related to the problem. I'm not saying logs are not important, they could help quite a bit to understand - just "as much as possible" sounds a bit too much oriented in the amount and not the quality of the information ;)

A good old tutorial that often helps me to explain this is maybe this one: 
https://www.chiark.greenend.org.uk/~sgtatham/bugs.html

litui: @henning strong points all 'round. I'll attempt to reword, and probably include that link as well!

litui:
Implemented changes suggested in previous reviews. Merging.

Issue #25, opened 2023-06-07
Alternate Tunings


dctucker:
## Description

Some synths and electronic pianos have the ability to add user-defined scales that differ from 12-tone equal temperament (12TET). This is usually enabled by allowing the user to set an offset (in cents) for each of the twelve notes in an octave (`N = 12`), and additionally by sending SysEx messages to the instrument.

I suspect I'll mostly be taking the lead on implementing this, but feedback/advice is welcome.

Incorporating some of the feedback received, the acceptance criteria section has been updated.

## Acceptance criteria
- [ ] A menu available in Song mode will allow specifying the tuning of each note, and the desired root note
- [x] Pitch adjustment applies when performing any synth instrument (Subtractive/FM/Pulse)
- [ ] Persist tuning banks to SD card
- [ ] Persist select tuning bank when saving a song
- [ ] Read tuning bank into memory when loading a song if specified and present, reset to 12TET otherwise and warn if bank not present
- [ ] Reset tuning data to default (12TET) in new song

## Stretch goals
- [ ] Pitch adjustment applies when performing synth sample
- [ ] Pitch adjustment applies to outgoing CV
- [ ] Ability to set a tuning bank for a single instrument track
- [ ] SysEx Implementation following [MIDI Tuning specification](https://www.midi.org/specifications/midi1-specifications/general-midi-specifications/general-midi-2/midi-tuning-updated)
    - [ ] `SCALE/OCTAVE TUNING DUMP, 2 byte format` or `SCALE/OCTAVE TUNING 2-BYTE FORM (NON REAL-TIME)`
    - [ ] `KEY-BASED TUNING DUMP`
    - [ ] `SINGLE NOTE TUNING CHANGE`
- [ ] Support for N divisions per octave where `N != 12`
  - This is currently included under stretch due to the amount of changes required for `N = 12` being much smaller than `N != 12` owing to `noteWithinOctave` being quite prevalent within `voice.cpp`.

## Open questions
- ~Should we enable persistence of tuning banks beyond Song mode by leveraging Flash storage or prefer loading from SD card?~ **Answered**: We will persist tuning banks as Scala files on the SD card.
- No menu is currently available from Song mode, so adding it here seems logical.
  - The argument could be made to somehow involve the Scale mode button, though this might confuse the two features. I'm open to feedback.
  - The current implementation adds a submenu to Settings, which is convenient because it can be accessed from keyboard mode where you can trigger notes while altering the tuning.
[...]

PR #39, opened 2023-06-09
[Draft] Tunings

dctucker:
Watch this space for ongoing progress. This branch may get deleted or squashed before merging to minimize commit message noise.

https://github.com/SynthstromAudible/DelugeFirmware/issues/25


src/MenuItemTuning.cpp
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright © 2018-2023 Synthstrom Audible Limited
+ * Copyright © 2023 Casey Tucker
```
dctucker, line 3: I'd like some feedback regarding what is expected and appropriate regarding attribution. Now that this is a community project, there will be more authors, and while attribution can seen in git commit messages, maybe when new files are introduced or heavily modified we would want to include authors in the copyright notices. Adding my name below the existing attributions seemed to be the recommended way after a cursory web search, but I'd like to know what the maintainers think.

litui: I think attribution makes sense, but Copyright is tricky territory in open source and I would not accept any PRs with a unique license term or copyright without running it by the Synthstrom team.

The bulk of the community source code is, as I understand it, *Copy-left* and licensed under the GNU General Public License with some pieces being under license from Renesas or available under the public domain. I would consider a "copyright" notice from Synthstrom of this sort to be overridden by the General Public License, effective after June 5, 2023 but this is probably territory for @ian-jorgensen to weigh in on.

src/SessionView.cpp
```diff
@@ -362,6 +368,16 @@ int SessionView::buttonAction(int x, int y, bool on, bool inCardRoutine) {
 		}
 	}
 
+	else if (x == scaleModeButtonX && y == scaleModeButtonY) {
```
dctucker, line 371: I was originally going to implement the menu on the Select encoder button, but I noticed [a few lines above](https://github.com/dctucker/DelugeFirmware/blob/bdd2a6333a6147b36a61b3c86d34e125fadd66b9/src/SessionView.cpp#L353-L367) already want to make use of this to cancel arming and launching during playback, and I didn't want to break that, so it's going on the Scale button until we figure out a better way.

Alternatively we could only show the song menu on Select encoder when playback is inactive?

samcar17: Hmm... sopping playback to menu dive isn't ideal, so my vote would be sticking with the scale button for the moment. It does pose an interesting question of how to make things consistent with the current UI, so maybe in future it'd be worth doing a rewrite with the song-mode menu on the select knob and the canceling launch feature can be a different button press/combo?

dctucker: Yeah, I think a bit of community input on how many folks get a lot of use out of the current "select knob cancels" would help make this decision clearer. Having slept on it, I'm now thinking of adding a "Song" submenu to either the global settings menu and/or the sound editor menu.

I kinda want to be able to send note on events while modifying the tuning, kinda like a guitar player would, ya know?

samcar17: Hard agree w/r/t note on messages while tuning. Personally I like the settings sub-menu idea, but maybe the forum/discord is a better place to get community-wide opinions?

src/TuningSystem.cpp
```diff
@@ -0,0 +1,93 @@
...
+	cents += offsets[noteWithinOctave] / 100.0;
+	double frequency = referenceFrequency * pow(2.0, cents / 1200.0);
+
+
+	double value = frequency / 1378.125;
+	value *= pow(2.0, 32);
+	tuningFrequencyTable[noteWithinOctave] = lround(value);
+
+	value = pow(2.0, noteWithinOctave / 12.0) * pow(2.0, 30);
+	tuningIntervalTable[noteWithinOctave] = lround(value);
```
dctucker, line 42: This should be readily adaptable to precise tuning across the entire range, and this function is only invoked by turning the knob so no performance hit expected. I've verified (on x86_64 at least!) that when offsets are all zero, the same values are generated as in lookuptables.cpp. I'd like to add unit tests for things like this, but my current test source isn't pretty enough to work on everyone's machine, and it'd be better if I were compiling for the same target and running the test binary on QEMU.

m-m-adams:
Is this still ongoing?

PaulFreund:

dctucker:
I haven't had the time to work on any of this. I'd like to have had this shipped, but my branch isn't the only one that plays with alternate tunings, others have expressed interested in doing work on this feature, and there seem to be conflicting opinions on how this feature should work. When community consensus is established regarding the minimum viable requirements I would be happy to contribute to this feature.

PR #42, opened 2023-06-09
fix: enable proper initialization of c++ global objects

bfredl:
not done yet, this is just a demo that it work.

most of the details here were provided by @gpanders , I just adapted them to the deluge code base.

### DIFF
--- a/linker_script_rz_a1l.ld
+++ b/linker_script_rz_a1l.ld
@@ -186,32 +186,27 @@ SECTIONS
 		*(.data.*)
 		_edata = .;
 		
-		
-		/* Uncommenting this section back in allows the section at the top of the main C++ file to be uncommented...
-		But I'm not sure if this is really the right place for this. */
-		
-		/*
-		__CTOR_LIST__ = .;
-		. = ALIGN(2);
-		__ctors = .;
-		*(.ctors)
-		__ctors_end = .;
-		__CTOR_END__ = .;
-		__DTOR_LIST__ = .;
-		___dtors = .;
-		*(.dtors)
-		___dtors_end = .;
-		__DTOR_END__ = .;
-		. = ALIGN(2);
-		_mdata = .;
-		*/
-		
 		*(.got.plt) 
 		*(.got)
 		.	= ALIGN(0x8);
 		address_end_data_ROM = .;
 	} > RAM012L
-	
+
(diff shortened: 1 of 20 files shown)


bfredl:
some references:

 -  https://interrupt.memfault.com/blog/how-to-write-linker-scripts-for-firmware
 -  https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html

bfredl:
Note: there are more global constructors in `SoundEditor::SoundEditor()` but I haven't touched these as then the logic to assign names to menu items would need to be changed. This should be done as a separate task (likely involving other refactors of the menu system), and is out scope for this PR. For now, letting these objects be re-initialized is perfectly fine (there is no memory leak, as the default constructors do not allocate memory).

jamiefaye:
I checked out this PR and built it for my 7 Seg Deluge and encountered a problem when I scrolled to a synth and pushed the select button. This caused the system to freeze. Hanging up at MenuItemSubmenu::drawPixelsForOled() at the call to drawName().

bfredl:
> I checked out this PR and built it for my 7 Seg Deluge and encountered a problem when I scrolled to a synth and pushed the select button. This caused the system to freeze. Hanging up at MenuItemSubmenu::updateDisplay() at the call to drawName().

Could reproduce some crashes on the OLED as well. Pushed a workaround for those.

jamiefaye:
Issue is still present on 7SEG.

linker_script_rz_a1l.ld
```diff
@@ -186,32 +186,27 @@ SECTIONS
...
-		_mdata = .;
-		*/
-		
 		*(.got.plt) 
 		*(.got)
 		.	= ALIGN(0x8);
 		address_end_data_ROM = .;
 	} > RAM012L
-	
+
```
stellar-aria, line 194: I think you'll also need a `.preinit_array` section in here to deal with vtable stuff (validation and maybe init?)

References: 
[Picolibc docs (but also applies for newlib)](https://github.com/picolibc/picolibc/blob/main/doc/init.md)
[Info on the SysV ABI](https://stackoverflow.com/a/60178049)

bfredl: thanks, done

bfredl:
superseeded by #104

PR #44, opened 2023-06-11
Disabled RTT in OpenOCD

litui:
RTT is available over USB serial CDC when using the DelugeProbe, however the OpenOCD RTT support does not seem to want to cooperate with the Deluge. Removing the exposed RTT port in OpenOCD configurations for now so as not to confuse folks.

Please use the DelugeProbe's serial RTT output.


litui:
Self-approving since this one is minimal and low impact.

Issue #19, opened 2023-06-07
Implement multiple note layouts in keyboard View


MilesParker:
The current keyboard view (i.e. "playing" Deluge notes ) supports a +5 semitone (fourth) row offset / tuning. This task is to allow users to set different layouts for display. Reasonable options run from +3 to +8 (octave).

UX TBD, but one possibility is shift + piano_keys + <> (| ^v). (Alternatives?)

(We could also consider other irregular options such as guitar, with the differing G-B offset, or a whole tone layout. As that looks like a more complex change it probably makes more sense as a separate task.)

For a similar implementation, see "Row Offset" in Linnstrument options  #below:
https://www.rogerlinndesign.com/support/linnstrument-support-panel-settings

MilesParker:
(Enhancement, can't seem to add label..)
Willing to take a go at this one but won't be able to make progress until my Deluge gets back from upgrade. But feel free to tackle it.

sichtbeton:
> The current keyboard view (i.e. "playing" Deluge notes ) supports a +5 semitone (fourth) row offset / tuning. This task is to allow users to set different layouts for display. Reasonable options run from +3 to +8 (octave).
> 
> UX TBD, but one possibility is shift + piano_keys + <> (| ^v). (Alternatives?)
> 
> (We could also consider other irregular options such as guitar, with the differing G-B offset, or a whole tone layout. As that looks like a more complex change it probably makes more sense as a separate task.)
> 

Would it be possible to have a note layout for kits so you can play all kit items? maybe even include something like a 16 levels approach for certain kit items?

MilesParker:
That's a really neat idea, kind of a pads mode. Though not really in scope for this particular issue.

bfredl:
> Willing to take a go at this one but won't be able to make progress until my Deluge gets back from upgrade. But if anyone else is inspired, go for it. :)

I took stab at this (at least the variable fixed offset part) as I really need it with the ongoing microtuning work: https://github.com/SynthstromAudible/DelugeFirmware/pull/46 .

PaulFreund:
@MilesParker Do you still want to work on this? I'm planning to start work on a PR with a multi layout (piano, isomorphic, fixed), configurable (rows, columns, in-key only, chords) keyboard that should implement #19 and #55 extending on #46 and don't want to collide with anybody

MilesParker:
[...]

PR #46, opened 2023-06-11
keyboardScreen: support multiple note layouts

bfredl:
addresses https://github.com/SynthstromAudible/DelugeFirmware/issues/19

shift + <|> to change row offset. Limited to offsets between 1-12 seems reasonable (max value gives octave shift).

allowing offset=0 would make sense together with rows being mapped to velocity instead (could be a followup).

- [x] TODO: save/load this as part of saving a song.
- [x] UI fixes, like a popup showing the value, and fix colors not always updating

### DIFF
--- a/src/definitions.h
+++ b/src/definitions.h
@@ -615,7 +615,7 @@ typedef enum SyncLevel_ {
 
 #define ALLOW_SPAM_MODE 0 // For debugging I think?
 
-#define KEYBOARD_ROW_INTERVAL 5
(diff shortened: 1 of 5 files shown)


bfredl:
Updated. row spacing is stored as part of an instrument clip, just like the overall scroll position of the keyboard. 
I tested loading a song with such clips on a older firmware without this change, song loads fine with this new setting being ignored.

UI now works fine on the OLED with a popup. This could use a quick test from someone with a 7SEG as well.

PR #56, opened 2023-06-14
Requirements, workflow and governance guidelines and definitions

PaulFreund:
As a foreword, please read this (draft) PR as being written with the best intentions. I don't want to overreach any boundarys or step on anyones toes, I want to give a starting point for discussion that is free to change in any dimension according to our needs. 

Target is to give the greatest amount of people possible the chance to contribute to the project while knowing how their contributions will be handled while making sure we give the users of the community firmware a great experience with new features and stable performance.

Regarding governance the idea is to define the rules and workflow on how it can be handled in this PR and let our project management write down the actual governance and persons in the [CODEOWNERS](https://github.blog/2017-07-06-introducing-code-owners/) file so it is automatically taken into account in GitHubs PR system.

- [x] CONTRIBUTING.md
- [x] GOVERNANCE.md

EDIT: I decided to remove the Ownership section from GOVERNANCE.md for now and leave it for someone else to add it.

Pull request #59 should follow this Pull request as it implements a mechanism fore the proposed requirement of [runtime feature settings](https://github.com/SynthstromAudible/DelugeFirmware/pull/56/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R76)

PR #59, opened 2023-06-15
Community feature settings in main settings menu

PaulFreund:
In accordance with the requirement in PR #56 for Pull requests altering user facing behavior or have a massive impact in constant cycle load ([see here](https://github.com/SynthstromAudible/DelugeFirmware/pull/56/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R76)) this Pull request aims to implement:

* A new menu item "Community features" in main menu settings (Shift+Pressing Selection knob) above Firmware version
* Once opened the user shall be presented with a list of all community features that can be enabled/disabled or selected (in case features are not compatible with each other, for example Row style vs Column style clip launching)
* It shall be very easy and worry free for Pull requests to implement new settings, there will be a module global instance of the settings that can be accessed through it's header file in all other objects
* The settings shall be stored as an XML file on the SD card instead of the flash for easy access in case manual intervention is required (e.g. testing a beta firmware and going back to an official release without a certain feature). 
* The XML file shall be upward and downward compatible as much as possible. For this reason all settings unknown to the firmware will persist reading and writing a new config file.

As always all feedback is welcome!

Usage for feature devlopers currently looks like this:
* To add a setting add a new block in ```RuntimeFeatureSettings.h``` along with new enum defines for the possible values: 
```C++
[RuntimeFeatureSettingType::FileFolderSorting] =  { 
    .displayName = "File/Folder sorting",
    .xmlName = "fileFolderSorting",
    .value = RuntimeFeatureStateToggle::Off, // Default value
    .options = { 
        { .displayName = "Off", .value = RuntimeFeatureStateToggle::Off }, 
        { .displayName = "On", .value = RuntimeFeatureStateToggle::On },
        { .displayName = NULL, .value = 0 }
    }
},
```
The flags can be easily checked anywhere in the code by including ```RuntimeFeatureSettings.h``` and adding a condition like this:
```C++
runtimeFeatureSettings.get(RuntimeFeatureSettingType::FileFolderSorting) == RuntimeFeatureStateToggle::On
```

Example file:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<runtimeFeatureSettings
	firmwareVersion="4.1.4-alpha"
	earliestCompatibleFirmware="4.1.3">
	<setting name="settingA" value="1"></setting>
	<setting name="settingB" value="2"></setting>
	<setting name="settingC" value="0"></setting>
	<setting name="settingZ" value="9"></setting>
</runtimeFeatureSettings>
```


jamiefaye:
Testing on a 7SEG here. When I click on Comm, I see an L and nothing else. If I twist the select knob, it will show SONG. Am I seeing the right thing?

PaulFreund:
Hi @jamiefaye, I have to admit I didn't test the menu empty on 7seg @sichtbeton was so nice to test it for me. Can you please uncomment the example block in RuntimeFeatureSettings.h as can be seen in the description? Since there are no settings yet I left it out.

sichtbeton:
Any build i can test?

jamiefaye:
Please update your MR as needed. Then I won't mess things up by uncommenting stuff and losing sync.

PaulFreund:
The menu only makes sense as soon as other developers add new features so right now it is intentionally empty. For testing to see an entry and check XML generation and parsing I left an entry as a comment but that should not be in the merged code.

Since I expected the menu to ship only as soon as we have runtime options like the new sorting behavior I didn't test it empty. Do you want me to add code to not display it if the list is empty?

src/RuntimeFeatureSettings.h
```diff
@@ -0,0 +1,95 @@
...
+public:
+	void readSettingsFromFile();
+	void writeSettingsToFile();
+
+protected:
+	RuntimeFeatureSetting settings[RuntimeFeatureSettingType::MaxElement] = {
+
+	    //// @TODO: Remove example on first use
+	    // [RuntimeFeatureSettingType::FileFolderSorting] =  {
+	    //     .displayName = "File/Folder sorting",
```
jantuitman, line 70: if this should be readable on 7 segment, names should be a single case alphanumeric only, name and the display should scroll like it does with file open if the filename is too long. Preferable the name should be very short. "FSORT" could be the better name in this example.

PaulFreund: The strings should be automatically displayed in the best readable casing and from the video I can confirm that they automatically scroll. Regarding length, for now this can be customized with a preporcessor ifdef in the definition, in the long run there was the idea to refactor displayable Strings into containers to also allow runtime switching. 

I can not Test with 7SEG but here is a link to the test video mentioned above https://discord.com/channels/608916579421257728/1107026299945496577/1120693871282757683. The menu had 3 test entries at the time

PaulFreund:
From my side this PR is still ready to merge. Here is a link to the binary I created with some mock options as can be seen in the video above: https://discord.com/channels/608916579421257728/1107026299945496577/1120692404031340594. I'm 90% sure it goes back to SONG because there is no entry in the menu.

As soon as the first feature is merged like PR #66 with an option to enable/disable the menu starts to make sense and give value.

sichtbeton:
this is great! so now feature developers could use this option mechanism?

PR #60, opened 2023-06-16
Compile-time signed_saturate bits parameter

stellar-aria:
This PR is to fix a build issue I found.

## Background
Unlike all of the other inline-assembly functions that are provided by `functions.h`, signed_saturate has one crucial difference:

The ARMv7A `ssat` instruction takes an immediate, not a register, for its second operand.

## The Problem
At sufficiently high optimization levels, this doesn't present a problem, as compile-time constants are combined and reduced down to a single immediate value before reaching the assembler. However, this is technically just hiding the problem that the signed_saturate function is taking what is potentially a run-time variable and trying to insert it into assembly code.

The assembler is, expectedly, not very happy about this when it tries to happen:
```
DelugeFirmware/src/functions.h:108:9: warning: 'asm' operand 1 probably does not match constraints
  108 |         asm("ssat %0, %1, %2" : "=r"(out) : "I"(bits), "r"(val));
      |         ^~~
DelugeFirmware/src/functions.h:108:9: error: impossible constraint in 'asm'
```
If the constraint is expanded to 'X' (any), we can see the source of the issue at the assembly stage:
```
Error: constant expression required -- `ssat r3,[fp,#-9],r3'
```
Which is, of course, that the compiler inserted a reference to a stack variable and the assembler has no idea what to do with it.

## The Solution
Thankfully, C++ already has facilities for strongly enforcing compile-time code evaluation of constants: templates.

By simply moving 'bits' from a function parameter to a template parameter, the compiler does this calculation before the generated assembly is passed off and everything is kept happy.

This does unfortunately trickle down to any functions that pass their own parameters to `signed_saturate`, but that's how compile-time constraints are meant to work in the first place.

### DIFF
--- a/src/functions.cpp
+++ b/src/functions.cpp
@@ -164,7 +164,7 @@ void functionsInit() {
 int32_t getFinalParameterValueHybrid(int32_t paramNeutralValue, int32_t patchedValue) {
 	// Allows for max output values of +- 1073741824, which the panning code understands as the full range from left to right
 	int32_t preLimits = (paramNeutralValue >> 2) + (patchedValue >> 1);
-	return signed_saturate(preLimits, 32 - 3) << 2;
+	return signed_saturate<32 - 3>(preLimits) << 2;
 }
 
 int32_t getFinalParameterValueVolume(int32_t paramNeutralValue, int32_t patchedValue) {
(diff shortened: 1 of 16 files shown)


litui:
!!! Thank you for this!  I think this is also related to the Assembler compile issues I was having in arm-none-eabi-gcc 12.2. I'll take this for a spin and see how it goes!

litui:
Functionally, this seems to be solid. I built successfully under both arm-none-eabi-gcc 9.2.1 and 12.2 and tested both in debug builds on the Deluge 7seg under e2studio in Windows.

File sizes are slightly smaller under 12.2 which is nice to see. Great PR!

Just the clang formatting check needs to pass before I can give the merge go-ahead.

stellar-aria:
Patched!

src/functions.cpp
```diff
@@ -211,8 +212,8 @@ int32_t getFinalParameterValueLinear(int32_t paramNeutralValue, int32_t patchedV
 
 	// positivePatchedValue's range is ideally 0 ("0") to 1073741824 ("2"), but potentially up to 2147483647 ("4"). 536870912 represents "1".
 
-	return lshiftAndSaturate(multiply_32x32_rshift32(positivePatchedValue, paramNeutralValue),
-	                         3); // Must saturate, otherwise sustain level can easily overflow
+	return lshiftAndSaturate<3>(multiply_32x32_rshift32(
+	    positivePatchedValue, paramNeutralValue)); // Must saturate, otherwise sustain level can easily overflow
```
bobtwinkles, line 216: nit: move the comment above so we get slightly better wrapping behavior here

bobtwinkles:
Minor nit that you absolutely don't have to address if it's annoying, but otherwise this looks like a nice safety improvement.

src/functions.cpp
stellar-aria: yeah that's fair

litui (approved):

PR #64, opened 2023-06-17
doc: add pinmap diagram

bobtwinkles:
[Rendered](https://raw.githubusercontent.com/SynthstromAudible/DelugeFirmware/8f846189e50ae7981ff34bca83f9740fef2af172/contrib/doc/pinmap/pins.svg)

### DIFF
--- a/contrib/doc/pinmap/generate.py
+++ b/contrib/doc/pinmap/generate.py
@@ -935,9 +935,9 @@ def main():
     ET.register_namespace('', 'http://www.w3.org/2000/svg')
     ET.register_namespace('svg', 'http://www.w3.org/2000/svg')
     root = ET.Element('svg')
-    root.attrib['width'] = '1750'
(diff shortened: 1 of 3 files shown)


contrib/doc/pinmap/generate.py
```diff
@@ -0,0 +1,1239 @@
+import io
```
dctucker, line 1: This is impressive.

Maybe throw a shebang at the top and `chmod` the file `0755` for us *nix users.

```suggestion
#!/usr/bin/env python3

import io
```

bobtwinkles:
sure, I'll do that. I also noticed a couple of minor errors (SSI0 is too
big) so I should fix those before this is merged.

bobtwinkles:
I think something is still wrong with the SSI0 but I'm too tired to fix it right now, consider this as still a draft.

PR #66, opened 2023-06-17
Browser sorts by folders then files alphabetically

dctucker:
Does what it says on the tin.

Implementation of https://forums.synthstrom.com/discussion/5368/show-folders-before-files-on-oled


src/FileItemArray.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef FILEITEMARRAY_H_
+#define FILEITEMARRAY_H_
+
+#include "r_typedefs.h"
+#include "ResizeableArray.h"
+#include "FileItem.h"
+
+class FileItemArray : public ResizeableArray {
+public:
+	FileItemArray(int newElementSize) : ResizeableArray(newElementSize) {}
```
bobtwinkles, line 27: Is there a reason to take the element size explicitly if we always use `sizeof(FileItem)`?

src/Browser.cpp
```diff
@@ -43,7 +43,7 @@ extern "C" {
 String Browser::currentDir;
 bool Browser::qwertyVisible;
 
-CStringArray Browser::fileItems(sizeof(FileItem));
+FileItemArray Browser::fileItems(sizeof(FileItem));
```
bobtwinkles, line 46: Assuming my grep skills are OK (which given the hour I'm typing this at they may not be) this was the only thing using `CStringArray`. If you want to rename that type instead of copying it that may not be the worst idea.

src/FileItemArray.h
dctucker: I'll take another shot at trying to factor this out. It's currently this way because CStringArray did it this way, but I also don't see why it's necessary.

dctucker:
I've converted this back to draft because I discovered a bug when loading songs.

weavermedia:
Could this be a preference in main settings?

jamiefaye:
Please fix conflicts.

PaulFreund:

PaulFreund:
I honestly didn't follow up on that and I think there is no ticket for it. If you have the motivation to find out and update your changes we can reopen the PR :)

PR #69, opened 2023-06-19
Deluge Build Tool (DBT) - Cross-platform, Low dependency, SCons-based build system

litui:
It's here, it's basic, it does the bare minimum!  It's the Deluge Build Tool, inspired (and, in part, borrowed) from the Flipper Build Tool (FBT) for Flipper Zero.

The current iteration runs just dandy and auto-pulls the latest arm-none-eabi-gcc 12.2 toolchain on the following architectures:
- win32-x86_64 (tested in Windows 11)
- linux-x86_64 (tested in Ubuntu)
- linux-arm64 (untested)
- darwin-x86_64 (Intel Mac)
- darwin-arm64 (Mac M1/M2 - No need for Rosetta, in theory!)

The toolchain is downloaded by the dbt script from [litui/dbt-toolchain](https://github.com/litui/dbt-toolchain/releases). Fully extracted it consumes around 1.2GB of storage on Mac (darwin-arm64). This breaks down as:

- openocd (9.2M) - xpack v0.12.0 - included for future use with hardware debuggers
- cmake (62M) - xpack v3.23.5 - included for optional use by those more acquainted with cmake; this is not used by DBT
- python (101M) - Standalone Install v3.11.3 - included for use in SCons scripting
- arm-none-eabi-gcc (831M) - xpack v12.2.1 - ARM cross-compile toolchain

As of right now, once the toolchain is in place, DBT parse the .cproject XML for details on many (but not all) of the build parameters, will build a specified target (or the default `dbt-build-release-oled`) from objects to firmware binaries (ELF, BIN, HEX), and drop them into a target build folder, very much like e2 Studio does. It can also (mostly) clean up after itself! Yay!

Please take this for a spin on your supported platform of choice and let me know here what works and what doesn't. I'll fix up bugs in this PR but will reserve new features for additional PRs (and I have quite a list of things I want to gradually add into this tool).

### DIFF
--- a/scripts/toolchain/dbtenv.sh
+++ b/scripts/toolchain/dbtenv.sh
@@ -1,0 +1,345 @@
+#!/bin/sh
+
+# shellcheck disable=SC2034,SC2016,SC2086
+
+# public variables
+DEFAULT_SCRIPT_PATH="$(pwd -P)";
+DBT_TOOLCHAIN_VERSION="${DBT_TOOLCHAIN_VERSION:-"6"}";
+
+if [ -z ${DBT_TOOLCHAIN_PATH+x} ] ; then
+    DBT_TOOLCHAIN_PATH_WAS_SET=0;
+else
+    DBT_TOOLCHAIN_PATH_WAS_SET=1;
+fi
+
+DBT_TOOLCHAIN_PATH="${DBT_TOOLCHAIN_PATH:-$DEFAULT_SCRIPT_PATH}";
+DBT_VERBOSE="${DBT_VERBOSE:-""}";
+
+dbtenv_show_usage()
+{
+    echo "Running this script manually is wrong, please source it";
+    echo "Example:";
+    printf "\tsource scripts/toolchain/dbtenv.sh\n";
+    echo "To restore your environment, source dbtenv.sh with '--restore'."
+    echo "Example:";
+    printf "\tsource scripts/toolchain/dbtenv.sh --restore\n";
+}
+
+dbtenv_curl()
+{
+    curl --progress-bar -SLo "$1" "$2";
+}
+
+dbtenv_wget()
+{
+    wget --show-progress --progress=bar:force -qO "$1" "$2";
+}
+
+dbtenv_restore_env()
+{
+    TOOLCHAIN_ARCH_DIR_SED="$(echo "$TOOLCHAIN_ARCH_DIR" | sed 's/\//\\\//g')"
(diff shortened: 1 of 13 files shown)


litui:
Somehow my rebase put @bobtwinkles's stuff in here. *shrug* Git is mysterious in its ways.

m-m-adams:
> Somehow my rebase put @bobtwinkles's stuff in here. _shrug_ Git is mysterious in its ways.

Ya, when you merge they’ll get filtered out as duplicates

bfredl:
I get the following error when the script invokes any python command:

> ./toolchain/linux-x86_64/python/bin/python3.11: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory

The libcrypt.so.X file on the user's system might not be compatible. This file is supposed to be included in `./toolchain/linux-x86_64/python/lib/` and this could be considered a bug with indygreg/python-build-standalone .

Anyway, after working around this issue (installing all wheels into an env based on my system's python3.11), it goes all the way to building a firmware image, great progress! :sparkling_heart:

bfredl:
FWIW, just rebasing this branch again, on top of an up-to-date `origin/community` works to get rid of the extra commits.

scripts/toolchain/dbtenv.sh
```diff
@@ -0,0 +1,350 @@
...
+    fi
+    echo "Warning! DBT environment script was sourced more than once!";
+    echo "You might be doing things wrong, please open a new shell!";
+    return 1;
+}
+
+dbtenv_set_shell_prompt()
+{
+    if [ -n "${PS1:-""}" ]; then
+        PS1="[dbt]$PS1";
```
bfredl, line 115: Maybe a space between `[dbt]` and the user's default prompt?

litui:
> The libcrypt.so.X file on the user's system might not be compatible.

Ah interesting. I'll see what might be needed to bundle this into the toolchain packages for each arch.
 Hopefully I won't have to rebuild indygreg.

litui:
Oh I should also mention that I used *black* formatter actively while developing this so it won't be a mess if we ever add python formatting to the checks.

litui:
Ahhh... that's my bad. I forgot about the arm64/aarch64 naming weirdness. Let me think about ways to get around that.

m-m-adams:
On ARM with this fix I go the error mentioned above about missing libcrypt.so, it build successfully after installing libxcrypt-compat 

It's possible I already had that installed for something else on my x86 machine - probably worth noting it as a requirement for some linux distros in the documentation since this appears to affect Arch and WSL2 as well per this discussion in python-build-standalone: https://github.com/indygreg/python-build-standalone/issues/173

Otherwise I guess it will go away once everyone is running python 3.13

PR #71, opened 2023-06-20
3d-printable or CNC-able right panel.

litui:
- Goal is to have 3d-printable files that fit the Adafruit SWD breakout as soldered in the modding guide as well as CNC GRBL parameters for cutting existing wood panels.

I don't own a 3d printer but if folks with 3d printers could give these a try testing for fit in the Deluge and debug port dimensions I'd appreciate it.


litui:
image

Position is set up, in theory, for ideal placement between screws on the top and bottom of the deluge. Used my own right panel to confirm positioning and size with an Adafruit SWD breakout:

PXL_20230619_234223541
PXL_20230619_234200356

jamiefaye:
There are two objects in the STL file. Should there only be one?

litui:
Ahahah. whoops. Forgot about the hidden one. I'll re export.

litui:
Set up a GRBL for making changes to existing panels. I'll probably set up another to lock-in spacing once test panels arrive.

bobtwinkles:
It's probably wise to squash out the janky 3d scanned version before this is merged so we don't carry that data in the git history forever =)

jamiefaye:
I just checked the 3d print against my Deluge. The 3D print is about 15mm too long

litui:
So weird. I'll fix that. I swear I measured twice. Thanks for checking!

litui:
@bobtwinkles I thought I had... granted I was very tired and that's probably what I accidentally pushed to my other branch 😓

litui:
Oh I see what you mean with the squash. Eh, I'm not too bothered by it being there. If someone wants a flintstones aesthetic panel, they're welcome to it.

jamiefaye:
> So weird. I'll fix that. I swear I measured twice. Thanks for checking!

If you need me to, I can tweak the TinkerCad.

litui:
I reduced the overall length and re-uploaded, but I'll also need to double check that the remainder of my measures are correct. I might have to compensate a bit before this will be usable.

The CNC gcode should still work as the length isn't a factor.

jamiefaye:
still 5mm too big (measuring the part that fits inside the Deluge case).

litui:
@jamiefaye Okay, I'm just gonna ditch it and go from scratch again. I've got all the panels from Ian to start on the CNC stuff so let me make perfectly sure it's right before I waste any more of your filament.

litui:
image

Manual busting out of the caliper before taking another stab at it.

PR #72, opened 2023-06-20
Build action

bobtwinkles:
This adds a matrix build action that builds all 4 major configurations ({debug,release}-{oled,7seg}) based on @litui's toolchains and [my own dockerfile around those](https://github.com/bobtwinkles/deluge-ci-images/pkgs/container/deluge-ci-images).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,36 @@
+on: [push, pull_request]
+jobs:
+  build:
+    name: Build ${{ matrix.profile }}-${{ matrix.hardware }}
+    strategy:
+      matrix:
+        profile: [ 'debug', 'release' ]
+        hardware: [ 'oled', '7seg' ]
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


litui:
Can we tweak the actions so they only affect pull requests to `community`? It doesn't make much sense to run these against PRs from `community` to elsewhere. Mainly I bring this up because under the current maintainer settings I'm not able to 'sync' my branches to `community`... I have to PR and merge 😅 . I should probably just get Ian to fix that ;)

bobtwinkles:
That makes sense. I've limited both workflows (format check and builds) to `community` and PRs against that branch. The build job also runs for `synthstrom-official` though the short retention period probably makes that one less useful (and I don't think the workflow configs are even synced to that branch anyway).

jamiefaye:
> I'll wait til my daytime to merge so I can be awake to support if anything goes wrong. Looks good!

Please merge when ready.

PR #73, opened 2023-06-20
Generate compile commands

bobtwinkles:
Generates a clang CompilationDB, handy for editors that are using clangd/ccls/etc.


litui:
Was wondering if anyone would want to enable that! Very cool.

litui (approved):
Looks good! Love that it's so easy to flip on in SCons.

Issue #70, opened 2023-06-19
Next steps for DBT


litui:
Now that DBT is in `community` there is a laundry list of stuff I want to add to it but I'll start with the following:

- [x] Disable default build in favour of listing targets.
- [ ] Build stats on completion, sizing, etc.
    - [ ] Danger-checking/warning for .bin size 0 or oversize .bin
    - [x] Output .siz file as e2studio does
- [ ] Test for or just provide some notice that libcrypt.so.1 is needed, maybe with some instructions.
- [ ] Dynamic version string (eg: "c20230619-a1b2c3d-d" in the scheme: &lt;c for community&gt;&lt;yyyyMMdd&gt;-&lt;git_commit&gt;-&lt;build type: "r" or "d"&gt;)
    - [ ] Eclipse macro/variable-based
    - [ ] DBT equivalent to Eclipse-based approach for same behaviour
- [x] Direct target support / multitarget (eg: `./dbt dbt-build-debug-oled dbt-build-release-oled` etc.)
    - [x] Fix compilation db (compile_commands.json) to output to the build directories
    - [ ] Add a --cdb_symlink=&lt;build&gt; param for mapping a specific compile_commands.json to be the default.
- [ ] Debugger shortcuts
    - [ ] DelugeProbe/CMSIS-DAP and FTDI compatible openocd + gdb shortcuts
    - [ ] JLink support with whatever tools they use (know very little about JLink, but have some starting points from over at the Flipper Zero repo)
    - [ ] Quick, commandline "install" support via debugger-based hardware flashing
- [x] Restructure to make more modular in prep for all the above ^^^^^
- [x] Touch .map file so it gets cleaned up with the rest of the targets.

More to be added.

Playing around with Kconfiglib is on my to-do list here as well, but I'm not sure I have a specific goal in mind with it yet or idea of how it fits. Once I do, expect some checkboxes to be added.

Pie-in-the-sky / requires a lot of other things to be in place:

- Test and/or test framework runner
- UF2 image support (should work with DelugeProbe and maybe even USB Mass Storage eventually...)
- USB interface?
    - Maybe some midi sender/receiver mocking capability for putting the Deluge through its paces
    - DFU loader support

m-m-adams:
[...]

PR #74, opened 2023-06-20
DBT additions, odds, and ends

litui:
See #70 for details on what's planned for this PR

Made some tired mistakes in my git workflow. Sorry about the mess.

**Edit: cleaned up the mess with rebase -i.**

Summary:

- In short, the big deal is simpler targeting and multi-targeting in more of a traditional makefiley manner. Eg:  `./dbt dbt-build-release-oled`
- No more default targeting (it was kinda confusing)
- -e2_target= still works when no direct targets are specified in the above manner, so this won't break the build scripts
- Restructured a bunch of stuff to allow for more modularity
- Probably more I'm forgetting.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -74,20 +74,50 @@ if GetOption("base_config") == "e2_xml":
     if GetOption("e2_target"):
         Default(GetOption("e2_target"))
 
+    from dbt.project import E2Project
+
+    # Crank up the e2studio XML parsing class
+    cproject = E2Project(log, GetLaunchDir())
+
+    # Brand and sort things appropriately so we follow the general scheme of,
+    # but don't overlap with e2 build directories.
+    e2_build_targets = list(cproject.get_build_targets())
+    dbt_build_targets = [p.replace("e2-build-", "dbt-build-") for p in e2_build_targets]
+    e2_build_targets.sort()
+    dbt_build_targets.sort()
+
+    if not BUILD_TARGETS:
+        print(
+            "Cannot build without a specified target. Try one or more of the following:"
+        )
+        print("    Direct Targets:")
+        for d_t in dbt_build_targets:
+            print("        {}".format(d_t))
+        # print("    Special Targets:")
+        # print("        all (builds all targets, sequentially)")
+        print("\nTo clean a build, specify a target and use the flag '-c'.")
+        Return()
+
     # Launch and configure a new environment for each target BUILD_TARGETS
     # contains plain commandline targets or the contents of -e2_target
     # ONLY IF there are no commandline targets.
-    for target_entry in BUILD_TARGETS:
+    for target_raw in BUILD_TARGETS:
         # Convert entry to plain string so there are no differences
+        target = str(target_raw)
         # between plain commandline targets and "option" targets
-        target = str(target_entry)
         multienv[target] = env.Clone(
             BUILD_LABEL=target, BUILD_DIR=os.path.abspath(target)
         )
 
         if not multienv[target].SConscript(
             os.path.join("site_scons", "e2_xml.scons"),
-            exports={"env": multienv[target], "log_parent": log},
+            exports={
+                "env": multienv[target],
+                "log_parent": log,
+                "cproject": cproject,
+                "e2_build_targets": e2_build_targets,
+                "dbt_build_targets": dbt_build_targets,
+            },
         ):
             success = False
             Return("success")
--- a/SConstruct
+++ b/SConstruct
@@ -15,6 +15,7 @@ REL_SOURCE_DIR = "src"
 PROJECT_SOURCE_DIR = os.path.abspath(REL_SOURCE_DIR)
 
 print("\nDeluge Build Tool (DBT)\n")
+print("For help use argument: --help\n")
 
 # Not really sure how the logger will be used yet, but wanted it
 # in here.
@@ -43,7 +44,7 @@ from dbt.util import (
 
 if GetOption("base_config") not in ["e2_xml"]:
     print("Error: invalid base config mode.")
-    Exit(1)
+    Return()
 
 # Start up a basic environment
 env = Environment(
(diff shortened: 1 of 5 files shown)


litui:
Sounds good. I'll put that into the post-virtualdir options.

litui:
And I'll add a -e2_cdb_main=<blah> flag to my to-do list for the symlinking. I'll have to remember how symlinking on windows platforms works and whether or not it needs admin. Been awhile.

litui:
Readying this batch of changes for review to try to get them in sooner rather than later, as they're prerequisites for the next bunch of stuff.

SConstruct
```diff
@@ -62,154 +81,98 @@ if GetOption("base_config") == "e2_xml":
...
+    VariantDir(
+        os.path.join(t_env["BUILD_LABEL"], REL_SOURCE_DIR), "#src", duplicate=False
+    )
 
-# Export a `compile_commands.json` for help
-env.Tool('compilation_db')
-env.CompilationDatabase()
+    # Export a `compile_commands.json` for help
+    # t_env["COMPILATIONDB_PATH_FILTER"] = t_env["BUILD_LABEL"]
+    t_env["COMPILATIONDB_USE_ABSPATH"] = True
```
bobtwinkles, line 140: Can move this to the default env, probably? I think the `Tool` invocation can be as well but I'm not a SCons expert.

site_scons/commandline.scons
```diff
@@ -17,10 +17,21 @@ AddOption(
 AddOption(
     "--e2_target",
     dest="e2_target",
-    default="dbt-build-release-oled",
     type="string",
     nargs=1,
     action="store",
     metavar="TARGET",
-    help='Build target directory as in e2studio, prefaced with "dbt-" instead of "e2-"',
+    help='Legacy single-target e2 option; target prefaced with "dbt-" instead of "e2-"',
```
bobtwinkles, line 24: If you want to nuke this in this PR, the build dockerfile uses `dbt` as an entrypoint so modifying the `run`  in the "Run build" step should be sufficient to keep the build working. Something like this:

```diff
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 6532849..f370332 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
             --user=$(id --user):$(id --group) \
             -v $(pwd):/src \
             ghcr.io/bobtwinkles/deluge-ci-images:main \
-            --e2_target=dbt-build-${{ matrix.profile }}-${{ matrix.hardware }}
+            dbt-build-${{ matrix.profile }}-${{ matrix.hardware }}
       - name: Publish built firmware (bin)
         uses: actions/upload-artifact@v3
         with:
```

If there's another reason to keep the option around we can do that too.

SConstruct
```diff
@@ -62,154 +81,98 @@ if GetOption("base_config") == "e2_xml":
...
+for target, t_env in multienv.items():
+    # Helper functions borrowed from FBT for wrapping lengthy commandlines into temp files.
+    # This is primarily useful in windows, but shouldn't hurt anything on other platforms.
+    wrap_tempfile(t_env, "LINKCOM")
+    wrap_tempfile(t_env, "ARCOM")
+
+    # VariantDir does the magic to ensure output goes to the dbt- whatever
+    # build directory. Careful: THIS CAN BE REALLY FINICKY if paths aren't set right
+    VariantDir(
+        os.path.join(t_env["BUILD_LABEL"], REL_SOURCE_DIR), "#src", duplicate=False
```
bobtwinkles, line 135: Is there a reason to use `BUILD_LABEL` instead of `BUILD_DIR` here? Given we're constructing a path the latter feels more appropriate (but it will generate a absolute path if that will cause the problems mentioned above?)

litui: VariantDir is, as noted VERY particular about what types of paths go where. I spent most of last weekend trying to get it to properly cooperate and this was the result. In this case, it prefers a path relative to the location of the SConstruct. It'd probably be a touch safer to os.path.relpath(t_env["BUILD_LABEL"]) so it tweaks the slashes as needed in case we ever have slashes (sub-subdirectories) in a label but that seems like a pretty edge case.

It's also very touchy if it gets run from within an SConscript file or as part of an SConscript() function call. 😿

bfredl:
minor suggestion:

```diff
diff --git a/scripts/dbt/util.py b/scripts/dbt/util.py
index 7e3f521..ede6d83 100644
--- a/scripts/dbt/util.py
+++ b/scripts/dbt/util.py
@@ -6,7 +6,7 @@ from SCons.Errors import StopError
 from SCons.Subst import quote_spaces
 
 WINPATHSEP_RE = re.compile(r"\\([^\"'\\]|$)")
-SOURCEWALK_RE = re.compile(r"\.[cs][p]{0,2}", re.IGNORECASE)
+SOURCEWALK_RE = re.compile(r"\.[cs][p]{0,2}$", re.IGNORECASE)

 # Used by default when globbing for files with GlobRecursive
 # Excludes all files ending with ~, usually created by editors as backup files
```

this prevents things like swapfiles (`.voice.cpp.swp`) to be detected as source files and cause trouble.

litui:
> minor suggestion

Will fix. That was an oversight.

site_scons/commandline.scons
litui: No reason to keep it around, but for any workflows others might have developed around it. I think I'd rather wait just as a show that I'm not out to introduce something only to break it the next day ;). Working with backward compatibility in mind and avoiding breaking legacy stuff too soon is probably a good way to approach this in governance. Maybe in a couple weeks now that I've changed the text to indicate that it's legacy.

SConstruct
litui: It *can* definitely go into the default env, but my hope was to use the default env as a starting point for other functionality like debugging that may not be handling a target in the "build" sense. Since it's build target related, I put it in the t_env (badly named "generic target environment").

litui:
Thanks for the review, @bobtwinkles, @bfredl! Going to go ahead and merge.

PR #75, opened 2023-06-20
Fix undeclared function warnings

bobtwinkles:
Fixes about 80 warnings. 

Bonus editorconfig change. It's still not perfect, but hopefully it generates slightly less spurious changes now.


bobtwinkles:
Someone with e2 definitely needs to test this one, I edited the XML by hand in a way that DBT accepts but it might not be kosher for e2.

PR #80, opened 2023-06-21
Fix warnings

bobtwinkles:
This cleans up the majority of the remaining warnings in the debug-oled build, with the exception of:
- Some `-Wstringop-overflow` warnings in the Segger code, which look like it's the compiler losing track of buffer sizes and not an actual problem. Further investigation required.
- `-Winvalid-offsetof` warnings from the `Patcher` code. I think that logic can be turned in to C++ templates to retain the same machine code while better explaining what we're doing to the compiler, but it will be pretty subtle work with far-reaching implications so I'm saving it for a followup.
- 160+ `-Wstack-usage` warnings in debug mode. I'm not sure what to do with these really, it's probably good to keep tabs on it but with that many warnings it's basically just noise. Maybe we can do something with `-fstack-usage` instead?

The release-oled build has a bunch of `-Wstringop-overflow` warnings in the PIC UART interaction that require separate investigation. It seems like a similar situation where the warning infrastructure loses the plot after some optimizations are applied and thinks the destination is of size zero.

This is probably easiest to review change by change, they don't really build on top of eachother just fix different warnings.

### DIFF
--- a/src/AudioClip.cpp
+++ b/src/AudioClip.cpp
@@ -1124,7 +1124,7 @@ void AudioClip::setPos(ModelStackWithTimelineCounter* modelStack, int32_t newPos
 
 uint64_t AudioClip::getCullImmunity() {
 	uint32_t distanceFromEnd = loopLength - getLivePos();
-	bool doingTimeStretching =
-	    (voiceSample && voiceSample->timeStretcher); // We're gonna cull time-stretching ones first
-	return ((uint64_t)voicePriority << 33) + ((uint32_t)!doingTimeStretching << 32) + distanceFromEnd;
+	// We're gonna cull time-stretching ones first
+	bool doingTimeStretching = (voiceSample && voiceSample->timeStretcher);
+	return ((uint64_t)voicePriority << 33) + ((uint64_t)!doingTimeStretching << 32) + distanceFromEnd;
 }
--- a/src/functions.h
+++ b/src/functions.h
@@ -166,6 +166,15 @@ inline int32_t lshiftAndSaturateUnknown(int32_t val, uint8_t lshift) {
 	return signed_saturate_operand_unknown(val, 32 - lshift) << lshift;
 }
 
+static constexpr uint32_t charsToIntegerConstant(char a, char b, char c, char d) {
+	return (static_cast<uint32_t>(a)) | (static_cast<uint32_t>(b) << 8) | (static_cast<uint32_t>(c) << 16)
+	       | (static_cast<uint32_t>(d) << 24);
+}
+
+static constexpr uint16_t charsToIntegerConstant(char a, char b) {
+	return (static_cast<uint16_t>(a)) | (static_cast<uint16_t>(b) << 8);
+}
+
 int32_t stringToInt(char const* string);
 int32_t stringToUIntOrError(char const* mem);
 int32_t memToUIntOrError(char const* mem, char const* const memEnd);
--- a/src/AudioFile.cpp
+++ b/src/AudioFile.cpp
@@ -96,7 +96,7 @@ int AudioFile::loadFile(AudioFileReader* reader, bool isAiff, bool makeWaveTable
 			switch (thisChunk.name) {
 
 			// Data chunk - "data"
-			case 'atad': {
+			case charsToIntegerConstant('d', 'a', 't', 'a'): {
 				foundDataChunk = true;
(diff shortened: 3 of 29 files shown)


src/AudioClip.cpp
```diff
@@ -1126,5 +1126,5 @@ uint64_t AudioClip::getCullImmunity() {
 	uint32_t distanceFromEnd = loopLength - getLivePos();
 	bool doingTimeStretching =
 	    (voiceSample && voiceSample->timeStretcher); // We're gonna cull time-stretching ones first
-	return ((uint64_t)voicePriority << 33) + ((uint32_t)!doingTimeStretching << 32) + distanceFromEnd;
+	return ((uint64_t)voicePriority << 33) + ((uint64_t)!doingTimeStretching << 32) + distanceFromEnd;
```
bobtwinkles, line 1129: This is a functional change but is probably a bugfix.

src/functions.h
```diff
@@ -166,6 +166,15 @@ inline int32_t lshiftAndSaturateUnknown(int32_t val, uint8_t lshift) {
...
 
+static constexpr uint32_t charsToIntegerConstant(char a, char b, char c, char d) {
+	return (static_cast<uint32_t>(a)) | (static_cast<uint32_t>(b) << 8) | (static_cast<uint32_t>(c) << 16)
+	       | (static_cast<uint32_t>(d) << 24);
+}
+
+static constexpr uint16_t charsToIntegerConstant(char a, char b) {
+	return (static_cast<uint16_t>(a)) | (static_cast<uint16_t>(a) << 8);
+}
+
```
bobtwinkles, line 177: I tested these both using godbolt and by manually inspecting the generated assembly so I'm 85% sure these generate equivalent values to the multichar constants, but someone should definitely look at them with a fresh pair of eyes and thorough consideration of endianness of the target platform.

We could use them to replace a bunch of magic constants through the XML parsing logic, but I'll leave that for someone more ambitious than I in a followup change.

src/AudioClip.cpp
litui: Can you confirm for me if this piece of code still works as intended? I'm with you that it's probably a bug fix or oversight originally, but on the off chance it was intentional.

bobtwinkles: I don't have a project around that would hit this code. Should I just rip this commit out to a separate PR so we can consider it in isolation? It'll be easy to test once I can set breakpoints.

jamiefaye:
Lets split out that problematic line as suggested. Then we can merge the rest in.

src/AudioClip.cpp
litui: Probably safest to do that.

src/functions.h
litui: These seem like a reasonable alternative approach to me and I trust your testing on them for these cases.

litui (approved):
Haven't tested on hardware yet, but these all seem reasonable. There are a couple cases here that raise my eyebrow about whether they will run identically on the actual hardware. It's nice seeing these clear up in the build output too ^_^

I suggest we give it a run on hardware before merging, otherwise we good.

bobtwinkles:
If someone has a diversity of song files to hand to test loading with/without this change that's probably advisable.

jamiefaye:
Sounds strange over on my 7LED. Even a 1 clip, 1 note on synth 0 produces a
lingering "outter space" sound.

I just took the branch as it was about 45 minutes ago.

– Jamie

jamiefaye:
It now sounds OK. Has anyone tested on OLED? If it works there too, we can merge.

bobtwinkles:
I'll do some more thorough testing on OLED in a few hours.

jamiefaye:
As for the stack size warnings, there is a setting in e2Studio that sets the limit. We could set that to something bigger than 100.

bobtwinkles:
OK, I think I've convinced myself this works OK on an OLED Deluge. The only thing I haven't thoroughly tested is the wavetable loading code changes since I don't have a good idea of what impact those tags are supposed to have on the flow.

bobtwinkles:
I've added the clip culling commit back because after checking where it's used, the priority computed here appears to be completely ignored. In general it appears that the whole clip priority system is broken for audio clips:

https://github.com/SynthstromAudible/DelugeFirmware/blob/ff8e15741eb381caa794ae1b8933b9e0100a567a/src/song.cpp#L4493-L4505

PR #81, opened 2023-06-21
doc: document what version of clang-format we use

bobtwinkles:


litui:
Just gonna bundle clang (or at least the clang-format and utils part of it) into dbt-toolchain for simplicity. 15.0.7 is the latest xpack.

bobtwinkles:
If we bump to 15 we'll need to do that synchronously with a minor reformatting PR, as `soundeditor.cpp` triggered a bug in 14.

litui:
No worries, I'll load in 14 for now.

PR #87, opened 2023-06-22
DBT-Toolchain 8

litui:
- Enable DBT to pull new version of toolchain
- Toolchain adds: clang-format 14
- Toolchain removes: cmake
- Toolchain fixes: mac dotfiles should no longer be tarred.


litui:
*looks at workflow failures* Ha... WELL, this is going to pose problems innit.

litui:
Submitted PR https://github.com/bobtwinkles/deluge-ci-images/pull/1 to get the CI's DBT upgraded to 8 so we can upgrade `community`'s DBT to 8.  Maybe @bobtwinkles and I should look at integrating dbt-toolchain and the deluge-ci-images repos so this can be done in one pass. 🤔

bobtwinkles:
The other thing I should fix is find a way to publish to tags other than `main` in the ci-images repo, so we can pull specific versions of the toolchain (instead of just pulling `main` all the time)

bobtwinkles:
#89 switches the check workflow to use a versioned toolchain so things can be moved in lockstep.

PR #89, opened 2023-06-22
workflow: use versioned toolchain

bobtwinkles:
This way commits that require atomic updates across `dbt` and the toolchain can work.


litui (approved):
I'm not a big fan of how many places we need to specify the version right now, but it's okay for the moment until we think of a better place/approach to storing and propagating the version.

PR #68, opened 2023-06-18
[RFC] Project structure/reorganization

stellar-aria:
This draft is an open RFC towards introducing a global file structure to the project. I've included the preliminary work I've done so far. There's probably some mistakes, but it's very late and I'll try to find them tomorrow and finalize this v1 draft.

Otherwise, I'm actively looking for comments/opinions on naming, location, critiques on the structure, etc.

No files have been edited, nothing has been renamed. There are definitely some things that _should_ be edited (autogenerated empty constructer/destructors that could be placed in the headers and allowing for removal of the corresponding `.cpp` file, renaming `AudioSample.hpp` to `StereoSample.hpp` to match the class in it, etc.), but that can be left for later.

This is **not** a working build. Build files will have to be updated accordingly when/if the file structure is finalized.

PR #96, opened 2023-06-23
toolchain: single source of truth for required version

bobtwinkles:
this adds a file, `toolchain/REQUIRED_VERSION`, which acts as the source of truth for DBT and CI to pull the required toolchain version from.

I haven't tested this on Windows or OSX because I don't have those platforms, but at least the general shape should work.


wright:
This builds OK on OSX.

litui:
I can test it in windows later today. Thanks for doing this!

litui (approved):
Appears to work fine in Windows!

jamiefaye:
Lets do this one and #68 soon!

PR #104, opened 2023-06-25
Linkerscript rework and proper initialization

stellar-aria:
This completely reworks the linkerscript to be more in line with Renesas' own examples (from their software pack) and other embedded ARM systems. This supplants PR #42

## Added:
`NOLOAD` added to the `.bss` and `.heap` sections, making debug-loading _much_ faster.
Initialization of global variables now functions properly, removing the need for dedicated initialization.
Uniform initialization syntax (braces) for all of the global variables, to make it clear they're being initialized when they're declared.

## Changed:
`main1` is now `main`
`main2` is now `deluge_main` (as it is a 'main' function living in deluge.cpp...)

# Notes
This does _not_ make any changes to `SoundEditor`'s menu system. That refactoring is coming in a future PR.


src/Deluge.cpp
```diff
@@ -502,7 +502,7 @@ extern "C" volatile uint32_t usbLock;
 
 extern "C" void usb_main_host(void);
 
-extern "C" int main2(void) {
+extern "C" int deluge_main(void) {
```
dctucker, line 505: :sparkling_heart:

stellar-aria: thaaaanks

stellar-aria: must have missed that somehow 💯

stellar-aria: oh wait lol I misunderstood I thought I'd forgotten to add that chunk XD

stellar-aria:
Did a double-check and found a potential issue. Everything else looks good, so this just needs testing now.

jamiefaye:
I do see the e2studio debugger stop twice now, once at the entry point and once at the first line of the new main()
Not sure why, the second breakpoint isn't shown in the GUI. Is it baked in somewhere?

stellar-aria:
most debug GUIs have a built-in "break at main()", yeah, so e2studio might be doing that now

PR #108, opened 2023-06-26
FIX: LFO sync menu working on 7SEG; working on OLED;

### LINKED ISSUE #106: LFO sync menu options on 7SEG display are broken
Using 7SEG release bin from latest community build https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5359807075

## Steps to reproduce (7SEG only):
 - load latest community build (from above)
 - power cycle (to reset defaults for  menu option later)
   should be in a **new synth 0**
 - choose menu shortcut LFO1-SYNC 
   should display `[oFF ]`
 - (select, scroll left) `[28t ]`
 - (select, scroll left) `[4th ]`
 - (select, scroll left) `[2nd ]`
 - (select, scroll left) `[6th ]`
 - (select, scroll left) `[th  ]`
 - (select, scroll left) `[th  ]`
 - (select, scroll left) `[nd  ]`

Note that (select, scroll right) the first few options look OK. But eventually it'll loop around to the newer options, as above.

## Expected behaviour 7SEG
 - [x] all the menu options should be readable
 - [x] any menu options longer than 4 characters should scroll to show the full string (these are best avoided if possible).
tested by @chrisbc on &7SEG see #108 

## Expected behaviour OLED
 - [x] all the menu options should be readable

chrisbc:
if someone can test OLED  please? DONE (see below)

closes #106


phfalk:
I gave it a quick test on OLED and the descriptions are a little to wide to fit the screen and don't scroll. That's why my code replaced the `-notes` with `-tplts` or `-dtted`. I wasn't really convinced that that was the right thing to do, but also didn't really see a better way to do it and still stay compatible with how the text is generated.

I think maybe a compromise between the two solutions would work best. We could get rid of all the `#ifdef`s by making the text scroll on 7seg, but keep the replacement to at least somewhat shorten the text. The text would still be descriptive enough in my opinion and the `-notes` part doesn't really add much information anyway.

ok-reza:
Perhaps dotted SYNC stuff on 7SEG could just use a period to reduce scrolling?
eg. `[16TH.]` `[ 2nd.]`

It would simplify the problem if `-TH`, `-ND`, `-BAR` got removed too, at least for 7SEG. That way `[ 128]` is 128th notes and `[128t]` is 128th triplets. But some confusion may occur too so it's not a perfect solution.

I would be curious to see how scrollable menu items feel on 7SEG. It is weird (and inconsistent) that menu items aren't manually scrollable in the first place. Eg. why can't we manually scroll from `[rAnd]` for arp mode and reveal `[ndoM]`?

sichtbeton:
> Perhaps dotted SYNC stuff on 7SEG could just use a period to reduce scrolling? eg. `[16TH.]` `[ 2nd.]`
> 
> It would simplify the problem if `-TH`, `-ND`, `-BAR` got removed too, at least for 7SEG. That way `[ 128]` is 128th notes and `[128t]` is 128th triplets. But some confusion may occur too so it's not a perfect solution.
> 
> I would be curious to see how scrollable menu items feel on 7SEG. It is weird (and inconsistent) that menu items aren't manually scrollable in the first place. Eg. why can't we manually scroll from `[rAnd]` for arp mode and reveal `[ndoM]`?

agree on all of it. also: on my 7seg it doesnt scroll at all.

chrisbc:
Thanks all!

@sichtbeton ....It def scrolls, so you must not have the correct bin file. I'll add a link below. Apologies I should have made this easier!

@phfalk in my PR "-notes" is gone. You don't have my code if you're seeing that still. AFAIK I've only made it shorter, not longer for OLED. It's is longer for 7SEG, but scrolls, so .....

@ok-reza Perhaps you can try it on 7SEG too? When I link the  binary

**Notes:** 
 -  I decided for first attempt I'd make the descriptions identical for OLED & 7SEG and see if scrolling menu works.
  - all that non-indented `#if HAVE_OLED` were making my brain suffer  - best avoided at all cost IMHO .
  - I found scrollingText() is used in a couple of other menuItems (MidiDevices, Range)  so felt it was a reasonable option here....
  - IMHO scrollingText work fine for this, but others might not agree :) . V happy to refine after we have some solid  test feedback

chrisbc:
OK - I have my fork running the build scripts for testing...
 - [7SEG release](https://github.com/mupaduw/DelugeFirmware/suites/13881950866/artifacts/771699539)

phfalk:
> @phfalk in my PR "-notes" is gone. You don't have my code if you're seeing that still. AFAIK I've only made it shorter, not longer for OLED. It's is longer for 7SEG, but scrolls, so .....

Are you sure about this? As far as I can see, you have deleted the lines (70-71) that replace `-notes` with the new type. I double checked that I checked out the correct branch, but it's easy to get lost with the different forks.

> all that non-indented `#if HAVE_OLED` were making my brain suffer  - best avoided at all cost IMHO .

I generally agree. That code is a little bit ugly and it would be much better if there was some abstraction that handles this differently. But for now, the compile time macro is what we have and the established style (as far as I can see) in this project is to not indent the `#if`s.

> IMHO scrollingText work fine for this, but others might not agree :) . V happy to refine after we have some solid  test feedback

I don't really have an opinion on scrolling text on the 7seg, but my idea was to fit all the info in four characters like this:
```
Even sync:    [ BAR] [ 2ND] [ 4TH] ... [16TH] ... [128T] [256T]
Triplet sync: [TBAR] [T2ND] [T4TH] ... [T16T] ... [T128] [T256]
Dotted sync:  [DBAR] [D2ND] [D4TH] ... [D16T] ... [D128] [D256]
```

But I think @ok-reza's idea could also work to avoid the need for scrolling. The problem I see with that approach is that [128T] is ambiguous with what people are used to.

sichtbeton:
> > @phfalk in my PR "-notes" is gone. You don't have my code if you're seeing that still. AFAIK I've only made it shorter, not longer for OLED. It's is longer for 7SEG, but scrolls, so .....
> 
> Are you sure about this? As far as I can see, you have deleted the lines (70-71) that replace `-notes` with the new type. I double checked that I checked out the correct branch, but it's easy to get lost with the different forks.
> 
> > all that non-indented `#if HAVE_OLED` were making my brain suffer  - best avoided at all cost IMHO .
> 
> I generally agree. That code is a little bit ugly and it would be much better if there was some abstraction that handles this differently. But for now, the compile time macro is what we have and the established style (as far as I can see) in this project is to not indent the `#if`s.
> 
> > IMHO scrollingText work fine for this, but others might not agree :) . V happy to refine after we have some solid  test feedback
> 
> I don't really have an opinion on scrolling text on the 7seg, but my idea was to fit all the info in four characters like this:
> 
> ```
> Even sync:    [ BAR] [ 2ND] [ 4TH] ... [16TH] ... [128T] [256T]
> Triplet sync: [TBAR] [T2ND] [T4TH] ... [T16T] ... [T128] [T256]
> Dotted sync:  [DBAR] [D2ND] [D4TH] ... [D16T] ... [D128] [D256]
> ```
> 
> But I think @ok-reza's idea could also work to avoid the need for scrolling. The problem I see with that approach is that [128T] is ambiguous with what people are used to.

I agree, a solution without scrolling would be very good. The deluge scrolls way too much anyways :-)

thank you for doing this!

chrisbc:
> Are you sure about this? As far as I can see, you have deleted the lines (70-71) that replace `-notes` with the new type. I double checked that I checked out the correct branch, but it's easy to get lost with the different forks.

NO it turns out I didn't GROK this, sorry. I'll reinstate that approach as it's probably least problematic. Thanks @phfalk !

chrisbc:
closes #106 

OK, this is now tested and accepted on OLED and 7SEG. See end of https://discord.com/channels/608916579421257728/1123078039518056468 for that.

@jamiefaye @litui can you review this PR now pls?

Big thanks to @phfalk @alex-r-t & @ok-reza for the testing and patience.

PR #117, opened 2023-06-28
Linkerscript rework and global initializations - Attempt 2

### LINKED ISSUE #113: Commit 884da39 wouldn't boot
Using 7seg deluge, tested before and after this merge. Loading firmware from SD card resulted in unresponsive Deluge on commits after 884da398a3b622bdab414b420041cac867b26008 until the change was reverted.

@stellar-aria has a bug fix in the works.

stellar-aria:
Despite the builds working completely fine over hardware debuggers, @litui found an issue where the alignment and positioning of certain linkerscript sections made it so the deluge would not boot when the firmware was written to the QSPI ROM.

This is and should only ever be an issue with the linkerscript, so at least now we know the root cause, (and honestly should never have to edit the thing again).

There seemed to be an issue with initialization order (from my panicked attempts at fixing things last night), but methodically updating blocks of code to match the breaking commit eventually revealed nothing, so that might have been an issue with the linkerscript as well.

I've tested this on a 7seg Deluge with the bin on the SD card to replicate @litui's problems, and everything seems to be working well now.

Closes #113


bfredl:
This seems to work on the OLED as well, but I noticed one issue:

In the global settings menu (shift-select), the last two menu items (the newly added Runtime Features, as well as Firmware version) are both blank. you can still scroll to them and open them, they just have no text.

PR #119, opened 2023-06-29
proof of concept: Wren

dctucker:
[Wren](https://wren.io/) is a classy little scripting language that fits neatly into the Deluge. This branch demonstrates automatically loading and running a "hello world" from the SD card.

Must be built using `ENABLE_WREN=1`:
```
dbt dbt-build-debug-oled ENABLE_WREN=1
```

Create a file called `scripts/init.wren` on your SD card with the following contents:
```wren
Deluge.onInit {
    Deluge.print("HELLO WORLD!")
    Deluge.press(Buttons.keyboard)
}

Deluge.bind(Buttons.syncScaling) {|down|
  if (down) {
    Deluge.print("Sync scaling")
  }
}
```
With firmware from this branch it should load this file and you should it start in keyboard mode and "HELLO WORLD" pop up briefly.

Documentation has been added [here](https://github.com/dctucker/DelugeFirmware/tree/feature/wren/src/deluge/vm/README.md)

It's possible another VM will prove itself a better fit than Wren. [Jerryscript](https://jerryscript.net/) and [LuaJIT](https://luajit.org/performance_arm.html) have also been proposed, though there are additional challenges to get them to build for Deluge.

The long-term vision for implementing a scripting engine would be allowing users to load scripts from the SD card that would hook into functions within the Deluge, manipulate songs and parameters, respond to events to enable macros. It's also possible that it could serve as an easier-to-use front-end for things like custom menus or interfaces. Hooking into audio subroutines may be possible, though quite likely at the cost of performance. In any case, the goal is to lower the barrier to entry for folks who want to write programs that can control the Deluge, and being able to quickly see results without having to do a firmware upgrade helps with that.

Thanks @stellar-aria for working on the linker script and memory allocator working.


jamiefaye:
Wren would be invaluable as a scheme for automated testing too. Which we will need.

sichtbeton:
super interesting. how would one start with this? how would this scripting language interface with the deluge?

dctucker:
> how would one start with this?

Currently only the Hello World mentioned in the PR description has been implemented, and even that's subject to change since this is in such an early stage.

> how would this scripting language interface with the deluge?

I'm interpreting this question very generally since I'm not exactly sure how to answer this, but feel free to ask a followup question if I missed the mark 😅.

Long-term, there will be API documentation with examples of what is possible. The overall workflow is write a script, copy it to the`scripts/` folder on the SD card, and the Deluge would load the script at power on (possibly when a card is inserted if folks want that). I'm thinking this might be useful for folks who want to build a custom user interface, so we'll enable callbacks like `onPad` and `onSelect`, and expose foreign functions that allow the script to do things like `Deluge.startPlayback()` or `ui.menu.enterSubmenu()`. 

Hopefully this topic also prompts discussion more generally about how we want to refactor the firmware code. Just a fuzzy idea of where I think this could be headed.

sichtbeton:
> Long-term, there will be API documentation with examples of what is possible. The overall workflow is write a script, copy it to the`scripts/` folder on the SD card, and the Deluge would load the script at power on (possibly when a card is inserted if folks want that). I'm thinking this might be useful for folks who want to build a custom user interface, so we'll enable callbacks like `onPad` and `onSelect`, and expose foreign functions that allow the script to do things like `Deluge.startPlayback()` or `ui.menu.enterSubmenu()`.
> 

So this would mean, I could just write stuff that would use the language with these callback. got it. I would see this be handy for stuff like changing values for a lot of clips or moving them around randomly or something like this. 

so my question is: can it only be run at startup? bc this would limit the usability a bit. if on the other hand it can be called while the deluge is running (not neccessarily playing) it could be much more applicable. I could also imagine small workflow enhancers like "flush all unused kit items from kit" which could be written with the language. a kind of "mini-plugin" system.

and to make it more accessible to musicians / non coders is a very good thing. I guess the easier it is the better.

great!

dctucker:
> can it only be run at startup?

Not necessarily, but it will likely be most useful there. You could create a new submenu of different actions to perform, and add this submenu as an item within one of the existing Settings / Sound Editor / Feature menus. Loading at startup is preferable because parsing the script may take some CPU power. Either way I think we'll want to have a menu that lets folks select a script file to load and run immediately.

> a kind of "mini-plugin" system.
> easier it is the better.

Very much this.

stellar-aria (approved):

sichtbeton:
is there a .bin for 7seg that could be tested?

dctucker:
https://dctucker.com/DelugeFirmware/wren-firmware.tar.gz

dctucker:
@sichtbeton https://dctucker.com/DelugeFirmware/wren-firmware.tar.gz

src/deluge/vm/wrenimpl.cpp
```diff
@@ -0,0 +1,238 @@
...
+	//UINT written;
+	//FIL fil;
+	//auto res = storageManager.createFile(&fil, "SCRIPTS/log.txt", true);
+	//size_t len;
+
+	switch (errorType) {
+	case WREN_ERROR_COMPILE:
+		VM::print("E compile");
+
+		uartPrint(mod);
```
bfredl, line 58: it would be useful to switch to the new highlevel `Debug::print()` family of functions to allow debugging wren over USB.

PaulFreund:

dctucker:
If there's demand there is for this as a feature, I could try to get a commit pushed that resolves these conflicts in the next few days.

PaulFreund:
I think ultimately it is up to you if you want to see it in the codebase. Unfortunately though we already have a lot of problems with our binary size that grew between 350 and 450k and the performance of samples is very noticeably degraded, see issue #350. We also can't use -O0 builds as they don't fit in memory anymore. If you want to put it in the codebase we need to make sure it is only linked with a compile switch.

PaulFreund:
Closing for now but we can reopen any time :)

PR #121, opened 2023-06-30
Towards a unified Display interface

stellar-aria:
A lot of discussion has happened around making the "display" types both refactored towards a common interface, and eventually selectable at runtime, both for ease of distributing unified binaries and for scripting/testing purposes.

- [x] Remove OLED use in segmentdriver
- [x] Unified Display interface (DisplayBase)
- [x] Derived objects OLED and SevenSegment
- [x] Remove HAVE_OLED macro
- [x] Replace use of segmentdriver throughout codebase
- [x] Some kind of XML flag or device detection to figure out if OLED or 7SEG for general users
- [x] Runtime string selection (will allow for both runtime screen type selection and localization)
- [x] Unify build configs
~~- [ ] 7SEG emulator Display object for testing use on an OLED~~ Will be done at a later PR
~~- [ ] Factor out common code~~ Will be done later as part of the general continuing work in this area of the codebase


src/deluge/hid/matrix/matrix_driver.cpp
```diff
@@ -23,7 +23,7 @@
 #include "gui/ui/browser/sample_browser.h"
 #include "hid/matrix/matrix_driver.h"
 #include "util/functions.h"
-#include "hid/display/numeric_driver.h"
+#include "hid/display.h"
```
bobtwinkles, line 26: unused, I think. But if we unify all UI output contexts then we do though.

src/deluge/testing/hardware_testing.cpp
```diff
@@ -23,7 +23,7 @@
 #include <string.h>
 #include "io/midi/midi_engine.h"
 #include "hid/encoders.h"
-#include "hid/display/oled.h"
```
bobtwinkles, line 26: unused I think

src/deluge/util/container/array/c_string_array.cpp
```diff
@@ -20,7 +20,7 @@
 #include <string.h>
 #include "util/d_string.h"
 #include "util/functions.h"
-#include "hid/display/numeric_driver.h"
+#include "hid/display.h"
```
bobtwinkles, line 23: unused, i think

bobtwinkles (commented):
some notes, many of which you're probably aware of

bfredl:
This hangs at the boot screen for me (after completing the logo animation). oled-debug build.

I checked out the second last commit (i e without the "Attempt to only use as OLED when detected." change) and that works fine for me. did a quick run through the sound and settings menus and didn't find any issues.

src/deluge/gui/l10n/l10n.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
```
bfredl, line 1: as we don't use `.h` vs `.hpp` for the most part for C vs C++ headers, perhaps could be called out more explicitly in the name as well, like `l10n_for_c.h`. And/or explained in a short comment. I was confused at first "why just four strings??".

stellar-aria: Yeah, this is only really used by the RZA1 library in a couple places anyways. Maybe I should just move it over there, since that's all C anyways? Something like RZA1/extern/deluge_l10n.h

Common C headers are something I'd usually put in {project_root}/include, especially if they needed to be accessed by other internal-project libraries.

stellar-aria: or heck they should go in deluge.h probably

stellar-aria: Fixed

src/definitions_cxx.hpp
```diff
@@ -549,7 +549,7 @@ enum class LFOType {
 	RANDOM_WALK,
 };
 
-constexpr int32_t kNumLFOTypes = util::to_underlying(LFOType::RANDOM_WALK);
+constexpr int32_t kNumLFOTypes = util::to_underlying(LFOType::RANDOM_WALK) + 1;
```
stellar-aria, line 552: Just for reference, this was a bug I spotted while doing other things. It doesn't need to wait for this PR, but it is a bug.

src/deluge/gui/views/audio_clip_view.h
```diff
@@ -51,9 +51,8 @@ class AudioClipView final : public ClipView, public ClipMinder {
 	uint32_t getMaxLength();
 	uint32_t getMaxZoom();
 
-#if HAVE_OLED
+	// OLED ONLY
```
bfredl, line 54: I think comment like this is redundant when all functions/variables below already have "oled" in the name.

src/deluge/gui/menu_item/gate/selection.h
```diff
@@ -45,22 +47,25 @@ class Selection final : public menu_item::Selection<5> {
...
-
-		return {"Gate output 1", "Gate output 2", "Gate output 3", "Gate output 4", "Minimum off-time"};
-#else
-		return {"Out1", "Out2", "Out3", "Out4", "OFFT"};
-#endif
+	static_vector<std::string_view, capacity()> getOptions() override {
+		using enum l10n::String;
+		auto gate_output_fmt_string = l10n::getView(STRING_FOR_GATE_OUTPUT_N);
+		return {
+		    fmt::vformat(gate_output_fmt_string, fmt::make_format_args(1)),
```
bfredl, line 63: same issue here (use after free; menu gets funky on OLED)

src/deluge/gui/menu_item/cv/selection.h
```diff
@@ -41,19 +45,27 @@ class Selection final : public menu_item::Selection<2> {
...
+		//     std::views::iota(1_u32, capacity())      //<
+		//     | std::views::transform([](size_t idx) { //<
+		// 	      return fmt::vformat(l10n::get(STRING_FOR_CV_OUTPUT_N), fmt::make_format_args(idx));
+		//       })
+		//     | std::views::common;
+
+		// return {strings.begin(), strings.begin() + capacity()};
+
+		return {
+		    fmt::vformat(l10n::getView(STRING_FOR_CV_OUTPUT_N), fmt::make_format_args(1)),
```
bfredl, line 66: and another one

src/deluge/gui/menu_item/midi/devices.cpp
```diff
@@ -121,26 +123,21 @@ MIDIDevice* Devices::getDevice(int32_t deviceIndex) {
...
-#if HAVE_OLED
-	midiDeviceMenu.title =
-	    soundEditor.currentMIDIDevice->getDisplayName(); // A bit ugly, but saves us extending a class.
-#endif
 	return &midiDeviceMenu;
 }
 
-#if HAVE_OLED
-
 void Devices::drawPixelsForOled() {
```
bfredl, line 139: This menu (the one with multiple devices) also looks distorted on OLED. although when going into a specific device, the device name renders fine as a title.

bfredl:
I've done a bit of testing on OLED and most views/popups/etc work just fine. I've commented on some specific menu items which have issues tho.

src/deluge/deluge.cpp
```diff
@@ -543,26 +566,28 @@ extern "C" int32_t deluge_main(void) {
...
+	if (have_oled) {
+		// If OLED sharing SPI channel, have to manually control SSL pin.
+		setOutputState(SPI_SSL.port, SPI_SSL.pin, true);
+		setPinAsOutput(SPI_SSL.port, SPI_SSL.pin);
 
-#endif
+		setupSPIInterrupts();
+		oledDMAInit();
+		setupOLED(); // Set up OLED now
+		display = new deluge::hid::display::OLED;
```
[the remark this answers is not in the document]
stellar-aria: I think an easier way of doing a combined interface would be to simply make an aggregate type, and make the "type" check that's currently happening throughout the codebase instead a virtual function call like `display.have_oled()` that the aggregate can override true for both have_oled() and have_7seg().

I can try and work that into this PR now actually.

src/deluge/gui/l10n/strings.h
```diff
@@ -0,0 +1,647 @@
...
+	STRING_FOR_CANT_GRAB_TEMPO_FROM_CLIP,
+	STRING_FOR_SYNC_NUDGED,
+	STRING_FOR_FOLLOWING_EXTERNAL_CLOCK,
+	STRING_FOR_SLOW_SIDECHAIN_COMPRESSOR,
+	STRING_FOR_FAST_SIDECHAIN_COMPRESSOR,
+	STRING_FOR_REBOOT_TO_USE_THIS_SD_CARD,
+	STRING_FOR_LARGE_ROOM_REVERB,
+	STRING_FOR_MEDIUM_ROOM_REVERB,
+	STRING_FOR_SMALL_ROOM_REVERB,
+	STRING_FOR_USB_DEVICES_MAX,
```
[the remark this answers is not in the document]
stellar-aria: I was planning on sorting these alphabetically at some point, and wanted to keep the "USB_DEVICES" related stuff together, via common leading substring

src/deluge/gui/views/audio_clip_view.h
stellar-aria: Fixed.

src/deluge/gui/l10n/strings.h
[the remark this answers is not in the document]
bfredl: sure, but to match them the one in deluge.h could be changed just as well.

src/deluge/gui/menu_item/modulator/destination.h
```diff
@@ -25,8 +27,12 @@ class Destination final : public Selection<2> {
 	using Selection::Selection;
 	void readCurrentValue() override { this->setValue(soundEditor.currentSound->modulator1ToModulator0); }
 	void writeCurrentValue() override { soundEditor.currentSound->modulator1ToModulator0 = this->getValue(); }
-	static_vector<std::string, capacity()> getOptions() override {
-		return {"Carriers", HAVE_OLED ? "Modulator 1" : "MOD1"};
+	static_vector<std::string_view, capacity()> getOptions() override {
+		using enum l10n::String;
+		return {
+		    l10n::getView(STRING_FOR_CARRIERS),
+		    fmt::vformat(l10n::getView(STRING_FOR_MODULATOR_N), fmt::make_format_args(1)),
```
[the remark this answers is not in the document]
stellar-aria: Fixed.

src/deluge/gui/menu_item/gate/selection.h
stellar-aria: Fixed.

src/deluge/gui/menu_item/cv/selection.h
stellar-aria: Fixed.

src/deluge/gui/menu_item/midi/devices.cpp
stellar-aria: This has been fixed upstream via #344

stellar-aria: I'll merge that in

src/deluge/gui/l10n/strings.h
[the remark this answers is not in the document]
stellar-aria: Fixed

bfredl:
Thanks, those are fixed. One that I didn't check before: the sub-menu for a specific gate (V-trig vs S-trig) doesn't have any text at, neither title nor text on the items (you can still select between them). Perhaps something weird is going about with `options_[2]` which is dynamically populated (or deleted) for specific gates.

bfredl (approved):

PR #122, opened 2023-06-30
Drum Randomizer

alter-alter:
In KIT mode, while holding down the audition pad(s), pressing the [RANDOM] pad will randomly select a sample from the folder containing each sample.

Demo: https://youtu.be/gw_8WSeZs8o
\* The button combinations in the video are different from the current implementation.

**Known Issues:**
- Undo functionality is not available.
- There is a limit on the number of sample files within a folder (currntly 25 files) for processing speed.

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -74,6 +74,8 @@
 #include "gui/ui_timer_manager.h"
 #include "modulation/patch/patch_cable_set.h"
 #include "model/drum/midi_drum.h"
+#include "storage/multi_range/multi_range.h"
+#include "storage/audio/audio_file_holder.h"
 
 #if HAVE_OLED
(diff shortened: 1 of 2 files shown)


m-m-adams:
Since this changes user facing behaviour could you place it in the community settings menu - #59 ? There's still ongoing discussion here so the contributing guide isn't in the main repo yet but the discussion can be found in #56

alter-alter:
Thanks for the information. Should this feature be turned Off by default?

bobtwinkles:
It doesn't override any existing behavior right, so I think having it enabled by default is OK.

alter-alter:
> It doesn't override any existing behavior right, so I think having it enabled by default is OK.

Thank you!

litui (approved):
If this works on the hardware, we're good to go.

alter-alter:
Thank you all!

Issue #123, opened 2023-06-30
USB midi host does not work with multiple ports


m-m-adams:
[Issue from discord](https://cdn.discordapp.com/attachments/1107026299945496577/1124127368596947114/IMG_5341.png)

Midi hosting is not working properly for devices with multiple ports. Current implementation checks the port in the message and only passes if it's a port that has an associated profile, but the USB host driver side can currently only connect to the first port

PR #124, opened 2023-06-30
fix for devices with more ports than deluge supports

m-m-adams:
Fix for issue #123 by treating higher ports than the connected device as port 0. This matches the behaviour from the initial community release


m-m-adams:
I don't have any multi port devices recognized by Deluge to test but this is confirmed working by Sean.Good.Vibes in the Deluge discord

This change has no regression in behaviour of differentiating ports when used as a peripheral

litui (approved):

PR #129, opened 2023-07-01
Quantize

alter-alter:
This is a feature that enables quantization after recording.

In the Clip view, press the Note pad + shift + ◄► knob combination to switch between the NUDGE, QUANTIZE, and QUANTIZE ALL ROW modes. While pressing the note pad(s), press+turn the ◄►knob to the right to apply quantize, and to the left to apply humanize.

Demo: https://youtu.be/iAL447a1jD4

Issue:
- The switch operation between the Nudge feature and the Quantize feature is somewhat difficult.

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -543,7 +543,7 @@ doOther:
 	else if (x == xEncButtonX && y == xEncButtonY) {
 
 		// If user wants to "multiple" Clip contents
-		if (on && Buttons::isShiftButtonPressed()) {
+		if (on && Buttons::isShiftButtonPressed() && !isUIModeActiveExclusively(UI_MODE_NOTES_PRESSED)) {
 			if (isNoUIModeActive()) {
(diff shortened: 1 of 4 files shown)


bobtwinkles (commented):
This looks fine from a functional perspective. I think the user interface could use some iteration (it feels somewhat clunky to select). Spitballing some ideas:
  - We could steal the tempo encoder while a pad is pressed, but mis-input could be very destructive (though it is undoable, I think).
  - shift + keyboard to enter a quantize mode where tapping notes quantizes them and tapping the audition pad quantizes the whole row. This has the disadvantage of making the UI even more modal than it already is, but might be justified.

alter-alter:
Changed the quantize UI to use the tempo encoder. I think it is now more intuitive.

jamiefaye:
Forgive my ignorance, but what is the Note Pad?

jamiefaye:
OK, I see that you have to "turn on" this feature from a global menu.

alter-alter:
> Forgive my ignorance, but what is the Note Pad?

I apologize for the lack of clarity. It's the pad on which the note is placed.

PR #140, opened 2023-07-03
Add Debug launch profiles for VSCode

stellar-aria:
This adds VSCode launch profiles for debugging using a JLink running a remote GDB Server. There's also now a python script in the scripts/tasks folder that supports launching a debug server using either JLinkGDBServer or OpenOCD.

I personally have a script called task.py at the toplevel of the git directory that also lets me call any tasks in that folder as subcommands, (i.e. `python task.py debug` but I don't want to add too much stuff to the git root if we don't have to, unless there's approval/interest.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,55 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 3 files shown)


litui (approved):
Easy merge. 0 impact on core code or workflow outside of vscode.

PR #141, opened 2023-07-03
Horizontal Shift for Audio Clips

topisani:
I have always wanted to be able to change the start point of audio clips, like the end can be changed using the red bar. The current "easy" way to do it is: reverse clip, change end, reverse back.

I first looked at implementing a similar green start bar (which would probably be the most intuitive), but i understand why rohan left it out now - it introduces some complexity with being able to scroll left of zero etc. I do have a partial implementation of some of this, and if i can find good, intuitive behavior for it, i might submit that as well.

But for now, i chose to instead implement horizontal shift for audio clips. This is the feature where `▲▼` is held while `◀▶` is turned. It works the same as for instrument clips, except it does not wrap around, only shift along the underlying audio, and it cannot go left of zero (will display a "CAN'T SHIFT PAST START" message)


src/deluge/gui/views/clip_view.cpp
```diff
@@ -29,16 +29,23 @@
 #include "model/clip/clip_minder.h"
 #include "gui/views/view.h"
 #include "definitions.h"
+#include "model/consequence/consequence_clip_horizontal_shift.h"
+#include "memory/general_memory_allocator.h"
+#include <new>
+
+static Clip* getCurrentClip() {
+  return currentSong->currentClip;
+}
```
bobtwinkles, line 38: If it's not too much trouble, having this refactor pulled out to a separate kit would be nice (then when tracing history we can differentiate between the feature work and the refactor).

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -1122,6 +1122,29 @@ void AudioClip::setPos(ModelStackWithTimelineCounter* modelStack, int32_t newPos
...
+	
+  int64_t newStartPos = int64_t(sampleHolder.startPos) + getSamplesFromTicks(amount);
+  if (newStartPos < 0) {
+    return false;
+  }
+  
+  uint64_t length = sampleHolder.endPos - sampleHolder.startPos;
+  
+  sampleHolder.startPos = newStartPos;
+  sampleHolder.endPos = newStartPos + length;
```
bobtwinkles, line 1139: I'm not 100% sure, but I think we need to also update `sampleholder.clustersForStart` or at least run `SampleHolder::claimClusterReasons` -- otherwise we could have an underrun when RAM is scarce.

sichtbeton:
Wow, very cool! Would this work for loops or leave a gap at the start?

topisani:
> Wow, very cool! Would this work for loops or leave a gap at the start?

Not entirely sure actually, it depends on whether loops are recorded to a single file, or multiple. It will work the exact same way as moving the start point using the waveform editor or the reverse->length->reverse flow

src/deluge/gui/views/clip_view.cpp
topisani: Just a separate commit, same PR?

sichtbeton:
> > Wow, very cool! Would this work for loops or leave a gap at the start?
> 
> Not entirely sure actually, it depends on whether loops are recorded to a single file, or multiple. It will work the exact same way as moving the start point using the waveform editor or the reverse->length->reverse flow

Nice!

src/deluge/model/clip/audio_clip.cpp
topisani: Ill call claimClusterReasons - i also cannot figure out exactly whether or not this is necessary, 
The sample editor calls `claimClusterReasons(reversed, CLUSTER_LOAD_IMMEDIATELY_OR_ENQUEUE)` after changing start/end points, I assume this should do the same?

src/deluge/gui/views/clip_view.cpp
bobtwinkles: Sounds great, thanks!

src/deluge/model/clip/audio_clip.cpp
bobtwinkles: That sounds reasonable to me -- I haven't read the cluster management code in enough detail to be 100% confident but it's definitely the right shape.

sichtbeton:
alright couple of findings:

- it works! finally we can shift an audio clip
- if the audio ends but the clip itself is longer its silent (as expected)
- would be cool if it wraps around (but you said its complicated)
- if you shift all the way to the left (no audio anymore visible) it crashes
- couldnt tell if it respects the margin feature yet

very cool indeed! thx for your work on this. a lot of people asked for this.

topisani:
@sichtbeton helped me test this on a 7seg unit. The latest changes fixed some edge cases. I believe this now works as intended.
- Fixed shifting past the end of the sample
- unassigning and reassigning the voice fixed some issues that happened when shifting during playback.
- Fix shift direction so it is consistent with instrument clips
- Fix shifting of automation so it is consistent with audio.

Currently, the automation will be wrapped around like in instrument. I think this behavior is acceptable, as the use case probably is not that important, and it is still consistent, usable behavior. Also, This feature will probably often be used to either trim dead space off the start, in which case the automation will be removed from the end afterwards, or to shift the whole clip to the next loop, in which case keeping the automation in sync might make sense. I am not completely sure how automation is currently handled when an audio clip is reversed and trimmed anyway, but i think this makes at least as much sense as that. Do tell me if you disagree, i can look at implementing another solution, but id prefer not to let this minor UI improvement propagate all the way through the paramManager etc.

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -1123,26 +1123,40 @@ void AudioClip::setPos(ModelStackWithTimelineCounter* modelStack, int32_t newPos
...
 bool AudioClip::shiftHorizontally(ModelStackWithTimelineCounter* modelStack, int amount) {
+	// No horizontal shift when recording
+	if (recorder) return false;
+	// No horizontal shift when no sample is loaded
+	if (!sampleHolder.audioFile) return false;
+
+	int64_t newStartPos = int64_t(sampleHolder.startPos) - getSamplesFromTicks(amount);
+	uint64_t sampleLength = ((Sample*)sampleHolder.audioFile)->lengthInSamples;
+
+	if (newStartPos < 0 || newStartPos > sampleLength) {
```
bobtwinkles, line 1134: `startPos == sampleLength` is OK?

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -1127,16 +1127,18 @@ bool AudioClip::shiftHorizontally(ModelStackWithTimelineCounter* modelStack, int
...
-  int64_t newStartPos = int64_t(sampleHolder.startPos) + getSamplesFromTicks(amount);
-  if (newStartPos < 0) {
-    return false;
-  }
-  
-  uint64_t length = sampleHolder.endPos - sampleHolder.startPos;
-  
-  sampleHolder.startPos = newStartPos;
-  sampleHolder.endPos = newStartPos + length;
+  sampleHolder.claimClusterReasons(sampleControls.reversed, CLUSTER_LOAD_IMMEDIATELY_OR_ENQUEUE);
```
bobtwinkles: Style check is going to get upset at the missing space here.

bobtwinkles (commented):
I think style check is going to get mad (see below) but this mostly LGTM. Thanks!

src/deluge/model/clip/audio_clip.cpp
topisani: I had the same thought, but the sample marker editor allows it, so i assume it is ok
https://github.com/SynthstromAudible/DelugeFirmware/blob/9701434a2fd5c229f0a88ab8db43a9ed200092aa/src/deluge/gui/ui/sample_marker_editor.cpp#L220-L221

CommunityFeatures.md
```diff
@@ -45,6 +45,9 @@ Synchronization modes accessible through the "LFO SYNC" shortcut.
 
 ### Kit Keyboard View
  - ([#112]) All-new use for the "keyboard" button in kit clips, uses the main pad grid for MPC-style 16 level playing. Horizonatal encoder scrolls by one pad at a time, allowing positioning drums left to right, and vertical encoder jumps vertically by rows.
+ 
+### Audio Clip View
+ - ([#141]) Holding the vertical encoder down while turning the horizontal encoder will shift the clip along the underlying audio file, similar to the same interface for instrument clips.
```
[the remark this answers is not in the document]
topisani: Oh right, i forgot this obviously doesnt have the github magic links

topisani:
Addressed review comments, I have played around with this a bit more, and see no issues

bobtwinkles (approved):

bobtwinkles:
this was probably broken by the 2 big merges recently actually, but other than maybe fixing some style troubles this LGTM

PR #145, opened 2023-07-03
MenuItem rework (splitting)

stellar-aria:
This is the follow up to PR #117, removing the `new (addr) Type` initializations from the SoundEditor constructor and placing the global declarations in their own files (as well as the necessary reordering to do so).

This PR also removes all in-source class definitions for the various menu item types, and places them in their own directory structure following the one-class-per-file guideline. This understandably makes SoundEditor much short and easier to understand now. All menu items now live in the menu_item namespace, with sub namespaces as necessary.

### Next Steps once this is merged:
Refactor the use of "soundEditor.currentValue" into appropriate member variables in the MenuItem inheritance hierarchy (for ex. menu_item::Value or menu_item::Integer.) This will also ease the transition to enum classes for selection options

Organize the menus themselves better in an intuitive way, as they're all currently in the menus.cpp file.

Gradually remove the HAVE_OLED macro once DisplayText (from the Display refactor #121) is in the trunk


bobtwinkles:
Testing this on my OLED revealed at least some breakage. The following sound parameters resulted in freezes on selection in the subtractive synth:
  - OSC level
  - Pulse Width
  - Patch cable strength (use SHIFT combo to select Vibrato, then press SELECT). I don't actually know what this is supposed to do, but it's definitely not hang the systems :smile: 
  - Mod FX type allows scrolling past the end of the list. 
  - Envelope parameters (ADSR on both ENV1 and ENV2)
  - SVF is missing text in LPF mode menu

The main settings menu seems to still work, though I didn't test that it functions correctly. I don't see anything in this change that would indicate it's broken though

I did not test the following menus which we probably should try at least a little before landing this:
 - Any menus specific to synth modes other than subtractive (FM, ringmod)
 - No extensive testing of anything from the kit view, though some spot testing indicated it was fine.
 
Not caused by your change, but I also noticed that "OSC SYNC" and "RETRIGGER PHASE" use different UIs for seemingly no reason on the OLED. In general there's a lot of unification we can/should do on the UI side both internally and user-facing...

stellar-aria:
These were, in order:
Accidental infinite loop inherited from source::PatchedParam
Ditto
Ditto
Regression from beginning this work before alter-alter's stereo chorus was merged in (so the list was exactly as long as it should have been 😅 )
Infinite loop
Regression from beginning before SVF got merged

stellar-aria:
As the seemingly reigning template queen I'm also going to make it so that template declarations are on their own line lol

bfredl:
bug report (for the push 7 minutes ago):

- shift+synth to create new synth
- press select
- press select again to enter osc1
- scroll down one step and press select again to choose "level"
- press select to enter "modulate with menu"
- now selecting any mod source doesn't work (it just jumps back to the "modulate with" menu with env1 select).

bfredl:
Did a roundthrough of _most_ menus/sub-menus of the sound editor and global settings menus and couldn't find any more bugs :+1:  (for the debug-OLED version)

stellar-aria:
I'll do one final check-through on my end and then open this for final review

bobtwinkles:
LGTM, if there's any remaining bugs we can squash those in followups. This is disruptive enough I'm inclined to get it in early (though we should merge #143 and then this one, probably? I'm almost tempted to say this and 143 should get squashed together so we can break all the PRs exactly once, atomically).

PR #147, opened 2023-07-04
global CC support

m-m-adams:
This enables learning CCs to the global midi commands (play, stop, loop, overdub, etc.).  This improves compatibility with footswitches that can't send midi notes. 

As the existing learnt midi structures don't differentiate between notes and CCs, it works by adding 18 to the channel number if the incoming learnt midi is a CC. This puts CCs from channels 0-MPE_UPPER on 18-36 respectively. This is slightly hacky but avoids rewriting all midi learn related structures and functions to specify whether they want to check notes or CCs.


src/deluge/io/midi/learned_midi.h
```diff
@@ -81,7 +81,8 @@ class LearnedMIDI {
 
 	MIDIDevice* device;
 	uint8_t
-	    channelOrZone; // In addition to being set to channel 0 to 15, can also be MIDI_CHANNEL_MPE_LOWER_ZONE or MIDI_CHANNEL_MPE_UPPER_ZONE
+	    channelOrZone; // In addition to being set to channel 0 to 15, can also be MIDI_CHANNEL_MPE_LOWER_ZONE or MIDI_CHANNEL_MPE_UPPER_ZONE. C
+	                   // Channels 18-36 signify CCs on channels 1-16+MPE respectively. Check with the constant IS_A_CC
```
bobtwinkles, line 85: So this now represents one of 3 things?
  - Channel (0 - 15)
  - MPE lower zone (16)
  - MPE upper zone (17)
  - CC channel (18-34)
  - MPE lower zone CC (35)
  - MPE upper zone CC (36)

If you're up for it, it's possible to create this arrangement in an enum so we get a little more type safety, but it will still be janky. Something like this:

```
class MIDITargetType {
    enum TargetType: uint32_t {
        CHANNEL_0 = 0,
        // Must keep the next 15 clear, 
        MPE_LOWER = 16,
        MPE_UPPER = 17,
        CC_CHANNEL_0 = 18,
        CC_MPE_LOWER = 34,
        CC_MPE_UPPER = 35,
    } _value;

    MIDITargetType(uint8_t channelOrMpeZone, bool isCC) :
      _value(static_cast<TargetType>(isCC ? channelOrMpeZone + CC_CHANNEL_0 : channelOrMpeZone)) {

      }

    uint8_t getChannelOrMPEZone() {
        if (isCC()) {
            return static_cast<uint8_t>(this->_value - CC_CHANNEL_0);
        } else {
            return static_cast<uint8_t>(this->_value);
        }
    }

    bool isCC() {
        return static_cast<uint8_t>(this->_value) > 18;
    }
};
```

As long as we're touching this line, it would be nice to

m-m-adams: Thanks for the suggestion! I hated myself as I wrote the code but couldn't think of a safer way. I'll give this a shot and see how it goes

bobtwinkles: Some self-review feedback on that class, `TargetType` should probably be inherited from `uint8_t` but yeah hopefully it's a little smoother.

jamiefaye:
Awaiting resolution of conflicts here.

src/deluge/io/midi/learned_midi.h
m-m-adams: This causes other breaks that I've been unable to chase down - there are more magic numbers for midi channel in use to represent gold knob mappings and likely other things too. The current behaviour doesn't interfere but a refactor is non trivial due to the number of different interfaces with the midi learn structure and the send/receive functionality

m-m-adams:
Fixed now

PR #149, opened 2023-07-04
Add VSCode tasks for building and debug server launch

stellar-aria:
This adds the relevant building tasks for 7SEG/OLED Debug/Release builds, as well as launch tasks for running the debug servers in VSCode in such a way that the previously added debug profiles (#140) can connect.


m-m-adams:
Confirmed these all work!

PR #152, opened 2023-07-05
Partial fix for delugeprobe.

litui:
- SWD config fix working as well as jlink (more or less)
- JTAG config remains to be worked on and tested.


litui:
This is no longer necessary. Rolled these commits into the DBT update #88 .

PR #153, opened 2023-07-05
Refactor UART message to button/pad logic

dctucker:
Simplifies and abstracts the calculations for determining which pad or button has been pressed in the main loop.
Before:
```c
if (x == synthButtonX && y == synthButtonY)
```
After:
```c
if (b == SYNTH)
```


src/deluge/hid/buttons.h
```diff
@@ -22,12 +22,20 @@
 
 namespace Buttons {
 
+struct button_s {
```
bobtwinkles, line 25: I think this can be a class, with `charToButton` as the constructor and `buttonToChar` and `isButton` as (static) member functions.

src/deluge/hid/matrix/matrix_driver.h
```diff
@@ -24,6 +24,10 @@
 
 class AudioClip;
 
+struct pad_s {
+	int x, y;
+};
```
bobtwinkles, line 29: Similar to above, this could be a class with `charToPad` as the constructor and `padToChar`/`isPad` as (static) member functions.

src/deluge/deluge.cpp
```diff
@@ -324,42 +324,25 @@ bool readButtonsAndPads() {
...
-			if (y < displayHeight * 2) {
-				x <<= 1;
-				if (y >= displayHeight) {
-					y -= displayHeight;
-					x++;
-				}
+			int result;
+			if (MatrixDriver::isPad(value)) {
+				auto p = MatrixDriver::charToPad(value);
+				result = matrixDriver.padAction(p.x, p.y, thisPadPressIsOn);
```
bobtwinkles, line 337: if you're feeling brave, the structure can be plumbed all the way down the call stack.

we should double check the generated assembly to make sure we're not spilling to RAM in that case, though I don't know how much it would really matter.

src/deluge/hid/buttons.h
dctucker: I do want this to interface with C. If I understand correctly, as long as the class doesn't need a vtable the only difference should be default visibility.

jamiefaye:
I am going to approve this on the fast track because it touches in many places.

topisani:
Any reason why this was not done using a `enum struct`? that would be a significantly more type-safe way to do this

dctucker:
@topisani I think there was a permutation of this code that used an `enum class` but it never made it into a commit since I wasn't sure how to access the underlying type. There's probably room for a follow-up PR that changes this and adds `static_cast` where necessary.

PR #157, opened 2023-07-05
sound editor: "mod matrix" overview over all active patch cables

bfredl:
This provides a quick overview over all modulation which is active as part of a sound, similar to a "mod matrix" screen present in many synths.

TODO:
- [x] fine tune the UI for OLED. When a cable only have one source, we could fill in the current strength value in the "src2" slot
- [x] fine tune the UI for 7SEG. We want to use the blinkenlights to show source(s) and destination as the information doesn't fit on the screen.
- [ ] shift-delete for quick deletion?

### DIFF
--- a/docs/community_features.md
+++ b/docs/community_features.md
@@ -98,6 +98,9 @@ Synchronization modes accessible through the "LFO SYNC" shortcut.
 #### Sample Waveform View
  - ([#293]) When a sample has loop start and loop end points set, holding down loop start and tapping loop end will lock the loop points together. Moving one will move the other, keeping them the same distance apart. Use the same process to unlock the loop points. Use SHIFT+TURN<> to double or half the loop length.
 
+#### Sound Editor
+  - ([#157]) Add a "Mod Matrix" entry to the sound editor menu which shows a list of all currently active modulations.
+
 #### Takeover Mode
 
  - ([#170]) The Takeover menu consists of three modes that can be selected from:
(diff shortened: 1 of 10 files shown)


m-m-adams:
What about highlighting destinations when a source shortcut is clicked, the inverse of the current click destination to highlight sources behaviour?

bfredl:
Now it is usable across devices. the screen scroll on 7seg is maybe not very readable (would need to invent yet a table of short parameter names), but you can navigate using shortcut blinks which is shown now as part of the menu.

m-m-adams (changes requested):
Code looks good to me!

I'm not sure about the multi level modulations, the way they show up without a level looks like a bug compared to the other rows.

Maybe try like this?

src1[4] src2[4] intensity[4] dest[6],

with a space on both sides and between everything, and no scrolling. I think 6 characters is clear for all the modulatable params but I admittedly haven't looked that hard

bfredl:
> src1[4] src2[4] intensity[4] dest[6],

I tried it but 6 chars for the dest is just too cramped, I am gonna use max then and for two sources shift everything and accept that those ones might need to scroll.

Too much weirdness around strings right now. (if not crashes, menu items randomly taking the wrong value). making draft until I figure out something which works.

bfredl:
Did a bit more extensive testing, and for better and worse, the remaining crashes I see are the same as already on community firmware.

bfredl:
Seems stable now after rebasing on the change to use `std::string_view` for menu items.

m-m-adams (approved):

PR #162, opened 2023-07-06
Change row color in session_view/clipView

Queroland:
Change row color in clipView/sessionView by holding row & shift & scrollVertical (shortcut like in clip view).


Queroland:
It`s tested with OLED release configuration and works. 
Same shortcut as in clip view (press row & shift &vertical scroll).  People will love this feature - making it a lot easier to set up the color and keep track on what is what.

PR #165, opened 2023-07-06
Add names to CI workflows

trappar:
This just makes navigating the GitHub UI slightly easier. Funny enough, despite the actions requiring approval to run, GitHub immediately renames the actions in their UI just because I created this PR.

That definitely couldn't be abused at all! _/sarcasm_

**Where this is seen:**
image

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,5 @@
+name: Build
+
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


bobtwinkles (approved):
Thanks!

bobtwinkles:
huh, I thought maybe it renamed them just for you but no it renames them globally. That's... probably not good lol.

PR #169, opened 2023-07-07
ci: automatically push clang-format fixes

bobtwinkles:


bobtwinkles:
Example commit of this working: https://github.com/bobtwinkles/DelugeFirmware/commit/1866cf1263e0989975cb639da17d0f942a918b6b

I'm not 100% sure if it'll work on a commit someone else makes yet.

bobtwinkles:
Marking as draft for now because it doesn't work

trappar:
IMO CI should block when there are linting/formatting issues rather than fixing them. Maybe this would be better as a pre-commit hook?

bobtwinkles:
I did get a little caught up with about whether I could and didn't consider whether I should. The idea is to make it easier for people who don't want to deal with the formatter. 

This would likely be better served by a pre-commit hook, but those don't get installed by default (for good reason). I'll try to sketch one out to see how it goes.

jamiefaye:
What I would like (personally, for me) is to have the system report that my
formatting needs to be fixed and to then have a button I push that goes
ahead and fixes it.

– Jamie

m-m-adams:
I'd like to revisit this discussion - what about installing pre commit hooks via dbt and then having format be one of them?

PaulFreund:

PR #170, opened 2023-07-07
Midi Takeover Mode

seangoodvibes:
**Midi Takeover Mode**

A new item has been added to the Midi menu: **Takeover**

The Takeover menu consists of three modes that can be selected from:

> **Jump:** This is the default mode for the Deluge. As soon as a Midi Knob/Fader position is changed, the Deluge's internal Knob position/Parameter value jumps to the position of the Midi Knob/Fader.
> 
> **Pickup:** The deluge will ignore changes to its internal Knob position/Parameter value until the Midi Knob/Fader's position is equal to the Deluge Knob position. After which the Midi Knob/Fader will move in sync with the Deluge.
> 
> **Scale:** The deluge will increase/decrease its internal Knob position/Parameter value relative to the change of the Midi Knob/Fader position and the amount of "runway" remaining on the Midi controller. Once the Midi controller reaches its maximum or minimum position, the Midi Knob/Fader will move in sync with the Deluge. The Deluge will value will always decrease/increase in the same direction as the Midi controller.

Summary of changes to Deluge's code:

**Expanded MidiKnob class to include previous position info:**
```
Deluge/src/deluge/modulation/knob.h

class MIDIKnob : public Knob {
public:
	MIDIKnob() {}
	bool isRelative() { return relative; }
	bool is14Bit() { return (midiInput.noteOrCC == 128); }
	bool topValueIs127() { return (midiInput.noteOrCC < 128 && !relative); }
	LearnedMIDI midiInput;
	bool relative;
	bool previousPositionSaved;
	int previousPosition;
};
```

**Added code for Pickup and Takeover Modes**
```
Deluge/src/deluge/model/model_controllable/mod_controllable_audio.cpp

bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber,
                                                          uint8_t value, ModelStackWithTimelineCounter* modelStack,
                                                          int noteRowIndex) {

***[...]***

else {
	if (midiEngine.midiTakeover == MIDI_TAKEOVER_JUMP) {//Midi Takeover Mode = Jump
		newKnobPos = 64;
		if (value < 127) {
			newKnobPos = (int)value - 64;
		}
		knob->previousPositionSaved = false;
	}
	else { //Midi Takeover Mode = Pickup or Value Scaling
		/*
		Step #1: Convert Midi Controller's CC Value to Deluge Knob Position Value

		- Midi CC Values for non endless encoders typically go from 0 to 127
		- Deluge Knob Position Value goes from -64 to 64

		To convert Midi CC Value to Deluge Knob Position Value, subtract 64 from the Midi CC Value

		So a Midi CC Value of 0 is equal to a Deluge Knob Position Value of -64 (0 less 64).

		Similarly a Midi CC Value of 127 is equal to a Deluge Knob Position Value of +63 (127 less 64)

		*/

		int midiKnobPos = value - 64;

		//Save previous knob position for first time
		//The first time a midi knob is turned in a session, no previous midi knob position information exists, so to start, it will be equal to the current midiKnobPos
		//This code is also executed when takeover mode is changed to Jump and back to Pickup/Scale because in Jump mode no previousPosition information gets saved

		if (!knob->previousPositionSaved) {
			knob->previousPosition = midiKnobPos;

			knob->previousPositionSaved = true;
		}

		//adjust previous knob position saved

		//Here we check to see if the midi knob position previously saved is greater or less than the current midi knob position +/- 1
		//If it's by more than 1, the previous position is adjusted.
		//This could happen for example if you changed banks and the previous position is no longer valid.
		//By resetting the previous position we ensure that the there isn't unwanted jumpyness in the calculation of the midi knob position change amount
		if (knob->previousPosition > (midiKnobPos + 1) || knob->previousPosition < (midiKnobPos - 1)) {

			knob->previousPosition = midiKnobPos;
		}

		//Here we obtain the current Parameter Value on the Deluge
		int32_t previousValue =
			modelStackWithParam->autoParam->getValuePossiblyAtPos(modPos, modelStackWithParam);

		//Here we convert the current Parameter Value on the Deluge to a Knob Position Value
		int knobPos =
			modelStackWithParam->paramCollection->paramValueToKnobPos(previousValue, modelStackWithParam);

		//Here is where we check if the Knob/Fader on the Midi Controller is out of sync with the Deluge Knob Position

		//First we check if the Midi Knob/Fader is sending a Value that is less the current Deluge Knob Position
		//If less, check by how much its less. If the difference is greater than 1, ignore the CC value change (or scale it if value scaling is on)
		if (midiKnobPos == (knobPos - 1)) {
			newKnobPos = knobPos - 1;
		}

		//Next we check if the Midi Knob/Fader is sending a Value that is greater than the current Deluge Knob Position
		//If greater, check by how much its greater. If the difference is greater than 1, ignore the CC value change (or scale it if value scaling is on)
		else if (midiKnobPos == (knobPos + 1)) {
			newKnobPos = knobPos + 1;
		}

		else {

			//if the first two conditions fail and pickup mode is enabled, then the Deluge Knob Position (and therefore the Parameter Value with it) remains unchanged
			if (midiEngine.midiTakeover == MIDI_TAKEOVER_PICKUP) {//(runtimeFeatureSettings.get(RuntimeFeatureSettingType::MidiTakeoverMode) == RuntimeFeatureStateToggle::PU) { //Midi Pickup Mode On
				newKnobPos = knobPos;
			}
			//if the first two conditions fail and value scaling mode is enabled, then the Deluge Knob Position is scaled upwards or downwards based on relative
			//positions of Midi Controller Knob and Deluge Knob to min/max of knob range.
			else { //Midi Value Scaling Mode On
				//Set the max and min of the deluge midi knob position range
				int knobMaxPos = 64;
				int knobMinPos = -64;

				//calculate amount of deluge "knob runway" is remaining from current knob position to max and min of knob position range
				int delugeKnobMaxPosDelta = knobMaxPos - knobPos; //calculate how far delugeKnobPos is from Max - Positive Runway
				int delugeKnobMinPosDelta = knobPos - knobMinPos; //calculate how far delugeKnobPos is from Min - Negative Runway

				//calculate amount of midi "knob runway" is remaining from current knob position to max and min of knob position range
				int midiKnobMaxPosDelta = knobMaxPos - midiKnobPos; //calculate how far midiKnobPos is from Max - Positive Runway
				int midiKnobMinPosDelta = midiKnobPos - knobMinPos; //calculate how far midiKnobPos is from Min - Negative Runway

				//calculate by how much the current midiKnobPos has changed from the previous midiKnobPos recorded
				int midiKnobPosChange = midiKnobPos - knob->previousPosition;

				//Set fixed point variable which will be used calculate the percentage in midi knob position
				int midiKnobPosChangePercentage;

				//if midi knob position change is greater than 0, then the midi knob position has increased (e.g. turned knob right)
				if (midiKnobPosChange > 0) {
					//fixed point math calculation of new deluge knob position when midi knob position has increased

					midiKnobPosChangePercentage = (midiKnobPosChange << 20) / midiKnobMaxPosDelta;

					newKnobPos = knobPos + ((delugeKnobMaxPosDelta * midiKnobPosChangePercentage) >> 20);
				}
				//if midi knob position change is less than 0, then the midi knob position has decreased (e.g. turned knob left)
				else if (midiKnobPosChange < 0) {
					//fixed point math calculation of new deluge knob position when midi knob position has decreased
					
					midiKnobPosChangePercentage = (midiKnobPosChange << 20) / midiKnobMinPosDelta;

					newKnobPos = knobPos + ((delugeKnobMinPosDelta * midiKnobPosChangePercentage) >> 20);
				}
				//if midi knob position change is 0, then the midi knob position has not changed and thus no change in deluge knob position / parameter value is required
				else {
					newKnobPos = knobPos;
				}

			}
		}

		//save the current midi knob position as the previous midi knob position so that it can be used next time the takeover code is executed
		knob->previousPosition = midiKnobPos;
	}
}

***[...]***

}
```

**Defined Takeover menu values and number of modes** 
```
Deluge/src/definitions.h

#define MIDI_TAKEOVER_JUMP 0
#define MIDI_TAKEOVER_PICKUP 1
#define MIDI_TAKEOVER_SCALE 2
#define NUM_MIDI_TAKEOVER_MODES 3
```

**Added Takeover Modes to Midi Engine (to store Takeover Mode selection from Midi menu)**
```
Deluge/src/deluge/io/midi/midi_engine.h

class MidiEngine {
public:
uint8_t midiTakeover;
}

Deluge/src/deluge/io/midi/midi_engine.cpp

MidiEngine::MidiEngine() {
midiTakeover = 0;
}
```

**Defined Takeover Menu**
```
Deluge/src/deluge/gui/menu_item/midi/takeover.h

#pragma once
#include "gui/menu_item/selection.h"
#include "gui/ui/sound_editor.h"

namespace menu_item::midi {
class Takeover final : public Selection {
public:
	using Selection::Selection;
	void readCurrentValue() { soundEditor.currentValue = midiEngine.midiTakeover; }
	void writeCurrentValue() { midiEngine.midiTakeover = soundEditor.currentValue; }
	char const** getOptions() {
		static char const* options[] = {"Jump", "Pickup", "Scale", NULL};
		return options;
	}
	int getNumOptions() { return NUM_MIDI_TAKEOVER_MODES; }
};
} // namespace menu_item::midi
```

**Added Midi Takeover submenu**
```
Deluge/src/deluge/gui/ui/menus.cpp

#include "gui/menu_item/midi/takeover.h"

// MIDI Takeover
midi::Takeover midiTakeoverMenu{HAVE_OLED ? "TAKEOVER" : "TOVR"};

//MIDI menu
MenuItem* midiMenuItems[] = {&midiClockMenu, &midiThruMenu, &midiTakeoverMenu, &midiCommandsMenu, &midiInputDifferentiationMenu, &midi::devicesMenu, NULL};
```

**Added read/write to save Midi Takeover Settings to XML file**
```
Deluge/src/deluge/storage/flash_storage.cpp

void resetSettings() {
midiEngine.midiTakeover = 0;
}

void readSettings() {
midiEngine.midiTakeover = buffer[113];
}

void writeSettings() {
buffer[113] = midiEngine.midiTakeover;
}
```


weavermedia:
Could value scaling be added to this, similar to Ableton Live and many other MIDI controllable devices?

From the Ableton Live docs:

> Value Scaling — This option ensures smooth value transitions. It compares the physical control’s value to the destination parameter’s value and calculates a smooth convergence of the two as the control is moved. As soon as they are equal, the destination value tracks the control’s value 1:1.

Or maybe as a future enhancement?

seangoodvibes:
> Could value scaling be added to this, similar to Ableton Live and many other MIDI controllable devices?
> 
> From the Ableton Live docs:
> 
> > Value Scaling — This option ensures smooth value transitions. It compares the physical control’s value to the destination parameter’s value and calculates a smooth convergence of the two as the control is moved. As soon as they are equal, the destination value tracks the control’s value 1:1.
> 
> Or maybe as a future enhancement?

Hey absolutely that is possible.

I would make it selectable. So you would have Midi Pickup Mode with or without scaling

I would just need to understand the intended behaviour.

If we take an example, could you please describe how you would expect the Deluge to react?

Let’s say the Deluge Parameter Value is 50 and the Midi Parameter Value is 0.

So Deluge’s Knob Position is all the way right and the MIDI Knob Position is all the way left.

When I turn the MIDI Knob right, how should the Deluge Parameter Value change?

Thanks!

weavermedia:
While I'm sure there are 'proper' algorithms for this value scaling behavior this would be my two cents:

----

Example 1 (knob turns *towards* the current Deluge value)

Yes, exactly what you described. As the knob turns right the same value difference is deducted from the Deluge value, bringing the two values towards each other.

Knob moves 0 to 2 👉 Deluge value moves 50 to 48
Knob moves 2 to 10 👉 Deluge value moves 48 to 40
Knob moves 10 to 25 👉 Deluge value moes 40 to 25
After this point the values track 1:1

----

Example 2 (knob turns *away from* current Deluge value)

Here I think we use a multiple that takes into account how far is left to the end of the scale for each value.

Deluge value = 25
Knob value = 40
Deluge has 25 values left to end of scale (in the direction of change)
Knob has 10 values left to ned of scale (in the direction of change)
Ratio of change = 25:10 = 2.5:1
For every 1 value change by knob the Deluge value changes by 2.5x

Knob moves 40 to 42 👉 Deluge moves 25 to 30 (knob change 2 = Deluge change 2 x 2.5 = 5)
Knob moves 42 to 46 👉 Deluge moves 30 to 40 (knob change 4 = Deluge change 4 x 2.5 = 10) 
Knob moves 46 to 50 👉 Deluge moves 40 to 50 (knob change 4 = Deluge change 4 x 2.5 = 10) 

Of course this wouldn't be implemented in steps of 0-50 because the Deluge doesn't work at that resolution internally (I don't think?) so the full scale of vales would be used.

----

Although both of these examples will produce potentially unexpected behavior - example 1 would initially turn *down* a filter cutoff as you turn the knob *up* - in a live or performance situation I would still prefer this than an abrupt jump in value.

This boils down to 2 points on a line. When point A moves how should point B move in order to smoothly meet point A. The only two states are point A is moving *towards* point B or it's moving *away from* point B.

We could always do some tests in Ableton Live and adopt their approach. This will likely not only be a dependable and considered function but also one already familiar to the huge amount of Live users.

seangoodvibes:
Perfect thanks for the explanations Dan! @weavermedia 

this should be relatively simple for me to implement. I’ll be back home Sunday so I’ll start tackling it then :)

Cheers

seangoodvibes:
Changing this PR to draft as I'm working to incorporate the changes Dan requested above.

m-m-adams:
I don't think the scaling describes above is correct - increasing the a MIDI value should always increase the deluge value. Dans suggested approach would cause the deluge value to drop while the knob is increased and that's almost certainly worse than the current behaviour.



What you want to do is scale the ranges


E.g. 

Deluge value at 25
MIDI value at 10
Turn knob to 23 (+13)

Behaviour should be that the 10% (13/127) change in position brings the value 10% closer, leaving it at 27.5 (25+.1*(50-25))

With this behaviour the pot direction and parameter direction always match, and their values match at either end. Once their values match they will always stay in sync

seangoodvibes:
> 

Thanks for your input! I’m going to double check how ableton does it one more time and I will post the results here. I’ll test a bunch of different scenarios.

seangoodvibes:
**Ableton Value Scaling Scenarios:**

**Scenario #1:** _Increase Midi Value to 127 when Ableton Value is 0_

Current Values:

Ableton Value = 0
Midi Value = 126

New Values:

Ableton Value = 127 _(+127)_
Midi Value = 127 _(+1)_

**Scenario #2:** _Decrease Midi Value to 64 when Ableton Value is 0_

Current Values:

Ableton Value = 0
Midi Value = 126

New Values:

Ableton Value = 0 _(no change)_
Midi Value = 64 _(-62)_

**Scenario #3:** _Increase Midi Value to 70 when Ableton Value is 0_

Current Values:

Ableton Value = 0
Midi Value = 64

New Values:

Ableton Value = 12 _(+12)_
Midi Value = 70 _(+6)_

**Scenario #4:** _Decrease Midi Value to 64 when Ableton Value is 12_

Current Values:

Ableton Value = 12
Midi Value = 70

New Values:

Ableton Value = 11 _(-1)_
Midi Value = 64 _(-6)_

**Scenario #5:** _Decrease Midi Value to 10 when Ableton Value is 11_

Current Values:

Ableton Value = 11
Midi Value = 64

New Values:

Ableton Value = 1 _(-10)_
Midi Value = 10 _(-54)_

**Scenario #6:** _Increase Midi Value to 64 when Ableton Value is 1_

Current Values:

Ableton Value = 1
Midi Value = 10

New Values:

Ableton Value = 59 _(+58)_
Midi Value = 64 _(+54)_

seangoodvibes:
If the Ableton implementation is not the desired "method" let me know. Maybe we can improve upon it? Or we implement Ableton's method and then add a second "value scaling" method to offer users more choice?

weavermedia:
Agreed, the value scaling method used in Ableton Live is better than my initial thoughts. The only issue I have with it is if the Deluge value is close to one end of the scale and the MIDI controller is close to the other end, the movement of the MIDI control might seem to have no effect on the value until the values match. That would bug me but it's the behavior known to all the Live (and likely other DAW) user out there.

m-m-adams:
Awesome work @seangoodvibes! I think the ableton behaviour looks good, that description seems to match the hydrasynth as well which is the only synth I've used midi scaling with

seangoodvibes:
Hi :) I've updated the description for this commit above. It's ready to be tested and merged I think! @weavermedia @m-m-adams @litui @trappar 

Attached compiled firmware binaries here :)
[Takeover Mode - Compiled.zip](https://github.com/SynthstromAudible/DelugeFirmware/files/12033345/Takeover.Mode.-.Compiled.zip)

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1412,15 +1414,156 @@ bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice
...
+					if (midiEngine.midiTakeover == MIDI_TAKEOVER_JUMP) {//Midi Takeover Mode = Jump
+						newKnobPos = 64;
+						if (value < 127) {
+							newKnobPos = (int)value - 64;
+						}
+						knob->previousPositionSaved = false;
+					}
+					else { //Midi Takeover Mode = Pickup or Value Scaling
+						/*
+						MIDI TAKEOVER MODES:
```
m-m-adams, line 1426: This high level description probably belongs in community features.md rather than the code

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1412,15 +1414,156 @@ bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice
...
+
+									newKnobPos = knobPos + ((delugeKnobMaxPosDelta * midiKnobPosChangePercentage) / 1000000);
+								}
+								//if midi knob position change is less than 0, then the midi knob position has decreased (e.g. turned knob left)
+								else if (midiKnobPosChange < 0) {
+									//fixed point math calculation of new deluge knob position when midi knob position has decreased
+
+									midiKnobPosChangePercentage = (midiKnobPosChange * 1000000) / midiKnobMinPosDelta;
+
+									newKnobPos = knobPos + ((delugeKnobMinPosDelta * midiKnobPosChangePercentage) / 1000000);
```
m-m-adams, line 1547: Style: use <<20 and >>20 instead of the multiplication/division by 1000000, it makes it clearer how many bits are being used for the integer and fractional parts


The performance change would be negligible here but it's better practice

m-m-adams (approved):
Code looks good and I appreciate the clear commenting, left a style suggestion and I think the big description should be moved to community feature 


Note - untested in practice since I have no MIDI knobs this week

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
seangoodvibes: I have made the recommended change and moved the Takeover Mode description to community features.md. See most recent commit.

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
seangoodvibes: I have made the recommended change and replaced the multiplication/division with << 20 and >> 20 respectively. See most recent commit.

jamiefaye:
Please fix format issues and I will merge this one in.

seangoodvibes:
> Please fix format issues and I will merge this one in.

I fixed it last night :)

jamiefaye:
The fix may have not made it all the way to
https://github.com/SynthstromAudible/DelugeFirmware/pull/170. Or at least
that is what the "pre merge check" thinks.

jamiefaye:
You may need to deconflict this one too as github thinks there are
conflicts too.

jamiefaye:
[reformatted.txt](https://github.com/SynthstromAudible/DelugeFirmware/files/12045669/reformatted.txt)
Tests still failing. See above.
https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5549625586/jobs/10133928478?pr=170

seangoodvibes:
> [reformatted.txt](https://github.com/SynthstromAudible/DelugeFirmware/files/12045669/reformatted.txt) Tests still failing. See above. https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5549625586/jobs/10133928478?pr=170
> 
> Apologies for all the trouble here. Its possible some of our test scripting is wrong.

No worries, let's try again!

seangoodvibes:
@jamiefaye I think we're good to merge now!

PR #172, opened 2023-07-07
Editor setting improvements

PaulFreund:
- [x] Make sure Git always checks in and out with LF line endings
- [x] Changed editorconfig and VSCode settings to always generate final newline
- [x] Improve tasks and debugger config in VSCode


PaulFreund:
Would be really nice if someone could test Load/Run on Linux. I had to make a hack for windows and tried to not duplicate the customLaunchSetupCommands I graciously got from litui with minimally different contents.

PR #174, opened 2023-07-08
midi: allow bigger sysex messages to be sent, such as the OLED image

bfredl:
Some simplifications were made to the driver code by Rohan under the
assumption we were only going to queue up to 32 usb-midi messages
for a device per "tick" of the main routine.
These had to be reconsidered, like there is now again a callback from the usb driver
interrupt back to midi_engine to check for more data to send.

It doesn't seem to be possible to increase the amount of data sent in
each call to the driver more than 32 messages (or 128 bytes).
Instead, implement a ring buffer for midi data waiting to be sent over USB.
When we get an interrupt for usbSendComplete, we can check the ring
buffer again and restart the low-level send until the ring buffer is
empty.

Implement a sysex request for sending the contents of the OLED display
as a sysex reply message. This needs more work before being
generally useful as an external display (not to mention client code for multiple platforms)
, but serves as a test case for larger sysex messages. This also tangents #121 , an unified build
would eventually allow developers with a 7SEG display to test how their code
would look on a virtual OLED display.


bfredl:
I have tested these for
- [x] serial midi
- [x] deluge acting as a USB device
- [ ] deluge acting as USB host

I will try to find a way to test the last one before marking this as ready

jamiefaye:
Looking forward to this one going in.

bfredl:
Update: sending as a host to devices now works for most devices. a buffer limit of 64 bytes seems common, so lets use that for devices. some devices have even lower limits tho, like my MIDIMATE II serial MIDI to USB converter which only accepts sending one (1) midi packet at a time. (We could try to figure this out from the configuration data, I tried with the `bMAXPacketSize` field but it doesn't seem to be the right one).

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -530,17 +531,60 @@ void readAHostedDeviceFromFile() {
 
 void ConnectedUSBMIDIDevice::bufferMessage(uint32_t fullMessage) {
 	// If buffer already full, flush it
-	if (numMessagesQueued >= MIDI_SEND_BUFFER_LEN) {
-		midiEngine
-		    .flushUSBMIDIOutput(); // TODO: this is actually far from perfect - what if already sending - and if we want to wait/check for that, we should be calling the routine.
-		                           // And ideally, we'd be able to flush for just one device.
-		numMessagesQueued = 0;
+	// if (numMessagesQueued >= MIDI_SEND_BUFFER_LEN) {
```
m-m-adams, line 534: Remove instead of commenting out

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -530,17 +531,60 @@ void readAHostedDeviceFromFile() {
...
+	// 	midiEngine
+	// 	    .flushUSBMIDIOutput(); // TODO: this is actually far from perfect - what if already sending - and if we want to wait/check for that, we should be calling the routine.
+	// 	                           // And ideally, we'd be able to flush for just one device.
+	// 	numMessagesQueued = 0;
+	// }
+	//
+	int queued = ringBufWriteIdx - ringBufReadIdx;
+	// if (queued > some_smaller_number) {
+	//   if (notAlreadySending) {
+	//     startTheSend();
```
m-m-adams, line 544: Can you add a comment on where the sends is actually called?

src/deluge/io/midi/midi_device.h
```diff
@@ -155,11 +163,10 @@ class MIDIDeviceUSBHosted final : public MIDIDeviceUSB {
 
 class MIDIDeviceUSBUpstream final : public MIDIDeviceUSB {
 public:
-	MIDIDeviceUSBUpstream(uint8_t portNum = 0) { portNumber = portNum; }
+	MIDIDeviceUSBUpstream(uint8_t portNum = 0) : MIDIDeviceUSB(portNum) {}
```
m-m-adams, line 166: Is the portnum still necessary after giving it a default value in the mididevice USB class?

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -530,17 +531,60 @@ void readAHostedDeviceFromFile() {
...
 	anythingInUSBOutputBuffer = true;
 }
 
+bool ConnectedUSBMIDIDevice::hasRingBuffered() {
+	// must me the same unsigned type as ringBufWriteIdx/ringBufReadIdx
+	uint32_t queued = ringBufWriteIdx - ringBufReadIdx;
+	return queued > 0;
+}
+
+bool ConnectedUSBMIDIDevice::consumeBytes() {
```
m-m-adams, line 564: Can you add an explanatory comment on what this functions purpose is and when it should be called?

src/deluge/io/midi/midi_device_manager.h
```diff
@@ -30,10 +30,14 @@ struct MIDIDeviceUSB;
 #endif
 
 // size in 32-bit messages
-// TODO: increasing this even more doesn't work. For now this gives
-// maximum SysEx send size of 96 bytes including start/end bytes
-// (3 payload bytes per USB-MIDI message)
-#define MIDI_SEND_BUFFER_LEN 32
+// NOTE: increasing this even more doesn't work.
+// Looks like a hardware limitation (maybe we more in FS mode)?
```
m-m-adams, line 34: What happens?

src/deluge/io/midi/midi_engine.cpp
```diff
@@ -82,8 +82,24 @@ void usbSendComplete(int ip) {
 	bool inHostMode = (g_usb_usbmode == USB_HOST);
 
 	// If in host mode, see if we want to send to another device
+	// NB: this only gets called in host mode anyway, refactor!
 	if (inHostMode) {
 
+		// check if there was more to send on the same device
+		bool has_more = connectedDevice->consumeBytes();
+		if (has_more) {
```
m-m-adams, line 90: Should this modify flushUSBMidiToHostedDevice instead of being added inline?

src/deluge/io/midi/midi_engine.cpp
```diff
@@ -82,8 +82,24 @@ void usbSendComplete(int ip) {
...
+			// TODO: do some cooperative scheduling here. so if there is a flood of data
+			// on connected device 1 and we just want to send a few notes on device 2,
+			// make sure device 2 ges a fair shot now and then
+
+			g_usb_midi_send_utr[USB_CFG_USE_USBIP].tranlen = connectedDevice->numBytesSendingNow;
+			g_usb_midi_send_utr[USB_CFG_USE_USBIP].p_tranadr = connectedDevice->dataSendingNow;
+			int pipeNumber = g_usb_hmidi_tmp_ep_tbl[USB_CFG_USE_USBIP][midiDeviceNum][0];
+			usb_send_start_rohan(&g_usb_midi_send_utr[USB_CFG_USE_USBIP], pipeNumber, connectedDevice->dataSendingNow,
+			                     connectedDevice->numBytesSendingNow);
+			return;
```
m-m-adams, line 100: How does this interact with the previous code to flush all devices? Is that now unreachable?

src/deluge/io/midi/midi_device.cpp
```diff
@@ -349,6 +353,51 @@ void MIDIDeviceUSB::sendMessage(uint8_t statusType, uint8_t channel, uint8_t dat
...
+	if (len < 3 || data[0] != 0xf0 || data[len - 1] != 0xf7) {
+		return;
+	}
+
+	int ip = 0;
+	ConnectedUSBMIDIDevice* connectedDevice = NULL;
+
+	for (int d = 0; d < MAX_NUM_USB_MIDI_DEVICES; d++) {
+		if (connectionFlags & (1 << d)) {
+			connectedDevice = &connectedUSBMIDIDevices[ip][d];
```
m-m-adams, line 366: I don't believe this works properly, won't it just send to the first connected device? Since each ConnectedUsbMidiDevice holds an array of 4 USBMidiDevices it should check that the current device is actually held by the connecteddevice it finds

m-m-adams (changes requested):
I've added some comments where I think this could be improved a bit. 

The only big issue is that the current behaviour of send all messages to all devices doesn't work well for sysex since it's much more data and sysex parsing bugs are very common (especially with older gear). In peripheral mode that's fixable by enabling the third upstream port (already in the config), in host mode it would need some extra code to check that the device being sent to is the one requesting data

bfredl:
> The only big issue is that the current behaviour of send all messages to all devices doesn't work well for sysex since it's much more data and sysex parsing bugs are very common (especially with older gear). In peripheral mode that's fixable by enabling the third upstream port (already in the config), in host mode it would need some extra code to check that the device being sent to is the one requesting data

This is already what this branch implements. Sysex replies are being sent to the specific device that sent a request. not sure what makes it seem like otherwise. 

I agree the thrird port should be added at this point, but differentiation between virtual ports is already implemented (request the oled data on virtual midi out 2; get a reply on virtual midi in 2).

src/deluge/io/midi/midi_device_manager.h
bfredl: I've just been reading the the usb midi spec and we might be able to use the maximum buffer size specified by a downstream device instead of this guesswork.

src/deluge/io/midi/midi_engine.cpp
bfredl: it still works with the caveat mentioned in the comment above. I might restructure this function to make it clearer as it is now explicit that this function is only used for host mode.

src/deluge/io/midi/midi_device.cpp
bfredl: This follows the existing logic in sendMessage and it does seem to work. `connectionFlags` is private to the inner USBMIDIDevice and should only have a flag set for the specific connected device. It's a bit weird way of structuring things tho.

What I think we don't support at all is virtual port/cable differentiation for a specific connected device. That would be an important follow-up refactor.

src/deluge/io/midi/midi_engine.cpp
bfredl: yes, I will do so

bfredl:
Thanks for the review. I tried to make the naming of new stuff more consistent and added a bit more comments.

bfredl:
> My assumption was based on the original firmware behaviour on flush to send the buffered midi to all devices .

what happened is that all devices gets flushed at the same time even though one device's buffer was full. I'd like to untangle that _temporal_ mix-up at some point as well. But there was still a dedicated buffer for each physical device so it was always possible to send messages to a specific device (even though with note/cc data they often contained exactly the same contents).

bfredl:
Added the third port dedicated for sysex data (it never sends channel/MPE data).

src/deluge/hid/hid_sysex.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+void HIDSysex::sendOLEDData(MIDIDevice* device, bool rle) {
+	// TODO: in the long run, this should not depend on having a physical OLED screen
+#if HAVE_OLED
+	const int block_size = 768;
+	const int packed_block_size = 878;
+
+	rle = false; // not yet implemented
+
+	uint8_t reply_hdr[5] = {0xf0, 0x7e, 0x02, 0x40, rle ? 0x02 : 0x01};
```
dctucker, line 37: If I'm reading [this](https://www.midi.org/specifications-old/item/table-4-universal-system-exclusive-messages) correctly, the message being sent is "Universal Non-realtime sample data packet".

Should this be changed to use `0x7d` as the second byte, or is there a specific reason this message type was chosen?

bfredl: no you are right, it is supposed to be `0x7d`

PR #178, opened 2023-07-10
Fine tempo knob

suivaht:
Hello everyone, I'm trying this again...  This build allows the user to switch the tempo knob behavior.  New default behavior is that turning the tempo knob changed BPM in increments of 1.  Pushing and turning the knob increments like the old default (typically by 4).  This can be enabled/disabled in the runtime features menu (finally figured out how to do this after taking some good advice and checking out how @alter-alter added features to that menu... great community here).  Please let me know your thoughts...

### DIFF
--- a/src/deluge/model/settings/runtime_feature_settings.h
+++ b/src/deluge/model/settings/runtime_feature_settings.h
@@ -35,6 +35,7 @@ enum RuntimeFeatureSettingType : uint32_t {
 	DrumRandomizer,
 	MasterCompressorFx,
 	Quantize,
+	FineTempoKnob,
 	MaxElement // Keep as boundary
 };
 
@@ -105,6 +106,14 @@ protected:
 	         .options = {{.displayName = "Off", .value = RuntimeFeatureStateToggle::Off},
 	                     {.displayName = "On", .value = RuntimeFeatureStateToggle::On},
 	                     {.displayName = NULL, .value = 0}}},
+
+		[RuntimeFeatureSettingType::FineTempoKnob] =
+			{.displayName = "Fine Tempo Knob",
+			 .xmlName = "fineTempoknob",
+			 .value = RuntimeFeatureStateToggle::On, // Default value
+			 .options = {{.displayName = "Off", .value = RuntimeFeatureStateToggle::Off},
+						 {.displayName = "On", .value = RuntimeFeatureStateToggle::On},
+						 {.displayName = NULL, .value = 0}}},
 
 	};
 
--- a/src/deluge/playback/playback_handler.cpp
+++ b/src/deluge/playback/playback_handler.cpp
@@ -60,6 +60,7 @@
 #include "gui/ui/load/load_song_ui.h"
 #include "gui/ui_timer_manager.h"
 #include "hid/display/oled.h"
+#include "model/settings/runtime_feature_settings.h"
 
 extern "C" {
 #include "RZA1/gpio/gpio.h"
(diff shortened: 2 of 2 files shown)


litui:
Why is this change being put forward as the default? Is this something desired by most users? Is it likely to interfere with existing workflows?

Personally, I would err on the side of the default being the previous, stock, expected behaviour rather than the new approach.

src/deluge/playback/playback_handler.cpp
```diff
@@ -1883,46 +1884,90 @@ void PlaybackHandler::tempoEncoderAction(int8_t offset, bool encoderButtonPresse
...
-				// Fine tempo adjustment
-				uint32_t tempoBPM = calculateBPM(currentSong->getTimePerTimerTickFloat()) + 0.5;
-				tempoBPM += offset;
-				if (tempoBPM > 0) {
-					currentSong->setBPM(tempoBPM, true);
-					displayTempoBPM(tempoBPM);
+		else {
+
+			if (runtimeFeatureSettings.get(RuntimeFeatureSettingType::FineTempoKnob)
+		    		== RuntimeFeatureStateToggle::On) {
```
m-m-adams, line 1890: Store this in a variable to check later

src/deluge/playback/playback_handler.cpp
```diff
@@ -1883,46 +1884,90 @@ void PlaybackHandler::tempoEncoderAction(int8_t offset, bool encoderButtonPresse
...
+
+				if (!(playbackState & PLAYBACK_CLOCK_EXTERNAL_ACTIVE)) {
+
+					if (Buttons::isButtonPressed(hid::button::TEMPO_ENC)) {
```
m-m-adams, line 1894: Compare this to the feature toggle instead of nesting if/else

E.g. if tempo_button_pressed == tempo_press_coarse_enabled {}

src/deluge/model/settings/runtime_feature_settings.h
```diff
@@ -106,6 +107,14 @@ class RuntimeFeatureSettings {
 	                     {.displayName = "On", .value = RuntimeFeatureStateToggle::On},
 	                     {.displayName = NULL, .value = 0}}},
 
+		[RuntimeFeatureSettingType::FineTempoKnob] =
+			{.displayName = "Fine Tempo Knob",
+			 .xmlName = "fineTempoknob",
+			 .value = RuntimeFeatureStateToggle::On, // Default value
```
m-m-adams, line 113: Default to off as suggested

m-m-adams (changes requested):
The feature setting check would be cleaner if you compare it to the button press result instead of duplicating code

Also I agree with litui on the default value, I think this is better but until we have a good process for testing and reviewing UI changes it should be opt in

suivaht:
@litui I understand your point and I can change the default if it is an issue. This new behavior is the default on nearly every other machine and is my preferred default (also had a thread on the forums with people saying they'd prefer this behavior). But it's probably more consistent to keep the default so I will do that

src/deluge/playback/playback_handler.cpp
suivaht: Thanks for the tip... I'm learning as I go...

litui:
> @litui I understand your point and I can change the default if it is an issue. This new behavior is the default on nearly every other machine and is my preferred default (also had a thread on the forums with people saying they'd prefer this behavior). But it's probably more consistent to keep original default behavior so I will do that

I'm not necessarily saying /don't/ make it the default, I'm just suggesting that if you're going to do something bold like making it the default you need to show your work. Link the forum thread for a start, and let's have a friendly debate on the merits of making it the default ;).  Perhaps it's justified, but just putting in a PR with a changed functionality as the default doesn't reach the bar in my opinion.

suivaht:
I will tinker and clean this up but any other thoughts/feedback are welcome... Also if there is any github protocol I am ignoring please let me know (gently, haha!)... Thanks for the feedback

suivaht:
@litui Got it, no worries...  I was just approaching it the same way I was with my own personal builds but I now understand that I need to think about all the other users not expecting a drastic change when they turn on their machines.  This is all helpful feedback.

The forum post I referred to was not quite as robust as I remember but it did get a few responses...

https://forums.synthstrom.com/discussion/comment/22865#Comment_22865

I'll revisit this tomorrow after some sleep and research...

weavermedia:
Would it be possible to have settings for both tempo knobs states (turn and push+turn)? I ask because there's a third option I'd love to see of 0.1 BPM increments. If I had the choice I'd have turn change by 1 BPM and push+turn change by 0.1 BPM.

We could add settings like:
- TEMPo > TURN > LARG/SMAL/FINE (default LARG)
- TEMPo > PUSH > LARG/SMAL/FINE (default SMAL)

Where:
- LARG = 4 BPM (or whatever tempo-related amount it is)
- SMAL = 1 BPM
- FINE = 0.1 BPM

Or just call SMAL "1" and FINE "0.1" of course.

So we cold select which increment we want for both actions.

m-m-adams:
That's a larger change and I think out of scope for this PR - this swaps the current behaviour, while your suggestion would require adding an entire new mode

weavermedia:
@m-m-adams Agreed 👍 I'll look at it when I can.

suivaht:
Hello everyone, thank you for all the feedback...  I changed the default tempo behavior setting to the normal default (turn to increment in 4, push turn to increment by 1), tried to clean up the code a bit, but one piece of advice I wasn't able to get working was  to store the 'runtimeFeatureSetting' value in a variable...  The updated code works as expected though...  Let me know if there are any issues with this update...

m-m-adams:
Looks good! I just meant to check later, putting the check inline instead of a variable is totally valid

suivaht:
I'm looking at the workflow conflicts and the changes in the latest community build and it looks like the approach for adding the runtime feature settings has totally changed.  I can't really resolve these conflicts until I see how I need to implement this again...  Looking through changes now and will update when I believe it is ready to review again.

bobtwinkles:
Sorry for the churn, I've sent a PR to your branch at https://github.com/suivaht/DelugeFirmware/pull/1 which should take care of the change.

suivaht:
@bobtwinkles the pull request was very much appreciated and has just been merged.  Thank you

litui:
Gotcha covered, @suivaht. The checks have to be triggered by maintainers.

suivaht:
@litui thank you... This is a great community... I appreciate all of the help from everyone who commented

4321-AJ:
Hello all, very new to all this so apologies in advance, but I concur with weavermedia in respect to increments of 0.1.
I come from the DJ world and having just got hold of a deluge, the first thing I was interested to see was if the fine pitch/bpm control from a record deck could be emulated, allowing any tune made on the deluge to be perfectly bpm matched to other records in a dj set. I understand this requires a completely new feature/mode, but thought I would just confirm why this would be very useful. 
I am I  awe of this community and it seems the possibilities are endless for this hardware with all the amazing members, working to add all these features.

PR #180, opened 2023-07-10
Refactor indicator LEDs

dctucker:
Similar to #153, this PR reduces the number of parameters from two to one for affecting the state of indicator LEDs.


src/deluge/hid/led/indicator_leds.h
```diff
@@ -22,30 +22,91 @@
...
+#if DELUGE_MODEL >= DELUGE_MODEL_144_PAD
+	return uartBase + x + y * 9;
+#else
+	return uartBase + x + y * 10;
+#endif
+}
+
+typedef uint8_t IndicatorLED;
+
+enum KnownIndicatorLEDs : IndicatorLED {
```
bobtwinkles, line 44: To avoid the style errors here, you can use

```
// clang-format off
..... this code
// clang-format on
```

src/deluge/hid/led/indicator_leds.h
```diff
@@ -22,30 +22,91 @@
 
 #define numLedBlinkers 4
 
+namespace IndicatorLEDs {
```
bobtwinkles, line 25: as long as we're in the area, should probably use the `snake_case` namespace names (to match the menu item refactor)

bobtwinkles (changes requested):
LGTM functionally, will test on hardware tomorrow if nobody else gets to it.

Minor style comments otherwise.

src/deluge/hid/led/indicator_leds.h
bobtwinkles: and apparently the `deluge::` root namespace too

PR #183, opened 2023-07-11
hid::Button is now an enum class

dctucker:
> > [This comment](https://github.com/SynthstromAudible/DelugeFirmware/pull/153#issuecomment-1624226635) 

> @topisani I think there was a permutation of this code that used an `enum class` but it never made it into a commit since I wasn't sure how to access the underlying type. There's probably room for a follow-up PR that changes this and adds `static_cast` where necessary.

### DIFF
--- a/src/deluge/deluge.cpp
+++ b/src/deluge/deluge.cpp
@@ -308,12 +308,12 @@ bool readButtonsAndPads() {
 
 			Uart::println("");
 			Uart::println("undoing");
-			Buttons::buttonAction(hid::button::BACK, true, sdRoutineLock);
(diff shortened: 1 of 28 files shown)


topisani:
Nice! I think this makes it easier to avoid mistakes when calling functions that accept a button (and when refactoring). We could add helpers like the following to clean the implementation up even further, but I may open a PR with that myself instead of just continuing to comment on your work :angel: 

```cpp
uint8_t toCode(Button b) {
  return static_cast<uint8_t>(b);
}

Button fromCode(uint8_t c) {
  return static_cast<uint8_t>(c);
}
```

PaulFreund:

dctucker:
Closing because the codebase has evolved significantly.

PR #185, opened 2023-07-11
refactor: remove unmaintained code for the 40-pad deluge

bfredl:
While this code certainly has some nostalgic value, the code code currently does not compile, and is not directly useful to the community firmware project. The effort to bring this code back to a workable state would be large, and there is likely a lot of subtle bugs creeping in even if one were to get it into a compilable taste.

Not to mention, the number of people this effort would benefit and also would be able to help with the forward-going maintenance is rather small.

For the cases where consulting these differences is useful, e.g. for reverse engineering, one can look into an older state of the git repo. For most cases, like when I was trying to get an overview of the PIC UART protocol, having half of the bufferPICUart() calls being dead code is just a distraction.


jamiefaye:
I should have reverted this. And the gitiot here is me.

jamiefaye:
Can you resubmit this. I really botched my attempt at fixing the merge conflicts and had to revert this. I do like the idea of getting rid of 40 key stuff.

PR #192, opened 2023-07-13
sysex: add 7SEG over sysex and improve OLED

bfredl:
Run Length Encoding avoids wasting a lot of bandwidth on large empty regions.


ok-reza:
amazing work, super handy for referencing OLED gui for making mock ups.

i noticed that if i turn off my deluge while on the webpage, google chrome completely freezes up, is there something that can be done to avoid the crash?

PR #196, opened 2023-07-15
Fill clips

0beron:
Draft PR to keep track of what I'm up to with fill clips.

- Fill clips.
  - A fill clip has altered scheduling behaviour to make varied fills at the end of a section easier to trigger.
    - [x] Arming a fill clip during playback will schedule it to _finish_ at the end of the next loop.
    - [x] If the time interval to the end of the loop is less than the fill clip length, it is started immediately, and at the correct position to end at the loop boundary.
    - [x] A fill clip, once playing will arm itself to stop at the end of the clip (and hence the loop boundary).
    - [x] Disarming a fill clip when armed to start will cancel it.
    - [x] Disarming a fill clip when playing will stop immediately.
    - [x] Fill clips will align to repetitions of the main loop if done by turning 'select'.
    - [x] Fill clips must obey the one clip per output convention, otherwise this would require major arrange alterations.
    - ~~Fill clips that use the same kit/synth/midi channel as any other clip, can play simultaneously.~~
      - ~~Maybe allow this to be customised or add a third launch style to make fills mutually exclusive with other fills, allowing easy switching between alternatives during the fill for better 'playability'~~.
    - [x] Fill clip status pads are re-coloured, orange/cyan for stopped/playing.
      - ~~Maybe add this to the PADS colour menu to allow customisation.~~
- Section scheduling
  - [x] Scheduling a section will schedule the non-fill clips as normal, and thereafter be equivalent to scheduling each of the fill clips in that section individually.
    - [x] Need to figure out the meaning of a shared or repeated section.
    - Repeats may be disallowed for fills (not sure it makes sense?)
    - Shared is pretty straightforward and may 'just work', since it will only affect the non-fill clips.
- [x] Verify all this scheduling behaviour plays nicely with SOLO and recording to the arranger.
- [x] Verify all this works when clips are playing in reverse or ping-pong.
- [x] Mechanism to change the launch style of a clip.
  - Currently done by holding the launch/status pad of the relevant clip and pressing select to access a menu of launch types
  - This is extensible to other clip launch behaviours that may arise in future.
  - Open to other UX suggestions here. The current mechanism requires that you hold the status pad, which triggers immediately on press so has the 'side effect' of launching the clip. I anticipate that the clip types will mainly be changed outside of playback and then not changed.
- [x] Save and load per-clip launch style to/from the song file.

All of the above up for debate and improvement.


sichtbeton:
Does this work in tandem with the new Ableton grid view? If so, then this would be a game changer. 

Second question: does this record to arranger?

0beron:
Record to arranger works with the same limitation as in the default firmware, which is that a clip in the arranger cannot start midway through.
Not tried in conjunction with the grid view yet, but there is a high likelihood it will just work, if pressing a grid item is equivalent to to pressing the status pad. Only thing that might not work without some changes is changing the clip type from default -> fill or vice versa, you may have to do this from song mode. After that I would expect them to work in the grid.

sichtbeton:
> Record to arranger works with the same limitation as in the default firmware, which is that a clip in the arranger cannot start midway through.

Which has nothing to do with your fantastic PR but is something that holds the del back.

 Not tried in conjunction with the grid view yet, but there is a high likelihood it will just work, if pressing a grid item is equivalent to to pressing the status pad. Only thing that might not work without some changes is changing the clip type from default -> fill or vice versa, you may have to do this from song mode. After that I would expect them to work in the grid.

right. this might be implemented in the grid code maybe someday.

src/deluge/playback/mode/session.cpp
```diff
@@ -191,9 +191,36 @@ bool Session::giveClipOpportunityToBeginLinearRecording(Clip* clip, int32_t clip
 	return begun;
 }
 
-void Session::doLaunch() {
-
-	view.flashPlayDisable();
+/**
```
m-m-adams, line 194: nit: remove instead of comment

m-m-adams (approved):

soymonitus:
Ey @0beron I have cloned your branch, merging it with the latest from community branch and fixed a couple of minor conflicts, and I have tested how it works with the new Grid ableton view and it works as expected. the clips launch as they should. Regarding fill clips that are clones one of each other with different section colors, if you engage both, only one plays which I guess is expected behavior. So from my side everything looks quite good

docs/community_features.md
```diff
@@ -75,6 +75,12 @@ Here is a list of features that have been added to the firmware as a list, group
  - ([#221]) The normal behavior of the Deluge is to try to keep up with 'in progress' notes when instant switching between clips by playing them late. However this leads to glitches with drum clips and other percussive sounds. Changing this setting to OFF will prevent this behavior and *not* try to keep up with those notes, leading to smoother instant switching between clips.
 	- This feature can be turned ON/OFF in the Runtime Settings (Community Features) Menu (accessed by pressing "SHIFT" + "SELECT"). 
 
+#### 4.1.5 - Fill Clips
```
seangoodvibes, line 78: Just a small suggested change.

Can you move your addition up so that it becomes #### 4.1.3 - Fill Clips and then renumber the other two so they become #### 4.1.4 - New Grid Layout and #### 4.1.5 - Catch Notes

Thanks

seangoodvibes, line 78: Just a small suggested change.

Can you move your addition up so that it becomes #### 4.1.3 - Fill Clips and then renumber the other two so they become #### 4.1.4 - New Grid Layout and #### 4.1.5 - Catch Notes

seangoodvibes (changes requested):
Nvm

seangoodvibes (approved):
Approving changes to community features!

hexagon5un:
Right place for a bug report?  If so:

"Disarming a fill clip when playing will stop immediately." seems not to work.  Instead, if you re-arm it while the fill is playing, it loops the fill nonstop.  This can be triggered by the mute/launch button, by the selection button that the fill belongs to, or in grid mode.

Steps to reproduce: 

1) Make a 4-bar loop
2) Make a 1-bar fill
3) Arm fill -- it plays in 4th bar as it should
4) Arm fill again while still playing -- the fill goes into infinite (1-bar) loop.
5) Arm again makes it stop, but only if you use the mute/grid button -- doing it via section-select in either row or grid mode doesn't stop it.

That said, if you don't double-tap, I'm totally digging the way the section-select arms the fill.  This makes them super useful for lead-in transitions to that section.  Very musically useful!  So thanks a ton for the feature.  :)

hexagon5un:
Also, on "Mechanism to change the launch style of a clip" -- works in row mode, but not in grid mode.  

That is, holding mute/launch and pressing the select knob gives you the fill/default choice as expected, but I would also expect it to work when holding the grid equivalent, but it doesn't.

PR #198, opened 2023-07-16
Manual slicing

alter-alter:
In the Slicer view, press the ◀︎▶︎ knob. When you press the green pad in the bottom left, it starts playing the first slice, and pressing an empty pad creates a new slice. Turning the Select knob to the left or pressing the Pad + DELETE button allows you to delete a slice.

Turning the ◀︎▶︎ knob allows you to adjust the start point of the slice. Additionally, turning the ▼︎▲︎ knob transposes the slice.

Pressing the Select knob finalizes the slicing process.

https://youtu.be/w_1FBTHhPNs

### DIFF
--- a/src/deluge/gui/ui/slicer.cpp
+++ b/src/deluge/gui/ui/slicer.cpp
@@ -500,6 +500,7 @@ int Slicer::padAction(int x, int y, int on) {
 		int slicePadIndex = (x % 4 + (x / 4) * 16) + ((y % 4) * 4); //
 
 		if (slicePadIndex < numManualSlice) { //play slice
+			bool closePopup=(currentSlice != slicePadIndex);
 			currentSlice = slicePadIndex;
 			if (slicePadIndex + 1 < numManualSlice) {
 				preview(manualSlicePoints[slicePadIndex].startPos, manualSlicePoints[slicePadIndex + 1].startPos,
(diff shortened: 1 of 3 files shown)


weavermedia:
Is it possible for the displayed transpose value to follow the last played pad? Currently if you play a pad and transpose it, say -5, that -5 will continue to flash on the display after you play a different pad which is confusing.

Could the display be updated on every pad press to show the transpose value for the current pad? If no transpose is set for the current pad it could just show the pad number, like normal.

alter-alter:
To avoid confusion, the slice number pops up when the pad is pressed.

weavermedia:
> To avoid confusion, the slice number pops up when the pad is pressed.

Yes indeed, but on 7-seg it stays on the display even after you press another pad. Regardless of whether that new pad has a different transpose settings or no transpose setting at all.

In the video watch the display continues to show the value even after a different pad is pressed.

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/73efc5b4-1d12-4cb1-a029-f4e59a87dc07

weavermedia:
@alter-alter Could you do a build in your own fork for these latest changes? That's where I'm previewing this feature from. I don't know how to checkout a PR on my local machine.

weavermedia:
@alter-alter Never mind, I figured it out 😀 Testing the new changes and it's better that the old transpose values don't show but the new way has problems too. Showing the slice number on every pad press means that if you're tapping a pad while transposing (to hear the effect) you don't really get to see the transpose value properly on the screen. This is because every tap of the pad replaces the transpose value with the slice number.

Is it possible maybe to hold the transpose value on the display for a second before changing back to displaying the slice number? Basically ignore any display updates called by pad presses for the duration of 1 second after every change to the transpose amount. Just a thought.

alter-alter:
Thank you for the feedback. 

- Transpose popup is now turned off when another pad is pressed.
- The slice number popup is no longer displayed.

sichtbeton:
Couldn't check it out yet, but does it convert the chops to kit rows? Because then you can alter the sound of each chop further. And adding to that: is it possible to clone a chop?

alter-alter:
@sichtbeton Yes, it is possible to convert the chops to kit rows. Press the select encoder in the same way as in slicer mode.
However, cloning a chop is not currently supported.

weavermedia:
@alter-alter Excellent! Feels really smooth now 👏

sichtbeton:
one issue: when audio preview is enabled in the menu (preview=on) the sample will play completely through. chops can be made but the first triggered preview of the sample is playing underneath make it impossible to work.

alter-alter:
@sichtbeton Thank you for the feedback. Currently, when entering Slicer mode, the preview gets cut off by pressing Shift + Select, but it seems that the preview is not cut off when entering from the menu by long-pressing the Select encorder. I will fix this issue.

PR #200, opened 2023-07-16
Fix MasterCompressor by considering masterVolumeAdjustment value

alter-alter:
Fixed master compressor. The masterVolumeAdjustment value is now considered in the process. With this change, threshold value are now displayed correctly.

This modification affects songs that had saved master compressor settings.

### DIFF
--- a/src/deluge/dsp/master_compressor/master_compressor.cpp
+++ b/src/deluge/dsp/master_compressor/master_compressor.cpp
@@ -61,8 +61,10 @@ void MasterCompressor::render(StereoSample* buffer, uint16_t numSamples, int32_t
 				r = rawr * (1.0 - wet) + r * wet;
 			}
 
-			thisSample->l = l * adjustmentL * 2147483647;
-			thisSample->r = r * adjustmentR * 2147483647;
+			thisSample->l = l * 2147483647;
+			thisSample->r = r * 2147483647;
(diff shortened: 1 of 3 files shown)


src/deluge/dsp/master_compressor/master_compressor.cpp
```diff
@@ -30,14 +30,21 @@ MasterCompressor::MasterCompressor() {
...
+		double adjustmentR = (masterVolumeAdjustmentR) / 4294967296.0;
+		if (adjustmentL < 0.000001)
+			adjustmentL = 0.000001;
+		if (adjustmentR < 0.000001)
+			adjustmentR = 0.000001;
 		do {
-			double l = thisSample->l / 2147483648.0;
-			double r = thisSample->r / 2147483648.0;
+			double l = thisSample->l / 2147483648.0 / adjustmentL;
+			double r = thisSample->r / 2147483648.0 / adjustmentR;
```
bobtwinkles, line 47: Instead of converting the master volume adjustments to floats, I think we can apply it directly to `thisSample->l` and `thisSample->r` to get everything done in integers (potentially requires `multiply_32x32_rshift32`?)

alter-alter: At this stage, floating point is required to process without clipping with masterVolumeAdjustment taken into account. Additionally, some of the operations have been modified to use integers.

PR #202, opened 2023-07-16
Add generic container specializations

stellar-aria:
This adds the specializations required to use the STL containers with fallback_allocator, as well as adds a copy constructor for fallback_allocator which is required for std::map

### DIFF
--- a/src/deluge/util/containers.h
+++ b/src/deluge/util/containers.h
@@ -1,0 +1,52 @@
+#pragma once
+#include "memory/fallback_allocator.h"
+#include <deque>
+#include <forward_list>
+#include <list>
+#include <map>
+#include <queue>
+#include <stack>
+#include <unordered_map>
+#include <vector>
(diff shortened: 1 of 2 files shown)


src/deluge/util/containers.h
```diff
@@ -0,0 +1,52 @@
...
+
+// Hash map
+template <typename Key, typename T, typename Alloc = memory::fallback_allocator<std::pair<const Key, T>>>
+using unordered_map = std::unordered_map<Key, T, std::hash<Key>, std::equal_to<Key>, Alloc>;
+
+// Stack
+template <typename T, typename Alloc = memory::fallback_allocator<T>>
+using stack = std::stack<T, deque<T, Alloc>>;
+
+// Stack
```
bobtwinkles, line 49: nit: comment

bobtwinkles (approved):

PR #204, opened 2023-07-16
[DRAFT] row layout option for kit keyboard

weavermedia:
This is an optional row layout for the kit keyboard. Velocities are mapped to rows instead of 4x4 grids.

- Pressing the keyboard button in a kit shows the row layout instead of 4x4 grid layout.
- Velocity ranges 7-127 (same as the 4x4 grid) from left to right
- Color intensity ranges from 8-248 (same as 4x4 grid) from left to right
- Vertical scroll changed from 4 rows at at time to 1 row at a time to match the kit row layout

This initial code is just a proof of concept with the original code commented out.

**Might be considered too niche for the community, as such I'm happy for it to be rejected.**

_Also not sure if this should start as a PR or as an issue. Since I have some initial code I decided to do a PR._

If this goes ahead there are a few things to do:

- Add a community feature toggle for grid or row mode, something like KIT KEYBOARD > GRID / ROW
- OR make both views available by using SHIFT+KEYBOARD for the row layout
- Add the conditionals to switch behavior based on feature toggle or shift key
- Re-enable the mute and audition pads on the right
- Change the un-mapped pads to blank instead of dim white used by the 4x4 grid layout

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/e5b3f448-257f-4856-a779-e20d6ea1b1fa


PaulFreund:
Hi @weavermedia, so the first thing you should do if this is still work in progress is to switch the PR to a Draft PR to indicate that it is not ready for merging. 

Regarding the feature itself I would like to suggest you wait until I finished the [keyboard layout API](https://github.com/SynthstromAudible/DelugeFirmware/pull/138/files#diff-038c7c2950fd00088b4cb004457cba67020bbd82767578ba069ec5b8803710a7R33) in #138. You should then be able to either copy the drum keyboard and add a setting to it or make a duplicate that uses your new calculation :)

weavermedia:
@PaulFreund thanks for the advice! I converted this PR to a draft but would it now be better for me to close it and wait until I can submit a new PR using your finished keyboard layout API?

PaulFreund:
I'd suggest to leave the Draft PR open to indicate to others that you want to implement this/work on this and wait until I finished if that is okay with you :) You then apply your changes in the new structure, should be as simple as before :)

weavermedia:
@PaulFreund Will do 🙌

soymonitus:
Why not using the existing shortcut for synth's keyboard mode which is holding Keyboard button and turning Select knob? right now that shortcut is doing nothing on Kit view and to make UI consistent I guess that is the best solution to change between grid and row mode

PR #209, opened 2023-07-19
Refactor out CacheManager from MemoryRegion

stellar-aria:
This extracts a CacheManager object from Memory Region and refactors them, in preparation for replacing the custom allocators.

I'm requesting both code review and testing, since this is pretty fragile stuff to be messing with at the moment.

Edit: I've tested this with a sample that is longer than the memory space (114MiB) and it ran fine, so I _think_ things are okay with this.


src/deluge/memory/cache_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <tuple>
+
+extern bool skipConsistencyCheck;
+uint32_t currentTraversalNo = 0;
+
+// Size 0 means don't care, just get any memory.
+uint32_t CacheManager::ReclaimMemory(MemoryRegion& region, int totalSizeNeeded, void* thingNotToStealFrom,
+                                     int* __restrict__ foundSpaceSize) {
+
+#if TEST_GENERAL_MEMORY_ALLOCATION
```
m-m-adams, line 15: Have you tried running with this flag? No idea what it is

src/deluge/memory/cache_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+			// If that one Stealable alone was big enough, that's great
+			if (amountToExtend <= 0) {
+				found = true;
+				break;
+			}
+
+			// Otherwise, see if available neighbouring memory adds up to make enough in total
+			NeighbouringMemoryGrabAttemptResult result = region.attemptToGrabNeighbouringMemory(
+			    stealable, spaceSize, amountToExtend, amountToExtend, thingNotToStealFrom, currentTraversalNo, true);
```
m-m-adams, line 121: I might be missing it but what is thing not to steal from?

src/deluge/memory/cache_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+			// If that couldn't be done (in which case the original, central Stealable won't have been stolen either), move on to next Stealable to assess
+			if (!result.address) {
+				if (result.longestRunFound > longestRunSeenInThisQueue) {
+					longestRunSeenInThisQueue = result.longestRunFound;
+				}
+				stealable = static_cast<Stealable*>(reclamation_queue_[q].getNext(stealable));
+				continue;
+			}
+
+			newSpaceAddress = result.address;
```
m-m-adams, line 139: Just to confirm - this is only reachable if we've either stolen enough memory at once or if we've successfully also stolen neighbouring memory?

src/deluge/memory/cache_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+	if (!(found || stolen)) {
+#if TEST_GENERAL_MEMORY_ALLOCATION
+		skipConsistencyCheck = false;
+#endif
+		AudioEngine::logAction("/CacheManager::reclaim nope");
+		return 0;
+	}
+
+	if (found && !stolen) {
+		// Warning - for perc cache Cluster, stealing one can cause it to want to allocate more memory for its list of zones
```
m-m-adams, line 162: What does this mean? And why does it send up an error i007?

src/deluge/memory/general_memory_allocator.cpp
```diff
@@ -22,7 +22,7 @@
 #include <new>
 #include "io/debug/print.h"
 #include "util/functions.h"
-#include <string.h>
+#include <cstring>
```
m-m-adams, line 25: Do we already depend on this somewhere? Just thinking of binary size

src/deluge/memory/cache_manager.cpp
stellar-aria: It's a way to pass around a pointer to the current object or an object that should not be stolen from at any cost when requesting more memory.

src/deluge/memory/cache_manager.cpp
stellar-aria: Correct, or we've found memory that can be stolen but has not yet been (the found boolean flag).

src/deluge/memory/cache_manager.cpp
stellar-aria: It means that a block of potentially free-able memory has been found, but not yet stolen. Honestly don't know why it triggers that error, but it's a direct translation of the `foundIt` label here:
https://github.com/SynthstromAudible/DelugeFirmware/blob/6c915942ee0383c93f27f2eff8bfd9bc353532c7/src/deluge/memory/memory_region.cpp#L382

src/deluge/memory/general_memory_allocator.cpp
stellar-aria: `cstring` is just the C++ wrapper for `string.h`. You're not supposed to use C standard headers directly in C++ code, so the C++ standard library provides wrappers that handle all the weird stuff that can crop up.

stellar-aria: It's header-only so there's no overhead

src/deluge/memory/cache_manager.cpp
stellar-aria: Nope, it theoretically tests the memory but I have no idea if the code in the "tests" are still valid for this version.

jamiefaye:
Is this one ready now? It passes my smoke test.

stellar-aria:
> Is this one ready now? It passes my smoke test.

Yep!

m-m-adams:
Oops totally forgot to follow up, didn't notice a performance hit!

m-m-adams (approved):

PR #210, opened 2023-07-19
Fix: Stop audio preview from the sample browser when entering Manual slicing

alter-alter:
Changed to stop audio preview (played by the sample browser) when entering manual slicer mode.
https://github.com/SynthstromAudible/DelugeFirmware/pull/198#issuecomment-1640987706


alter-alter:
Thank you!

sichtbeton:
nice, thats a good addition!

PR #211, opened 2023-07-19
Feature: FILL iteration mode for notes

weavermedia:
This is a draft PR for a new note iteration mode called FILL that only plays the note if the FILL function is currently enabled by some key press.

Todo

- [x] Choose proper key combo to activate the fill mode (preferably MIDI-assignable) = assignable in community settings
- [x] Decide where to put the FILL setting in the probability/iteration list = zero position, before 5%
- [x] Properly light the SYNC_SCALING LED while held (if set in community settings)
- [x] ~~Add a NOT FILL mode where notes *won't* play when FILL is active~~ 👉 Future enhancement
- [x] Add community setting that allows Fill Mode to be chosen as a primary action for the Sync Scaling button
- [x] ~~Implement FILL. (dot) value as a NOT FILL note mode~~ 👉 Future enhancement
- [x] Add FILL to Global MIDI Commands, note on activates and note off deactivates
- [x] ~~Change community setting from simple on/off to hold/toggle/once~~ 👉 Future enhancement

_I realize there's an existing and much more complex 'Fill clips' PR [here](https://github.com/SynthstromAudible/DelugeFirmware/pull/196) and this may cause some confusion so I'm open to whatever needs to be done to fix that. UPDATE: communicated with the PR author and hopefully we can treat these as two separate features._

~~For this simple proof of concept I hijacked the 95% note probability setting as FILL and the SHIFT+TRIPLETS key combo to activate the fill notes.~~

The FILL setting has been added to the start of the probability/iteration list:
FILL, 5%, 10%, ...

This seems a logical place because [1] apart from putting it next to 100% this is the least scrolling to get to it (less than after 8of8) and [2] in a way you are actually setting it to 0% chance of playing during normal playback.

And, in no small part, putting it at position zero in the list was by far the easiest and least intrusive way to code it! 🙌

Video attached because it's easier to see it in action. In the video the majority of notes in the drum pattern are set to FILL therefore do not sound during normal playback. I activate fill mode by holding ~~SHIFT+TRIPLETS~~ (since changed to SYNC_SCALING). While those keys are pressed the FILL notes will sound.

This should work in all modes - Clip View, Song View and Arrange View - both as a performance tool and for re-recording/resampling new variations in the studio.

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/90e6a2ff-ce50-4a08-8d50-a92d27a1c22f


weavermedia:
Need help figuring out how to integrate a FILL value into the current list of probability/iteration values.

weavermedia:
I believe this is ready for review. If anyone has any feedback or spots and bugs please let me know.

weavermedia:
Actually looks like there's competition for the sync scaling button so let's put it in a preference. So far I think the idea is something like this:

Button Functions > Sync Scaling > Sync Scaling / Fill Mode / etc (where other functions can be added here)

Does this look like what we talked about @chrisbc ?

Will we also eventually need another menu entry for shift+ actions too?

src/deluge/playback/playback_handler.cpp
```diff
@@ -2506,6 +2511,27 @@ bool PlaybackHandler::tryGlobalMIDICommands(MIDIDevice* device, int channel, int
 	return foundAnything;
 }
 
+// Returns whether the message has been used up by a note-off command
+bool PlaybackHandler::tryGlobalMIDICommandsOff(MIDIDevice* device, int channel, int note) {
+
+	bool foundAnything = false;
+
+	for (int c = 0; c < kNumGlobalMIDICommands; c++) {
```
m-m-adams, line 2519: For now should it just check the fill command instead of all of them?

.gitignore
```diff
@@ -21,3 +21,4 @@ __pycache__
 .sconsign.dblite
 /dbt_options.py
 /scripts/debug/openocd.log
+.vscode/*
```
m-m-adams, line 24: remove this file from your commit (git rm --cached I think)

src/deluge/playback/playback_handler.cpp
weavermedia: Yeah, good point. I just used a switch statement so the next person to touch this knows the convention. I'll do a single check for now with a note to use switch for extra checks 👍

.gitignore
weavermedia: Werps! That was from merging my own community into this branch to get up to date. So what's the best way for me to keep this line locally but not have to keep excluding it from pushes?

src/deluge/playback/playback_handler.cpp
m-m-adams: Oh I meant the for loop - we only care about the full command so we can skip the checks to other global midi commands

src/deluge/playback/playback_handler.cpp
```diff
@@ -2489,6 +2489,11 @@ bool PlaybackHandler::tryGlobalMIDICommands(MIDIDevice* device, int channel, int
 				}
 				break;
 
+			case GlobalMIDICommand::FILL:
+				currentSong->fillModeActive = true;
+				indicator_leds::setLedState(IndicatorLED::SYNC_SCALING, true);
```
soymonitus, line 2494: Why do you light up the SYNC_SCALING led? What if the user didn't enable to use the sync scaling button for fills, and instead uses it for the default behaviour or a different community feature?

src/deluge/playback/playback_handler.cpp
```diff
@@ -2506,6 +2511,27 @@ bool PlaybackHandler::tryGlobalMIDICommands(MIDIDevice* device, int channel, int
...
+bool PlaybackHandler::tryGlobalMIDICommandsOff(MIDIDevice* device, int channel, int note) {
+
+	bool foundAnything = false;
+
+	for (int c = 0; c < kNumGlobalMIDICommands; c++) {
+		if (midiEngine.globalMIDICommands[c].equalsNoteOrCC(device, channel, note)) {
+			switch (static_cast<GlobalMIDICommand>(c)) {
+			case GlobalMIDICommand::FILL:
+				currentSong->fillModeActive = false;
+				indicator_leds::setLedState(IndicatorLED::SYNC_SCALING, false);
```
soymonitus, line 2524: Same here as my comment above

src/deluge/playback/playback_handler.cpp
weavermedia: Yeah, the sync scaling code is leftover from the initial plan to take over the sync scaling button. Next step after global MIDI command is to add a community setting to choose which physical button triggers fill mode. Or none, of course, if the user just wants to use it via MIDI.

src/deluge/playback/playback_handler.cpp
weavermedia: Yep, same. It's leftover from the initial implementation. Next task is to add the community setting.

src/deluge/playback/playback_handler.cpp
weavermedia: I'm not sure I understand how to go about this. Do you mean I should get rid of the for loop and then in the if statement replace [c] with the actual index of the FILL command, like [17]?

weavermedia: Latest commit should be better.

src/deluge/model/settings/runtime_feature_settings.cpp
```diff
@@ -59,6 +59,28 @@ static void SetupOnOffSetting(RuntimeFeatureSetting& setting, const deluge::stri
...
+	setting.options[1] = {
+	    .displayName = "FILL",
+	    .value = RuntimeFeatureStateSyncScalingAction::Fill,
+	};
+
+	setting.options[2] = {
+	    .displayName = NULL,
+	    .value = 0,
+	};
+}
```
weavermedia, line 82: How does this look as a way to allow users to assign a specific front panel  button as FILL?

weavermedia: https://discord.com/channels/608916579421257728/1118654563596128336/1139997858028867667

weavermedia:
@m-m-adams This is ready for review now, when you have a moment. Thanks for the help debugging the weird build errors.

src/definitions_cxx.hpp
```diff
@@ -720,8 +720,9 @@ enum class GlobalMIDICommand {
 	LOOP_CONTINUOUS_LAYERING,
 	UNDO,
 	REDO,
+	FILL,
 };
-constexpr auto kNumGlobalMIDICommands = util::to_underlying(GlobalMIDICommand::REDO) + 1;
+constexpr auto kNumGlobalMIDICommands = util::to_underlying(GlobalMIDICommand::FILL) + 1;
```
m-m-adams, line 725: I'd suggest adding a LAST member to the enum and taking it's value, less likely to lead to mistakes down the road

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2074,7 +2074,7 @@ void InstrumentClipView::adjustProbability(int32_t offset) {
 
 					// Incrementing
 					if (offset == 1) {
-						if (probabilityValue < kNumProbabilityValues + 35) {
+						if (probabilityValue < kNumProbabilityValues + kNumIterationValues) {
```
m-m-adams, line 2077: Probability value represents fill, a probabilty, or an iteration right? I'd suggest renaming it to triggerCondition or something to make it less misleading

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2170,7 +2171,7 @@ void InstrumentClipView::adjustProbability(int32_t offset) {
 		// Decide the probability, based on the existing probability of the leftmost note
 		probabilityValue = editPadPresses[leftMostIndex].intendedProbability & 127;
 		probabilityValue += offset;
-		probabilityValue = std::clamp<int32_t>(probabilityValue, 1, kNumProbabilityValues + 35);
+		probabilityValue = std::clamp<int32_t>(probabilityValue, 1, kNumProbabilityValues + kNumIterationValues);
```
m-m-adams, line 2174: Is it possible to have a probability of 0? If not please change line 2100 to be >=1 instead of >0

src/deluge/model/note/note_row.cpp
```diff
@@ -2965,7 +2965,7 @@ int32_t NoteRow::readFromFile(int32_t* minY, InstrumentClip* parentClip, Song* s
 				if (velocity == 0 || velocity > 127) {
 					velocity = 64;
 				}
-				if ((probability & 127) > (kNumProbabilityValues + 35) || (probability & 127) == 0
+				if ((probability & 127) > (kNumProbabilityValues + kNumIterationValues)
```
m-m-adams, line 2968: is dropping the (probabilty & 127) == 0 condition intentional?

m-m-adams (commented):
Generally looks good - I've just got a couple small questions where I think code clarity could be improved

src/definitions_cxx.hpp
weavermedia: Agreed. This seemed like a very brittle way to do it in the original code. Maybe this is the first new global MIDI command to be added.

src/deluge/gui/views/instrument_clip_view.cpp
weavermedia: Yes, seems like probabilityValue was named before the iteration feature was introduced. Iteration values are a separate range but all the probability values and iteration values are referenced in one long series.

I like the idea of trigger Condition but it feels like quite a big change to rename it. It's not just renaming the variable name, it would also mean possibly changing lots more function names and code comments to get the clarity to want. The word 'probability' seems to be used interchangeably for either actual percentage probability or the full range of trigger conditions.

Is this the right time to make that change?

src/deluge/gui/views/instrument_clip_view.cpp
weavermedia: I have repurposed 0 as the fill value. I like 0 for fill because [1] it's the cleanest way to implement with the least code disruption and [2] conceptually the fill notes *do* have a zero percent chance of playing during normal playback.

src/deluge/model/note/note_row.cpp
weavermedia: Yes, for the same reason as above - 0 is the value fro fill.

sichtbeton:
will this make the feature cutoff? would be super cool.

src/deluge/gui/views/instrument_clip_view.cpp
m-m-adams: No lets do it in a separate PR for safety

weavermedia:
> ... would be nice to add a LAST member to global midi commands but it's not critical.

I added a LAST member to global MIDI commands as suggested.

src/definitions_cxx.hpp
weavermedia: Done 👍

sichtbeton:
Could this get merged in before the cutoff?

weavermedia:
> Could this get merged in before the cutoff?

Hope so! 😀

soymonitus:
> > Could this get merged in before the cutoff?
> 
> 
> 
> Hope so! 😀

Please fix conflicts. Hope this is merged soon

PR #213, opened 2023-07-19
Fix for missing menu items on 7seg

m-m-adams:
The midi devices menu and MPE zone menu weren't properly showing on 7seg due to what seems to be initialization order bugs. This moves their declarations into menus.cpp

### DIFF
--- a/src/deluge/gui/menu_item/midi/devices.cpp
+++ b/src/deluge/gui/menu_item/midi/devices.cpp
@@ -26,8 +26,6 @@ extern menu_item::Submenu midiDeviceMenu;
 
 namespace menu_item::midi {
 
-extern Devices devicesMenu{"Devices"};
-
 static const int lowestDeviceNum = -3;
 
 void Devices::beginSession(MenuItem* navigatedBackwardFrom) {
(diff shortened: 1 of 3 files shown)


src/deluge/gui/menu_item/midi/devices.cpp
```diff
@@ -26,7 +26,7 @@ extern menu_item::Submenu midiDeviceMenu;
 
 namespace menu_item::midi {
 
-Devices devicesMenu{"Devices"};
+extern Devices devicesMenu{"Devices"};
```
stellar-aria, line 29: Could you remove the brace-initializer here so we know it's not getting used? And for the other extern?

src/deluge/gui/menu_item/mpe/direction_selector.cpp
```diff
@@ -22,7 +22,7 @@
 
 namespace menu_item::mpe {
 
-DirectionSelector directionSelectorMenu{"MPE"};
+extern DirectionSelector directionSelectorMenu{"MPE"};
```
stellar-aria, line 25: and here

stellar-aria (commented):
LGTM pending brace-init stuff.

stellar-aria (approved):

Issue #84, opened 2023-06-22
Snap-to-zero-crossing in waveform editor


sapphire-arches:
From a discussion on Discord earlier today: https://discord.com/channels/608916579421257728/1118654563596128336/1121185820854980678

In the waveform editor, the following shortcuts should work:
- Vertical scroll encoder should swap the currently selected endpoint in the order
  1. start
  2. loop start if pulled out
  3. loop end if pulled out
  4. end
- Holding down the select encoder while moving it should snap to zero crossings

weavermedia:
Could we also add a highlight to the zero crossing row in the waveform display? Probably only really useful when zoomed all the way in to single-sample level. I do single cycle looping on samples sometimes and having the zero crossing sample column brighter (or another color) than the others would be helpful.

sichtbeton:
and maybe to have it backwards compatible, this could be made optional as a "community setting".

weavermedia:
Looked into zero-crossing a little. There's a common method for finding where the audio sample level crosses the 0 line which just compares the sign of the integer. When the sign changes, that's a zero-cross point.

But we have to remember we're dealing, for the most part, with stereo samples. We probably can't just look for zero-crossing because for both stereo channels to cross zero on the same sample is far less likely.

Maybe a combination of doing an average of both channels with some low threshold might work?

Probably best to actually find out how other systems do it. Maybe look at open source DAWs and/or sample editors.

sichtbeton:
maybe this could be turned into a discussion as its not really an issue rather then an enhancement?

weavermedia:
@sichtbeton good call 👍

PR #214, opened 2023-07-20
[DRAFT] Sample loop lock

weavermedia:
Allows you to lock the sample loop points movements together.

1. Set sample LOOP START and LOOP END as usual
2. Hold LOOP START and tap LOOP END to engage LOCK
3. Move either loop point and they're locked together
4. Turn SCROL ▲▼ right to DOUBle the loop length
5. Turn SCROL ▲▼ left to HALF the loop length
6. Hold LOOP START and LOOP END again to release LOCK (FREE)

Something I missed from my old hardware samplers is the ability to lock the loop start and end points together. So when you move one the other moves too so they stay the same distance apart. When working with rhythmic samples this gets some really cool results.

Still stuff to do on this feature and proper considerations to make for the controls.

- [ ] Prevent scrolling the loop outside the sample start and end points
- [ ] Prevent doubling the loop longer than the sample
- [ ] Prevent halving the loop to 0
- [ ] Ensure playback behaves well when scrolling the loop past the current playback position
- [ ] Should loop lock boolean become a property of the sample?
- [ ] Probably don't want to hijack the vertical encoder just for doubling/halving the loop length

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/92f3391d-ff65-495e-a7d5-a71190b319d3


sichtbeton:
This is great! My only gripe with the loop points in the deluge right now is that they are not really usable because they introduce rather nasty clicks. It's time-consuming to get rid of them, and sometimes it doesn't work at all. I remember we already talked about this somewhere else. 

To make a long story short: adding an optional zero-crossing feature would be awesome in combination with your loop lock. It would probably be a separate PR (and probably a complex one too).

weavermedia:
@sichtbeton Yes, zero-crossing will be a very useful feature. Does seems fairly complex though, as you said. There's a [PR](https://github.com/SynthstromAudible/DelugeFirmware/issues/84) for it that we both commented on. I did a bit of research so I'll drop my notes there.

weavermedia:
Reimplemented as #293

PR #219, opened 2023-07-20
Refactor definitions.h for modern C++

stellar-aria:
Ongoing efforts to refactor definitions.h to be more in line with modern C++ standards.

~~Still some work to be done (such as renaming the constexpr variables now that they're no longer macros), and smoke testing/bugfixes.~~

This is now ready for testing and needs a great deal of it.


src/definitions_cxx.hpp
```diff
@@ -493,13 +493,14 @@ constexpr auto kNumPolyphonyModes = util::to_underlying(PolyphonyMode::CHOKE) +
 
 #define NUM_PHYSICAL_MOD_KNOBS 2
 
-enum LPFMode {
-	LPF_MODE_12DB,
-	LPF_MODE_TRANSISTOR_24DB,
-	LPF_MODE_TRANSISTOR_24DB_DRIVE,
-	LPF_MODE_SVF,
-	NUM_LPF_MODES,
+enum class LPFMode {
```
m-m-adams, line 514: This is the same name as the class used in the filter menu - can that cause any issues given that soundEditor.currentModControllable->lpfMode is referenced within the menu_item::LPFMode class and will now be an LPFMode as well?

stellar-aria: Not with proper namespace scoping. See https://github.com/SynthstromAudible/DelugeFirmware/blob/46df82da7fd44a447bded914a412edde63a217f6/src/deluge/gui/menu_item/filter/lpf_mode.h#L32

stellar-aria: Ideally when I get the Menu Item rework done, these will be templated over the values they contain (such as a LPFMode), and the need for casting to/from will be gone.

PR #221, opened 2023-07-21
Catch Notes setting to allow not catching existing notes.

weavermedia:
Small thing that always bugged me. When shift switching between clips there was code that looked for any notes in the new clip that would've been recently triggered and plays them. For drum clips this was especially noticeable - producing double hits and flams and didn't sound good at all.

This community setting I called Catch Notes is a simple on/off that adds a conditional to the code that finds and plays these 'already triggered' notes. Set to 'off' these notes will not be triggered and shift switching between drum clips now becomes flawless.

It defaults to 'on' which is the current Deluge behavior so no surprises for users.

Again, a video is worth a billion words. In the first half Catch Notes is on (current default behavior) and you'll hear lots of double hits and flamming. After I turn it 'off' switching clips is silky smooth.

_NOTE: I would've called this a bug fix until I found the code that plays these notes. It's very deliberate and intentional code so I wouldn't consider it a bug to fix. Just a behavior to optionally change._

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/f14b4985-0c57-4632-860e-af94d4f8e33d

### DIFF
--- a/src/deluge/gui/menu_item/runtime_feature/settings.cpp
+++ b/src/deluge/gui/menu_item/runtime_feature/settings.cpp
@@ -37,6 +37,7 @@ Setting menuMasterCompressorFx(RuntimeFeatureSettingType::MasterCompressorFx);
 Setting menuFineTempo(RuntimeFeatureSettingType::FineTempoKnob);
 Setting menuQuantize(RuntimeFeatureSettingType::Quantize);
 Setting menuPatchCableResolution(RuntimeFeatureSettingType::PatchCableResolution);
+Setting menuCatchNotes(RuntimeFeatureSettingType::CatchNotes);
 
 std::array<MenuItem*, RuntimeFeatureSettingType::MaxElement + 1> subMenuEntries{
     &menuDrumRandomizer,
@@ -44,6 +45,7 @@ std::array<MenuItem*, RuntimeFeatureSettingType::MaxElement + 1> subMenuEntries{
     &menuFineTempo,
     &menuQuantize,
     &menuPatchCableResolution,
+    &menuCatchNotes,
 
     nullptr,
 };
--- a/src/deluge/model/note/note_row.cpp
+++ b/src/deluge/model/note/note_row.cpp
@@ -239,8 +239,7 @@ addNewNote:
 		if (clipCurrentlyPlaying && !muted) {
 			((InstrumentClip*)modelStack->getTimelineCounter())->expectEvent();
 
-			if ((runtimeFeatureSettings.get(RuntimeFeatureSettingType::CatchNotes)
-					== RuntimeFeatureStateToggle::On)) {
+			if ((runtimeFeatureSettings.get(RuntimeFeatureSettingType::CatchNotes) == RuntimeFeatureStateToggle::On)) {
 				// If the play-pos is inside this note, see if we'd like to attempt a late-start of it
 				int actualPlayPos = getLivePos(modelStack);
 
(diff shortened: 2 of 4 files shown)


PaulFreund:
In reference to the discussion in #253 I found a way to reliably reproduce the issue:

1. With current Community HEAD open to a blank song
2. Make sure Catchnotes is turned on in community settings
4. Set tempo to 120 or higher (might make it easier)
3. Create a TR-808 clip with a kick in position 1, 5, 9 and 13 (four to the floor) and a snare in 3, 7, 11 and 15
4. Copy the clip to the next row
5. Hold shift and rapidly click between the two row arm buttons
6. You will hear a lot of double sounding clips

sichtbeton:
> In reference to the discussion in #253 I found a way to reliably reproduce the issue:
> 
> 1. With current Community HEAD open to a blank song
> 2. Make sure Catchnotes is turned on in community settings
> 3. Set tempo to 120 or higher (might make it easier)
> 4. Create a TR-808 clip with a kick in position 1, 5, 9 and 13 (four to the floor) and a snare in 3, 7, 11 and 15
> 5. Copy the clip to the next row
> 6. Hold shift and rapidly click between the two row arm buttons
> 7. You will hear a lot of double sounding clips
> 
> Very important edit: Looking at the code I just noticed that your default = ON is the official behavior and turning the setting OFF introduces your chaanges. Very good job fixing this issue :D !
> 
> @sichtbeton can you please describe in which state of the Catchnotes setting you are experiencing problems? If you had the same problem I had I think we make this the new default behavior. I also think maybe we should change the logic of the setting and call it "Fix double triggering" or something more appropriate so = ON indicates that the default behavior was changed.

with "off" it's working like expected. with "on" it's double triggering. so it's (if I wrap my head around this right) what you are seeing. naming it "off" when actually off is a sensible to do. 

good job @weavermedia !

weavermedia:
Haha, you're both absolutely correct @PaulFreund @sichtbeton 😀

I'll explain my thought process (and where it went wrong 😜):

'Catch notes' is what the original Deluge code does. It catches and plays the notes that would have been played it the clip was switched perfectly on the beat. That's what leads to the double hits.

My feature is a switch to disable this 'catch notes' behavior. But *(here's where it went wrong)* turning a feature 'on' to disable the behavior named in the feature is indeed completely the wrong way around!

I propose we continue to call the feature Catch Notes and keep the default to 'on' but simply flip the logic.

Then update the feature description to explain the default behavior and offer this switch as a way to disable that default behavior.

I'll submit a fix PR later today.

weavermedia:
Wow, I got really mixed up there for second! 😂 @PaulFreund @sichtbeton 

I'll explain my thought process:

'Catch notes' is what the original Deluge code does. It catches and plays the notes that would have been played it the clip was switched perfectly on the beat. That's what leads to the double hits.

My feature is a switch to disable this 'catch notes' behavior. Maybe it's not very intuitive to have a community feature that has a name of default behavior, rather the the name reflecting new behavior?

The setting defaults to 'on' because that's the original Deluge behavior.

weavermedia:
For reference here is Rohan's response to my FB post asking if the default behavior is a bug:
https://www.facebook.com/groups/DelugeBetaTesters/posts/1677334075806819/

sichtbeton:
> Wow, I got really mixed up there for second! 😂 @PaulFreund @sichtbeton
> 
> I'll explain my thought process:
> 
> 'Catch notes' is what the original Deluge code does. It catches and plays the notes that would have been played it the clip was switched perfectly on the beat. That's what leads to the double hits.
> 
> My feature is a switch to disable this 'catch notes' behavior, therefore preventing the double hits.
> 
> Maybe it's not very intuitive to have a community feature that has a name of default behavior, rather the the name reflecting new behavior?
> 
> The setting defaults to 'on' because that's the original Deluge behavior.

confusing. but you're right on the money there.

PR #222, opened 2023-07-21
UX guidelines

m-m-adams:
This PR introduces a UX guideline for new features in the deluge, and proposes a process for approving changes to the existing UI


Documentation/ux_principles.md
```diff
@@ -0,0 +1,50 @@
...
+    - A user who understands synthesis should immediately be able to edit the deluge
+* Favour simplicity
+    - Menus should never be mandatory
+    - Make the shortcuts and gold knob system feel like a feature and not a hindrance 
+    - The OLED screen is a bonus, not a requirement
+* Favour performance
+    - Entire parameter ranges should be useful and scaled for quick editing
+    - Gold knob twists control frequently edited parameters
+    - Gold knob presses should toggle, not scroll through lists
+    - Infrequently edited parameters belong on shortcut pads
```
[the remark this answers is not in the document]
m-m-adams: Yeah that's true - I thought I had a line saying everything should be in the menu but it looks like I deleted it.

I'll think on how to reword it - my intent is frequently edited on knobs, infrequently edited on shortcuts, set and forget in menus

weavermedia:
"Gold knob presses should toggle, not scroll through lists"

The knobs already scroll through lists, or am I misunderstanding? Filter LPF/HPF/EQ and Reverb Small/Medium/Large for example. They're different in that the reverb list is setting the size param in the background and the knob then just edits the amount but presses on the filter knob actually scrolls the filter through different modes, changing the action of the knob turns.

Then there's the master compressor feature, which uses knob presses to scroll through lists of editable parameters.

m-m-adams:
I think you're understanding it fine, I don't believe anything you've mentioned is a good UI pattern. The overloaded knob functions on turn make performance features like filter sweeps unpredictable - grabbing that knob could silently be in a state where you're affecting any of LPF/HPF/EQ. The compressor gold knob presses are both inconsistent with the existing UI and obtuse to track on 7seg. The compressor shortcuts do fit in with this UI guide.

The reverb size and filter type toggles are what I had in mind as a good UI pattern. I could reword it to say that presses should change modes and not change editor targets?

weavermedia:
Yes indeed, I've been caught out with unexpected knob turns when recording videos. Not as critical as playing live but still annoying.

I wonder if some kind of 'performance mode' might make sense? Some way to lock certain controls to avoid ambiguity when performing but still allowing complex multi-control editing in the studio. Maybe expecting the Deluge to be both things at the same time is unrealistic.

Maybe a per-song XML tag that locks one function per knob. The assignments would be optional, so anything unassigned would act as normal. This means you could lock the filter to LPF, of example, or prevent the delay knob push from changing PING/NORM, etc.

If I was serious about performing with the Deluge I would personally be assigning all critical performance controls out to a dedicated MIDI controller. Not always possible or desirable of course. Sometimes the whole point is that you're performing with **just** a Deluge.

Rewording the toggle reference might be good idea. I took from that the meaning of switching between two states. But if you mean stepping through a list of states that's cool.

Getting the terminology right feels tricky. Pushing the top filter knob changes which parameter that knob controls. Pushing the master compressor top knob also changes which parameter that knob controls. The difference is that pushing the filter knob changes the mode of the filter, whereas pushing the master compressor knob doesn't, it merely changes which parameter is being edited.

The implementation of the master compressor did concern me when I first saw it and to be honest I was a little surprised that it got merged with little discussion about this new UI pattern. The feature itself is very cool but I do think the UI and presentation could be brought more in line with current patterns.

So maybe something like this: "Gold knob presses should switch modes, not scroll through parameter lists."

Documentation/ux_principles.md
```diff
@@ -0,0 +1,51 @@
...
+## What the Deluge Isn't
+The deluge is not a computer, and should not feel like you're operating one. Direct complexity and options should be minimized as much as possible, instead allowing the user to create complexity through combining existing features. An example of this is envelope shape - the envelope attack shape is fixed, but through modulating the attack rate with the envelope it can be smoothly morphed from logarithmic to exponential.
+
+## Guidelines
+
+* Favour explicitness 
+    - Parameters describe their action on the waveform, not their result on the final sound
+    - A user who understands synthesis should immediately be able to edit the deluge
+* Favour simplicity
+    - Menus should never be mandatory
```
trappar, line 20: I think this is too strong of a statement. There's plenty of behavior where menus are mandatory. Device settings and community features is an example, but there's also been talk of things like autosampling and batch actions. We can't change the physical labels on the hardware so certain features will simply not be able to be exposed any way other than through menus, and I think that's okay.

Documentation/ux_principles.md
```diff
@@ -0,0 +1,51 @@
...
+    - Make the shortcuts and gold knob system feel like a feature and not a hindrance 
+    - The OLED screen is a bonus, not a requirement
+* Favour performance
+    - Entire parameter ranges should be useful and scaled for quick editing
+    - Gold knob twists control frequently edited parameters
+    - Shortcuts control infrequently edited paramters
```
trappar, line 26: I would add an item after this "Menu-only items are reserved for rarely edited parameters and should be used when no other reasonable implementation option exists"

Documentation/ux_principles.md
```diff
@@ -0,0 +1,51 @@
...
+    - Gold knob presses switch modes, not scroll through parameter lists
+* Favour consistency
+    - Actions which have similarity between views should have the same shortcut
+    - <>+back clears all notes in a clip or all clips in the arranger
```
trappar, line 30:
```suggestion
      - i.e. `<>` + `back` clears all notes in a clip or all clips in the arranger
```

Documentation/ux_principles.md
weavermedia: @trappar But that statement will become less and less true over time. As more high-impact features are added, with no shortcuts or reasonable button combos, they will have to live in menus, surely?

Documentation/ux_principles.md
m-m-adams: I think that's fine, by mandatory I mean that users shouldn't feel like they need to use the menus and are missing out if they don't. I'll change it to 'menus should never feel mandatory' to emphasize that it's about the user experience more than anything. 

E.g. right now all sound design parameters are accessible without using the menu system. Conversely the midi/MPE settings do require a menu, but the defaults are fine 90% of the time and there is some logic in the deluge to detect MPE devices and do reasonable changes to accommodate.

Documentation/ux_principles.md
m-m-adams: This is deliberate - it indicates that within the community firmware there will be an emphasis on UX, and a feature that is cool but not a good UX may not get merged

I think it's ok that for some things the answer is to take advantage of the deluge's fantastic sequencer and integrate something else to your setup!

m-m-adams:
I'm not sure what I've done wrong but there's another commit on my branch that doesn't seem visible in the pr

weavermedia:
Should the UX guide mention the implied shift action of the audition and note pads? It's something to take into account when designing new interactions.

- NOTE+SAVE deletes a pad row in a KIT, instead of NOTE+SHIFT+SAVE. Aside - NOTE+SAVE does nothing in SYNTH.
- AUDITION+grid shortcuts allow value editing, or open the browser or waveform view.

m-m-adams:
Yeah good idea, maybe a section on current UI paradigms? It could include shift access as well as what knobs are expected to do in different views

weavermedia:
It would be good to get some consistency on the usage of Runtime Feature Settings vs Community Features Menu. For regular users I think the second is best but in code it's referred to as the first.

Documentation/ux_principles.md
```diff
@@ -0,0 +1,52 @@
...
+This is a subset of issues where the current UX does not fully align with the principles above
+
+* Horizontal scrolling is column by column in the arranger but screen by screen in the clip or song
+* Clip nudging in the arranger is clip+ turn <>, note nudging in clip is note + press and turn <>
+
+## Unclear
+
+These are spots where there is ambiguity in the UI that could be improved
+
+* Clips in session view are not clearly linked to tracks in arranger view. Perhaps the audition pads in arranger view could show the row colour (#163) from session view
```
weavermedia, line 50: Should this use the term 'song view' to align with the front panel and the user manual?

Documentation/ux_principles.md
```diff
@@ -0,0 +1,52 @@
...
+* Linked/clone clips are not immediately visible in clip view - perhaps all linked clips should get the same row colour
```
weavermedia, line 51: Maybe change 'visible' to 'identifiable'?

Documentation/ux_principles.md
PaulFreund: Maybe we should refactor session to song view also in the code to remove the ambiguity

CONTRIBUTING.md
```diff
@@ -81,7 +81,17 @@ The following requirements must be fulfilled for a Pull request to be mergable t
...
     * If possible files and flash configuration created with the changes of the Pull request can be used with older firmwares not having the change (downward compatibility). Older firmwares must not break if a newer configuration was stored on the device.
-* If the Pull request changes end user behavoir or introduces new features a new entry in the [CommunityFeatures.md](CommunityFeatures.md) file needs to be created in the preexisting style describing the feature and it's options as a small manual to users. This includes all runtime and compile time flags which shall be named in respective sections.
+* If the Pull request changes end user behavoir or introduces new features a new entry in the [CommunityFeatures.md](Documentation/CommunityFeatures.md) file needs to be created in the preexisting style describing the feature and it's options as a small manual to users. This includes all runtime and compile time flags which shall be named in respective sections.
+
+### UI Changes
+
+Changes to the existing UI should follow the following process
+
+* Ensure they meet the guidelines set out in [UX Principles](Documentation/ux_principles.md)
```
PaulFreund, line 90: We should align CamelCase.md with UPPERCASE.md and snake_case.md. I don't know which is best but three different doesn't seem right :D

CONTRIBUTING.md
```diff
@@ -81,7 +81,17 @@ The following requirements must be fulfilled for a Pull request to be mergable t
...
+* Place the change in the community feature menu for one beta cycle
+* Following a full beta cycle and user feedback, open a poll on the feature
+* Given maintainer approval and positive poll results, the feature can be moved from the community menu to the default Ui
```
PaulFreund, line 93: I feel like this is not explicit enough. What is a positive poll result, more than 50%, 100%? I did originally intend all settings that change existing behavior to stay in community settings and everything else integrated in the main menu. If we want to make community settings new standard behavior I agree we need a mechanism. Suggestion: At the end of every beta cycle there is a community meeting on discord where people can put up stuff to vote on and 90% of the participants must agree. (I'm flexible on the 90%, I'm sure there is prior art for voting systems)

Documentation/ux_principles.md
m-m-adams: Maybe we should just embrace session view for song view and change the manual? I feel like the codebase is more correct here since it's the deluge version of ableton/logic session view


But yeah for right now I'll change to song view

CONTRIBUTING.md
m-m-adams: Yeah I left it somewhat open since I don't think we'll be able to prescribe a complete process until we actually try it out. I like the idea of the Discord community meeting though.

CONTRIBUTING.md
m-m-adams: UPPERCASE.md is what github wants for the top level documents (CONTRIBUTING/CODEOWNERS/ETC). The codebase is generally using snake_case for file names so I'll change community features to match

m-m-adams: Hmm that may not be mandatory - https://github.com/orgs/community/discussions/22673

PR #223, opened 2023-07-21
unison stereo spread

bfredl:
only works for basic waves (and wavetables) yet, not samples nor FM/RM. I'm gonna try refactor the mono vs stereo rendering logic, to not duplicate the same panning logic for samples, FM and ringmod.

For now the shortcut is to press SELECT inside the unison number menu, but we could also take the free spot above it (right of decimation).

### DIFF
--- a/src/definitions_cxx.hpp
+++ b/src/definitions_cxx.hpp
@@ -808,6 +808,7 @@ constexpr int kCacheByteDepth = 3;
 constexpr int kCacheByteDepthMagnitude = 2; // Invalid / unused for odd numbers of bytes like 3
 
 constexpr int kMaxUnisonDetune = 50;
+constexpr int kMaxUnisonStereoSpread = 50;
 
 // This is about right. Making it smaller didn't help. Tried it as 9, and I'm pretty sure some fast percussive details were lost in the output
 constexpr int kPercBufferReductionMagnitude = 7;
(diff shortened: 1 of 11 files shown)


bfredl:
Now it supports all synth types. Code could be simplified if `renderBasicSource` always took a `StereoSample *` buffer but I didn't want to change the flow of the algorithm yet for the case when this new functionality is turned off, without proper measurement of impact.

Note: the diff is slightly less scary than it looks since a lot of it is moving the code for stereo voice buffer from "subtractive" to the end of `voice::render` so it could be used by the other engines as well when needed.

entzmingerc:
Hey @bfredl, when I had oscillator sync on and I turn unison stereo spread to a nonzero value, the oscillator sync no longer applies. You can hear the sync happening with osc sync on, turning up the volume of oscillator 2, and modulating osc 2 transpose with an LFO and hear it the sync effects. Then with unison voice number set to 2 or more, when you turn up the unison stereo spread, the syncing effect is no longer heard.

I wasn't sure if this was a bug or if it's a limitation of the engine.

Also I think I've used unison stereo spread on 99% of synth patches I've made since you added this. Thank you very much for your work on Deluge.

bfredl:
@entzmingerc sounds like a bug. I'd suggest open a separate issue so it can be tracked properly.

PR #228, opened 2023-07-22
Fix for i008 in latest community

m-m-adams:
Error i008 is being triggered when browsing synth presets with latest community, this breaks out the error macro into a function to allow breakpoints for debugging.

Bug is narrowed down to a call to nodes.empty() on parameter automation when the preset is loaded for preview


m-m-adams:
Found a clue in auto_param.cpp!

	// Must first delete any automation because sometimes, due to that annoying support I have to do for late-2016 files, we'll be overwriting a cloned ParamManager, which might have had automation.

m-m-adams:
The freeze always seems to be while updating arp rate, line 3527 in sound.cpp

Issue #105, opened 2023-06-25
Filter Overhaul


m-m-adams:
The existing deluge filters have a fixed configuration of a low pass ladder filter into a high pass ladder filter. The goal of this issue is to switch it to something more like the hydrasynth, where the LPF would have a variety of models to choose from and the HPF is replaced by an SVF that's continuously variable from LP to band or notch to HP. This opens up a ton of possibilities for sound design by 

Main goals:
- [x] Augment HPF ladder filter with a morphable (low-band/notch-high) SVF. Maintain a classic switch to use the existing HPF
- [x] Enable series/parallel switching of the two filters
- [ ] ~~Enable oscillator routing separately into each filter~~ not feasible without overhauling oscillators
- [x] Add additional filter types to LPF slot (SVF, vocal, BPF bank, maybe more)

Side goals:
- [x] Refactor filter code to separate ladder filter code from filter configuration #304 
- [x] Refactor filter menu and configuration code so that a single list of filters can be maintained and automatically populate menus and configuration (mostly)

I'm transferring the progress tracker from my PR to this issue so that PRs can be incorporated when usable before the entire issue is complete.

m-m-adams:
PR #103 adds a lowpass SVF to the low pass filters, accessible through the existing filter selection button

bfredl:
> The goal of this issue is to switch it to something more like the hydrasynth, 

While I like the filters on the hydra and agree it is a good inspiration, I wonder if it not be simpler for us and more flexible to treat the two filter slots as symmetric and allow to pick any filter model in either slot?

m-m-adams:
Yeah that's a great point, I was mostly thinking about what the UI would look like. If the HPF slot is always an SVF then the HPF db/oct pad can become the morph control 

On the other hand there are the two unused pads below the current filter controls that we could expand too, but I think that would need a broader discussion on how we're handling UI changes

sichtbeton:
As already stated on the discord, there is a strange behaviour when using the filter for a complete kit. steps to replicate the bug:

- have one item or more in a kit
- enable "affect all"
- turn lpf to "svf"
- play a longer sound to hear the filter actually working
- turn the filter down.

what happens is that the moment the filter is engaged, there is an audible click coming from the filter and there is a very sudden jump in cutoff.

m-m-adams:
I'm currently looking at building a filter interface with three methods, init, setup and filter.
- Init - construct a new filter with cleared internal state
[...]

PR #233, opened 2023-07-25
refactor filter configs into seperate classes

m-m-adams:
No logic changes - this breaks the filter configuration setups out into their own class to aid in refactoring in support of #105 

This also adds some integer square root functions to fast_fixed_math.c, I ended up avoiding their use but they're left here in case they are useful to someone else

### DIFF
--- a/src/deluge/dsp/filter/filter_set.cpp
+++ b/src/deluge/dsp/filter/filter_set.cpp
@@ -26,7 +26,7 @@
 FilterSet::FilterSet() {
 }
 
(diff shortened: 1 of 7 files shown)


jamiefaye:
Perhaps we should hold off on fast_fixed_math.c stuff until somebody else needs it. Then we don't risk something unused hanging around forever.

m-m-adams:
Yeah that was my thinking too, if someone else needs it at least this way it's findable

PR #237, opened 2023-07-26
Enable dbt format to alphabetize includes

stellar-aria:
In the long-term this should reduce merge conflicts due to includes, as well as making clear exactly which files use what parts of the codebase.

Any PRs broken by this merge should run `dbt format`and _then_ perform a merge commit from community (if necessary).


.clang-format
```diff
@@ -19,4 +19,4 @@ AlwaysBreakTemplateDeclarations: Yes
 
 # Not sure about this one, it's a lot of changed lines but I also like
```
bobtwinkles, line 20: can delete this comment

src/RZA1/intc/devdrv_intc.h
```diff
@@ -32,6 +32,7 @@
 /******************************************************************************
 Includes   <System Includes> , "Project Includes"
 ******************************************************************************/
+#include <stdint.h>
```
bobtwinkles, line 35: Technically this should probably be `r_typedefs` (which also defines the integer types which is its own can of worms). I'm not opposed to using this as an excuse to move to the standard headers though.

m-m-adams (approved):
Build works and nothing looks glaringly wrong. We might have some bugs stemming from initialization order changes but given the lack of tests I don't know a better way to find them than merging this and seeing if bug reports come in

src/RZA1/intc/devdrv_intc.h
m-m-adams: Stdint is already in use in the project, we might as well commit to it

.clang-format
stellar-aria: Deleted

src/RZA1/intc/devdrv_intc.h
stellar-aria: Yeah I think I'd rather break away from using r_typedefs.h towards standard C headers.

bobtwinkles (approved):

PR #250, opened 2023-07-28
Highlight incoming MIDI notes and Layout option to display Norns grid input

PaulFreund:


src/deluge/gui/menu_item/runtime_feature/settings.cpp
```diff
@@ -57,10 +59,18 @@ Submenu subMenuAutomation{
...
+    &menuFineTempo,
+    &menuQuantize,
+    &menuPatchCableResolution,
+    &menuCatchNotes,
+    &menuDeleteUnusedKitRows,
+    &menuAltGoldenKnobDelayParams,
+    &subMenuAutomation,
+    &menuDevSysexAllowed,
+    &menuHighlightIncomingNotes,
+    &menuDisplayNornsLayout};
```
stellar-aria, line 73: trailing comma should probably be here so we get the dangling brace

src/deluge/gui/ui/keyboard/layout/norns.h
```diff
@@ -0,0 +1,47 @@
...
+
+constexpr int32_t kMinNornsRowInterval = 1;
+constexpr int32_t kMaxNornsRowInterval = 16;
+
+class KeyboardLayoutNorns : public KeyboardLayout {
+public:
+	KeyboardLayoutNorns() {}
+	virtual ~KeyboardLayoutNorns() {}
+
+	virtual void evaluatePads(PressedPad presses[kMaxNumKeyboardPadPresses]);
```
stellar-aria, line 32: If these are overriding a parent's virtual method they should be marked override instead of virtual.

stellar-aria (changes requested):
Other than a few style-based things, the logic of this looks good.

I'm a bit concerned about the number of nested ifs in melodic_instrument, but that's a problem for another time 😅

src/deluge/gui/menu_item/runtime_feature/settings.cpp
PaulFreund: Fixed :)

src/deluge/gui/ui/keyboard/layout/norns.h
PaulFreund: Fixed in all layouts :)

stellar-aria (approved):

jamiefaye:
I went out of order on the PRs (again) and this one picked up conflicts from one that should have happened later. Sorry about that.

entzmingerc:
Here's what I've got so far trying to get this to work.
midigrid is the library that acts as a fake grid interface for norns script (I'm not currently using polygrid)
awake is the norns script I'm using
7seg deluge USB connection to norns

### Norns code changed:
To get awake/awake.lua to work with midigrid, I placed ``local grid = util.file_exists(_path.code.."midigrid") and include "midigrid/lib/mg_128" or grid`` above the line ``g = grid.connect()`` around line 42 as advised by the midigrid documentation and forum discussion. 

I added deluge as a supported device in midigrid. I modified the file midigrid/lib/supported_devices.lua and added ``{ midi_base_name= 'deluge 1',      device_type='deluge' },`` to the table of supported_devices. Norns detects Deluge 1, Deluge 2, Deluge 3 as midi devices, but it will only be a midigrid "supported device" if the midi device name appears in this table. 

Attached is the lua file for the deluge device type for midigrid. I placed this in midigrid/lib/devices. Github doesn't let me upload .lua files, so I've saved it as a .txt. Please delete TEXT from the filename and save as a .lua file if you use it. Following the example of other device type files, I use generic_device.lua use the default specification, then change it to fit for deluge (width 16, note placement, etc.)
[delugeTEXT.txt](https://github.com/SynthstromAudible/DelugeFirmware/files/12389406/delugeTEXT.txt)

Currently, I can press pads and send messages from deluge to norns using a midi clip set to channel 16. I see the LEDs updating on deluge if I learn the incoming midi data from norns to deluge using a different clip, but problems happen when I learn the incoming data to the midi clip sending midi messages out on channel 16. 

### Problem:
- midigrid sends noteON midi messages to deluge to update LEDs using channel 1, note = location, velocity = brightness
- deluge sends noteON midi messages to norns as grid commands (I've currently hardcoded this to channel 16)
- we have to use a midi clip on deluge to send midi messages, we have to learn the midi clip to the incoming midigrid data to update the LEDs, and this triggers 128 noteON events in the midi clip sent out from deluge to norns, which triggers norns to refresh the grid LEDs, and this feedback fills the midi buffers locking everything
- We have to use noteON from deluge to norns because a midi clip pad press sends a noteON message

Also, each clip on Deluge currently needs to learn the incoming data to have the LEDs update, but this doesn't matter imo because only one clip should be "the monome grid mode clip".

If norns-view is enabled in features menu, we could restrict triggering outgoing noteON midi messages from incoming noteON events on channel 1 since midigrid will consume this channel anyway with grid updating stuff, but I'm not sure if that's possible. We'd still want to send outgoing noteON messages, but only if we press a pad on deluge. (I like this but don't know if it's possible)

Another solution I don't like as much because it's a weird workaround would be changing the deluge device type file in midigrid to send and receive control change messages instead of noteON for updating deluge LEDs so it'd be (channel 1, controller, value) but then it'd be different from your PR trying to highlight incoming notes, and then deluge midi clip would have to send a midi CC message if it's on channel 1 when a pad is pressed. (this feels like it'd cause problems, but I'm not sure)

Thanks for reading! That's as far as I've gotten.

entzmingerc:
Midi differentiation seems like it might help a bit, with norns connected I've got din, COMPUTER 1, and COMPUTER 2 showing up as midi devices. I think norns is COMPUTER 1, but I don't know for sure. Here's an updated deluge device type file for midigrid with a commented out section for sending midi CC messages instead of using note_on. 
[deluge TEXT.txt](https://github.com/SynthstromAudible/DelugeFirmware/files/12393209/deluge.TEXT.txt)

Could use a different message like poly aftertouch, I'm not sure which type of midi message interferes the least with other deluge operations. Just need one argument to be 0-127 and the other to have at least 16 values we can map to brightness levels.

PaulFreund:
I now understand the problem, learning the track means all input is mirrored to the output of the MINI channel. I'll think about how this can be solved in an elegant way

PR #252, opened 2023-07-29
Universal DBT building via CMake

stellar-aria:
This migrates e2 from using the internal toolchain to the DBT provided toolchain. It also migrates the underlying build system from SCons to CMake, which should make our lives a little bit easier in the long run.

## User facing changes
A whole bunch of new `dbt` commands! Oh boy!
- `dbt build` now has expanded argument details and does not proxy to SCons.
  - `-m` enables metadata tagging on every output file. Now we'll know when/where a firmware .bin/.elf comes from!
- `dbt nuke` for absolutely, positively, erasing your build folder(s) from existence, beyond the scope of `dbt clean`.
- `dbt configure` for manually configuring CMake if you're into that sort of thing (but honestly it gets called by `dbt build` if the build folder isn't found anyways)
- ~~`dbt buildgen` for auto-magically regenerating the `CMakeLists.txt` that now live under `/src/*/`. Add a new `.cpp` file? Don't worry about editing the `CMakeLists.txt` by hand, just run `dbt buildgen`!~~

## Under the hood changes
CMake/Ninja instead of SCons. SCons is nice, but CMake is the standard for C/C++ projects, and will make any external library integrations (such as [fmt](https://github.com/fmtlib/fmt)) that much easier. It's known by most C/C++ developers, and modern CMake fairly easy to understand and parse.  It is _very_ well documented, making finding variables and settings via a web search easy.

Ninja has proper output interleaving, which also enables us to add `problemMatcher` support to VSCode, and simply allow e2's GCC problem regex matcher to work as well. It's simple, it's fast (automatic multi-processor support), and it supports multi-config builds.

There is now a global `compile_commands.json`, under `/build` where most editors expect it.

## What does this mean
It means we now have a single target toolchain and a single build configuration, for every build, on every platform. We are now exclusively using GCC 12.1, which also will allow us to finally officially move to C++20, with such niceties as `consteval`, ranges, and the constraint system.


CMakeLists.txt
```diff
@@ -0,0 +1,232 @@
...
+    # Warnings
+    # $<$<CONFIG:DEBUG>:-Wextra> # TODO: Enable when ready
+    # $<$<CONFIG:DEBUG>:-Wall> # TODO: Enable when ready
+    # $<$<CONFIG:DEBUG>:-Wpedantic> # TODO: Enable when ready
+    $<$<CONFIG:DEBUG>:-Wnull-dereference>
+    $<$<CONFIG:DEBUG>:-Wno-psabi>
+
+    # Stack usage
+    $<$<CONFIG:DEBUG>:-Wstack-usage=100>
+    $<$<CONFIG:RELEASE>:-Wstack-usage=1000>
```
bobtwinkles, line 181: is this the new hotness on how to conditionally add flags vs. `if (){}` and appending to the language/build specific flag lists?

README.md
```diff
@@ -68,44 +70,38 @@ DBT must always be run from the *root directory* of the repository. The remainde
...
-* `--spew` - This is shorthand for `--verbose=3` and `--nowo`, each described above.
+* `-S` or `--no-status` - this disables the `ninja` status line while building. Output from compilation commands (such as warnings and errors) will still be printed.
+* `-v` or `--verbose` - this prints greater level of detail to the console (i.e. exactly the compiler/linker commands called)
+* `-c` or `--clean-first` - clean before building 
 
-`clean` - Clean up all supported targets (dbt-build-release-oled, dbt-build-release-7seg, dbt-build-debug-oled, and dbt-build-debug-7seg)
-* Variables described above under `all` also apply to `clean`.
+#### CMake custom arguments
+
+Any additional arguments to CMake may be transparently passed via `dbt build`.
```
bobtwinkles, line 96: Document what these are (ENABLE_RTT and FORCE_COLORED_OUTPUT are the 2 I can see immediately) would probably be good.

bobtwinkles (commented):
Minor nits and we should have windows (DBT and E2), OSX, and Linux E2 (if there are any left?) developers test this branch locally to ensure we're not breaking those workflows.

CMakeLists.txt
stellar-aria: lol yes. Generator expressions are 10 years old (CMake 3.0) but have only gotten more powerful over time

README.md
stellar-aria: Those aren't arguments, they're configure options, and they can be listed with the typical `cmake -B build -LH` or edited with cmake-gui, and are internal use anyways. Arguments are things listed by `cmake --build build`, and the only relevant one I can see is the number of parallel jobs, which I'll add here.

m-m-adams:
Works for me on Fedora running E2

PR #253, opened 2023-07-29
Add Catch Notes feature to CommunityFeatures.md and update feature list.

weavermedia:
Added the Catch Notes feature to CommunityFeatures.md. Updated the feature list. Added missing link for MIDI Takeover Mode.

### DIFF
--- a/CommunityFeatures.md
+++ b/CommunityFeatures.md
@@ -16,7 +16,7 @@ Here is a list of general improvements that have been made ordered from newest t
 
 # Added features
 
-Here is a list of features that have been added to the firmware as a list ordered from newest to oldest:
+Here is a list of features that have been added to the firmware as a list, grouped by category:
 
 ## Synthesizer features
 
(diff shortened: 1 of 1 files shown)


sichtbeton:
Can this be set off by default? It interferes with normal operation of the section changes and would be good to have off by default.

weavermedia:
@sichtbeton I'm totally happy for this to be off by default. That was my first instinct since it changes current behavior. I asked the question (in the Discord I think) and there was mention of whether that would actually be considered a bug fix rather than a new feature.

If it interferes with section changes it should definitely be off by default. Have you confirmed this feature is interfering with section changing behavior? What behavior is being altered? Does turning off the feature return the section behavior back to normal?

All this feature does is adds a conditional around the code that attempts to play 'late notes' after changing clips. I'm not sure how it could cut off samples already playing, unless there's some way that carry-over notes are passed from one clip to the next that I haven't seen?

For reference I asked about the playback glitches on Facebook a couple of years ago and Rohan responded basically saying it's a feature not a bug: https://www.facebook.com/groups/DelugeBetaTesters/posts/1677334075806819/

sichtbeton:
> Reference

It came up when legato triggering sections with #251. it double triggered notes and when having long releases, the notes got cutoff.

Toggling this to off made it normal again.

weavermedia:
@sichtbeton What do you mean by 'legato triggering sections'?

sichtbeton:
its a new mode that was introduced within the new grid view (i linked the PR). but it also did this when doing a normal section change.

weavermedia:
@sichtbeton I can't reproduce with normal section changes. Can you give me more details or an example SONG file? or the samples that get cut off, what mode are they set to? ONCE/CUT/LOOP/STREtch?

sichtbeton:
Now when trying to replicate it, I fail too. This is strange.

weavermedia:
@sichtbeton OK, let me know if it comes up again. In the meantime I do personally think this feature should be off by default anyway, considering Rohan's reasoning in 2021. I'll submit a PR to update the default to OFF and ask for review.

sichtbeton:
> @sichtbeton OK, let me know if it comes up again. In the meantime I do personally think this feature should be off by default anyway, considering Rohan's reasoning in 2021. I'll submit a PR to update the default to OFF and ask for review.

sorry for stirring the pot here. I should be more careful when testing this stuff.

Issue #217, opened 2023-07-20
no scrolling on 7seg display


sichtbeton:
for now, the scrolling of the 7seg display is not working anymore. sure, it was distracting to repeat the scrolling forever.  but having not even one scroll initially and being forced to scoll manually makes the display even harder to read then with scrolling.

I suspect this behaviour was introduced in either of these PRs:

https://github.com/SynthstromAudible/DelugeFirmware/pull/176
https://github.com/SynthstromAudible/DelugeFirmware/pull/178

> I also disabled OLED text scrolling because i find it incredibly distracting. I simply commented out a line in View.cpp.  

Suggestion: 

- revert to the original scrolling by default
- have an option to scroll only once and then stop
- another option to never scroll
- additionally: scroll speed setting for slow / medium / fast

m-m-adams:
Confirmed - this should definitely be fixed, it makes 7seg extraordinarily difficult to navigate

m-m-adams:
Neither of those PRs is responsible, the change to view.cpp wasn't merged

sichtbeton:
> Neither of those PRs is responsible, the change to view.cpp wasn't merged 

This is weird. Then how come?

m-m-adams:
Who knows! It's a bug

stellar-aria:
This was done by Rohan, see:

m-m-adams:
So trying both, I think reverting this change by Rohan is an improvement for my own workflow but it does make #99 a significant issue for people who rely on the file postfix behaviour. Basically the behaviour should be:

- If we arrived at the file by typing:
    1. Show the typed in text and whatever afterwards without scrolling
- Else
    1. If the text does not end with _number, scroll normally
    2. If the test does end with _number, show the last 4 characters first but then scroll normally

I'm gonna work on a fix later today

Issue #101, opened 2023-06-24
Scrolling through Position of Song Title Inconsistency


ok-reza:
This is part of the pre-OLED to post-OLED firmware inconsistencies I've noticed after loading the open-source community firmware onto a 7-SEG display Deluge. (See #76, #97, #98, #99, #100)

### Scrolling through title position now displays keyboard.

Video showing the inconsistency: https://youtu.be/zBvPcSGLN04

- Scrolling through a title on 7SEG will then prompt the alphanumeric keyboard to pop up. This wasn't the case prior.

### Extent of issue
This is confirmed with:
[X] Folders
[X] Songs
[X] Kits (not an issue: alphanumeric keyboard will always be on display when loading a kit)
[X] Synths (not an issue: alphanumeric keyboard will always be on display when loading a kit)
[    ] Sample file browsing (does not display when scrolling)
[    ] Sample folder browsing (does not display when scrolling)

### Replicable on OLED?
It would be helpful to know if the behavior is the same on the OLED Deluge.

### Should this new behavior revert or stay?
If a user is scrolling through the title, it might mean they want to type something or delete from that point where they scrolled to to find a similar file. However, sometimes the user just wants to see the rest of the song title. By popping up the keyboard when scrolling, it adds to visual noise.

Issue #100, opened 2023-06-24
Searching for Non-Existent Songs/Folders Inconsistency


ok-reza:
This is part of the pre-OLED to post-OLED firmware inconsistencies I've noticed after loading the open-source community firmware onto a 7-SEG display Deluge. (See #76, #97, #98 #99, #101)

### Unable to search for non-existent songs/folders anymore

Video showing the inconsistency: https://youtu.be/PXD9Psp94UI

- Unable to type characters into keyboard if a project or folder doesn't start with them.
- Deluge points you to a song closest to the input but 'rounds down'.
(No songs starting with '6', you have a song starting with '4' and '7':
If you type 6 it will reveal song starting with '4'. It doesn't reveal song starting with '7' since it 'rounds down'.
- (Not in video) Inconsistent 'suggestions' based on last input.

### Extent of issue
This is confirmed with:
[X] Folders
[X] Songs
[X] Kits
[X] Synths
[X] Sample file browsing
[X] Sample folder browsing

### Replicable on OLED?
Can you enter non-existing song/folder names when searching via keyboard after hitting Load? OLED displays 3 items at a time, does it round down based on what you give it?

### Should this new behavior revert or stay?
I prefer to know if a song I thought existed doesn't exist on my SD card rather than inputting characters which then won't display on screen. However maybe this behavior has value since it will always point you to somewhere?

Issue #97, opened 2023-06-24
Kit/alphanumeric Order Inconsistencies


ok-reza:
This is part of the pre-OLED to post-OLED firmware inconsistencies I've noticed after loading the open-source community  firmware onto a 7-SEG display Deluge. (See #76, #98, #99, #100, #101)

### Kit/alphanumeric order inconsistencies and crashes.

Video showing the problem: https://www.youtube.com/watch?v=iq0kG6glywE

- The alphanumeric ordering seems to be changed

1. Expected: A1, A2, A3, B1, B2, B3, C1, etc.
2. Observed: A1, B1, C1, loops through a shortened list of kits, A2, B2, C2...inconsistent and unpredictable.

- A crash can occur when messing with the improper order found in the community firmware.

1. Certain kits aren't showing up at all, but can be revealed when scrolling backwards then forwards.
2. Doing so enough times causes a crash.

### Replicable on OLED?

It would be helpful to know if this issue persists on an OLED Deluge. I would suggest adding kits A1, A2, A3 through H1, H2, H3 yourself and sharing the behavior.

litui:
Thanks for the push @rohanhill!  Slight slap on the wrist due for not using a Pull Request 😉, but I appreciate your attention to these issues. 🙇‍♀️

PR #254, opened 2023-07-30
Improve 7seg text scroll and edit behaviour

### LINKED ISSUE #98: Saving New Iteration of Song Inconsistency
This is part of the pre-OLED to post-OLED firmware inconsistencies I've noticed after loading the open-source community  firmware onto a 7-SEG display Deluge. (See #76, #97, #99, #100, #101)

### Saving new iterations of songs inconsistencies

Video showing the problem: https://youtu.be/Vx7kILI1NwQ

- The iteration will now generally add a '_1' to the end of the title instead of ' 2' for the first iteration. 
(However, you can sometimes experience it adding a '_2' at the end instead by loading/reloading/trying to save again as show around :52 of video).
- Sometimes after hitting save, the display won't skip to the end of the title as expected and just show the start of the title (7-SEG display, title would have to be greater than 4 characters)
- Underscores are present between song title and iteration number, wasn't present before unless currently scrolled to that space's position.

### Extent of issue
This is confirmed with:

[X] Song iterations
[?] Synth iterations
[?] Kit iterations
[?] Folder iterations

### Replicable on OLED?
It would be helpful to know if the behavior is the same on the OLED Deluge. Does the OLED scroll to the end of a long title title that extends beyond the display? What number does it start with for the first iteration? Do underscores appear between title and the iteration number? etc

### LINKED ISSUE #99: Browsing Folders with Different Iterations Inconsistency
This is part of the pre-OLED to post-OLED firmware inconsistencies I've noticed after loading the open-source community firmware onto a 7-SEG display Deluge. (See #76, #97, #98, #100, #101)

### Browsing through folders with different iterations 

Video showing the problem: https://youtu.be/F6mVQvCr_nc

- Display doesn't skip to the end of the title anymore to reveal which iteration (When title is greater than 4 characters). Instead we see the same start of the title when scrolling through different versions.

### Extent of issue
This is confirmed with:
[X] Folders
[X] Songs
[X] Kits
[X] Synths
[?] Sample file browsing
[?] Sample folder browsing

### Replicable on OLED?
It would be helpful to know if the behavior is the same on the OLED Deluge. Does the OLED scroll to the end of a long title title that extends beyond the display?

m-m-adams:
Changes the 7seg browsing UI to the following logic:
- If we arrived at the file by typing:
    - Show the typed in text and the deluge prediction without scroling
- Else
    - Check first different character from previous filename
    - If the first difference is less than 3:
        - Scroll from beginning
    - Else:
        - Scroll from first difference -2

This fixes #99, #217, and #101 

Sets the save UI edit position to zero after setting scrolling text, allowing the automatically added _number postfix to appear first. Typing at this point still starts from nothing instead of appending to the end.

This fixes #98. 

Additionally #97 and #100 were fixed by @rohanhill's commit 4500701

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -1390,7 +1390,6 @@ void Browser::displayText(bool blinkImmediately) {
 	renderUIsForOled();
 #else
 	if (arrivedAtFileByTyping) {
-doQWERTYDisplay:
 		QwertyUI::displayText(blinkImmediately);
 	}
 	else {
@@ -1412,13 +1411,12 @@ doQWERTYDisplay:
 
 			else {
 nonNumeric:
-				goto doQWERTYDisplay; // Abandon the below for now.
 				numberEditPos = -1;
 				if (qwertyVisible) {
-					goto doQWERTYDisplay;
+					QwertyUI::displayText(blinkImmediately);
 				}
 				else {
-					scrollingText = numericDriver.setScrollingText(enteredText.get(), numCharsInPrefix);
+					scrollingText = numericDriver.setScrollingText(enteredText.get(), enteredTextEditPos);
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


sichtbeton:
This is awesome. One thing: could there be a "Scroll one option"? It's kind of distracting to have it scroll endlessly but helpful to once.

m-m-adams:
Yeah start it as a discussion feature request, it's probably possible but it's not obvious how to do that. This pr is just aimed at fixing bugs

PR #267, opened 2023-07-31
Save empty song UI

### LINKED ISSUE #240: When a song is created by just booting your Deluge, if you want to save it, folders don't show up, just the root directory and the songs contained there
On the other hand, if you load an existing song or a template song that you may have, when you click Save, you can see the folders to save the song inside a folder

m-m-adams:
When a song is saved with no default name available folder content wasn't loaded properly in the browser unless text is entered with the qwerty keyboard. This fixes that codepath by including subfolders in the folder read. 

Closes  #240


soymonitus:
Thanks for looking into this!

jamiefaye:
Are the CMake changes also to go in?

PR #268, opened 2023-07-31
Turn the GeneralMemoryAllocator into a singleton

stellar-aria:
Unfortunately, with the state of the heaps and the way global variable initialization order is undefined behavior, the only way to guarantee the GMA is up and running the first time it's accessed by any global variables is to make it a singleton.

This does that, though I expect this to eventually be swapped out when the memory allocator overhaul takes place.

### DIFF
--- a/src/deluge/memory/general_memory_allocator.cpp
+++ b/src/deluge/memory/general_memory_allocator.cpp
@@ -35,8 +35,6 @@ char emptySpacesMemoryInternal[sizeof(EmptySpaceRecord) * 1024];
 extern uint32_t __heap_start;
 extern uint32_t __heap_end;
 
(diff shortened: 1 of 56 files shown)


src/deluge/memory/general_memory_allocator.cpp
PaulFreund, line 1: general_memory_allocator.cpp#38 still contains a global object instance:
```C++
GeneralMemoryAllocator generalMemoryAllocator{};
```
Not exactly sure what effects this will have but if it's constructor is executed after the static instance in the singleton ::get() it might overwrite some previous setup.

stellar-aria: Fixed. Thank you so much!

PaulFreund (approved):

PR #273, opened 2023-07-31
Format task - mode to just check instead of write

bobtwinkles:
i found the incantation to make this work

cc @stellar-aria


stellar-aria (approved):
Looks pretty straightforward to me :D

Issue #274, opened 2023-07-31
Align master compressor settings with Deluge UX


PaulFreund:
Currently master compressor has two pahts for the settings which are both shortcuts.

Disclaimer: For grid style session view #251 I would like to claim Shift+Pad for instant change of playback state.

On the other hand the current display menu for the compressor parameters is a departure from current look and feel of the OLED screen (while being gorgeous of course). It also masks Affect Entire Sidechain although I don't think anybody needs it. 

My current suggestion would be:
* In Session and Arranger mode while Affect Entire is off non of the FX controls are used if no clip is held
* With Affect Entire disabled use the three rightmost FX buttons together with the gold knobs similar to all the other FX parameters
* Remove the fancy menu but show the currently edited parameter in the display while turning gold knob similar to synth parameters
* Remove Shift+Pad shortcuts. They could stay in Row display (won't work in Grid display) but the manual will get more complicated (no strong opinion here though)

m-m-adams:
Another possibility is just improving the deluge sidechain instead of maintaining two compressors. A possible approach is:

1. Feed sidechain compressor with each clips RMS level post-fx, calculated when it's copied to the song buffer 

2. Enable sidechain from synth clips and audio clips 

3. Change song level affect entire sidechain knob to increase send level and sidechain->level amount (e.g make it a one knob compressor)

weavermedia:
@m-m-adams I like having a 'proper' compressor with independent controls for threshold, attack, release, ratio and makeup gain. Feels like a step backwards to take those away now.

It's a nice bonus to have the mix control too, for parallel compression.

m-m-adams:
It could stay in the community feature menu or remain editable in more detail via the menu system. I really feel like it's a poor fit for the deluge UI by default, especially since it's exclusively a master effect - you can't send it different levels per track. 

Technically the only thing you would lose in my suggestion is makeup gain - the sidechain envelope already has controls for threshold (shape), attack, release, and ratio (vol duck). I could add auto makeup gain and then the level can be fine tuned with the affect all level

weavermedia:
I'd be very happy for the master compressor editing to take place in the menu system, especially if this also brought about editing access to all Affect Entire settings in the menu system.

Definitely interested in hearing an improved sidechain as you described but I'm not convinced it would sound as good as the master compressor.

[...]

PR #275, opened 2023-07-31
[Draft] Show value (expressed in the range 0-127) of golden knobs (anywhere: affect-entire song mode, synths, midi clips..)

soymonitus:
Useful to get a quick and dirty reference of where the hell are you. Besides, it matches the value that you would set if you midi learn that golden know to a Midi CC. A must for setting delay amount just max at value 63 (64 will have feedback till you get deaf…). Good also for changing global song-level affect-entire golden knobs which you can’t see the values they have anywhere.

https://youtu.be/npixC97l-uw


PaulFreund:
Hey really love this! I know numericDriver is really easy but any chance we use the same display as editing synth parameters here? It would also show you what is edited. Your video also matches what I propose in #274

soymonitus:
> Hey really love this! I know numericDriver is really easy but any chance we use the same display as editing synth parameters here? It would also show you what is edited. Your video also matches what I propose in #274

I found that getting the real value was quite hidden and that the modEncodetAction only has information about the knob position and the converted int32 value (which has nothing to do with the range 0-50 seen in some parameters.
So i though that maybe to simplify code and not build a custom function for every different parameter in every different screen (sounds like a lot of work), the possibility to show just the encoder value in its range (which is -64 to 64, which is converted to 0-127 for readability and similarity with Midi CC values) . In fact, what I only did was move the code that showed midi cc for midi clips to the parent so it works also for every other clip type.

soymonitus:
I still need to fix that when you are editing a parameter, for example Pan, and it appears in the screen with values -32 to 32, if you turn the corresponding golden knob it shows the popup over the real value, and it should not show the popup if the real value is behind. Maybe I need to take a look to a more complex implementation where it takes into account the parameter being edited and shows the same UI as when you do shift+shorcut pad

seangoodvibes:
Damn this is awesome and coincidentally in line with exactly what im doing in my automation lanes PR. i am showing 0-127 any time you turn gold knobs.

soymonitus:
I think so too, showing the actual value and parameter name needs lots of coding

soymonitus:
Maybe someone could help me add to the popup the name of the parameter (only for OLED, as a two lines popup, and just the value for 7seg)

seangoodvibes:
@soymonitus I looked at your code. I don't think this is the right way to implement this change.

The function you modified is used everywhere in the codebase to do calculations of the current parameter's knob position/parameter value. 

This will have unwanted consequences across the codebase any time this calculation function is called. We will get popups like crazy.

I don't recommend merging this until further thought is taken as to how to make such a change. 

Sorry this has wider implications!

soymonitus:
I will try to address it in a different way. Thanks for the feedback

seangoodvibes:
@soymonitus could you also change your PR to draft until it is ready to be merged :)? Thanks!

soymonitus:
Can someone help pointing me in the right direction. I am in the modEncoderAction function in view.cpp and I feel the code should be there but I need to find which parameter is the golden knob editing, if there is an object I can access to know the printable name and the value range

seangoodvibes:
> Can someone help pointing me in the right direction. I am in the modEncoderAction function in view.cpp and I feel the code should be there but I need to find which parameter is the golden knob editing, if there is an object I can access to know the printable name and the value range

Hey so I'm not sure about the value range, but this is one function I use get to the parameter display name, but its only for Patched Params..(there's unpatched ones too...)

char const* getPatchedParamDisplayNameForOled(int p) {

You have to pass it a Parameter ID

Midi would be a bit different since you have to pass it the Midi CC control number, and this function helps draw what Midi CC you're editing:

void InstrumentClipMinder::drawMIDIControlNumber(int controlNumber, bool automationExists) {

---

I think view.modEncoderAction is the right function, so you're on the right track

newKnobPos + 64 is going to hold the value you want to display

---

To get the ParamID/controlNumber...

There are a bunch of functions to get the paramID from the knob, I'm not exactly sure which one you'll need, since I know you can also learn params to those knobs which makes thing tricker. And there's different types of params (patched vs unpatched) And if you're in a Midi Clip it's different also.

These are some I found:

int GlobalEffectableForClip::getParameterFromKnob(int whichModEncoder) {
int GlobalEffectable::getParameterFromKnob(int whichModEncoder) {

Sorry I can't be anymore help right now.

Hope this guides you in the right direction!!

soymonitus:
> > Can someone help pointing me in the right direction. I am in the modEncoderAction function in view.cpp and I feel the code should be there but I need to find which parameter is the golden knob editing, if there is an object I can access to know the printable name and the value range
> 
> Hey so I'm not sure about the value range, but this is one function I use get to the parameter display name, but its only for Patched Params..(there's unpatched ones too...)
> 
> char const* getPatchedParamDisplayNameForOled(int p) {
> 
> You have to pass it a Parameter ID
> 
> Midi would be a bit different since you have to pass it the Midi CC control number, and this function helps draw what Midi CC you're editing:
> 
> void InstrumentClipMinder::drawMIDIControlNumber(int controlNumber, bool automationExists) {
> 
> I think view.modEncoderAction is the right function, so you're on the right track
> 
> newKnobPos + 64 is going to hold the value you want to display
> 
> To get the ParamID/controlNumber...
> 
> There are a bunch of functions to get the paramID from the knob, I'm not exactly sure which one you'll need, since I know you can also learn params to those knobs which makes thing tricker. And there's different types of params (patched vs unpatched) And if you're in a Midi Clip it's different also.
> 
> These are some I found:
> 
> int GlobalEffectableForClip::getParameterFromKnob(int whichModEncoder) { int GlobalEffectable::getParameterFromKnob(int whichModEncoder) {
> 
> Sorry I can't be anymore help right now.
> 
> Hope this guides you in the right direction!!

This is superhelpful. Thanks a lot for taking the time to answer

m-m-adams:
I don't really use the delay but does it start to feedback continuously past 63, and still have adjustments up until 127? That seems like maybe the range should be tweaked lol

m-m-adams:
I just want to leave a mention that this was an intentional design decision on the deluge, it's talked about in Rohan's interview in the book push, turn, move. The idea was that you would just see the LED bars and not need to look at the screen.


If the exact value makes that big a difference for the delay maybe it should be set to flash at 63 like pan and eq do?

sichtbeton:
> I just want to leave a mention that this was an intentional design decision on the deluge, it's talked about in Rohan's interview in the book push, turn, move. The idea was that you would just see the LED bars and not need to look at the screen.
> 
> If the exact value makes that big a difference for the delay maybe it should be set to flash at 63 like pan and eq do?

good idea. also maybe having an option to remove this (feedback part) would be good. it can get annoying pretty fast and can destroy a live set. so having this optional would not only prevent that but also provide a more detailed delay dialing mechanism (much larger knob reach).

seangoodvibes:
Just wanted to mention:

The knob position range is actually 0-128 (-64 to +64 internally)

soymonitus:
Closing as I will not work on this for a while

PR #278, opened 2023-08-01
Standard types, int32_t, and min/max

stellar-aria:
This PR covers three relatively minor whole-project refactors:

- eliminate the use of r_typedefs.h outside of RZA1, and make it use standard headers instead of custom typedefs.
- convert all `int` in the deluge/ source tree to `int32_t` (not actually changing the representation). There's already a very mixed usage of this, presumably because Rohan started out with `int`and later switched, this just enforces the expectation that all `int`s are actually `int32_t` for ARM32, in a way that makes the code more portable. 
- swap getMin/getMax for the std::min, std::max, and (where possible) std::clamp functions. These have the benefit of being strongly typed and enforcing type safety.

Extra:
- added typed integer literal suffixes (`_u8` through `_i64`) to util/misc.h


bobtwinkles (approved):
LGTM, assuming the int to int32_t conversion was done mechanically

stellar-aria:
> LGTM, assuming the int to int32_t conversion was done mechanically

It was, yeah, standard find/replace

src/deluge/gui/ui/keyboard/layout/in_key.h
```diff
@@ -71,7 +71,7 @@ class KeyboardLayoutInKey : public KeyboardLayout {
 		}
 		int32_t octave = (((note + kOctaveSize) - rootNote) / kOctaveSize) - 1;
 		// Make sure we don't go into negative because our root note is lower than C-2
-		return getMax(0, octave * scaleNoteCount + padScaleOffset);
+		return std::max<uint16_t>(0, octave * scaleNoteCount + padScaleOffset);
```
PaulFreund, line 74: Unfortunately using std::max<uint16_t> will return USHRT_MAX instead of 0

PR #282, opened 2023-08-02
Alternative golden knob delay params (ability to change encoder click behavior)

soymonitus:
Instead of changing PingPong On/Off and Type Analog/Digital, they will change SyncType and SyncLevel.
I have always preferred dotted 8th delays over the even 16th delays, probably because of the music genre. But the global effects from the Deluge (when affect-entire engaged in Kits and Song) default to Even 16th delays.
This PR allows users to change what parameters you change when clicking the golden knob encoders. The default behavior is changing Pingpong/Normal (upper knob) and Analog/Digital (lower knob), but with this "alternative" params mode, the upper and lower knobs will toggle between SyncType and SyncLevel values.

DEMO: https://youtu.be/GbhAAqT3_Bo

### DIFF
--- a/Documentation/community_features.md
+++ b/Documentation/community_features.md
@@ -73,6 +73,9 @@ Synchronization modes accessible through the "LFO SYNC" shortcut.
 ### Catch Notes
  - ([#221]) The normal behavior of the Deluge is to try to keep up with 'in progress' notes when instant switching between clips by playing them late. However this leads to glitches with drum clips and other percussive sounds. Changing this setting to OFF will prevent this behavior and *not* try to keep up with those notes, leading to smoother instant switching between clips.
 
+### Alternative Delay Params for golden knobs
+ - ([#281]) Ability to select, using a Community Features Menu, which parameters are controlled when you click the Delay-related golden knobs. The default (for upper and lower knobs) is PingPong On/Off and Type (Digital/Analog), and you can modify it so the knob clicks change the Sync Type (Even, Triplets, Even) and SyncLevel (Off, Whole, 2nd, 4th...) respectively.
+
 <h1 id="runtime-features">Runtime settings aka Community Features Menu</h1>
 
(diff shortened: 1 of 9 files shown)


soymonitus:
Good news is that if you ever need to change the all 4 paramaters for the affect-entire delay, for example, you usually change the SyncType and SyncLevel, so you have always the community feature toggle set to On, but one day you need to disable PingPong for a some kit in particular or for the song's delay, then, you can go to the menu, toggle the feature Off, adjust the delay pingpong for your kit or song, and go to the menu and toggle again to your most used params setting, as pingpong setting for that kit will remain as you set it, it's not dependent on the community feature toggle

PaulFreund:
@soymonitus Sorry to say but this PR changes default behavior in conflict with the silkscreen and it is default on, I think it can be added in general but as per comunity guidelines it would have needed to get a positive review first.

@jamiefaye  Can you please revert it and wait with new merges until they have a positive review? I think Bob sent you a mail outlining how to enable this as a merge requirement

soymonitus:
> @soymonitus Sorry to say but this PR changes default behavior in conflict with the silkscreen and it is default on, I think it can be added in general but as per comunity guidelines it would have needed to get a positive review first.
> 
> 
> 
> @jamiefaye  Can you please revert it and wait with new merges until they have a positive review? I think Bob sent you a mail outlining how to enable this as a merge requirement

Ey Paul, would if be enough to create a follow up PR to make the default community feature option to Off? Is that what you are saying?

soymonitus:
Follow-up fix here @PaulFreund  @jamiefaye  Is this what is needed to make PR 282 adhere to the guidelines?
https://github.com/SynthstromAudible/DelugeFirmware/pull/290

jamiefaye:
It went in OK according to my understanding.

soymonitus:
> It went in OK according to my understanding.

The functionality is well programmed and I tested it thoroughly, but I made the mistake to make it On by default when it changes the default behavior of golden knobs for Delay, when the guidelines seem to suggest to not set a community feature to On by default if it changes main main functionality

PaulFreund:
I reviewed the rest of the code and it is okay. The only thing missing according to guidelines was a review. The follow up PR is okay to fix it so we don't need to revert it if merged directly

PR #293, opened 2023-08-03
Feature: Waveform loop lock

weavermedia:
(Reimplementation of closed PR #214 with hopefully cleaner code)

Allows you to lock the sample loop points movements together in the waveform view.

1. Set sample LOOP START and LOOP END as usual
2. Hold LOOP START and tap LOOP END to engage loop lock [LOCK]
3. Move either loop point and they're locked together
4. SHIFT+TURN◀︎▶︎ clockwise to double the loop length [DOUB]
5. SHIFT+TURN◀︎▶︎ anti-clockwise to half the loop length [HALF]
6. Hold LOOP START and LOOP END again to release loop lock [FREE]

Usage notes
- You can't double the length past the sample end point [CANT]
- You can't half the length to less than 2 samples [CANT]
- You can't scroll a loop outside of the sample start/end bounds

Todo
- [x] Set loopLocked by holding loopStart marker and tapping loopEnd marker
- [x] Allow doubling and halving the loop length with SHIFT+TURN◀︎▶︎
- [x] Prevent scrolling the loop outside the sample start and end points
- [x] Prevent doubling the loop longer than the sample end point
- [x] Prevent halving the loop to 0

Questions for the community
- ✅ ~~What's the Deluge UX way of using SCROLL ▲▼ for loop doubling/halving?~~ SHIFT+TURN◀︎▶︎
- Should the loopLocked boolean be a permanent sample attribute or just live in the sample marker editor file?
- ✅ ~~Does this need a runtime feature setting?~~ No

Note the video shows outdated controls. Text here is up to date.

https://github.com/SynthstromAudible/DelugeFirmware/assets/3027787/0f7ca1a1-c63a-4601-aebb-51c016b8fc12


m-m-adams:
To maintain consistency with clip view I think doubling should go on shift+press <> and changing loop length should be on shift+turn <>. Neither of those are used in waveform edit view right now

weavermedia:
> To maintain consistency with clip view I think doubling should go on shift+press <> and changing loop length should be on shift+turn <>. Neither of those are used in waveform edit view right now

I moved doubling and halving the loop length to SHIFT+TURN◀︎▶︎  and returned the vertical encoder action back to default.

I'm not sure about using SHIFT+PRESS◀︎▶︎ for doubling the loop length. It is indeed consistent with the clip view action but since there's no clip view equivalent for halving the length it might be confusing.

Do you think changing loop length is OK just on SHIFT+TURN◀︎▶︎?

soymonitus:
This feature is awesome!

m-m-adams:
I think just turning to adjust is fine, having the double there but not a half also seems fine to me since it's like that for clips anyway

weavermedia:
@m-m-adams Do you think this feature needs a runtime setting? It doesn't change any default behavior, just adds some new behavior. It will be documented, if merged, but not sure it needs a runtime feature setting.

m-m-adams:
Nope! Just extending existing behaviour in a way that complements the existing UI

Awesome feature

weavermedia:
Documentation added to Community Features doc. I think is ready to go now. Can I get a review @m-m-adams?

src/deluge/gui/ui/sample_marker_editor.cpp
```diff
@@ -142,10 +145,28 @@ void SampleMarkerEditor::writeValue(uint32_t value, MarkerType markerTypeNow) {
 		getCurrentSampleHolder()->startPos = value;
 	}
 	else if (markerTypeNow == MarkerType::LOOP_START) {
-		getCurrentMultisampleRange()->sampleHolder.loopStartPos = value;
+		if (loopLocked) {
+			int intendedLoopEndPos = value + loopLength;
+			if (intendedLoopEndPos < getCurrentSampleHolder()->endPos) {
```
m-m-adams, line 150: What happens if the intended end pos is equal to the end of the sample? I think that's possible but unhandled

src/deluge/gui/ui/sample_marker_editor.cpp
```diff
@@ -142,10 +145,28 @@ void SampleMarkerEditor::writeValue(uint32_t value, MarkerType markerTypeNow) {
...
+		}
+		else {
+			getCurrentMultisampleRange()->sampleHolder.loopStartPos = value;
+		}
 	}
 	else if (markerTypeNow == MarkerType::LOOP_END) {
-		getCurrentMultisampleRange()->sampleHolder.loopEndPos = value;
+		if (loopLocked) {
+			int intendedLoopStartPos = value - loopLength;
+			if (intendedLoopStartPos > getCurrentSampleHolder()->startPos) {
```
m-m-adams, line 162: Same here

src/deluge/gui/ui/sample_marker_editor.cpp
```diff
@@ -623,6 +659,39 @@ static const uint32_t zoomUIModes[] = {UI_MODE_HOLDING_HORIZONTAL_ENCODER_BUTTON
 
 ActionResult SampleMarkerEditor::horizontalEncoderAction(int32_t offset) {
 
+	if (loopLocked && Buttons::isShiftButtonPressed()) {
+		if (offset > 0) { // turn clockwise
+			int end = getCurrentSampleHolder()->endPos;
+			int loopEnd = getCurrentMultisampleRange()->sampleHolder.loopEndPos;
+
+			if (loopEnd + loopLength < end) {
+				loopLength = loopLength * 2;
```
m-m-adams, line 668: is this right? This reads like it's doubling/halfing the loop length when you turn the knob instead of incrementing it which is not how I read the conversation

Documentation/community_features.md
```diff
@@ -67,6 +67,9 @@ Synchronization modes accessible through the "LFO SYNC" shortcut.
 ### Audio Clip View
  - ([#141]) Holding the vertical encoder down while turning the horizontal encoder will shift the clip along the underlying audio file, similar to the same interface for instrument clips.
 
+### Sample Waveform View
+ - ([#293]) When a sample has loop start and loop end points set, holding down loop start and tapping loop end will lock the loop points together. Moving one will move the other, keeping them the same distance apart. Use the same process to unlock the loop points. Use SHIFT+TURN<> to double or half the loop length.
```
m-m-adams, line 71: I think it should be shift+turn to edit loop length, shift+press to double. It feels weird having this little control over length if the loop is longer than a few samples

m-m-adams (changes requested):

m-m-adams:
Mostly good, but doubling and halfing by turning the knob feels wrong with longer loops where that adjustment range is really coarse. It also doesn't match the clip loop length adjustment. I think it should adjust length by one pad at the current zoom level, and if doubling length feels necessary for quick adjustments then put that on shift+push <>. 

Trying to move the loop by selecting a new start/end pad that would put loop end past sample end/start also feels weird. I think that should temporarily shorten the loop by truncating at the start/end position without changing the loop length

src/deluge/gui/ui/sample_marker_editor.cpp
weavermedia: I wasn't sure if the loop points could extend as far as sample star and end. I'll update so they can and test.

src/deluge/gui/ui/sample_marker_editor.cpp
weavermedia: Yes, I thought we might be at different viewpoints here. The intention is that you'll set a loop length related to the tempo of the sample or track then doubling or halving the loop length will keep it rhythmically intact. Being able to arbitrarily adjust the loop length with SSHIFT+TURN<> would be fine but I'd still want a want to exactly double or half it. It's a magic way to find cool and unusual looping sections within samples.

Documentation/community_features.md
weavermedia: See my response above about doubling/halving.

weavermedia:
The intention behind the feature is that you would set a loop length which is related tot he tempo of the sample, or the song, and then move the loop around to find cool looping sections. Doubling and halving the loop length keep the loop rhythmically intact so both of those features are crucial.

As such I wouldn't want to change the loop length without user intent either. If I got a good loop going then the loop length changed because I tried to move the loop out of bounds it would kill the flow. I'd rather just see a CANT message.

We can use SHIFT+TURN<> for continuous loop length adjustment and SHIFT+PUSH+TURN<> for doubling/halving.

m-m-adams:
Ahh ok that makes more sense to me then, I was playing with it more like a granular synth where the loop length wouldn't be linked to tempo (also made me really wish we could modulate start position!)

What about keeping your current doubling/halfing on the turn but allowing the start/end position to be moved freely by holding the marker and turning <> (like moving a clip in the arranger)? That would keep the tempo matching you want while also allowing the length of a loop to be adjusted more finely if needed

sichtbeton:
could this get in before the proposed feature cutoff this week for a comunity release? it looks ready but no pressure. it's a very cool feature.

src/deluge/gui/ui/sample_marker_editor.cpp
weavermedia: Addressed in https://github.com/SynthstromAudible/DelugeFirmware/pull/293/commits/1b6dc94235e14b644c91bfe5d96965dab55fe28e

src/deluge/gui/ui/sample_marker_editor.cpp
weavermedia: Addressed in https://github.com/SynthstromAudible/DelugeFirmware/pull/293/commits/1b6dc94235e14b644c91bfe5d96965dab55fe28e

weavermedia:
@sichtbeton Still trying to figure out how to assign the controls for fine-adjusting the loop length and doubling/halving the loop length. Any input is appreciated 😀

sichtbeton:
> @sichtbeton Still trying to figure out how to assign the controls for fine-adjusting the loop length and doubling/halving the loop length. Any input is appreciated 😀

turn <> to shift the loop
shift and turn <> to double or halven (analog to lenght of a clip)

or does this collide in any way with other functions?

m-m-adams:
Looking back at this I think fine adjustments could be a separate PR. Unlocking the loop, adjusting one end, and relocking is ok for now, we should just add this as a separate issue for our pre release bug fixes

m-m-adams (approved):
Future work: fine editing

@bobtwinkles suggestion to move the global to class member would be nice but there's only one loop editor at a time

The long term will likely involve moving that parameter into the sample synth as part of a bigger change to allow modulating the start/end/loop positions so I'm not particularly concerned about it's current location

weavermedia:
@m-m-adams What about the doubling and halving of the loop length? Would you rather I remove it from this PR and add as another follow up? Or leave it in here as-is? Currently using SHIIFT+TURN<> as in the first post.

weavermedia:
I'll fix the merge conflicts today.

m-m-adams:
The doubling and halting is fine, it's analogous to changing clip length with the same controls and just locked to sizes that are convenient for loop samples

PR #294, opened 2023-08-03
Switch to format_string, decrease fmtlib compiled size

stellar-aria:
This changes all title_format_str references to instead be of type fmt::format_string (instead of string), and adds two compile-time optimizations to bring the size of the static libfmt down (-Oz and FMT_STATIC_THOUSANDS_SEPARATOR)

(Accidentally closed this due to a branch rename)


stellar-aria:
yeah, I think limiting the number of template instantiations is reasonable, although if we find we're using it frequently, it might actually drop our size to switch back to `string` or `string_view`. I don't really see that being a _huge_ issue.

bobtwinkles (approved):
Confirmed size reduction and smoke tested some of the affected menus on OLED, LGTM

PR #298, opened 2023-08-03
[Draft] Convert kit's midi row into midi clips, so we can make use of arpeggiator and midi CCs

soymonitus:
Midi rows in kits currently are a bit dumb, they just trigger basic midi notes. I have always found myself reaching to the Shift + Arp mode to enable arp and discovered it is not possible.

Tasks

- [ ] Change midi kit rows to be internally a midi clip, unblocking them to be able to edit Arp settings (and capable also of all the other midi clip features like midi CCs for golden knobs)
- [ ] Implement the interface (button shortcuts) to edit Midi CCs on midi rows

### DIFF
--- a/src/deluge/model/drum/midi_drum.cpp
+++ b/src/deluge/model/drum/midi_drum.cpp
@@ -26,6 +26,8 @@ extern "C" {
 #include "util/cfunctions.h"
 }
 
+// Start here...
+
 MIDIDrum::MIDIDrum() : NonAudioDrum(DrumType::MIDI) {
 	channel = 0;
 	note = 0;


m-m-adams:
Semantically the right approach might be to make a kit midi row a container for a midi clip, that way kit midi won't get left behind in any future enhancements to midi clips in general.

That would make it easy to include CCs for #232 as well

soymonitus:
> Semantically the right approach might be to make a kit midi row a container for a midi clip, that way kit midi won't get left behind in any future enhancements to midi clips in general.
> 
> That would make it easy to include CCs for #232 as well

Wow, good point! because I also had in my TO DO list implementing Midi CC's for kits. Maybe I can work on this too as part os this draft

soymonitus:
Closing this as the scope seems way larger than what my current experience with the code can achieve. I will work on this in the future if nobody has already worked on it :)

soymonitus:
Closing this as the scope seems way larger than what my current experience with the code can achieve. I will work on this in the future if nobody has already worked on it :)

PR #299, opened 2023-08-04
ub: Remove illegal FAM from ParamSet

bobtwinkles:
ub: Remove illegal FAM from ParamSet

Flexible Array Members are not legal in C++ and their use here has already caused at least one "miscompilation" (see #230). This change replaces it with regular arrays in the children, with a pointer passed up to the parent. This costs about 1k extra bytes, but that seems likely to be code that was being compiled out when it shouldn't be.

Baseline (5edc967798635b7c0bf9f5e2c54b216f837ee9f0): 1540416 build/Release/delugeOLED.bin
With this change   : 1541344 build/Release/delugeOLED.bin

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -4272,7 +4272,7 @@ doNormal: // Wrap it back to the start.
 	    modelStack->addOtherTwoThingsAutomaticallyGivenNoteRow()->addParamCollection(mpeParams, mpeParamsSummary);
 
 	for (int32_t m = 0; m < kNumExpressionDimensions; m++) {
-		AutoParam* param = (m == 0 ? &mpeParams->params[m] : &mpeParams->fakeParams[m - 1]);
+		AutoParam* param = &mpeParams->params[m];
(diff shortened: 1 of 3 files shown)


m-m-adams:
This seems to have a minor performance hit. Using a test synth with some modulations setup I get an average of 45 voices on this branch vs 47 otherwise. I suspect it's the calls to get_whatever_paramSet in voice::render which get called 9 times per voice per buffer

m-m-adams:
Otherwise the implementation looks good and it does not have regressions on the MPE bug

m-m-adams (approved):
Good to go - performance is slightly worse in my first check but seems to be better in other cases so it's a wash

bobtwinkles:
I realized I forgot to initialize the params array for `ExpressionParamSet`, fixed in the latest push.

stellar-aria (approved):

seangoodvibes:
@bobtwinkles @trappar discovered an issue caused by this PR.

you can't apply different automation to the same instrument used in two different clips.

step's to reproduce:

1. create a clip
2. press play, record, edit filter automation on that clip
3. go back to song view
4. clone the first clip
5. switch to second clip and change automation there
6. switch back to first clip and check if the automation has also changed

PR #303, opened 2023-08-04
Merge Typed and untyped Enumeration and Selection MenuItems

stellar-aria:
This means there's no longer any type-instantiated templating occurring (it was only templated over enum class types anyways) and instead moves that logic to new `Value::get_value` and `Value::set_value`

Size reduction results:
Before:
```
$ arm-none-eabi-size build/Release/delugeOLED.elf
   text    data     bss     dec     hex filename
1536092    4316  345724 1886132  1cc7b4 build/Release/delugeOLED.elf
```

After:
```
$ arm-none-eabi-size build/Release/delugeOLED.elf
   text    data     bss     dec     hex filename
1515484    4316  345724 1865524  1c7734 build/Release/delugeOLED.elf
```
(20KiB saved)


src/deluge/gui/menu_item/value.h
```diff
@@ -31,12 +32,28 @@ class Value : public MenuItem {
 	void readValueAgain() override;
 	bool selectEncoderActionEditsInstrument() final { return true; }
 
+	void set_value(T value) { value_ = value; }
```
bobtwinkles, line 35: for consistency, these should be `camelCase`.

bobtwinkles (commented):
Minor style comment, otherwise LGTM (mechanical replacement).

src/deluge/gui/menu_item/value.h
stellar-aria: Fixed.

bfredl (approved):

PR #306, opened 2023-08-05
Define relwithdebinfo target

m-m-adams:
Creates a relwithdebinfo target that enables RTT and LTO compared to the default relwithdebinfo

This avoids a bizarre crash on voice stealing when O2 is used without LTO

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT error)
 if(IPO_SUPPORTED)
     set(CMAKE_INTERPROCEDURAL_OPTIMIZATION
         $<$<CONFIG:DEBUG>:OFF>
-        $<$<CONFIG:TEST>:ON>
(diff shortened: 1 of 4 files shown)


bobtwinkles (approved):
LGTM

CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ check_ipo_supported(RESULT IPO_SUPPORTED OUTPUT error)
 if(IPO_SUPPORTED)
     set(CMAKE_INTERPROCEDURAL_OPTIMIZATION
         $<$<CONFIG:DEBUG>:OFF>
+        $<$<CONFIG:TEST>:ON>
```
stellar-aria, line 17: is TEST an actual config or is this an artefact?

stellar-aria (commented):
LGTM pending comment response

CMakeLists.txt
m-m-adams: oops - artifact of trying to make a whole new config before deciding to just use relwithdebinfo instead

stellar-aria (approved):

PR #313, opened 2023-08-06
Only run pages deploy on community branch

trappar:
Currently pages tries to deploy every time build runs. This is a problem since:
* The build workflow currently runs on each commit
* The GH API has rate limiting
* We run more builds than would be allowed by rate limiting

This PR resolves this problem by only allowing the deploy pages job to execute on build workflow completion when the branch is "community"


bobtwinkles:
I think if we're going to merge this (and we probably should) it would be good to also limit the build list to the last N commits and remove the per-PR  section since there would no longer be a reasonable expectation of that section being up-to-date.

PR #315, opened 2023-08-07
Refactor PIC commands into driver, integrate RGB class

stellar-aria:
## Changes
- New PIC driver
- RGB color class integrated into the codebase
- Doxygen via CMake and DBT (doxygen target if doxygen is available, `dbt docs`)

## Details
This refactors all calls that send messages to the PIC to instead use a new PIC driver.

As a side effect, I got sidetracked into also integrating the RGB class throughout the system, so now the main pad image throughout both views and pad_leds are now a 2D array of RGB instead of a 3D array of uint8_t.

In the quest to get doxygen working in a sane way, I've moved the Doxygen config to being completely CMake-based (since we use CMake to get doxygen-awesome-css now as well rather than a submodule), and set up both DBT and CMake targets for the documentation, which gets generated into `build/html`

Size impact:
4KiB

Before:
```
$ arm-none-eabi-size build/Release/delugeOLED.elf
   text    data     bss     dec     hex filename
1523396    4316  345724 1873436  1c961c build/Release/delugeOLED.elf
```

After:
```
$ arm-none-eabi-size build/Release/delugeOLED.elf
   text    data     bss     dec     hex filename
1527788    4324  345692 1877804  1ca72c build/Release/delugeOLED.elf
```


src/deluge/gui/views/session_view.cpp
```diff
@@ -2203,7 +2194,7 @@ void SessionView::transitionToViewForClip(Clip* clip) {
 
 		else {
 			instrumentClipView
-			    .recalculateColours(); // Won't have happened automatically because we haven't begun the "session"
+			    .recalculateColors(); // Won't have happened automatically because we haven't begun the "session"
```
bobtwinkles, line 2197: nit: line end comment

src/deluge/util/lookuptables/lookuptables.cpp
```diff
@@ -685,4 +685,4 @@ const uint8_t noteCodeToNoteLetter[12] = {67, 67, 68, 68, 69, 70, 70, 71, 71, 65
 const bool noteCodeIsSharp[12] = {0, 1, 0, 1, 0, 0, 1, 0, 1, 0, 1, 0};
 
 
-const uint8_t defaultClipGroupColours[] = {102, 168, 24, 84, 186, 36, 126, 12, 147, 48, 72, 157};
+const uint8_t defaultClipGroupColors[] = {102, 168, 24, 84, 186, 36, 126, 12, 147, 48, 72, 157};
```
bobtwinkles, line 688: I assume these are actually hue values?

src/deluge/util/functions.h
```diff
@@ -413,64 +412,42 @@ int32_t stringToFirmwareVersion(char const* firmwareVersionString);
 
 // intensity is out of 65536 now
 // occupancyMask is out of 64 now
-inline void drawSquare(uint8_t squareColour[], int32_t intensity, uint8_t square[], uint8_t* occupancyMask,
-                       int32_t occupancyFromWhichColourCame) {
+inline RGB drawSquare(const RGB& squareColor, int32_t intensity, const RGB& square, uint8_t* occupancyMask,
```
bobtwinkles, line 415: This is only used in `pad_leds.cpp`, can probably move there

src/deluge/testing/hardware_testing.cpp
```diff
@@ -92,36 +93,42 @@ void setupSquareWave() {
 	int32_t count = 0;
 	for (int32_t* address = getTxBufferStart(); address < getTxBufferEnd(); address++) {
 		if (count < SSI_TX_BUFFER_NUM_SAMPLES) {
-			*address = 2147483647;
+			*address = std::numeric_limits<int32_t>::max();
 		}
 		else {
-			*address = -2147483648;
+			*address = std::numeric_limits<int32_t>::min();
+			;
```
bobtwinkles, line 100: extra `;`

bobtwinkles (commented):
Having the PIC protocol documented, a coherent RGB class, and 2k fewer LOC, and better doxygen is pretty nice. Would have been easier to review if it was 3 separate changes though :sweat_smile:. In particular the doxygen changes seem relatively easy to split out in to a separate PR and probably should be. The PIC protocol and color stuff is probably harder to separate, though having it separate would make bisecting potentially simpler. On that front too I'm not 100% sold on the colour -> color global find/replace, it adds a lot of changes without any increase in clarity.

stellar-aria:
> Having the PIC protocol documented, a coherent RGB class, and 2k fewer LOC, and better doxygen is pretty nice. Would have been easier to review if it was 3 separate changes though :sweat_smile:. In particular the doxygen changes seem relatively easy to split out in to a separate PR and probably should be. The PIC protocol and color stuff is probably harder to separate, though having it separate would make bisecting potentially simpler. On that front too I'm not 100% sold on the colour -> color global find/replace, it adds a lot of changes without any increase in clarity.

The PIC and color thing might be separable if I do some weirdness and stick a shim in the PIC PR. I'll give it a shot.

The global rename was because there had been some things introduced in the last three months that used "color" not "colour" (the keyboard view, particularly) and I thought it better to standardize on one over the other. I'm fine swapping to the other as long as we're consistent.

PR #317, opened 2023-08-07
Resolved Audition Pad Illumination Issue on Drum Keyboard Screen

alter-alter:
In the Drum Keyboard screen, I have fixed the issue where audition pads would remain illuminated after pressing pads 9 to 16 and then returning to the clip view.
https://youtu.be/TkzCt-VDFt4


PaulFreund:
Very nice catch! I will check your fix as soon as possible :) I have overlooked it because it didn't happen for my default kit with <= 8 elements

PaulFreund:
Also a good place to thank you for the reference implementation 😀 and also the very handy video explaining the behavior!

stellar-aria (approved):

PR #318, opened 2023-08-07
Add style recommendations to clang-tidy config

stellar-aria:
This is to push us towards a common style recommendation set. Having everything be camelCase is not good (esp. variables/functions both).

The key points:
classes are PascalCase
functions are camelCase
variables are snake_case

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -46,3 +46,14 @@ CheckOptions:
     value: google
   - key: modernize-replace-auto-ptr.IncludeStyle
     value: google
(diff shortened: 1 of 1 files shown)


jamiefaye:
Assuming recommendations != mandatory.

stellar-aria:
> Assuming recommendations != mandatory.

Correct.

The hope is to eventually have this run as part of a "pre-commit" task in dbt, but only on diffs , so that people are only responsible for the things they've touched.

PR #319, opened 2023-08-07
Add nightly builds to Github actions

stellar-aria:
This adds nightly builds to the GitHub actions via a "reuseable workflow". Unfortunately it also currently breaks our merge barrier, so we'll need to change that as well before merge.


bobtwinkles (approved):
LGTM. @jamiefaye  please coordinate with @stellar-aria to merge this, as it will require temporarily disabling the build required check and pointing it at the new name.

PR #332, opened 2023-08-10
Fix implicit casts for min/max/clamp

stellar-aria:
This fixes a number of improper implicit casts, primarily by removing the template specification and relying on deduction.


jamiefaye:
passes quick inspection and smoke test. approved.

PR #334, opened 2023-08-10
working intellisense configs without spurious warnings

m-m-adams:
Creates a cpp config without spurious warnings in vscode for mac

This is mac specific, I believe for windows it just needs the paths to the arm-gcc compiler updated however I do not know the correct paths. This config is marked as mac so should not error on other platforms


m-m-adams:
This is explicitly listing compile commands as the duplicate defines in compile-commands.json for HAVE_OLED confuse intellisense

seangoodvibes:
As discussed, I reviewed this on my Mac and it seems to be behaving as expected (showing the right errors).

seangoodvibes (approved):
As discussed, I have reviewed this PR on my Mac and it is behaving as expected. VSCode now show's the correct error messages.

PR #336, opened 2023-08-11
Morphable and Driveable and Parallel filters

m-m-adams:
This PR does the following:

1. Add an SVF option to the HPF slot, set for HPF mode by default. This can be modified through the menu, shortcuts, or clicking the lower knob while HPF is selected.

2. Adds a 3rd filter paramater called morph, placed in the menu and on the pads under the db/oct on filters. Morph smoothly morphs the SVF through LP-BP-HP, and smoothly increases drive in the ladder filters. This param is modulatable and automatable

3. Adds a setting to switch the filter order or run them in parallel. This setting is menu only and named ROUTE 

4. Refactors more of the ladder code to remove dead code paths


soymonitus:
Ey maybe too late to the party but i installed latest commit from community branch and when in HPF and mode is SVF as soon as i turn the golden knob a little bit i get no sound at all, like adding a little bit of cutoff in HPF-SVF breaks the audio output of that clip. Anyone experienced this?

PR #339, opened 2023-08-12
SVF Notch filter and improve default settings

m-m-adams:
- Adds an SVF notch mode (SV_Notch) alongside the SVF Band mode (SV_Band)
- Changes default SVF morph in HP slot to 50 (making it an HP filter)
- Tweaks default ladder saturation to match original firmware
- Applies ladder gain on input instead of feedback, avoiding interaction with resonance
- Tweaks SVF level to match ladder volumes
- Removes distinction between high and lowpass SVF enums so they can now be used in either slot


trappar:
I'm seeing the following issues. I don't know if these are all related to this PR, but they are at least issues with the new filter code:
* With HPF set to SV_BAND, morph is set to 0, and HPF frequency 0 it seems to not cut any frequencies. Switching to HPF frequency 1 switches to expected behavior.
* With LPF set to SV_BAND, morph set to 50, and LPF frequency set to 50, I can still hear highs - seems like the filter doesn't go high enough. I would expect the range to be the same as the HPF in SV_BAND with morph 0 but it is not.
* Getting artifacts when using the notch filter. This was created by creating a 4 voice supersaw with LPF frequency modulated by an envelope and no other modulation. Audio sample for this: [here](https://cdn.discordapp.com/attachments/1118654563596128336/1139803565527933018/notch_filter_weirdness.mp3)

m-m-adams:
First one is a bug, I need to update the logic that turns off filters

Second one is sort of a bug, max cutoff on the svf is 17khz right now but I should be able to fix that now. It should be the same on both hpf and lpf slots though other than the morph defaulting to 50 for the hpf they're identical

Third one I'm not sure about- the graph looks right but I can hear what sounds like voice culling. Can you share the synth preset?

m-m-adams:
These are now fixed. I can't replicate the artifacts you have with the notch, I suspect they're not related to the filter and it's just allowing them to be audible.

trappar:
> These are now fixed. I can't replicate the artifacts you have with the notch, I suspect they're not related to the filter and it's just allowing them to be audible.

I think I agree. It seems like it might be some sort of issue related to detune spread. I'll investigate further and open a new issue if I can get something solid.

soymonitus:
Added a bug about SVF here:
https://github.com/SynthstromAudible/DelugeFirmware/issues/355

m-m-adams:
Are you building from this PR? That bug should be fixed already

m-m-adams:
Latest commit enables hp filters while cutoff is zero/lp filters with cutoff 50 if morph is non zero. While I don't believe this is likely to be done intentionally (it is just a volume drop) it makes the behaviour more consistent as morph will now always behave the same through all possible frequency/resonance settings

m-m-adams:
Also confirmed this fixes #355 by providing a default morph value that acts as an HPF

src/deluge/dsp/filter/svf.h
```diff
@@ -43,6 +43,7 @@ class SVFilter : public Filter<SVFilter> {
 	inline q31_t doSVF(q31_t input, SVFState& state);
 	SVFState l;
 	SVFState r;
+	bool band_mode;
```
bobtwinkles, line 46: It's probably not super important here (since our minimum allocation granularity is 4 bytes anyway) but this forces the compiler to add 3 bytes of padding. If it's at the end the size can technically be 3 bytes lower.

src/deluge/dsp/filter/svf.cpp
```diff
@@ -86,14 +110,21 @@ inline q31_t SVFilter::doSVF(int32_t input, SVFState& state) {
...
+
 	//saturate band feedback
 	band = getTanHUnknown(band, 3);
-	//notch = high + low;
-	q31_t result = multiply_32x32_rshift32_rounded(lowi + low, c_low);
-	result = multiply_accumulate_32x32_rshift32_rounded(result, highi + high, c_high);
-	result = multiply_accumulate_32x32_rshift32_rounded(result, bandi + band, c_band);
 	//result = multiply_accumulate_32x32_rshift32_rounded(0, notchi+notch, c_notch);
-	result = 2 * result; //compensate for division by two on each multiply
+	result = 3 * result; //compensate for division by two on each multiply
```
bobtwinkles, line 127: Is the comment here stale now?

src/deluge/gui/menu_item/filter/hpf_mode.h
```diff
@@ -27,12 +27,12 @@ class HPFMode final : public Selection<kNumHPFModes> {
...
 	void readCurrentValue() override {
-		this->setValue(util::to_underlying(soundEditor.currentModControllable->hpfMode) - kNumLPFModes);
+		this->setValue(util::to_underlying(soundEditor.currentModControllable->hpfMode) - kFirstHPFMode);
 	}
 	void writeCurrentValue() override {
-		soundEditor.currentModControllable->hpfMode = static_cast<FilterMode>(this->getValue() + kNumLPFModes);
+		soundEditor.currentModControllable->hpfMode = static_cast<FilterMode>(this->getValue() + kFirstHPFMode);
 	}
-	static_vector<std::string, capacity()> getOptions() override { return {"HPLadder", "HPSV"}; }
+	static_vector<std::string, capacity()> getOptions() override { return {"SV_Band", "SV_Notch", "HPLadder"}; }
```
bobtwinkles, line 35: This should return a `string_view` probably, `std::string` will force an alloc + copy (even if it worked correctly which it won't until #354 is merged)

bobtwinkles (commented):
I'm going to assume all the DSP code works as intended and people with better ears than mine have validated it sounds good. Some minor functional comments but otherwise LGTM

src/deluge/gui/menu_item/filter/hpf_mode.h
bobtwinkles: Oh, if you rebase on to dfa7101e593498392e81615e653a3174ee767b39 the compiler will probably yell at you about this exact thing.

src/deluge/dsp/filter/svf.cpp
m-m-adams: Sort of, 2 is to compensate for the division on the multiply. I bumped it to 3 to match the volume of the ladder filters since it's easier than doing a *1.5 later 

I'll update the comment

bobtwinkles (approved):

PR #345, opened 2023-08-12
print used bytes in stack as well as free

m-m-adams:
Minor change, corrects the stack usage debug function and adds a line to print current usage as well as free space

### DIFF
--- a/src/deluge/memory/general_memory_allocator.cpp
+++ b/src/deluge/memory/general_memory_allocator.cpp
@@ -35,7 +35,7 @@ char emptySpacesMemoryInternal[sizeof(EmptySpaceRecord) * 1024];
 extern uint32_t __heap_start;
 extern uint32_t __heap_end;
 extern uint32_t program_stack_start;
-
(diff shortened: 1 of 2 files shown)


src/deluge/memory/general_memory_allocator.cpp
```diff
@@ -56,10 +56,11 @@ void GeneralMemoryAllocator::checkStack(char const* caller) {
 
 	char a;
 
-	int32_t distance = (int32_t)&a - (kInternalMemoryEnd - kProgramStackMaxSize);
+	int32_t distance = (int32_t)&a - (uint32_t)&program_stack_start;
```
stellar-aria, line 59: if this is the last use of kProgramStackMaxSize, could you also remove it from the definitions so we don't have a misleading constant around?

m-m-adams: Yeah good call

stellar-aria (approved):

PaulFreund (approved):

PR #349, opened 2023-08-13
Feature: Patchable wavefolding distortion

m-m-adams:
This PR adds a pre filter wavefolder, and the depth is patchable/automatable. The depth is accessible in both the menu and on the pad between saturation and LPF cutoff. The fold has no effect when set to 0 and removes itself from the signal path. 

Note that it has no effect on square waves, it's best with sines and triangles


m-m-adams:
Last commit increases the max range so it should be more noticeable

seangoodvibes (approved):
I have reviewed the code and did not notice anything out of place. I don't understand the math but it seems good to me. In areas where the new param was added I compared the code to other params and the additions are in line.

I am ok with this addition.

PR #352, opened 2023-08-13
fix: use string_view in more parts of the menu

### LINKED ISSUE #343: Cloning clip broken in latest nightly
On my OLED deluge, attempting to clone a clip yields a copy where the note positions are mangled.

[Here's a youtube video of what I'm seeing](https://youtu.be/N5y_0xC55Fc).

bobtwinkles:
Fixes #343 though probably not by actually fixing the underlying bug


stellar-aria (approved):
LGTM.

PR #354, opened 2023-08-14
Fix double-free in static_vector::non_trivial

### LINKED ISSUE #341: Pads menu settings bug
If you visit the Settings -> Pads menu off a fresh boot on an OLED deluge, you will see the following options:
* SHORTCUTS VERSION
* KEYBOARD FOR TEXT
* COLOURS

Scrolling down causes the menu items to get jumbled. One tick down shows:
* SHORTCUTS VERSION
* SHORTCUTS VERSION
* COLOURS

And scrolling back up shows:
* KEYBOARD FOR TEXT
* KEYBOARD FOR TEXT
* COLOURS

Once this state has been reached, it's like that until you reboot your deluge.

stellar-aria:
When I swapped the inner implementation of non_trivial objects from heap allocation (which had major memory alignment issues) to std::array, I didn't check the object destructor.

This resulted in some leftover code calling destructors on the inner objects of the array, right before the array itself would call those same destructors, resulting in a double-free for std::containers (such as std::string).

Ideally once the GMA has been fixed/replaced we can swap the non_trivial implementation back to heap allocation, but this is fine for the moment.

Huge thanks to @bobtwinkles for finding the double free causing this.

Fixes #341

### DIFF
--- a/src/deluge/util/container/static_vector.hpp
+++ b/src/deluge/util/container/static_vector.hpp
@@ -400,9 +400,11 @@ public:
 	/// Contract: the storage is not empty.
 	void pop_back() noexcept(std::is_nothrow_destructible_v<T>) {
 		SV_EXPECT(!empty() && "tried to pop_back from empty storage!");
-		auto ptr = end() - 1;
-		ptr->~T();
-		unsafe_set_size(size() - 1);
+		// auto ptr = end() - 1;
+		// ptr->~T();
+		//unsafe_set_size(size() - 1);
+
+		--size_;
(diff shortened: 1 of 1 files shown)


jamiefaye:
Can we have a comment in the code as to why code is commented-out? Just in case this stuff sticks around for awhile.

bobtwinkles (approved):
This leaves `static_vector` a bit of a leaky mess but it's better than heap corruption.

m-m-adams (approved):

PR #351, opened 2023-08-13
WIP: GeneralMemoryAllocator pretty-printer

bobtwinkles:
This adds a python script that can be used either within GDB or offline against a memory dump. To use it inside GDB, source the script and then run it from a context where the `globalMemoryAllocator` symbol is accessible:

```
b GeneralMemoryAllocator::get
c
source ./contrib/debug/heap_state.py
python parse_gma(gdb.inferiors()[0])
```

It should print out something like this:
```== Memory Region EXTERNAL: empty region list: 20044bec, count=2
--- Empty regions:
EmptyRegion{ addr:0c00808c, len:00000014 }
EmptyRegion{ addr:0c00808c, len:00000014 }
--- Stealable regions:
---- Stealables for stealable queue 0
---- Stealables for stealable queue 1
---- Stealables for stealable queue 2
---- Stealables for stealable queue 3
---- Stealables for stealable queue 4
---- Stealables for stealable queue 5
---- Stealables for stealable queue 6
Stealable node at: 0c0f12c4 for list 200300d8
Stealable node at: 0c0f933c for list 200300d8
Stealable node at: 0c1013b4 for list 200300d8
Stealable node at: 0c10942c for list 200300d8
Stealable node at: 0c1114a4 for list 200300d8
Stealable node at: 200300d8 for list 00000000
---- Stealables for stealable queue 7
---- Stealables for stealable queue 8
---- Stealables for stealable queue 9
== Memory Region INTERNAL: empty region list: 20042bf4, count=21
--- Empty regions:
EmptyRegion{ addr:201dcc00, len:00000010 }
EmptyRegion{ addr:201e2b5c, len:00000018 }
EmptyRegion{ addr:201e2c4c, len:00000018 }
EmptyRegion{ addr:201e2d3c, len:00000018 }
EmptyRegion{ addr:201e50f0, len:00000018 }
EmptyRegion{ addr:201e72dc, len:00000018 }
EmptyRegion{ addr:201e94c8, len:00000018 }
EmptyRegion{ addr:201eb6b4, len:00000018 }
EmptyRegion{ addr:201ed8a0, len:00000018 }
EmptyRegion{ addr:201efa8c, len:00000018 }
EmptyRegion{ addr:201f1c78, len:00000018 }
EmptyRegion{ addr:201fac60, len:00000018 }
EmptyRegion{ addr:201fcd7c, len:00000018 }
EmptyRegion{ addr:201fef68, len:00000018 }
[...]

PR #358, opened 2023-08-16
Operator New Alloc to external only

### LINKED ISSUE #350: Clip cloning leads to audible culling in simple songs
With very simple test song with one bassline of 16 steps (no detune) and a kit with 4 to the floor kick pattern and nothing else at 120 bpm I get audible culling and the following debug output:

```culled 1 voices. numSamples: 46. voices left: 1```

which can not be reproduced with official beta

### PR DESCRIPTION
**Pull request by m-m-adams**
Changes the new operator to allocate in external RAM. This frees a significant amount of internal RAM and regains some of our lost performance 

Note that with this change any performance critical objects must be manually allocated and initialized since new would place them on slower SDRAM. Luckily the audio related code already does this.

I believe this fixes  #350

### DIFF
--- a/src/deluge/memory/general_memory_allocator.cpp
+++ b/src/deluge/memory/general_memory_allocator.cpp
@@ -145,11 +145,10 @@ int32_t GeneralMemoryAllocator::getRegion(void* address) {
 		return MEMORY_REGION_INTERNAL;
 	}
 	else if (((uint32_t)address <= EXTERNAL_MEMORY_BEGIN) || (uint32_t)address >= EXTERNAL_MEMORY_END) {
-		numericDriver.freezeWithError("e999");
+		numericDriver.freezeWithError("M999");
 		return -1;
 	}
 	return MEMORY_REGION_SDRAM;
-	//return ((uint32_t)address >= (uint32_t)INTERNAL_MEMORY_BEGIN) ? MEMORY_REGION_INTERNAL : MEMORY_REGION_SDRAM;
 }
 
 // Returns new size
--- a/src/deluge/memory/general_memory_allocator.cpp
+++ b/src/deluge/memory/general_memory_allocator.cpp
@@ -38,6 +38,7 @@ extern uint32_t program_stack_start;
 extern uint32_t program_stack_end;
 GeneralMemoryAllocator::GeneralMemoryAllocator() {
 	lock = false;
+
 	regions[MEMORY_REGION_SDRAM].setup(emptySpacesMemory, sizeof(emptySpacesMemory), EXTERNAL_MEMORY_BEGIN,
 	                                   EXTERNAL_MEMORY_END);
 	regions[MEMORY_REGION_INTERNAL].setup(emptySpacesMemoryInternal, sizeof(emptySpacesMemoryInternal),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/deluge/memory/general_memory_allocator.cpp [hunk 1]
```diff
@@ -140,7 +141,15 @@ uint32_t GeneralMemoryAllocator::getAllocatedSize(void* address) {
 }
 
 int32_t GeneralMemoryAllocator::getRegion(void* address) {
-	return ((uint32_t)address >= (uint32_t)INTERNAL_MEMORY_BEGIN) ? MEMORY_REGION_INTERNAL : MEMORY_REGION_SDRAM;
+	if ((uint32_t)address >= (uint32_t)INTERNAL_MEMORY_BEGIN) {
+		return MEMORY_REGION_INTERNAL;
+	}
+	else if (((uint32_t)address <= EXTERNAL_MEMORY_BEGIN) || (uint32_t)address >= EXTERNAL_MEMORY_END) {
+		numericDriver.freezeWithError("e999");
```
bobtwinkles, line 148: I think we should start a convention of using "Mxxx" as the error codes in the allocator (the smoke test in #351  uses M000 for example)

src/deluge/memory/general_memory_allocator.cpp
```diff
@@ -140,7 +141,15 @@ uint32_t GeneralMemoryAllocator::getAllocatedSize(void* address) {
...
+		return -1;
+	}
+	return MEMORY_REGION_SDRAM;
+	//return ((uint32_t)address >= (uint32_t)INTERNAL_MEMORY_BEGIN) ? MEMORY_REGION_INTERNAL : MEMORY_REGION_SDRAM;
```
bobtwinkles, line 152: can delete this I think

src/deluge/memory/memory_region.h
```diff
@@ -38,7 +38,7 @@ struct NeighbouringMemoryGrabAttemptResult {
 #define SPACE_HEADER_ALLOCATED 0x80000000
 
 #define SPACE_TYPE_MASK 0xC0000000
-#define SPACE_SIZE_MASK 0x3FFFFFFF
+#define SPACE_SIZE_MASK 0x3FFFFFFFu
```
bobtwinkles, line 41: Should add the `u` suffix to TYPE_MASK as well, as long as we're in the area

src/deluge/memory/memory_region.h
```diff
@@ -54,7 +54,7 @@ class MemoryRegion {
 	void dealloc(void* address);
 	void verifyMemoryNotFree(void* address, uint32_t spaceSize);
 
-	OrderedResizeableArrayWithMultiWordKey emptySpaces;
+	//OrderedResizeableArrayWithMultiWordKey emptySpaces;
```
bobtwinkles, line 57: delete?

src/resetprg.c
```diff
@@ -144,6 +144,15 @@ void resetprg(void) {
...
 	__enable_irq();
 	__enable_fiq();
+	// Setup SDRAM. Have to do this before we init global objects
+	userdef_bsc_cs2_init(0); // 64MB, hardcoded
+#if !defined(NDEBUG)
+	const uint32_t SDRAM_SIZE = EXTERNAL_MEMORY_END - EXTERNAL_MEMORY_BEGIN;
+	uint16_t* sdram = (uint16_t*)EXTERNAL_MEMORY_BEGIN;
+	for (size_t i = 0; i < SDRAM_SIZE; i++) {
+		sdram[i] = 0x0000;
+	}
```
bobtwinkles, line 154: using SIMD or a `memset` might be faster here? Or is there a particular reason to use `uint16` sized writes

src/deluge/memory/memory_region.h
m-m-adams: Yeah sure, I left it since that one defaults to unsigned already

src/resetprg.c
m-m-adams: Memset is faster, I had it in one of earlier commits but I guess it got lost along the way. That explains why it got slower

PaulFreund:
I tested this PR in RelWithDebInfo and audibly it is the same unfortunately but in RTT instead of hard culling I get
```soft-culled 1 voice. voices now: 2```
scrolling through loading presets still gets me some hard culling. I think it is a step in the right direction and in combination with reducing image size could be enough. Once we get the same performance as our baseline we probably need to feed the audio rendering in those functions.

File: src/deluge/memory/memory_region.cpp [hunk 6]
```diff
@@ -98,7 +99,10 @@ static EmptySpaceRecord* recordToMergeWith;
 // Specify the address and size of the actual memory region not including its headers, which this function will write and don't have to contain valid data yet.
 // spaceSize can even be 0 or less if you know it's going to get merged.
 inline void MemoryRegion::markSpaceAsEmpty(uint32_t address, uint32_t spaceSize, bool mayLookLeft, bool mayLookRight) {
-
+	if ((address <= start) || address >= end) {
+		numericDriver.freezeWithError("e999");
```
bobtwinkles, line 103: -> `M998`

bobtwinkles (approved):
minor nitpick but otherwise LGTM

PR #347, opened 2023-08-12
[Ready to Merge] Automation View: Add new parameters

seangoodvibes:
As a follow-up to the initial Automation Instrument Clip View PR (#241), this PR adds some additional parameters to the automation view:

1. Automatable Parameters for Synths, Kits with affect entire DISABLED, and Midi

>The 4 new parameters added are:
>
> - **LPF** Morph
> - **HPF** Morph
> - **Distortion** Wavefolder
> - **Portamento**

2. Automatable Parameters for Kits with affect entire ENABLED

>The 12 new parameters added are:
>
> - **Master** Pitch
> - **EQ** Bass, Bass Freq, Treble, Treble Freq
> - **Sidechain** Shape
> - **Distortion** Decimation, Bitcrush
> - **Mod FX** Offset, Feedback
> - **Arp** Gate
> - **Portamento**

### Summary of Code Changes

1) Renamed NonGlobalEffectableParamsForAutomation to NonKitAffectEntireParamsForAutomation

2) Renamed GlobalEffectableParamsForAutomation to KitAffectEntireParamsForAutomation

3) Re-ordered the params in the functions.cpp switches to a more logical and easier to read order (same order as params are selected in the automation view)

4) Added new parameters to the Automation View

- Added LPF Morph, HPF Morph, Wavefolder, Portamento to the NonKitAffectEntireParams

- Added Bass, Bass Freq, Treble, Treble Freq, Sidechain Shape, Decimation, Bitcrush, Mod FX Offset, Mod FX Feedback, Arp Gate and Portamento to the KitAffectEntireParams

5) Updated the enums in Definitions_cxx.hpp to increase the number automatable params

6) Adjusted the Automation Overview, Single Pad Press and Select Encoder functions to recognize / be able to select the new Unpatched params added to the Kit Affect Entire view

PR #241, opened 2023-07-27
Automation Instrument Clip View

seangoodvibes:
# Latest Version for Testing & Release Notes

### Download Link (Updated Wednesday, August 11, 2023 at 6:50 PM Eastern / UTC-4)

https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/5837637150?pr=241

### Summary of changes since previous version:

- [x] Updated Automation Editor to display automation lanes at lower brightness when parameter is not currently automated and at regular brightness when parameter is currently automated
- [x] update cross-screen button to show message "Coming Soon (OLED) / SOON (7Seg)" as that functionality does not currently work in automation view
- [x] update CV instrument type to show message "Can't Automate CV (OLED) / CANT (7Seg)" as you cannot automate CV clips
- [x] fixed an issue with sidebar rendering when you exit the "load song" browser
- [x] fixed an issue with the select encoder not scrolling from last selected parameter when you use the pad shortcuts
- [x] disabled select encoder in kit view with affect entire disabled when a kit row has not been selected

# Description

Automation Instrument Clip View is a new view that complements the existing Instrument Clip View (which is accessed by pressing the Clip Button). It allows you to edit automatable parameters on a per step basis at any zoom level.

Automatable Parameters are broken down into two categories for Automation Instrument Clip View purposes:

1. Automatable Parameters for Synths, Kits with affect entire DISABLED, and Midi

>The 51 parameters that can be edited are:
>
> - **Master** Volume, Pitch, Pan
> - **LPF** Cutoff, Resonance
> - **HPF** Cutoff, Resonance
> - **EQ** Bass, Bass Freq, Treble, Treble Freq
> - **Reverb** Amount
> - **Delay** Rate, Amount
> - **Sidechain** Send, Shape
> - **Distortion** Decimation, Bitcrush
> - **OSC 1** Volume, Pitch, Phase Width, Carrier Feedback, Wave Index
> - **OSC 2** Volume, Pitch, Phase Width, Carrier Feedback, Wave Index
> - **FM Mod 1** Volume, Pitch, Feedback
> - **FM Mod 2** Volume, Pitch, Feedback
> - **Env 1** Attack, Decay, Sustain, Release
> - **Env 2** Attack, Decay, Sustain, Release
> - **LFO 1** Frequency
> - **LFO 2** Frequency
> - **Mod FX** Offset, Feedback, Depth, Rate
> - **Arp** Rate, Gate
> - **Noise**

[...]

PR #360, opened 2023-08-16
[Ready to Merge] Automation View: Long press bug fixes + Pad selection mode

seangoodvibes:
This PR is a follow up to PR #347 and the original Automation View PR #241

This PR resolves some bugs with entering automation long presses and also adds a new pad selection mode to offer the user even more precision automation editing.

### Fine tuning a long press

Added the ability to fine tune a long press's start and end values which automatically adjusts the interpolation between those values (without needing to enter the pad selection mode). 

Simply enter a long press and while continuing to hold the first pad in the long press, turn the gold knobs to adjust the start and end values of the long press.

When refining a long presses start/end values, each gold knob is used separately to adjust the start/end position values. 

The bottom left gold knob is used to adjust the start value. 

The top right gold knob is used to adjust the end value.

In a long press, you will also note that the LED indicators change to show the current value of the start and end position. Once you've let go of the long press, the LED indicators reset back to the parameters overall current value.

### Pad Selection Mode

image

Pad selection mode is accessed by pressing on either of the mod encoder buttons

In pad selection mode, you cannot edit the automation grid by pressing on the pads. 

With a single or multi pad (long) press you select the pad(s) to edit. 

A cursor is displayed on the grid to identify the single or multiple pads you've selected. Seeing the cursor means you are in pad selection mode.

Once you've made your pad selection, you can use the gold knobs to tweak the value of the singular pad or the value of the start/end pads in a long press.

When one pad is selected, one cursor is displayed on the grid and both led indicators show the value for that pad (cursor position).

When two pads are selected, two cursors are displayed on the grid and the lower led indicator shows the value of the left pad (left cursor), and the upper led indicator shows the value of the right pad (right cursor).

With this mode you can also press on each pad to see its current value without changing its value.

### Other changes / bug fixes

- Adjusted the max of the parameter value range to 128 from 127 by adjusting the kMaxKnobPos enum in definitions_cxx.hpp
- Fixed multi pad press bug that was calculating the wrong parameter values by adjusting the handleMultiPadPress function to use floats for the linear interpolation calculation and by adjusting the loop code to properly loop forwards/backwards depending on first pad pressed's horizontal position relative to the second pad pressed
- Removed calculateKnobPosForMultiPadPress function
- Fixed renderRow bug which was not displaying value ranges correctly
- Refactored code out to new getParameterKnobPos function
- Removed unnecessary setKnobIndicatorLevel calls
- Adjusted setParameterAutomationValue function so it doesn't displayParameterValue for multi pad press
- Minor clean up in function calculateKnobPosForSinglePadPress


m-m-adams (approved):
Looks good! This improves performance with lots of automation as well

seangoodvibes:
Thanks @m-m-adams 

This is ready to merge @jamiefaye :)!

PR #363, opened 2023-08-18
Grain FX

alter-alter:
Added Grain FX to Mod FX
https://youtu.be/9W70SZYMKXo

**Parameters:**
Mod Depth: Controls Grain Volume / Dry Wet Mix
Mod Offset: Adjusts Grain Size
Mod Rate: Sets Grain Rate
Mod Feedback: Selects Grain Type

**Note:**
This effect is somewhat resource-intensive, so please use only one instance per song.

### DIFF
--- a/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
+++ b/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
@@ -35,9 +35,9 @@
 #include "processing/engines/audio_engine.h"
 #include "processing/sound/sound.h"
 #include "storage/storage_manager.h"
+#include "util/fast_fixed_math.h"
 #include "util/misc.h"
 #include <string.h>
-
 extern "C" {}
 
 extern int32_t spareRenderingBuffer[][SSI_TX_BUFFER_NUM_SAMPLES];
@@ -47,6 +47,23 @@ ModControllableAudio::ModControllableAudio() {
 	// Mod FX
 	modFXBuffer = NULL;
 	modFXBufferWriteIndex = 0;
+
+	//Grain
+	modFXGrainBuffer = NULL;
+	modFXGrainBufferWriteIndex = 0;
+	grainShift = 13230; // 300ms
+	grainSize = 13230;  // 300ms
+	grainRate = 1260;   // 35hz
+	grainFeedbackVol = 161061273;
+	for (int i = 0; i < 8; i++) {
+		grains[i].length = 0;
+	}
+	grainVol = 0;
+	grainDryVol = 2147483647;
+	grainPitchType = 0;
+	grainLastTickCountIsZero = true;
+	grainLastDetune = 1;
+	grainInitialized = false;
 
 	// EQ
 	withoutTrebleL = 0;
@@ -74,6 +91,9 @@ ModControllableAudio::~ModControllableAudio() {
 	// Free the mod fx memory
 	if (modFXBuffer) {
 		GeneralMemoryAllocator::get().dealloc(modFXBuffer);
+	}
+	if (modFXGrainBuffer) {
+		GeneralMemoryAllocator::get().dealloc(modFXGrainBuffer);
 	}
 }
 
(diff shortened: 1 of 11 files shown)


sichtbeton:
Tested it with a simple field recording. Instant blade runner.

sichtbeton:
Some options can be modulated, some not. Why is that? Cpu?

alter-alter:
@sichtbeton 
> Some options can be modulated, some not. Why is that? Cpu?

Currently, it is following the standard behavior of Deluge. Changing that might be a bit challenging at the moment.

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -174,6 +194,60 @@ void ModControllableAudio::processFX(StereoSample* buffer, int32_t numSamples, M
...
+		else if (modFXType == ModFXType::GRAIN) {
+			if (!grainInitialized && modFXGrainBufferWriteIndex >= 65536) {
+				grainInitialized = true;
+			}
+			*postFXVolume = multiply_32x32_rshift32(*postFXVolume, 1518500250) << 1; // Divide by sqrt(2)
+			// Shift
+			grainShift = 44 * 300; //(kSampleRate / 1000) * 300;
+			// Size
+			grainSize = 44 * ((((unpatchedParams->getValue(Param::Unpatched::MOD_FX_OFFSET) >> 1) + 1073741824) >> 21));
+			grainSize = std::max<int32_t>(440, std::min<int32_t>(35280, grainSize)); //10ms - 800ms
```
m-m-adams, line 206: std::Clamp instead (applies in a bunch of places)

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -174,6 +194,60 @@ void ModControllableAudio::processFX(StereoSample* buffer, int32_t numSamples, M
 			modFXLFOWaveType = LFOType::SINE;
 			*postFXVolume = multiply_32x32_rshift32(*postFXVolume, 1518500250) << 1; // Divide by sqrt(2)
 		}
+		else if (modFXType == ModFXType::GRAIN) {
+			if (!grainInitialized && modFXGrainBufferWriteIndex >= 65536) {
+				grainInitialized = true;
+			}
+			*postFXVolume = multiply_32x32_rshift32(*postFXVolume, 1518500250) << 1; // Divide by sqrt(2)
```
m-m-adams, line 201: define a constant for root 2 in fixedpoint.h (ONE_OVER_SQRT2_Q31) and use that

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -174,6 +194,60 @@ void ModControllableAudio::processFX(StereoSample* buffer, int32_t numSamples, M
...
+			if (grainRate < 882) {                                            //50hz or more
+				grainSize = std::min<int32_t>(grainSize, grainRate << 3) - 1; //16 layers=<<4, 8layers = <<3
+			}
+			// Volume
+			grainVol = modFXDepth - 2147483648;
+			grainVol =
+			    (multiply_32x32_rshift32_rounded(multiply_32x32_rshift32_rounded(grainVol, grainVol), grainVol) << 2)
+			    + 2147483648; //Cubic
+			grainVol = std::max<int32_t>(0, std::min<int32_t>(2147483647, grainVol));
+			grainDryVol =
```
m-m-adams, line 248: It would be nice to have feedback scale with the mix level - grainFeedbackVol = grainVol >> 3; works well in my experimentation. The current feedback makes a pulsing sound with sustained notes and I'd like to be able to have lower mixes without it

m-m-adams (changes requested):
So first off this sounds great! Awesome job and thanks for contributing

Testing this it looks like there's 2 performance issues. Using a test song that can sustain 50 voices, I lose 5 for each clip I add a grain fx too and they stay lost until the modfx is turned off (not just away from grain). E.g. going from off->stereochorus will get 48 voices, going from grain to stereochorus will get 45. This is just caused by the memory for the grain buffer. A first fix is to deallocate that buffer when it's not needed. A potential further fix could be to allocate a single global grain buffer and treat it as a pseudo send fx. Each voice could pull from it and add to it with it's own depth and rate. This is more work to implement so I haven't tested it but I think it would work and still sound cool, if a bit less 'traditional' granular. 

Secondly the DSP code is more intensive than the other modfx. I think this is improvable, I've left some comments on easy wins but the bigger one is probably vectorizing the grain calculations. They're fully parallelizable so should be able to get on NEON for a 2-3x speedup

Third I think it is nicer to have the feedback scale with the mix level. There is a pulsing effect that becomes audible on sustained notes and it is alleviated with lower feedback settings. I've left a comment with the calculation I used for testing and for me it produces a more usable range of depth settings

m-m-adams:
It also might be good to swap the feedback and depth pads

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -204,6 +278,150 @@ void ModControllableAudio::processFX(StereoSample* buffer, int32_t numSamples, M
...
+						                  : grains[i].volScaleMax
+						                        - (grains[i].counter - (grains[i].length >> 1)) * grains[i].volScale;
+						int32_t delta = grains[i].counter * (grains[i].rev == 1 ? -1 : 1);
+						if (grains[i].pitch != 1024) {
+							delta = ((delta * grains[i].pitch) >> 10);
+						}
+						int32_t pos =
+						    (grains[i].startPoint + delta + kModFXGrainBufferSize) & kModFXGrainBufferIndexMask;
+
+						grains_l = add_saturation(
```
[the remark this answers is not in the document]
alter-alter: multiply_accumulate_32x32_rshift32_rounded (smmlar) does not seem to take saturation operations into account.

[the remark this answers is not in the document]
m-m-adams: It doesn't, but if this is overflowing then that should be fixed instead. The saturation in saturating arithmetic is hard clipping in audio terms and not really desirable

alter-alter:
@m-m-adams 
> It also might be good to swap the feedback and depth pads

Since the modulatable parameters are only Depth and Rate, I believe Depth would be suitable for controlling the volume of the grain.

m-m-adams:
You can just make it modulatable by changing it from a Param::Unpatched to a Param::Global. Automating it works fine with the current modfx so modulating should be fine as well

alter-alter:
@m-m-adams 
> You can just make it modulatable by changing it from a Param::Unpatched to a Param::Global. Automating it works fine with the current modfx so modulating should be fine as well

Thank you! However, the change seems to be static and permanent, and I don't want it to affect the entire system. Is there a way to dynamically change the patch type only when ModFxType is set to Grain?

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -394,27 +394,25 @@ void ModControllableAudio::processFX(StereoSample* buffer, int32_t numSamples, M
...
-						                             grains[i].panVolL))
-						        << 2,
-						    grains_l);
-						grains_r = add_saturation(
-						    (multiply_32x32_rshift32(multiply_32x32_rshift32(modFXGrainBuffer[pos].r, vol) << 1,
-						                             grains[i].panVolR))
-						        << 2,
-						    grains_r);
+						grains_l = multiply_accumulate_32x32_rshift32_rounded(
+						    grains_l, multiply_32x32_rshift32(modFXGrainBuffer[pos].l, vol) << 0, grains[i].panVolL);
```
m-m-adams, line 398: << 0 is a nop but I suspect it's optimized out anyway

m-m-adams (approved):
Good to go, the lshift by 0 should be removed but it's optimized out anyway

m-m-adams:
No problem - it can always be made patched in the future

PR #368, opened 2023-08-19
note row probability

m-m-adams:
This allows probability to be set for a whole note row at a time. The action is fully undoable

In combination with the euclidean sequencing this makes it very fast to get a beat going


seangoodvibes (approved):
I have reviewed and test this feature and the additions to the codebase make sense and I appreciate the refactoring in instrument clip view to segregate out some of the code into separate functions (display probability, set row probability).

The feature works as I would expect and is very intuitive.

great work :) approved

PR #369, opened 2023-08-19
New release web page pointing to nightlies

PaulFreund:
Here is how it will look like, I still need to figure out integration:

image

Feedback welcome!


jamiefaye:
It might be a good idea to put a link to the "official Synthstrom release build" firmware file(s) just so somebody who changes their mind can get ahold of the "antidote" if need be.

PaulFreund:
I updated the page to include more notes about the official firmware. We can continually improve the page especially after we discussed more complex scenarios and dynamic data. I expect it to just work but we might need to fix the workflow after the PR if something unexpectedly does not.

image

.github/workflows/build-pages.yml
```diff
@@ -4,12 +4,9 @@
...
 on:
-  # Automatically run after the build workflow completes, since that workflow
-  # must *always* run to get PRs approved.
-  workflow_run:
-    workflows: [Build]
-    types:
-      - completed
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["community"]
```
bobtwinkles, line 9: This should have a `paths:` filter so the workflow only runs on commits that touch the `pages/` directory I think

bobtwinkles (approved):
Minor nit (I doubt the build will take enough time for it to matter) but otherwise LGTM. Thanks for carrying the torch forward on this one!

.github/workflows/build-pages.yml
PaulFreund: Fixed :) Thanks!

PR #374, opened 2023-08-20
feature(sysex): more efficient screen updates + switch type at runtime

bfredl:
Instead of having the client poll the screen with regular intervals, this instead has the deluge push out updates whenever the screen changes. This should dramatically reduce the CPU overhead of this feature when there is no changes to the OLED display.

Implement throttling by checking the fullness of the sysex send buffer. This naturally reduces the refresh rate as needed when using serial MIDI, or some USB device which reading too slow

the client still has to continuously ask the deluge for updates, to stop sending after the client disconnects. This can be done with a much lower rate, one short message per second. The client can also force a full update so a dropped delta message does not cause the screen to be corrupt forever.

 In addition, based on the recent display unification work, it is now possible to switch the screen type at runtime from the client.

### DIFF
--- a/src/deluge/hid/hid_sysex.cpp
+++ b/src/deluge/hid/hid_sysex.cpp
@@ -1,9 +1,18 @@
 #include "hid/hid_sysex.h"
+#include "gui/ui_timer_manager.h"
 #include "hid/display/oled.h"
+#include "hid/display/seven_segment.h"
 #include "io/midi/midi_device.h"
 #include "io/midi/midi_engine.h"
+#include "memory/general_memory_allocator.h"
+#include "processing/engines/audio_engine.h"
 #include "util/pack.h"
 #include <cstring>
+
+MIDIDevice* midiDisplayDevice = nullptr;
+int32_t midiDisplayUntil = 0;
+uint8_t* oledDeltaImage = nullptr;
+bool oledDeltaForce = true;
 
 void HIDSysex::sysexReceived(MIDIDevice* device, uint8_t* data, int32_t len) {
 	if (len < 6) {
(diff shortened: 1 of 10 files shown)


soymonitus:
How do you, having an oled deluge, make it show the 7seg display? I couldn't find how. I have opened the website that connects to deuge and prints debug messages and shows the oled screen but the 7 seg display is not shown on the web

bfredl:
it needs an update on the client side as well. I've just been away traveling for some days but I will get back to it soon.

bfredl:
@soymonitus Testing branch of the web client which works with this branch
https://bfredl.github.io/delugeclient_testing/app.html

There is the magic "switch display type" button. Note that this does not cause the screen to immediately redraw/reset.
Typically you have to do some action (like scroll between presets, scroll between menu items/values, switch view) etc, in order for the display to actually update after switching between 7seg and oled.

src/deluge/hid/hid_sysex.cpp
```diff
@@ -28,10 +37,61 @@ void HIDSysex::requestOLEDDisplay(MIDIDevice* device, uint8_t* data, int32_t len
 	if (data[4] == 0 or data[4] == 1) {
 		sendOLEDData(device, (data[4] == 1));
 	}
+	else if (data[4] == 2 || data[4] == 3) {
```
PaulFreund, line 40: Nit: Currently the protocol parsing looks magic number heavy. Maybe add some enums for clarity?

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -565,11 +565,18 @@ void ConnectedUSBMIDIDevice::bufferMessage(uint32_t fullMessage) {
...
+	// must be the same unsigned type as ringBufWriteIdx/ringBufReadIdx
 	uint32_t queued = ringBufWriteIdx - ringBufReadIdx;
 	return queued > 0;
 }
 
+int ConnectedUSBMIDIDevice::sendBufferSpace() {
+	// must be the same unsigned type as ringBufWriteIdx/ringBufReadIdx
+	uint32_t queued = ringBufWriteIdx - ringBufReadIdx;
+	// each 4-byte MIDI-USB message contains 3 bytes of serial MIDI data
+	return (MIDI_SEND_BUFFER_LEN_RING - queued) * 3;
```
PaulFreund, line 577: Won't this overreport the buffer to be up to 3 times the allocated size? If used to pack the buffer it could lead to run over the array bounds.

PaulFreund (changes requested):

src/deluge/io/midi/midi_device_manager.cpp
bfredl: no, for the reason explained in the comment just above.

PaulFreund: Ah, I overlooked the buffer being 32 bit values and this function reporting bytes available not "slots". lgtm

PaulFreund (approved):

Issue #371, opened 2023-08-20
A song with just and audio clip, when loaded, has the clip button flashing


soymonitus:
@seangoodvibes found this accidentally but I think it's related to the Automation feature

seangoodvibes:
Hrm it is possible this was my mistake. Can you confirm it doesn't flash with normal firmware?

Good catch

soymonitus:
I think the deluge thinks wrongly that it is inside automation clip view. There must be a code path that leads to this behavior for audio clips

seangoodvibes:
> I think the deluge thinks wrongly that it is inside automation clip view. There must be a code path that leads to this behavior for audio clips

Yea you're right there's a bug. Adding it to my list of things to fix.

Thanks!

seangoodvibes:
@soymonitus This has been fixed. Can you test and close the issue if you are satisfied :)? Thx

PR #397, opened 2023-08-21
Fixed clip button blinking in Audio Clip View

seangoodvibes:
Fixed the clip button blinking in audio clip view as a result of the automation view implementation

This resolves issue #371

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
 			"label": "Build Debug",
 			"command": "./dbt",
 			"windows": {
-				"command": "./dbt.cmd"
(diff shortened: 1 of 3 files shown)


m-m-adams (changes requested):
This has a regression - holding clips that are in automation view no longer blinks the clip light

m-m-adams:
This now only somewhat fixes the bug - the clip button will flash while holding down an audio clip but stop when you enter it

seangoodvibes:
@m-m-adams attempt #2 to fix the regression has been committed!

m-m-adams (approved):

Issue #375, opened 2023-08-20
Fill notes feedback


trappar:
Going to leave my feedback here since [the PR](https://github.com/SynthstromAudible/DelugeFirmware/pull/211) already been merged.

Overall I found this feature initially confusing. I wasn't able to understand how it worked by looking at the community features. I had to read through the PR to try understand it. It feels like maybe some of this can be chalked up to bugs since the community features explanation seems to conflict with what I'm seeing on my OLED. For example, it mentions "The FILL setting is at the start of the probability range, before 5%.", but I see no such setting.

### Feedback
* The fact that you have to set a note to "PROBABILITY: 0%" in order for it to be a fill note is not intuitive. Why not have it say something like "PROBABILITY: 0% FILL ONLY". 
* What does the "PROBABILITY: 0% LATCHING" option do?
  * Was that always there?
  * I can't seem to find any mention of this anywhere.
  * It only seems to exist when there are multiple different samples playing in a single step, why?
  * It doesn't seem to influence behavior at all. These notes seem to always play.
* Is there any faster way to mark a note as a fill note than hold the note and scroll way left on the select encoder? This seems like it requires too much interaction on a per note basis, though it's not totally out of line with the existing level of interaction required for probability I suppose.
* No visual feedback to indicate which notes are fill notes makes patterns difficult to understand.
  * Could have them half-illuminated, but maybe we should reserve that sort of thing for something more important?
  * Could have these notes flash when holding the fill trigger?
* On OLED in the Sync Scaling Action community features menu it should say "SCALE" instead of "SCAL"

EDIT: Updated feedback below based on updates since this issue was created. ([here](#issuecomment-1691147165))

soymonitus:
I will add that "Note row probability" new feature reaches only till 5% probability and it would be awesome if FILL feature is also implemented for full rows. I can visualize myself cloning a row with the same sample as another one and setting Probability:0% to the whole row so i know that the snares i put in the first row are "normal" snares amd if i place any not on the new row they will be "fill" snares. Much quicker to work like this. Pleas @weavermedia can you address this?

soymonitus:
Nevermind @weavermedia it was a one-liner fix
https://github.com/SynthstromAudible/DelugeFirmware/pull/384

m-m-adams:
Yeah it's just a timing thing, when I did note row probability fill didn't exist and vice versa

Changing the label on 0% to fill makese sense to me

m-m-adams:
[...]

PR #399, opened 2023-08-22
Fill mode fixes

weavermedia:
Fixes for issue #375 

- show FILL conditional value instead of "PROBABILITY: 0%" (missed this during the last community merge)
- improve runtime setting for OLED, showing "Sync Scaling" instead of "SCAL"

I don't have the time to fix the latching fill value right now. I looked at the code but don't fully understand the logic behind `prevBase`.

@m-m-adams I added the iteration constant to your note row probability function. Shouldn't have done that in this PR really. Let me know if I should move it to a separate PR.


seangoodvibes:
@weavermedia please push fixes to release 1.0 branch. Thanks

weavermedia:
@seangoodvibes I don't know what that means or how to do it, sorry.

seangoodvibes:
> @seangoodvibes I don't know what that means or how to do it, sorry.

At the top of your PR it shows what you're merging commits from your branch to community.

Click on community and change it to Release 1.0.

As we are now prepping for our first release, a new branch was created as a copy of the community branch and all fixes are to be made in that branch.

The community branch is only for new features that are not part of the first release.

image

weavermedia:
@seangoodvibes Sorry, I don't have time to dig into this right now. If anyone can take over I'm fine with that.

seangoodvibes:
Also is there duplicate work in here with the work in this PR from @soymonitus 

https://github.com/SynthstromAudible/DelugeFirmware/pull/384

m-m-adams:
Yeah but it doesn't matter, they're the same change so no delta

m-m-adams (approved):

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2839,7 +2839,8 @@ void InstrumentClipView::setRowProbability(int32_t offset) {
 	uint8_t probabilityValue = noteRow->probabilityValue;
 	bool prevBase = false;
 	// Covers the probabilities and iterations
-	probabilityValue = std::clamp<int32_t>((int32_t)probabilityValue + offset, (int32_t)0, kNumProbabilityValues + 35);
+	probabilityValue = std::clamp<int32_t>((int32_t)probabilityValue + offset, (int32_t)1,
```
m-m-adams, line 2842: This needs to start from 0 to accomodate fill

m-m-adams (changes requested):

soymonitus:
Given that this hasn't been merged yet, is there time to include a fix for this in this same PR?

https://github.com/SynthstromAudible/DelugeFirmware/issues/412

trappar:
This seems to be hiding the fill latching issue in a way that will make it very hard to detect (this is a bad thing). When this it would have happened before and said "0% LATCHING" it now says "FILL", so in this case there are two entries in the probability list that say "FILL". Super hard to detect since it can appear to simply be a missed input. I don't think this should be merged in this state. I'd prefer it say "FILL LATCHING" even if that option shouldn't exist so that it's easier to fix in a future PR.

soymonitus:
I agree with @trappar in both things.
Fill latching should have its own name so we can address it in the future and so the user knows that only Fill (without latching) works
And agree to address my issue in a different PR

src/deluge/gui/views/instrument_clip_view.cpp
weavermedia: Fixed in https://github.com/SynthstromAudible/DelugeFirmware/pull/399/commits/22de11a907d187964e282964d95c2bb7e12dc217

weavermedia:
@trappar This fix wasn't meant to touch the latching behavior. Unfortunately I don't have time to look dig into it right now. I looked at the prevBase behavior but couldn't figure it out. Happy for someone else to jump in and resolve.

weavermedia:
@soymonitus Fill latching shouldn't exist at all. It's something to do with prevBase but I don't have time to dig into it right now.

And agreed, the note repeat issue should stay separate.

m-m-adams (approved):

m-m-adams:
> This seems to be hiding the fill latching issue in a way that will make it very hard to detect (this is a bad thing). When this it would have happened before and said "0% LATCHING" it now says "FILL", so in this case there are two entries in the probability list that say "FILL". Super hard to detect since it can appear to simply be a missed input. I don't think this should be merged in this state. I'd prefer it say "FILL LATCHING" even if that option shouldn't exist so that it's easier to fix in a future PR.
> 
> And @soymonitus it's probably easier if each issue can be addressed in an individual PR to keep the footprint small.

I think it's good to merge - while displaying fill for both is non-ideal I think it's better than displaying FILL latching when that doesn't exist. The prev base logic is complex and modifying it is better in a separate PR

m-m-adams:
Also on 7seg this still sets the latching dot anyway

sichtbeton:
Is there a setting  to mute a note when pressing “fill / sync scale”?

soymonitus:
> Is there a setting  to mute a note when pressing “fill / sync scale”?

This is known as Not-Fill but it is not implemented yet. It will be in the future

Issue #411, opened 2023-08-22
Automation view selection mode - double tap to switch from selected region to single column selection


trappar:
All of the following describes behavior in selection mode (after tapping on a mod encoder so that a cursor is shown).

Initially you can freely select columns short presses. Once you select a region with a long press (pressing two different columns), you can switch between regions but you can't switch back to selecting just a single column. To switch from region selection to single column selection you have to interrupt the region selection through some other interaction like scrolling, zooming, or exiting and reentering selection mode.

The reason it's like this is so that when a region is selected, you can poll the current value of columns, which is helpful for aligning a selected region with an adjacent column.

To make it possible for a user to easily switch back to selecting columns when a region is selected I'd like to propose that a double tap on a single column switches to selecting that column. So the first tap will show that column's value and the second tap will switch the cursor to that column.

PR #415, opened 2023-08-22
Automation View: Resolve Issue #411 : Pad selection mode enhancement to switch between short and long presses

seangoodvibes:
Per Issue #411 I have enhanced the Pad Selection Mode to permit you to switch back and forth between a Single and Long Press by simply pressing quickly on another column. This is a follow-up to PR #360 which introduced Pad Selection Mode.

If you press and hold another column while a Long Press is selected, it allows you to peak the value of that column.

FYI @trappar


PaulFreund:
Sounds like this should target community :)

trappar (approved):
Approving this from a functionality perspective. The code changes look safe enough too, but maybe someone else should take a look just in case.

m-m-adams (approved):
Looks good to me, and this feels smoother. I think we merge into release so we don't change the UI between this release and the next one

PR #430, opened 2023-08-24
Update issue template to be a form and removed checkboxes so it is no longer a task list

PaulFreund:


seangoodvibes (approved):
Approved, looks good :)

Issue #424, opened 2023-08-23
Automation View - Selection mode value popup can't be held open


trappar:
Currently in automation view in selection mode when a region is selected you can peak at a column's value by tapping/holding a pad within that column. The popup only stays open for a moment (maybe a second) even if you hold the pad.

It would be better if the popup stayed up for as long as the user held down the pad, allowing them to determine when they've absorbed the info rather than dismissing early and the user potentially having to retrigger the pad.

PR #438, opened 2023-08-24
[Ready for Review] Automation View: Resolve issue #424 : Selection mode value popup can't be held open

seangoodvibes:
Resolved issue whereby pop up value does not remain visible for duration that a pad is pressed in automation view

Reported by @trappar under issue #424


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -361,6 +361,7 @@ AutomationInstrumentClipView::AutomationInstrumentClipView() {
 	leftPadSelectedY = kNoLastSelectedPad;
 	rightPadSelectedX = kNoLastSelectedPad;
 	rightPadSelectedY = kNoLastSelectedPad;
```
seangoodvibes, line 363: create a new variable that remembers the knob position value for the last pad selected. This will be used in case the value pop-up needs to be re-drawn (e.g. if you press an audition pad and it hides the value popup by drawing note code / drum name, this will be used when the timer kicks in to re-draw the value for the pad selected).

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1488,6 +1489,11 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 
 			uiNeedsRendering(this);
 		}
+
```
seangoodvibes, line 1492: if you've let go of all pads, reset the lastPadSelectedKnobPos to the "unselected value" and re-draw the parameter name currently being edited

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2602,6 +2608,7 @@ void AutomationInstrumentClipView::selectEncoderAction(int8_t offset) {
 
 flashShortcut:
```
seangoodvibes, line 2610: reset the last selected knob position value when you've changed parameters

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2666,6 +2673,7 @@ void AutomationInstrumentClipView::initPadSelection() {
 	multiPadPressActive = false;
 	leftPadSelectedX = kNoLastSelectedPad;
 	rightPadSelectedX = kNoLastSelectedPad;
```
seangoodvibes, line 2675: initialize

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -3223,6 +3231,13 @@ bool AutomationInstrumentClipView::isOnParameterGridMenuView() {
 //displays patched param names or midi cc names
 void AutomationInstrumentClipView::displayParameterName(int32_t paramID) {
```
seangoodvibes, line 3233: when drawing the parameter name, check that a pad is not currently being pressed and that the last pad selected knob's position was recorded

if a pad is pressed and knob position recorded, display parameter value popup instead of drawing the parameter name and exit the function

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -3272,17 +3290,28 @@ void AutomationInstrumentClipView::displayParameterName(int32_t paramID) {
 //display parameter value when it is changed
 void AutomationInstrumentClipView::displayParameterValue(int32_t knobPos) {
```
seangoodvibes, line 3292: whenever a new parameter value is set, the display parameter value function is called
thus, at the same time, store that value as the last pad selected's value so it can be used for re-drawing the knob position if necessary

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -3272,17 +3290,28 @@ void AutomationInstrumentClipView::displayParameterName(int32_t paramID) {
...
 
+	lastPadSelectedKnobPos = knobPos;
+
 	char buffer[5];
 
 	intToString(knobPos, buffer);
```
seangoodvibes, line 3297: if a pad is currently being held, then we need to use different drawing functions for the popup to prevent the popup from disappearing while we're still holding the pad

otherwise if not being held, use the regular displayPopup function.

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -3272,17 +3290,28 @@ void AutomationInstrumentClipView::displayParameterName(int32_t paramID) {
...
+	}
+	else {
+		display->displayPopup(buffer);
+		setDisplayParameterNameTimer();
+	}
 
 	if (padSelectionOn && !multiPadPressSelected) {
 		indicator_leds::setKnobIndicatorLevel(0, knobPos);
 		indicator_leds::setKnobIndicatorLevel(1, knobPos);
 	}
```
seangoodvibes, line 3314: moved up above

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3486,11 +3486,15 @@ void InstrumentClipView::someAuditioningHasEnded(bool recalculateLastAuditionedN
 		exitUIMode(UI_MODE_AUDITIONING);
 		auditioningSilently = false;
```
seangoodvibes, line 3488: this function is called when an audition pad is let go. if a current pad is being held in the automation view, we don't want the auditioning ended code to override the value popup that is currently being displayed. So I added a check: if you're in the automation view, and a pad is currently being held, do not reset the display.

trappar:
Just tested and it looks like it's working perfectly! 🤩

m-m-adams (approved):

Issue #433, opened 2023-08-24
Automation View - Very rapid region entry/selection not registered


trappar:
### Please describe the problem:

When doing the "long press" action (entering a transition over a region, or selecting a region) if the two button presses come in practically simultaneously then it acts as though the user short pressed the final pad.

### What is the expected behavior?

Pressing two grid pads should always be treated as a "long press" regardless of timing. If a grid pad is being held down and another input comes in, it shouldn't matter if that second press happened 2ms later or 1000ms later.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

Both the nightly and deluge-v1_0_0-dev+2023_08_24-5bef78b.bin

### If possible provide the steps to reproduce the issue and upload additional media:

The easiest way to reproduce this is to just his two pads at the same time.

trappar:
Just checked and it seems like this is working great! 👍

PR #441, opened 2023-08-25
[Ready for Review] Automation View: Resolve Issue #433 : Very rapid region entry / selection not registered

seangoodvibes:
Addressing an issue reported by @trappar with Issue #433 

Refactored code for a single press to its own separate function

Recorded a second edit pad entry into the editPadPress struct for the second pad press in a long press which should resolve issues with deselection.

Also allows for further control of long presses by giving the user freedom to let go of the first pad and continue holding the second pad and still be able to edit the long press.

Fixed a couple errors in the parameter value being displayed in a short press selection versus long press selection.


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1358,8 +1358,7 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 		}
 		// If this is a automation-length-edit press...
 		//needed for Automation
```
seangoodvibes, line 1360: Removed time check between presses, which allows you to press two pads consecutively and it will select those two pads as a long press

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1377,6 +1376,8 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 			}
 
 			if (firstPadX != 255 && firstPadY != 255 && firstPadX != xDisplay) {
```
seangoodvibes, line 1378: record the second press in a long press into the editPadPress struct so that you it will remain active if you let go of the first press.

this enables you to enter a very long continuing sequence of long presses

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1420,33 +1421,11 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 
 		// Or, if this is a regular create-or-select press...
 		else {
```
seangoodvibes, line 1423: refactored this code out into the "recordSinglePadPress" function

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1478,7 +1455,7 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 		}
 		//switch from long press selection to short press selection in pad selection mode
 		else if ((clip->lastSelectedParamID != kNoLastSelectedParamID) && padSelectionOn && multiPadPressSelected
```
seangoodvibes, line 1457: make sure a pad is not currently being held before switching to single press selection

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1491,6 +1468,37 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 	}
 }
```
seangoodvibes, line 1470: new function created as a result of refactoring above. this records a pad press into the editPadPress array struct.

trappar:
Tested and it looks solid to me 👍

Issue #434, opened 2023-08-24
Automation View - Auditioning doesn't follow value from last input


trappar:
### Please describe the problem:

**The three following symptoms are related and take a little pre-explanation.** When not in selection mode, it seems like auditioning always plays the preview at the last entered value. For short presses you hear that value. For long presses it plays at the right press's value. This concept of previewing the last value interacted with seems helpful as it can help you dial in the sound you're looking for with your ears instead of your eyes.

That said, I'm noticing some mildly unexpected behavior surrounding using the audition pads for parameter value preview:

1. When in selection mode, if you select a column it doesn't immediately allow you to preview the value at that selected column. It instead only allows you to preview it if you adjust the value via the mod encoders.
2. When not in selection mode and entering a long press, it always previews the final value. It seems like it would make more sense for it to preview the last entered value - so if you enter the right value and then the left value, then the left would be the one previewed. Having seen the code I know this might be challenging, and it's also so minor it might not be worth it on its own... but...
3. When in selection mode, if you select a region and then edit either the start or end value, it always previews the end value. This is the case even if you are actively adjusting the region start value. This one seems more important than the 2nd item, and probably is due to the same mechanism.

### What is the expected behavior?

The current parameter value should always follow the last touch, regardless of if it's an input due to short press, long press, column selection, region selection, etc...

This allows the audition to be used to preview the sound at the entered value, which has fairly significant utility within this view.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

deluge-v1_0_0-dev+2023_08_24-5bef78b.bin

Issue #435, opened 2023-08-24
Automation View - Sidebar presses not registered while holding automation grid pads


trappar:
### Please describe the problem:

It's not possible to hold down an automation step and then use the sidebar in any way. This issue affects both in and out of selection mode.

(Just to be super clear it works great if you audition and **then** adjust automation. Order is important here)

### What is the expected behavior?

The sidebar should be fully usable while manipulating the automation grid.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

deluge-v1_0_0-dev+2023_08_24-5bef78b.bin

### If possible provide the steps to reproduce the issue and upload additional media:

Hold down a grid cell then try to audition.

PR #445, opened 2023-08-25
[Ready for Review] Automation View: Resolve Issues #434 and #435 : Use of Sidebar + Auditioning of last selected value

seangoodvibes:
This PR resolves two issues reported by @trappar

#434 : audition pads follow value from last input / selection
#435 : sidebar unusable while holding automation pads first

### DIFF
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -1426,7 +1426,7 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 					knobPos = getParameterKnobPos(modelStackWithParam, squareStart) + kKnobPosOffset;
 					indicator_leds::setKnobIndicatorLevel(1, knobPos);
 
-					if (!currentSong->isClipActive(clip)) {
+					if (!playbackHandler.isEitherClockActive()) {
 						if (modelStackWithParam->getTimelineCounter()
 						    == view.activeModControllableModelStack.getTimelineCounterAllowNull()) {
 
@@ -2235,7 +2235,7 @@ void AutomationInstrumentClipView::modEncoderAction(int32_t whichModEncoder, int
 
 						indicator_leds::setKnobIndicatorLevel(whichModEncoder, newKnobPos + kKnobPosOffset);
 
-						if (!currentSong->isClipActive(clip)) {
+						if (!playbackHandler.isEitherClockActive()) {
 							if (modelStackWithParam->getTimelineCounter()
 							    == view.activeModControllableModelStack.getTimelineCounterAllowNull()) {
 
@@ -3034,7 +3034,7 @@ void AutomationInstrumentClipView::handleSinglePadPress(ModelStackWithTimelineCo
 
 			if (modelStackWithParam && modelStackWithParam->autoParam) {
 
-				if (!currentSong->isClipActive(clip)) {
+				if (!playbackHandler.isEitherClockActive()) {
 					if (modelStackWithParam->getTimelineCounter()
 					    == view.activeModControllableModelStack.getTimelineCounterAllowNull()) {
 
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -97,7 +97,7 @@ const uint32_t auditionPadActionUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDI
 
 const uint32_t editPadActionUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDITIONING, 0};
 
-const uint32_t mutePadActionUIModes[] = {UI_MODE_AUDITIONING, 0};
+const uint32_t mutePadActionUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDITIONING, 0};
 
 static const uint32_t verticalScrollUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDITIONING, UI_MODE_RECORD_COUNT_IN,
                                                  0};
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -92,12 +92,16 @@ extern "C" {
 
 using namespace deluge::gui;
```
seangoodvibes, line 94: added UI_MODE_NOTES_PRESSED to the auditionPadAction and mutePadAction UI arrays so that you can use mute and audition pads while holding down a note (or in this case, an automation lane)

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1405,12 +1409,36 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
 					    getParameterKnobPos(modelStackWithParam, getPosFromSquare(leftPadSelectedX)) + kKnobPosOffset;
 					indicator_leds::setKnobIndicatorLevel(0, knobPos);
```
seangoodvibes, line 1411: for the right pad in a long press, the value to display is the value at the very last node, previously I was getting it at the starting node of the last pad.

leveraged the same code I use elsewhere in this view to get the knob position of the last node.

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -1405,12 +1409,36 @@ void AutomationInstrumentClipView::editPadAction(bool state, uint8_t yDisplay, u
...
+					}
+					else {
+						//this will differ for a kit when in note row mode
+						effectiveLength = clip->loopLength;
+					}
+
+					int32_t squareRightEdge = getPosFromSquare(rightPadSelectedX + 1);
+					uint32_t squareStart = std::min(effectiveLength, squareRightEdge) - kParamNodeWidth;
+					knobPos = getParameterKnobPos(modelStackWithParam, squareStart) + kKnobPosOffset;
 					indicator_leds::setKnobIndicatorLevel(1, knobPos);
```
seangoodvibes, line 1428: This sets the mod position to grab the value at the position of the second pad pressed (last pad pressed). This is needed so that when you use the audition pads, it grabs the last param value set.

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2205,6 +2233,13 @@ void AutomationInstrumentClipView::modEncoderAction(int32_t whichModEncoder, int
 
 						indicator_leds::setKnobIndicatorLevel(whichModEncoder, newKnobPos + kKnobPosOffset);
```
seangoodvibes, line 2235: When holding down a long press or selecting a long press (in pad selection mode) and then subsequently adjusting the values of that long press with the mod encoders, you need to update the mod position to grab the param values at the node last edited so that the changes can be heard while auditioning.

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2968,15 +3003,41 @@ void AutomationInstrumentClipView::handleSinglePadPress(ModelStackWithTimelineCo
 				effectiveLength = clip->loopLength;
 			}
```
seangoodvibes, line 3005: adjusted the peek value code while in pad selection mode

when a long press is selected, the peek value at the start position should be the value at the very first node
when a long press is selected, the peek value at the end position should be the very at the very last node
when a long press is selected, the peek value at positions between start and end will be the value at the middle node position

for non-long press selections, the peek value is always the middle node value

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2968,15 +3003,41 @@ void AutomationInstrumentClipView::handleSinglePadPress(ModelStackWithTimelineCo
...
+			else {
+				squareStart = getPosFromSquare(xDisplay);
+				uint32_t squareWidth = instrumentClipView.getSquareWidth(xDisplay, effectiveLength);
+				if (squareWidth != 3) {
+					squareStart = squareStart + (squareWidth / 2);
+				}
 			}
 
 			int32_t knobPos = getParameterKnobPos(modelStackWithParam, squareStart);
 			displayParameterValue(knobPos + kKnobPosOffset);
```
seangoodvibes, line 3030: for the pad selected in pad selection mode, set the mod position to the position corresponding to the value that was displayed. this is needed so that the selected pad's value can be auditioned

seangoodvibes:
Update: need to fix issue reported by @trappar whereby mod position for auditioning should not change during playback as it affects the audio. Just need to add a simply playback check.

trappar:
Seeing two issues currently:
1. Holding an audition pad down and then moving the selection cursor doesn't seem to update the parameter value at all.
2. When not in selection mode, holding both a grid pad + audition and making further adjustments doesn't update the automation value. This also doesn't work when the interaction order is reversed (holding audition, then holding grid pad, and updating value)

seangoodvibes:
> Seeing two issues currently:
> 
> 1. Holding an audition pad down and then moving the selection cursor doesn't seem to update the parameter value at all.
> 2. When not in selection mode, holding both a grid pad + audition and making further adjustments doesn't update the automation value. This also doesn't work when the interaction order is reversed (holding audition, then holding grid pad, and updating value)

Fixed. new version uploading now. Please re-test :D Thanks!

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2120,7 +2150,7 @@ void AutomationInstrumentClipView::modEncoderAction(int32_t whichModEncoder, int
 	ModelStack* modelStack = setupModelStackWithSong(modelStackMemory, currentSong);
```
seangoodvibes, line 2151: updated this UI mode check because it wasn't working if another UI (e.g. auditioning) was active at the time as a note is pressed

trappar:
Retested and it looks like everything is working great! 🚀

m-m-adams (approved):

Issue #447, opened 2023-08-25
Automation View - No Synth/Kit parameter names displayed for 7Seg


seangoodvibes:
### Please describe the problem:

On 7seg, when you've selected a parameter for editing in the automation editor, no parameter name is displayed on the screen. It defaults to scrolling the preset name.

### What is the expected behavior?

Same behaviour as OLED is expected - Display parameter name on the screen when parameter is selected and you're in the automation editor.

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

nightly, Built from release/1.0

### If possible provide the steps to reproduce the issue and upload additional media:

1. Turn on deluge
2. Enter clip
3. Press clip button to enter automation view
4. Select a parameter
5. Look at the 7Seg screen and see that parameter name is not displayed

PR #449, opened 2023-08-25
[Ready for Review] Automation View: Resolve Issue #447 - No Synth/Kit parameter names displayed for 7Seg

seangoodvibes:
This PR addresses issue #447 

As part of working to display parameter names on 7seg, I identified some issues with the strings being displayed on OLED as well. The big string refactoring exercise didn't keep some of the same string changes I had made as part of my Automation View PR.

So as part of this PR, I did some cleanup of the parameter strings and added strings for 7Seg as well (note: strings for 7Seg are the same at the moment as OLED a.k.a. too long for 7Seg, so they need to be shortened.).

I also consolidated all the patched/unpatched param strings in the same section and broken the strings down into patched vs unpatched strings.

Updated: for 7seg, parameter name strings are set to scroll which makes the longer strings easier to understand.


src/deluge/model/settings/runtime_feature_settings.cpp
```diff
@@ -89,9 +89,9 @@ void RuntimeFeatureSettings::init() {
 	                  RuntimeFeatureStateToggle::On);
 	// PatchCableResolution
 	SetupOnOffSetting(settings[RuntimeFeatureSettingType::PatchCableResolution], "Mod. depth decimals",
-	                  "ModDepthDecimals", RuntimeFeatureStateToggle::On);
+	                  "modDepthDecimals", RuntimeFeatureStateToggle::On);
```
m-m-adams, line 92: Won't changing the xml name break existing settings?

seangoodvibes: Removed as discussed

m-m-adams (approved):

PR #450, opened 2023-08-25
Bugfixes for fill mode

### LINKED ISSUE #375: Fill notes feedback

### LINKED ISSUE #385: Note row probability: new inputed notes don't inherit the note Row probability set
This one I guess is for @m-m-adams who developed the feature. When setting note row probability, that value is saved in the NoteRow, but it is only applied to existing notes. I would expect new inputed notes to have the same probability as the one set for the row, that is, inheriting the probability from the NoteRow object.

My assumption is that there must be somewhere in the code, which I don't know, some function that is called when a new note is added to a kit row, so we should take that opportunity to there get the NoteRow they belong to, and see if there is a probability set for that row, and if so, set it to the newly added note

### LINKED ISSUE #442: Fill MIDI command doesn’t turn off with CC value 0

There is an expectation in midi that note on with 0 velocity is treated as note off. The fill global midi command doesn’t turn off from a zero velocity note 

Fill would deactivate

_No response_

7SEG

Built from release/1.0

_No response_

_No response_

### LINKED ISSUE #451: Adding notes with fill mode on does not create fill notes

notes added while fill is held are not fill notes

When adding notes while holding fill they should be created as fill notes

_No response_

7SEG

Built from release/1.0

_No response_

_No response_

m-m-adams:
Fixes #442 
Fixes #451 
Fixes #375
Fixes #385 

Fill notes can no longer have a latching value (it didn't do anything)
Fill notes can now be quickly set by holding fill and adding notes
New notes inherit song fill status or row probability by default 
Fill mode can be properly deactivated over midi CC as well as note on/off

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -5266,7 +5266,7 @@ justDisplayOldNumNotes:
 						Note* note = newNotes.getElement(n);
 						note->pos = (uint32_t)(n * numStepsAvailable) / (uint32_t)newNumNotes * squareWidth;
 						note->length = squareWidth;
-						note->probability = kNumProbabilityValues;
+						note->probability = noteRow->getDefaultProbability(modelStack);
 						note->velocity = ((Instrument*)clip->output)->defaultVelocity;
 						note->lift = kDefaultLiftValue;
 					}
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -2068,7 +2068,7 @@ void InstrumentClipView::adjustProbability(int32_t offset) {
 							}
 							else {
 								// See if there's a prev-base
-								if (probabilityValue < kNumProbabilityValues
+								if (probabilityValue > 0 && probabilityValue < kNumProbabilityValues
 								    && getCurrentClip()->doesProbabilityExist(
 								        editPadPresses[i].intendedPos, probabilityValue,
 								        kNumProbabilityValues - probabilityValue)) {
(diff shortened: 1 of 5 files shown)


seangoodvibes:
Hi, I was able to test these two:

Fill notes can now be quickly set by holding fill and adding notes
New notes inherit song fill status or row probability by default

I think they're working as I expected.

I also reviewed the code and it looks ok. nothing stood out to me as out of place.

Maybe someone else can test the other items you fixed.

seangoodvibes (approved):

soymonitus:
Amazing work in this PR. Congrats!

PR #453, opened 2023-08-26
Bugfix - distinguish fill notes

m-m-adams:
Render fill notes in blue while fill mode active

### DIFF
--- a/src/deluge/util/functions.h
+++ b/src/deluge/util/functions.h
@@ -468,6 +468,39 @@ inline void getBlurColour(uint8_t rgb[], uint8_t fromRgb[]) {
 	rgb[2] = ((uint32_t)fromRgb[2] * 1 + averageBrightness) >> 5;
 }
 
+// clang-format off
+
+constexpr uint32_t IMat[4][4] = {{ONE_Q15, 0, 0, 0},
+							{0, ONE_Q15, 0, 0},
+							{0, 0, ONE_Q15, 0},
+							{0, 0, 0, ONE_Q15}};
+
+constexpr float c = cos(1.0f);
+constexpr float s = sin(1.0f);
+constexpr uint32_t RMat[4][4] = {{(uint32_t)(c*ONE_Q15), 		 			 0, (uint32_t)(s*ONE_Q15), 		 0},
+							 	 {(uint32_t)(s*ONE_Q15), (uint32_t)(c*ONE_Q15), 					0, 		 0},
+							 	 {		  			  0, (uint32_t)(s*ONE_Q15), (uint32_t)(c*ONE_Q15), 		 0},
+							 	 {		  			  0, 		 			 0, 					0, ONE_Q15}};
+
+// clang-format on
+
+inline void xformrgb(uint8_t rgb[], uint8_t fromRgb[], const uint32_t mat[4][4]) {
+
+	uint8_t r = fromRgb[0];
+	uint8_t g = fromRgb[1];
+	uint8_t b = fromRgb[2];
+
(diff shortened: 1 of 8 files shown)


trappar:
This is definitely an improvement, but it's also kind of chaotic. I wonder if just making fill steps the same color but reduced brightness would be better? Or maybe a single color could be dedicated to fill notes? Thinking like how there's the white blur color for notes that can't be accurately rendered at the current zoom level.

m-m-adams:
Yeah I agree, it's ok with a small number of fill notes but rough otherwise. Your first two suggestions don't work though

1. I tried this but they blend in with tails
2. All colours can get used by the main note colour

Making them only change colours while you hold fill might be the best way

seangoodvibes:
> Yeah I agree, it's ok with a small number of fill notes but rough otherwise. Your first two suggestions don't work though
> 
> 1. I tried this but they blend in with tails
> 2. All colours can get used by the main note colour
> 
> Making them only change colours while you hold fill might be the best way

I like option #3, call it "Peek at Fill Notes" hahah. I think we should make more use of this kind of "Peeking" UX to gain more insights into things.

src/deluge/util/functions.h
```diff
@@ -468,6 +468,39 @@ inline void getBlurColour(uint8_t rgb[], uint8_t fromRgb[]) {
...
+constexpr float c = cos(1.0f);
+constexpr float s = sin(1.0f);
+constexpr uint32_t RMat[4][4] = {{(uint32_t)(c*ONE_Q15), 		 			 0, (uint32_t)(s*ONE_Q15), 		 0},
+							 	 {(uint32_t)(s*ONE_Q15), (uint32_t)(c*ONE_Q15), 					0, 		 0},
+							 	 {		  			  0, (uint32_t)(s*ONE_Q15), (uint32_t)(c*ONE_Q15), 		 0},
+							 	 {		  			  0, 		 			 0, 					0, ONE_Q15}};
+
+// clang-format on
+
+inline void xformrgb(uint8_t rgb[], uint8_t fromRgb[], const uint32_t mat[4][4]) {
```
seangoodvibes, line 487: is there a top in this function name? should be "fromrgb" instead of "formrgb"?

m-m-adams: Ya

m-m-adams:
> > Yeah I agree, it's ok with a small number of fill notes but rough otherwise. Your first two suggestions don't work though
> 
> > 
> 
> > 1. I tried this but they blend in with tails
> 
> > 2. All colours can get used by the main note colour
> 
> > 
> 
> > Making them only change colours while you hold fill might be the best way
> 
> 
> 
> I like option #3, call it "Peek at Fill Notes" hahah. I think we should make more use of this kind of "Peeking" UX to gain more insights into things.

Option 3 is now implemented - they peek as blue to match the fill button

soymonitus:
Tested this branch. Works as expected

seangoodvibes (changes requested):
I don't want to scope creep, so I'm willing to approve this as is!

but just wanted to suggest an improvement first to the peeking logic:

1) if a non-fill note was entered on the grid, you then held sync scaling and pressed that non-fill note again, I expected to see that the note would change to fill instead of removing the note and requiring you to re-enter it as fill

2) if you're holding audition and turning select to change the row to fill, I would've expected the the entire row to change to blue while holding audition pad and then switch back to regular colour when you let go of audition

3) same as 1&2: when holding a single note and turning select to change to fill, I expected note colour to change to blue while holding the single note down

seangoodvibes (approved):
As discussed, approving this PR and will track suggested improvements separately.

Cheers

Issue #440, opened 2023-08-25
Automation View - Strange values while editing pitch bend / channel pressure


trappar:
### Please describe the problem:

On a MIDI track in automation view and editing pitch bend or channel pressure, I'm seeing unexpected values specifically in selection mode. See replication instructions for more details.

### What is the expected behavior?

When not in selection mode it looks like I can enter values 0 - 128. I would expect to be able to enter the same values when in selection mode

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

deluge-v1_0_0-nightly+2023_08_24-d114fd1.bin

### If possible provide the steps to reproduce the issue and upload additional media:

1. New song
2. Press MIDI
3. Press clip
4. Press Y (top right grid pad for channel pressure)
5. Press mod encoder to activate pad selection
6. Long press the first column and last column
7. Adjust mod encoders.

Depending on your starting values you will see strange and inconsistent values being reported. Depending on the values that existed before entering pad selection mode I'm seeing all sorts of strange ranges like 113-113 or 29-0. I've even seen situations where I can only enter negative numbers.

PR #456, opened 2023-08-26
[Ready for Review] Automation View: Resolve Issue #440 - Strange values while editing pitch bend / channel pressure

seangoodvibes:
In Issue #440 @trappar identified some weird pop up values being displayed on the screen when editing after touch and pressure. This is because the deluge was displaying pop up values at the same time as I was displaying pop up values in the deluge so it basically giving weird behaviour of overlapping popups and not showing the right value.

I've corrected this by putting a condition in the deluge function that was displaying the popup to say: "if in automation view, let the automation view handle the drawing of popups.

I did the same thing previously in another function called MidiParamCollection::knobPosToParamValue

I didn't realize then that there was another function being used in midi clips to display popups for Pitch Bend and After Touch (ExpressionParamSet::knobPosToParamValue)

### DIFF
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -3083,7 +3083,6 @@ void AutomationInstrumentClipView::handleMultiPadPress(ModelStackWithTimelineCou
 
 			//clear existing nodes from long press range
 			int32_t squareRightEdge = getPosFromSquare(secondPadX + 1);
-			//int32_t clearLength = std::min(effectiveLength, squareRightEdge) - getPosFromSquare(firstPadX);
 
 			//reset interpolation settings to default
 			initInterpolation();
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/automation_instrument_clip_view.cpp
seangoodvibes, line 1: just code cleanup unrelated to this PR, just removed a commented out line

src/deluge/modulation/params/param_set.cpp
```diff
@@ -553,16 +554,19 @@ void ExpressionParamSet::notifyParamModifiedInSomeWay(ModelStackWithAutoParam co
 // Displays text number. This will only actually end up getting used/seen on MIDI Clips, at channel/Clip level - not MPE/polyphonic.
 int32_t ExpressionParamSet::knobPosToParamValue(int32_t knobPos, ModelStackWithAutoParam* modelStack) {
```
seangoodvibes, line 556: If in automation view, don't display popup, let automation view handle it

trappar:
Just tested and it look like this fixes the issue 👍

m-m-adams (approved):

Issue #443, opened 2023-08-25
Automation View - Complex sequence of inputs causing unexpected behavior


trappar:
### Please describe the problem:

There's a complex series of actions which causes a bug in automation view:

1. Start with a new song.
2. Bring up automation view for any parameter.
3. Enter a long press/range starting with the right pad and then moving to the left pad on any pads from x:1 to x:13 (assuming columns are zero indexed).
4. Press any pad in the last column.

All columns before the left column of the range entered in step 2 suddenly shift to the value entered in the last column.

Take careful note of the conditions above. There are some interesting boundaries to these conditions, such as:
* If any note has been entered in the clip, then this bug does not occur.
* If the range in step 2 starts in the first column, then it's not possible to tell if there's any problem, since all columns that would be affected by the bug are already edited.
* If the range in step 2 extends to either of the last two columns, then the bug doesn't happen.
* The bug only seems to happen when all inputs prior to the final step are reverse direction long-presses. There can be multiple of them and the bug will still occur.
* SUPER WEIRDLY you can undo step 4 exactly once and do it again and the bug will happen. After repeating step 4 once, the bug does not occur again.
* Step 4 can also be a long press as long as the initial press is in the final column.

Hopefully all these conditions help to isolate where in the code this might be coming from...

### What is the expected behavior?

Step 3 above should not cause any values other than in the right most column to change.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

deluge-v1_0_0-nightly+2023_08_25-d3171eb.bin

### If possible provide the steps to reproduce the issue and upload additional media:

See instructions above.

trappar:
Just wanted to clarify that this issue is probably only worth fixing because it might point to some underlying problem which could potentially cause other problems. The conditions required to make this happen are so absurdly specific that if we could be 100% sure this didn't cause any other problems, then I'd say just leave it...

PR #458, opened 2023-08-27
[Ready for Review] Automation View: Resolve Issue #443 Interpolation Correction to Eliminate Unexpected Behaviour

seangoodvibes:
Minor change to interpolation before/after code when setting parameter automation values.

I wasn't creating a node immediately right of the node where the value is being set which is why I believe there was an interpolation bug being observed in a weird edge case.

This should resolve it.

### DIFF
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -2852,17 +2852,18 @@ void AutomationInstrumentClipView::setParameterAutomationValue(ModelStackWithAut
 	interpolationAfter = getNodeInterpolation(modelStack, squareStart, false);
 
 	//create a node to the left with the current interpolation status
-	if ((squareStart - kParamNodeWidth) >= 0) {
-		int32_t currentValue = modelStack->autoParam->getValuePossiblyAtPos(squareStart - kParamNodeWidth, modelStack);
-		modelStack->autoParam->setValuePossiblyForRegion(currentValue, modelStack, squareStart - kParamNodeWidth,
+	int32_t squareNodeLeftStart = squareStart - kParamNodeWidth;
+	if (squareNodeLeftStart >= 0) {
+		int32_t currentValue = modelStack->autoParam->getValuePossiblyAtPos(squareNodeLeftStart, modelStack);
+		modelStack->autoParam->setValuePossiblyForRegion(currentValue, modelStack, squareNodeLeftStart,
 		                                                 kParamNodeWidth);
 	}
 
 	//create a node to the right with the current interpolation status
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2852,17 +2852,19 @@ void AutomationInstrumentClipView::setParameterAutomationValue(ModelStackWithAut
 	interpolationAfter = getNodeInterpolation(modelStack, squareStart, false);
```
seangoodvibes, line 2853: For the node to the left, I just cleaned up the code a bit by creating a variable for the left adjacent node being edited.

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2852,17 +2852,19 @@ void AutomationInstrumentClipView::setParameterAutomationValue(ModelStackWithAut
...
 	//create a node to the left with the current interpolation status
-	if ((squareStart - kParamNodeWidth) >= 0) {
-		int32_t currentValue = modelStack->autoParam->getValuePossiblyAtPos(squareStart - kParamNodeWidth, modelStack);
-		modelStack->autoParam->setValuePossiblyForRegion(currentValue, modelStack, squareStart - kParamNodeWidth,
+	int32_t squareNodeLeftStart = squareStart - kParamNodeWidth;
+	if (squareNodeLeftStart >= 0) {
+		int32_t currentValue = modelStack->autoParam->getValuePossiblyAtPos(squareNodeLeftStart, modelStack);
+		modelStack->autoParam->setValuePossiblyForRegion(currentValue, modelStack, squareNodeLeftStart,
 		                                                 kParamNodeWidth);
 	}
```
seangoodvibes, line 2861: For the node to the right, I renamed the previous variable from squareRightEdge to squareNodeRightStart

also, my previous math for the right adjacent node was wrong. Where it said "squareStart + squareWidth" before, it should've said "squareStart + kParamNodeWidth."

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -3114,7 +3116,7 @@ void AutomationInstrumentClipView::handleMultiPadPress(ModelStackWithTimelineCou
 			                            effectiveLength, false);
```
seangoodvibes, line 3117: I don't think this caused any bugs before, but the statement was wrong. the subtraction of kParamNodeWidth should've been outside the std::min() brackets.

trappar:
Just confirmed this does resolve the behavior I observed 👍

m-m-adams (approved):

PR #465, opened 2023-08-28
Remove fmtlib

m-m-adams:
In patch cable set convert the value to an xx.xx float and directly insert to the string

In the menus replace asterisks with the single digit item number (env 1/2, LFO 1/2 etc.)

In the sysex menu convert the random int to a hex string


jamiefaye (commented):
How much impact does this have, in terms of code size reduction? Curious to see if the linker actually does strip out unused stuff.

m-m-adams:
None yet, once I've removed all the usage I can compare before and after removing the dependency

PaulFreund:
I tried that before and iirc without replacement the saving was about 160k but take with a grain of salt since it was some time ago

m-m-adams:
All told saves 100 kb, less than I was hoping but still worthwhile

src/deluge/gui/menu_item/formatted_title.h
```diff
@@ -10,7 +9,10 @@ class FormattedTitle {
 public:
 	FormattedTitle(l10n::String format_str) : format_str_(format_str) {}
 
-	void format(int32_t arg) { title_ = fmt::vformat(l10n::getView(format_str_), fmt::make_format_args(arg)); }
+	void format(int32_t arg) {
+		title_ = l10n::get(format_str_);
+		std::replace(title_.begin(), title_.end(), '*', (char)('0' + arg));
```
PaulFreund, line 14: Nit: The std::replace implementation takes 1696 bytes and if I read correctly does not support numbers > 9. Maybe we can replace it with something custom?

src/deluge/gui/menu_item/patch_cables.cpp
```diff
@@ -78,12 +77,10 @@ void PatchCables::renderOptions() {
...
 		int32_t param_value = cable->param.getCurrentValue();
 		int32_t level = ((int64_t)param_value * 5000 + (1 << 29)) >> 30;
-		if (level >= 100 || level <= -100) {
-			fmt::vformat_to_n(buf + off + 5, 5, "{:4}", fmt::make_format_args(level / 100));
-		}
-		else {
-			fmt::vformat_to_n(buf + off + 5, 5, "{:4}", fmt::make_format_args((float)level / 100));
-		}
+		char buffer[12];
+
```
PaulFreund, line 81: Nit: buffer is dead code

PaulFreund:
Maybe it was 160k for fmt + std::string but I don't remember correctly. 100k is definitely worth it

PaulFreund (approved):

src/deluge/gui/menu_item/formatted_title.h
m-m-adams: Yeah fair enough - I'll go back to a custom one

Issue #455, opened 2023-08-26
Arp triplet information doesn't get copied when copying a clip


pssssssst:
### Please describe the problem:

When i copy a clip using a triplet division on the arpeggiator the copied clip has the same division but without the triplet information.
Triplet divisions on the lfo and delay get copied correctly.

### What is the expected behavior?

The copied clip should have the exact same division 

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

1.0.0-nightly-25ABE47

### If possible provide the steps to reproduce the issue and upload additional media:

1. create a clip with an active arpeggiator utilizing a triplet division.
2. copy the clip in song mode
3. check arp division in new clip

PR #477, opened 2023-08-29
Fix clone clip arp settings

soymonitus:
This fixes bug:
https://github.com/SynthstromAudible/DelugeFirmware/issues/455


bobtwinkles (approved):

m-m-adams:
@soymonitus for future reference to auto link an issue it needs to pretty much exactly say fix/fixes #455 without other words in the middle

PR #480, opened 2023-08-30
Bugfix/global midi overdub

### LINKED ISSUE #407: Continuous overdub global midi command doesn't always overdub
If the audio clip to be looped into is armed red then the continuous overdub command does not actually overdub. The loop command works as expected regardless of the armed clip colour.

Fix: change that when the overdub midi command is received

m-m-adams:
Fixes #407

Enable audio output cloning on all continuous layering commands. This fixes a bug where the layer command only worked properly if the clip was magenta

It's still broken on midi clips (as is official), but just punching record in and out does continuous layering for midi anyway. A larger looper overhaul will be done later

Also makes an enum of looper modes to replace the define

### DIFF
--- a/src/deluge/model/clip/audio_clip.cpp
+++ b/src/deluge/model/clip/audio_clip.cpp
@@ -218,7 +218,7 @@ void AudioClip::finishLinearRecording(ModelStackWithTimelineCounter* modelStack,
 	recorder = NULL;
 }
 
-Clip* AudioClip::cloneAsNewOverdub(ModelStackWithTimelineCounter* modelStackOldClip, int32_t newOverdubNature) {
+Clip* AudioClip::cloneAsNewOverdub(ModelStackWithTimelineCounter* modelStackOldClip, OverDubType newOverdubNature) {
 
 	// Allocate memory for audio clip
 	void* clipMemory = GeneralMemoryAllocator::get().alloc(sizeof(AudioClip), NULL, false, true);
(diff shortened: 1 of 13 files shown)


src/deluge/model/clip/instrument_clip.h
```diff
@@ -248,7 +248,7 @@ class InstrumentClip final : public Clip {
 protected:
 	void posReachedEnd(ModelStackWithTimelineCounter* modelStack);
 	bool wantsToBeginLinearRecording(Song* song);
-	bool cloneOutput(ModelStackWithTimelineCounter* modelStack);
+	bool cloneOutput(ModelStackWithTimelineCounter* modelStack, OverDubType overdubnature);
```
PaulFreund, line 251: nit: capitalization

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -253,7 +253,8 @@ Clip* AudioClip::cloneAsNewOverdub(ModelStackWithTimelineCounter* modelStackOldC
 	return newClip;
 }
 
-bool AudioClip::cloneOutput(ModelStackWithTimelineCounter* modelStack) {
+bool AudioClip::cloneOutput(ModelStackWithTimelineCounter* modelStack, OverDubType overdubNature) {
```
PaulFreund, line 256: Is it intentional you added overdubNature to all cloneOutput functions but don't supply it in the call or use it?

PaulFreund (approved):
Wasn't able to perform a functional test so far but looks like it should work :)

src/deluge/model/clip/audio_clip.cpp
m-m-adams: No - remnant of an earlier approach that turned into a spaghetti mess

PR #487, opened 2023-08-31
Bin size/less inlining

m-m-adams:
GCC9 didn't inline by default and it is a major cause of the binary size change

LTO in combination with inlining can lead to massive size gains without an associated performance increase since the inlining thresholds are based on the number of inlinable functions (which is almost all of them with lto)

This disables inlining and then restores some performance by always inlining the filter code. Small functions and functions called once will still be inlined

### DIFF
--- a/src/deluge/dsp/filter/hpladder.cpp
+++ b/src/deluge/dsp/filter/hpladder.cpp
@@ -85,7 +85,7 @@ void HpLadderFilter::doFilterStereo(q31_t* startSample, q31_t* endSample) {
 		currentSample += 1;
 	} while (currentSample < endSample);
 }
-inline q31_t __attribute__((always_inline)) HpLadderFilter::doHPF(q31_t input, HPLadderState& state) {
+[[gcc::always_inline]] inline q31_t HpLadderFilter::doHPF(q31_t input, HPLadderState& state) {
 	//inputs are only 16 bit so this is pretty small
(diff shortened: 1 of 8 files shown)


src/deluge/dsp/filter/hpladder.cpp
```diff
@@ -85,7 +85,7 @@ void HpLadderFilter::doFilterStereo(q31_t* startSample, q31_t* endSample) {
 		currentSample += 1;
 	} while (currentSample < endSample);
 }
-inline q31_t HpLadderFilter::doHPF(q31_t input, HPLadderState& state) {
+inline q31_t __attribute__((always_inline)) HpLadderFilter::doHPF(q31_t input, HPLadderState& state) {
```
bobtwinkles, line 88: we should probably prefer the standards-compliant syntax for this:

```
[[gcc::always_inline]]
```

instead of

```
__attribute__((always_inline))
```

bobtwinkles (commented):
This definitely helps with the sample performance problems we've been having. Debug builds still cull, but release builds can at least play Stirman's test song (https://discord.com/channels/608916579421257728/991432258286268496/1146194461626474600)

PaulFreund:
Looks good especially with binary size, I didn't see a difference in performance (also backported it to 5f80c5c4). Once we found a fix for the performance delta I would like to add this on top if performance does not degrade from there.

bobtwinkles (approved):

PR #489, opened 2023-08-31
Fixes for SYSEX behavior

bfredl:
- `ENABLE_SYSEX_LOAD` did not take effect due to wrong cmake target type (no longer an `INTERFACE`)
- language got stuck on `seven_segment` after switching from `seven_segment` to `OLED` display type.


bfredl:
added some more logic to redraw the current menu or preset name immediately when pressing the "switch display type" button.

PaulFreund (approved):

PR #494, opened 2023-09-01
[Grid] Fix audio clip behavior

PaulFreund:
* Allow copying audio clips in the same track
* Allow copying audio clips to other audio tracks
* Allow copying audio clips to new tracks
* Fixed Learn/Input+Pad not opening input selector for audio clips


m-m-adams (approved):

soymonitus:
How is this feature working? Could someone elaborate?

PaulFreund:
In session grid so far you got "CANT" when trying to copy audio clips, now it works

soymonitus:
I saw that this allows you to clone an audio clip to a new track unlinked, so both can play at the same time. Something not possible in regular song mode

Issue #482, opened 2023-08-30
Grid View, entering clip flashes "Song" pad before switching back to "Clip" pad


ok-reza:
### Please describe the problem:

When you hold down a clip in Grid View, the 'clip' pad is illuminated. Once you hit 'clip' to enter it, during the animation of entering the clip, 'song' pad is illuminated briefly before permanently switching to 'clip' being illuminated.

### What is the expected behavior?

Like the behavior in Row Song View, the 'clip' pad should remain illuminated during animation into entering the clip.

### Is there a relevant Pull request?

#251

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

2023_08_28-ac1002e

### If possible provide the steps to reproduce the issue and upload additional media:

Described above

PaulFreund:
Please check #500, if I understood you correctly this shouldn't be an issue anymore since clip button is no longer used

stellar-aria:
Closing this since no longer able to replicate. (Fixed)

PR #500, opened 2023-09-02
Grid action modes improvement

### LINKED ISSUE #481: Grid View, display hangs a bit before showing Clip information.

When in Grid View, when you press and hold a clip, the title and type of clip is shown (via display and lit clip; synth, kit, midi, etc).

However, when compared to normal Song View (rows), this isn't as instantaneous, there is a little lag. This makes navigating clips feel more sluggish.

Clip info should appear as fast as Song View (rows).

#251

OLED

nightly

ac1002e

_No response_

PaulFreund:
After feedback in #365 I have been thinking about how to allow easy opening of clips and additional sub states of the grid. 

Here is the description from the new docs:

 - ([#251]) Add new grid session layout to "SONG" mode. All functionality from (classic) row layout applies except for the following:
	 - The data model of rows and grid mode are compatible, you can switch between them freely
	 - In grid mode you will not be able to see multiple clips that are in the same section, only the first one. To make them visible move the clips to other sections
	 - The colored coloumn on the right are all available sections, the columns are automatically filled with the tracks in the same order as in arrangement mode
	 - In session mode hold "SONG" and turn "SELECT" encoder to switch between row layout and grid layout
	 - Compared to rows layout overdub recording and copying clips to arranger is currently not supported
	 - A new menu to select the default Layout has been added in Shift+Selection Encoder -> Defaults -> UI -> Song -> Layout
	 - There are different interaction modes that change how the grid behaves
		- The mode can be changed by clicking on one of the colored pads in the Audition/Section column on the right
		- To permanently switch the mode click on a pad and release, to temporarily switch hold the mode pad and use the grid, the mode will snap back to the current permanent one
		- Green mode
		    - All main pads behave the same as the Mute/Launch pads in rows layout (arm/immediate launch/mute/record/MIDI learn arm)
			- Section pads (left sidebar column) behave the same as in rows layout, in addition Shift+Section will immediate launch all clips in that row
		- Blue mode
			- All main pads behave the same as the main pads in rows layout (open/select/create/delete/MIDI learn)
			- While holding a clip it can be copied to other empty slots by clicking on them, apart from audio/instrument conversion clips are automatically moved to that instrument/track and converted (e.g. Synth to MIDI target)
			- Section pads (left sidebar column) will allow changing repeat count while held

Fixes #481 and #482


soymonitus:
As part of this improvement, would these modes also separate copying clips from muting/unmuting? now when I use several fingers and press several pads at the same time to mute them i get a "TARGET FULL" message (which is because the Deluge thinks I want to copy from one slot to another).

PaulFreund:
Yes that will be separate and multi finger launching will be possible :)

PaulFreund:
PR functionality can now be tested for usability, I will do more thorough testing and feature documentation in the following days before converting from draft to open.

soymonitus:
Tested it and it is a much better experience now!

PaulFreund:
Fixed all found feedback, I'd like to merge it once reviewed to get more feedback and to start trying adding colors on top

soymonitus:
What about allowing tweaking the track effects in the green pad mode? I miss this. Seeing the blue mode as a "programming mode" and the green mode as a performance mode. In both modes editing the FX is useful, in programming for setting the starting settings and during performance for doing  a filter swipe while queueing tracks for mute or unmute.
My suggestion is that in green pad mode holding a pad you can edit FX.

PaulFreund:
Right now you can launch multiple clips at the same time in green mode and the arming or swithing (with shift) happens pad press rather than release. Having the ability to edit FX in green mode would require to change this so the arming/switching will happen on finger up (which would introduce a delay not present in rows layout and require the user to memorize how long to hold the pad until it will no longer arm if they only want to modify FX) and editing FX could only be done for the first or last pressed pad. 

I think these downsides outweight the upside so I didn't implement it. The alternative for performance is to stay in blue mode and hold the green mode pad for launching and then releassing it which will snap back to blue mode.

soymonitus:
Wow nice feature. I didn't know you can hold and release the green pad. I am gonna try this

stellar-aria (approved):
LGTM, no standout logical issues I could see.

wtkdwc:
@PaulFreund went here to ask about editing fx in green mode but see it already answered.. will try the suggested alternative.
anyway loving this mode, especially with the new track colors.
btw could in green mode pressing on empty clip stop the playing clip in that track, as in ableton/force clip modes?

PaulFreund:
Thank you for the feedback. For the next release I'm planning to add the following additional settings:
* Unarm with empty pads: Enable/Disable
* Allow selection in green mode: Enable/Disable (implies arm action invoked after finger up instead of down)
* Default mode: Green/Blue
* Permanent mode switching: Enable/Disable (mode always stays at default)

PR #508, opened 2023-09-04
Improve synth to kit row performance

m-m-adams:
Fixes a null pointer dereference in load_instrument_preset_ui

Adds more safety checks in d_string and memory_region (only for alpha/beta) 

Reduces memory usage for the load synth preset to kit row function


PaulFreund (approved):
Looks good, just don't know the system well enough to forsee if some other code might expect the param manager backup but I hope you checked that that won't crash somewhere else because of weird edge cases

m-m-adams:
> Looks good, just don't know the system well enough to forsee if some other code might expect the param manager backup but I hope you checked that that won't crash somewhere else because of weird edge cases

yeah, it's stored to save settings in case you later change back to the original drum. If you scroll through presets it creates a massive number of param managers though - they're not backed up in the equivalent synth instrument preset loading routine

Issue #506, opened 2023-09-03
MIDI Learn on kit tracks not working


trappar:
### Please describe the problem:

On my deluge, if I try to learn midi on synth track, everything works as expected, but when I try to do the same thing on a kit track, the UI shows that I've successfully mapped a row to a specific note, but then will not play when I trigger that note.

This happens for me uniformly on both DIN and USB inputs.

I asked about this in the nightly-testing channel. @ok-reza and @m-m-adams tested this and @ok-reza saw the same behavior I'm seeing while @m-m-adams was not able to reproduce the issue. We all installed the same version of the firmware on our devices and it didn't change anything. We also went through all of the settings in the MIDI menu and some other settings to see if it might be influenced by a setting stored in flash, but we weren't able to identify any differences between our devices' settings.

### What is the expected behavior?

I should be able to hold `Learn` + `Audition` in a kit and then trigger a note on a MIDI device to map that kit row to that note (this works).

Then I should be able to trigger that same MIDI note and have the kit row be triggered.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly, community, Built from release/1.0

ok-reza:
I can successfully make this work if the midi channel of the external device is set to 2-15, channels 1 & 16 cause the issue.

PR #512, opened 2023-09-04
Bugfixes - mpe channel handling and mono expression downgrading

### LINKED ISSUE #511: All NRPNs change MPE zone

Any NRPN received on channel 1 or 16 will change the number of MPE channels on the associated device

They get ignored (since we can't handle NRPNs yet)

no

OLED

nightly

_No response_

Open up midi MPE settings for a device
Send an RPN/NRPN from that device on channel 1 or 16
Observe the number of midi channels in that zone changing

m-m-adams:
Fix #511 

Also fix other bugs related to #506 - essentially channel 0 was being treated as MPE. The bug arises because if the last MPE channel is 0 there are no MPE channels, but if it is 1 then there are 2 (and opposite for ch 15). The existing code didn't handle this case and treated max MPE channel 0 as if it was a 1 channel MPE zone. These checks are now all done by the appropriate method from midi device. 

Also directs the monophonic expression event y axis to modwheel - this makes more sense than CC74 when collapsing an MPE channel to a single midi channel since most non MPE synths don't usefully use CC74. Similarly sends mod wheel to the y axis expression, allowing it to be mapped on the deluge


src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -226,7 +226,8 @@ void MIDIInstrument::monophonicExpressionEvent(int32_t newValue, int32_t whichEx
 	}
 
 	case 1:
-		midiEngine.sendCC(masterChannel, 74, (newValue >> 25) + 64, channel);
+		//send CC1 for monophonic expression - monophonic synths won't do anything useful with CC74
+		midiEngine.sendCC(masterChannel, 1, (newValue >> 25) + 64, channel);
```
bobtwinkles, line 230: this seems like it has potential to break setups, can/should we make it configurable?

m-m-adams: That's a fair point. I don't know of a non MPE synth that uses CC74 as a general mod destination but they might exist

docs/community_features.md
```diff
@@ -15,7 +15,9 @@ Here is a list of general improvements that have been made, ordered from newest
 - ([#17]) Increase the resolution of "patch cables" between mod sources and destinations.
 
 #### 3.2 - MPE
-- ([#29]) Bugfix to respect MPE zones in kit rows. In the official firmware kit rows with midi learned to a channel would be triggered by an MPE zone which uses that channel. With this change they respect zones in the same way as synth and midi clips.
+- ([#29]) Bugfix to respect MPE zones in kit rows. In the official firmware kit rows with midi learned to a channel would be triggered by an MPE zone which uses that channel. With this change they respect zones in the same way as synth and midi clips. ([512]) adds further fixes related to channels 0 and 15 always getting received as MPE.
```
bobtwinkles, line 18: `512` -> `#512`, and  you need to add an entry at the end of the document to the list of links

bobtwinkles (changes requested):

src/deluge/model/instrument/midi_instrument.cpp
bobtwinkles: (discussed on Discord) - That seems fine, we can add a menu item for this in the future. Required for nymphes, I think.

bobtwinkles (approved):

PR #519, opened 2023-09-06
Added menu to select default FX parameters (Delay: PingPong, Analog, Sync, Reverb: RoomPreset)

soymonitus:
the new menu is at:
Defaults -> FX ->
 Delay ->
    -> PingPong: Disabled, Enabled
    -> Type: Digital, Analog
    -> Sync: Off, 1-bar, 2nd-notes, ...256th-dtted
 Reverb ->
    -> Room size: Small, Medium, Large

This will override any new delay that is created. Instead of defaulting to a value dependent on resolution, for example, 32nd-even (if resolution is 384) or 16th-even (if resolution is 768), this gives control to the user to decide which is the delay setting for all newly created kit rows, synths, and also for global fx for new songs.
The reverb room preset selected will be applied to any newly created song.


soymonitus:
This is ready for review

soymonitus:
Added FIRMWARE_COMMUNITY version as suggested by @stellar-aria

soymonitus:
To add more context, as for the chat had in Discord:
- The values you set in Defaults for Delay and Reverb params are not overriding any older song that had saved
- They also don't affect already saved Synth presets and saved Kits
- They don't affect the kit rows added through synth presets (Audition + Synth within a kit)
- Those new params **only** apply to:
  - Global FX for newly created Songs (Shift + NEW SONG)
  - Affect-All FX for newly created Kits (Shift + Kit)
  - Delay params for newly added samples to a kit (Audition + Kit within a Kit)

m-m-adams:
I am very unsure about committing flash space to this. I think it would be better solved by a defaults.xml file or the ability to edit the kit/song level delay

soymonitus:
I see your point, a defaults.xml is more scalable to add other stuff.

m-m-adams:
Arguably all parameters - the synth created on new+synth and all the global parameters (filter cutoffs, modfx etc) would be useful to have a configurable default

soymonitus:
Closing this, as I will change it, as suggested by @m-m-adams to be several XML files where you can setup even more parameters: one xml for global FX, another for default kit row sounds (when creating a kit, doing shift + kit), and another for default synths (when doing Shift + Synth).

Issue #513, opened 2023-09-04
Iteration Dependence for Whole Kit Row Inconsistency


ok-reza:
### Please describe the problem:

When you engage in adjusting the iteration dependence (fill mode/probability/iteration dependence) the first turn of the encoder changes the value.

### What is the expected behavior?

The first turn of the encoder should actually just display the current value, the proceeding turns will change the value. This would be consistent with how individual note iteration dependence is handled as well as adjusting velocity per note.

### Is there a relevant Pull request?

#384

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

43800e07

### If possible provide the steps to reproduce the issue and upload additional media:

hold kit row audition pad, turn select encoder once. compare with hold a sequenced note, turn select encoder once.

PR #526, opened 2023-09-07
Fix row iteration dependency and also multinote (white note) iteration dependency

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/513

So it behaves the same way as single note iteration change, in which the first encoder turn just shows the current value, and the next encoder moves actually change the value.

Although the code seems to have lots of changes actually the change was quite simple, it was just a matter to use the same trick that Rohan used for the single-note change, which was checking for:

probabilityValue = .... // get current value

// If editing, continue edit
**if (display->hasPopup()) {
   ..... // the code that actually increases or decreases the probabilityValue and applies it to the notes
}**

displayProbability(probabilityValue, prevBase); // the popup is always painted even if the notes don't change, that's why the first encoder turn doesn't change it, just displays it, and the next encoder turns go into the IF statement

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -2161,58 +2161,63 @@ multiplePresses:
 
 		// Decide the probability, based on the existing probability of the leftmost note
 		probabilityValue = editPadPresses[leftMostIndex].intendedProbability & 127;
-		probabilityValue += offset;
-		probabilityValue = std::clamp<int32_t>(probabilityValue, 0, kNumProbabilityValues + kNumIterationValues);
 
-		Action* action = actionLogger.getNewAction(ACTION_NOTE_EDIT, true);
-		if (!action) {
-			return;
-		}
+		// If editing, continue edit
+		if (display->hasPopup()) {
+			Action* action = actionLogger.getNewAction(ACTION_NOTE_EDIT, true);
+			if (!action) {
+				return;
+			}
 
-		// Set the probability of the other presses, and update all probabilities with the actual notes
-		for (int32_t i = 0; i < kEditPadPressBufferSize; i++) {
-			if (editPadPresses[i].isActive) {
+			// Add the offset to the probability
+			probabilityValue += offset;
+			probabilityValue = std::clamp<int32_t>(probabilityValue, 0, kNumProbabilityValues + kNumIterationValues);
 
-				// Update probability
-				editPadPresses[i].intendedProbability = probabilityValue;
+			// Set the probability of the other presses, and update all probabilities with the actual notes
+			for (int32_t i = 0; i < kEditPadPressBufferSize; i++) {
+				if (editPadPresses[i].isActive) {
 
-				int32_t noteRowIndex;
-				NoteRow* noteRow =
-				    getCurrentClip()->getNoteRowOnScreen(editPadPresses[i].yDisplay, currentSong, &noteRowIndex);
-				int32_t noteRowId = getCurrentClip()->getNoteRowId(noteRow, noteRowIndex);
+					// Update probability
+					editPadPresses[i].intendedProbability = probabilityValue;
 
-				ModelStackWithNoteRow* modelStackWithNoteRow = modelStack->addNoteRow(noteRowId, noteRow);
+					int32_t noteRowIndex;
+					NoteRow* noteRow =
+					    getCurrentClip()->getNoteRowOnScreen(editPadPresses[i].yDisplay, currentSong, &noteRowIndex);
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2161,59 +2161,64 @@ void InstrumentClipView::adjustProbability(int32_t offset) {
...
+		if (display->hasPopup()) {
+			Action* action = actionLogger.getNewAction(ACTION_NOTE_EDIT, true);
+			if (!action) {
+				return;
+			}
 
-		// Set the probability of the other presses, and update all probabilities with the actual notes
-		for (int32_t i = 0; i < kEditPadPressBufferSize; i++) {
-			if (editPadPresses[i].isActive) {
+			// Add the offset to the probability
```
m-m-adams, line 2172: nit:redundant comment reduces readability

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2161,59 +2161,64 @@ void InstrumentClipView::adjustProbability(int32_t offset) {
 
 		// Decide the probability, based on the existing probability of the leftmost note
 		probabilityValue = editPadPresses[leftMostIndex].intendedProbability & 127;
-		probabilityValue += offset;
```
m-m-adams, line 2164: the row above this modifies the probability value

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2816,27 +2821,32 @@ void InstrumentClipView::setRowProbability(int32_t offset) {
...
 
 	uint8_t probabilityValue = noteRow->probabilityValue;
-	bool prevBase = false;
-	// Covers the probabilities and iterations
-	probabilityValue = std::clamp<int32_t>((int32_t)probabilityValue + offset, (int32_t)0,
-	                                       kNumProbabilityValues + kNumIterationValues);
 
-	noteRow->probabilityValue = probabilityValue;
+	// If editing, continue edit
+	if (display->hasPopup()) {
```
m-m-adams, line 2828: Doesn't work - the display always has a popup when audition is held

m-m-adams (changes requested):
Testing on OLED and this is non functional

Root problem is the display of the auditioned note code is a popup. You can test this by holding an audition pad and turning select, note that the value will immediately change.

src/deluge/gui/views/instrument_clip_view.cpp
soymonitus: We need to get the current probability and I think that line is the way of getting it. Then modified by the second line

src/deluge/gui/views/instrument_clip_view.cpp
soymonitus: Yeah I know but I don't know how t bypass that. Ideas?

src/deluge/gui/views/instrument_clip_view.cpp
soymonitus: Ok will delete

src/deluge/gui/views/instrument_clip_view.cpp
soymonitus: Replaced the check of display->hasPopup with a boolean variable check, which is reset at the same time you stop auditioning a row

m-m-adams:
This doesn't scale well. You should make an enum for popup type instead of the bool so it can be reused in other code

soymonitus:
> This doesn't scale well. You should make an enum for popup type instead of the bool so it can be reused in other code

Do you mean like an optional parameter to displayPopup() method which is the type of popup (enum), which if not set, it is considered as .Default or .Unknown or whatever, but could have other values like .EditProbability, etc? same optional parameter for hasPopup() function? if not passed, it checks for the .Default enum type, so the rest of the code is unaffected

m-m-adams:
For example make an enum with (none, general, probability). Accept an argument for the popup type on the popup method that defaults to general. When the popup is dismissed set the state to none. Have hasPopup return state!=none, and add a method getPopupType that returns the underlying value. 

This way if future code also needs to set and check specific popups we don't add more bools

soymonitus:
Added new method hasPopupOfType apart from hasPopup
hasPopup is used when you don't care about the popup type, and hasPopupOfType when you want to check if the popup present is of the specified type.
I also took the opportunity to change the adjustVelocity behavior to match the probability one, where if you don't release the note, the popup is not removed.

soymonitus:
@m-m-adams this is ready for review but I don't remember how to change it from Draft to normal

m-m-adams (approved):

Issue #491, opened 2023-08-31
Performance issues with noticable sample cutout


PaulFreund:
### Please describe the problem:

Performance seems to be degraded which is especially noticable for sample heavy tracks since sample cutout is very noticable.

Reproduction is  done with this song: [CORDAE](https://discord.com/channels/608916579421257728/991432258286268496/1146194461626474600)

### What is the expected behavior?

The song can be played in full without cutouts with current Official Beta and e2 generated v4.1.4-alpha3 commit. 

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly, community, Built from community, Built from release/1.0, other

### If possible provide the steps to reproduce the issue and upload additional media:

1. Copy https://discord.com/channels/608916579421257728/991432258286268496 to your Songs folder
2. Load either firmware with Load/Run OLED Release
3. Load song (CORDAE) - Arranger is opening
4. Play the song and listen, tracks are indexed from 1 top to bottom
    * Latest Rohan beta/release or 4.1.4-alpha3 with e2: All tracks play to the end without manual interaction with the device
    * fa9ab593: It is normal that track 3 (rap acapella) is cutting out 8/10 times all other sounds should continue to play through
    * 5f80c5c4: 
        * track 3 (acapella) is cutting out between pad 5 to 7 from the left
        * track 5 (bass) is cutting out at some point
        * track 6 (bells) is cutting out shortly after starting every time
        * track 7 (pad) can also cutout

PaulFreund:
There seem to be three known sources of decreased performance currently:
1. #350 seems to be happening once internal memory allocation for objects in the audio path fails and they are moved to SDRAM
2. From Scons (fa9ab59) to CMake (5f80c5c) there seems to be a quite big delta
3. From e2 (GCC9) to dbt (GCC10/12) there seems to have been a smaller but noticable delta too 

Better performance tracing will be of great use to get back or exceed preexisting performance

Icoustic:
Seeing the same when comparing with official beta on elaborate songs with lots of audio clips/ samples

m-m-adams:
Should we close this out or find a new reference song? Cordae no longer cuts out in nightly builds

stirman:
Confirmed. I can't reproduce this issue with the latest nightly build as of today. Thx devs!!! <3

Icoustic:
> Should we close this out or find a new reference song? Cordae no longer cuts out in nightly builds

[...]

Issue #488, opened 2023-08-31
resampling stops mid song


melokit:
### Please describe the problem:

So,when I want to load a song (load+select+scroll) I have to go from the end to the beginning (backward) because in the other direction (forward) the selection loops over 3/4 songs and cannot to go beyond..
there is also, as for other people apparently, the RESAMPLE function which stops working in full recording (to make a STEM for example). I have to restart the deluge, I can do 2/3 stems again and it starts again

### What is the expected behavior?

-to be able to navigate my songs in both directions
-that the resample function is stable and does not stop in the middle of a recording

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

1.0.0-nightly-b4aeb97

### If possible provide the steps to reproduce the issue and upload additional media:

next time

m-m-adams:
I can't reproduce the first one, maybe your encoder is faulty?


I also can't reproduce the second one, could you add some more detail on what you're resampling? It might be related to the song contents, especially if they're using large samples there are known bugs that were working on fixing

m-m-adams:
I was able to reproduce the second bug using more audio samples

m-m-adams:
This should be fixed now (or at least less common)

Issue #462, opened 2023-08-28
Tracks missing contents in old songs


Icoustic:
### Please describe the problem:

After loading a couple of old songs and playing back i.e. streaming audio files from SD, upon loading another, screen says "SD card error" then the song loads with tracks present but missing contents i.e. no files or colors in clips (when attempting to replicate, problem does not seem to arise when several old songs are loaded without actually playing inbetween). These are large songs with plenty of tracks and samples used.
Issue doesn't happen with a specific song every time, it happens after loading (and playing?) a certain amount of the songs consecutively.
Not having same issue with previous non-community official firmware, on same SD card.

Relevant info: Big SD card, 400GB

I can provide collect media save(s) for some songs if needed

### What is the expected behavior?

Old songs should load like they're supposed to

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

nightly

### What is the firmware name:

d56dfaa

### If possible provide the steps to reproduce the issue and upload additional media:

1. Power on 
2. Load song 1 in given setlist, play a few secs 
3. Load song 2 in setlist, play a few secs
4. Load song 3 in setlist, screen says "SD card error", song loads with missing contents

m-m-adams:
I can't reproduce this - it would be great if you could share some of the songs with issues

Icoustic:
> I can't reproduce this - it would be great if you could share some of the songs with issues

Sure, here's a wetransfer link to 3 songs which should be enough to replicate the issue. The issue can happen with any of the songs, depending on what order they are opened and played in.

I'm guessing this is some memory leak thing becoming apparent due to the songs being big and packed.

https://we.tl/t-h0jjB4pATW

m-m-adams:
Likely related to #350 and #491

Icoustic:
> Likely related to #350 and #491

Seems like it!

PaulFreund:
Can you please retest with tomorrows nightly (probably generated in ~2h)?

Icoustic:
> Can you please retest with tomorrows nightly (probably generated in ~2h)?

Sure 🙂

Icoustic:
> Can you please retest with tomorrows nightly (probably generated in ~2h)?

Yep, its working now. The track I linked in #491 plays like it should/does on the official firmware, and I’m no longer getting the SD card errors 👍

Icoustic:
[...]

PR #528, opened 2023-09-08
simple moving average smoothing on voice culler

m-m-adams:
Smooths the sample count with a two step lag to avoid culling voices due to cpu spikes

Lipstick on a pig fix for #491, #462, #350. Not a fix for #488, that one is caused by failure to allocate a new sample cluster in the recorder


jamiefaye:
This is worth a try. Does it help?

m-m-adams:
It makes a significant difference on the test songs we've been using, one of the main culprits is momentary delays while allocating space for new samples or starting several voices at once

It doesn't help for songs that are actually dsp or RAM limited though

jamiefaye:
> It makes a significant difference on the test songs we've been using, one of the main culprits is momentary delays while allocating space for new samples or starting several voices at once
> 
> It doesn't help for songs that are actually dsp or RAM limited though

Lets get it in then and keep working the other angles too.

PR #533, opened 2023-09-10
Add track coloring to grid view

PaulFreund:
Track color is generated randomly if unset, can be changed in edit (blue) mode by holding a populated clip and rotating the vertical encoder.

Changing track colors: Track color can be changed by holding any populated clip in a column and rotating the vertical encoder (v^). For fine changes to the color press the encoder while turning.

Thank you very much @stellar-aria for the cooperation and color mathing :)


stellar-aria (approved):
LGTM

PR #539, opened 2023-09-12
Randomization & Randomization Variants

tastycode:
This PR introduces randomization for several experiences in the Deluge.

- Synth Browser: Press `Y_ENC`, you'll jump a random number of spots ahead. This breaks up the monotony of hearing the same sequence of presets over and over again. I only get to around 40% of my library of presets because of this.
- Any integer control: e.g. Saturation, Env1Attack. Pressing Y_ENC once the control is on the screen jumps to a random value within the valid range of the control.
- Any list control: e.g. LFO1 Type, jumps to a random element in the list

**Not Implemented**
- Reset to original value. This is tricky, I just tried to make it reset to the original value at the beginning of the session (when beginSession) is called, and can't seem to get that to work. 
- Decimal control randomization. This one is tricky for some reason.
- There's some debug code still laying around somewhere

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/db7ec804-52a3-4d0e-8e32-f58fc3c7cbaf

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -1474,49 +1474,9 @@ ActionResult Browser::buttonAction(hid::Button b, bool on, bool inCardRoutine) {
 		}
 	} else if (b == X_ENC) {
 		if (on && currentUIMode == UI_MODE_NONE) {
-			fileIndexSelected = floor((random() * fileItems.getNumElements()) - 1);
-
-
-	if (scrollPosVertical > fileIndexSelected) {
-		scrollPosVertical = fileIndexSelected;
-	}
-	else if (scrollPosVertical < fileIndexSelected - NUM_FILES_ON_SCREEN + 1) {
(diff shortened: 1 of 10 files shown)


src/deluge/gui/ui/browser/browser.cpp
```diff
@@ -1472,14 +1478,33 @@ ActionResult Browser::buttonAction(hid::Button b, bool on, bool inCardRoutine) {
...
+
+			int32_t jumpCount = validFileItemsCount > 30 ? 30 : validFileItemsCount;
+
+			int32_t nextFileIndexSelected = random() % (jumpCount - 1);
+			int32_t offset = nextFileIndexSelected - fileIndexSelected;
+			fileIndexSelected = nextFileIndexSelected;
+			while (offset != 0) {
+				if (offset < 0) {
+					offset += 1;
+					selectEncoderAction(-1);
```
tastycode, line 1493: I know this is gross. Every other method I tried would result in a corrupted file list. The only choice I had was to pretend that the user scrolled for some random number of times.

tastycode:
Closing to prevent clutter.

PR #542, opened 2023-09-12
Add timestamp functions to Debug::print.

jamiefaye:
Code that prepends a 32 bit timestamp to Debug::print messages. Also includes a RoutineTimer that works like a stopwatch and includes durations.


src/deluge/drivers/uart/uart.h
```diff
@@ -32,7 +32,7 @@
  * and set HAVE_RTT to 0. Though this configuration hasn’t been tested for a while...
  */
 #ifndef ENABLE_TEXT_OUTPUT
-#define ENABLE_TEXT_OUTPUT 0
+#define ENABLE_TEXT_OUTPUT 1
```
m-m-adams, line 35: Is this intentional or did it slip in? The cmake config assumes the default is 0 and sets it to 1 in debug and relwithdebinfo builds

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -247,7 +247,7 @@ Voice* cullVoice(bool saveVoice, bool justDoFastRelease) {
 
 	uint32_t bestRating = 0;
 	Voice* bestVoice = NULL;
-
+	Debug::println("cull");
```
m-m-adams, line 250: Doesn't this lead to duplicated lines? There's printing in the routine function for culls as well

jamiefaye: It is a duplicate that won't go in the final.

src/deluge/drivers/uart/uart.h
jamiefaye: We can drop this part too and let cmake handle matters.

jamiefaye (commented):
Don't be afraid to tear into this with your comments.

src/deluge/io/debug/print.cpp
```diff
@@ -27,11 +27,109 @@ extern "C" {
 
 namespace Debug {
 
+constexpr uint32_t kNumSamplesBetweenReports = 44100;
+
+constexpr bool SysExLoggingEnabled = true;
```
m-m-adams, line 32: can this be a pre processor define? It'll make it easier to configure

src/deluge/io/debug/print.cpp
```diff
@@ -67,4 +166,306 @@ void print(int32_t number) {
...
+    : label(label), timeBase(timeBase), active(false), accumulator(0), count(0), t0(0) {
+}
+
+void AverageVOT::note(uint32_t value) {
+#if ENABLE_TEXT_OUTPUT
+	if (!active) {
+		count = 0;
+		accumulator = 0;
+		active = true;
+		asm volatile("MRC p15, 0, %0, c9, c13, 0" : "=r"(t0) :);
```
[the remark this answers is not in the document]
jamiefaye: It already has one! I should be using that.

src/deluge/io/debug/print.cpp
jamiefaye: I did add a #define SYSEX_LOGGING_ENABLED which is defaulted to false if not defined by the build configuration

src/deluge/io/debug/print.cpp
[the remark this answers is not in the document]
jamiefaye: I added an "always inlined" function that uses a reference to the variable to be updated.

src/deluge/io/debug/print.cpp
m-m-adams: Thanks! I just want to make it easy to find if it needs to be turned on/off

m-m-adams (approved):

PR #548, opened 2023-09-14
GMA related bugfixes

m-m-adams:
Fixes:

- Fixes heap_state.py to handle GMA::get() being inlined
- Fixes a memory leak in loading synths to kit rows
- Fixes the GMA test to work properly again. Confirms that the GMA works in isolation
- Fixes grabNeighboringMemory to steal stealables beyond empty regions instead of just consecutive stealables
- Fixes extend/extendright to not change the stealable header to allocated

Other changes:

- Adds a sequentiality test to ordered multi word arrays (to confirm empty regions are properly ordered)
- Adds additional audioengine logs around allocations and dsp heavy fx

Future work:
- I think grabneighboringmemory can be rewritten to be more efficient


bob_heapstate.patch
bobtwinkles, line 1: Can probably delete this file :sweat_smile:

src/deluge/util/container/array/ordered_resizeable_array_with_multi_word_key.cpp
```diff
@@ -119,3 +119,29 @@ bool OrderedResizeableArrayWithMultiWordKey::deleteAtKeyMultiWord(uint32_t* __re
 		return false;
 	}
 }
+
+void OrderedResizeableArrayWithMultiWordKey::testSequentiality(char const* errorCode) {
```
bobtwinkles, line 123: Can we share the code with `ResizeableArray` somehow instead of duplicating it?

bobtwinkles (commented):
I think the changes look reasonable, but I don't know these systems that well. I think this PR would benefit from a pass to reorganize it in to individually meaningful commits so we can merge without a squash, since there's a large number of changes to core systems that would be helpful to bisect within.

sichtbeton:
> Fixes:
> 
> * Fixes heap_state.py to handle GMA::get() being inlined
> * Fixes a memory leak in loading synths to kit rows
> * Fixes the GMA test to work properly again. Confirms that the GMA works in isolation
> * Fixes grabNeighboringMemory to steal stealables beyond empty regions instead of just consecutive stealables
> * Fixes extend/extendright to not change the stealable header to allocated
> 
> Other changes:
> 
> * Adds a sequentiality test to ordered multi word arrays (to confirm empty regions are properly ordered)
> * Adds additional audioengine logs around allocations and dsp heavy fx
> 
> Future work:
> 
> * I think grabneighboringmemory can be rewritten to be more efficient

Dont understand a word, but it sounds like progress to get the deluge to purr again!

src/deluge/util/container/array/ordered_resizeable_array_with_multi_word_key.cpp
m-m-adams: Not without a restructure to make ResizeableArray inherit from ResizeableArrayWithMultiWordKey instead of the other way around - technically this code would work for single word keys as well, but the single word version doesn't work for multi words

m-m-adams:
> I think the changes look reasonable, but I don't know these systems that well. I think this PR would benefit from a pass to reorganize it in to individually meaningful commits so we can merge without a squash, since there's a large number of changes to core systems that would be helpful to bisect within.

yeah sure, that's a good point

src/deluge/dsp/timestretch/time_stretcher.cpp
```diff
@@ -979,6 +979,8 @@ bool TimeStretcher::hopEnd(SamplePlaybackGuide* guide, VoiceSample* voiceSample,
 bool TimeStretcher::setupNewPlayHead(Sample* sample, VoiceSample* voiceSample, SamplePlaybackGuide* guide,
                                      int32_t newHeadBytePos, int32_t additionalOscPos, int32_t priorityRating,
                                      LoopType loopingType) {
+	//clear the current reasons since setting up the clusters will add new ones
+	voiceSample->unassignAllReasons();
```
bobtwinkles, line 983: Should we do this right at the end and then re-add a single reason? Otherwise there's a window if time where a cluster we're about to need could get stolen, I think.

bobtwinkles: The `setupClusterForPlayFromByte` logic does go right in to assigning reasons so it's probably OK, but it might be worth leaving a note.

m-m-adams: Will do - I think it's ok right now since clusters can't be allocated by the interrupts but a warning is a good idea

m-m-adams: Done

PR #553, opened 2023-09-17
Use l10n strings for mod encoder press actions

### LINKED ISSUE #532: Inconsistency in the display name for the stereo chorus effect on 7SEG units

On a 7SEG unit, when selecting the stereo chorus effect via the TYPE shortcut the display reads "S.CHo". However, when selecting stereo chorus by pressing the gold knob, it reads "StER". 

Stereo chorus should be labelled the same in both cases (doesn't matter which, as long as they are consistent)

_No response_

7SEG

nightly

deluge-v1_0_0-nightly+2023_09_09-799dc8b

_No response_

stellar-aria:
This fixes an oversight where I missed some hard-coded strings in the GlobalEffectable system which were used when the Mod Encoder (gold knobs) were pressed to change FX subtype.

Closes #532.

### DIFF
--- a/src/deluge/model/global_effectable/global_effectable.cpp
+++ b/src/deluge/model/global_effectable/global_effectable.cpp
@@ -17,6 +17,7 @@
 
 #include "model/global_effectable/global_effectable.h"
 #include "definitions_cxx.hpp"
+#include "gui/l10n/l10n.h"
 #include "gui/views/view.h"
 #include "hid/buttons.h"
 #include "hid/display/display.h"
@@ -33,6 +34,7 @@
 #include "storage/storage_manager.h"
 #include "util/misc.h"
 #include <new>
+
 using namespace deluge;
 
 GlobalEffectable::GlobalEffectable() {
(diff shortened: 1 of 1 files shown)


src/deluge/model/global_effectable/global_effectable.cpp
```diff
@@ -183,21 +187,21 @@ bool GlobalEffectable::modEncoderButtonAction(uint8_t whichModEncoder, bool on,
...
+					displayText = l10n::getView(STRING_FOR_HPF);
 					break;
 
 				case FilterType::EQ:
-					displayText = "EQ";
+					displayText = l10n::getView(STRING_FOR_EQ);
 					break;
 				}
-				display->displayPopup(displayText);
+				display->displayPopup(displayText.data());
```
bobtwinkles, line 204: maybe out of scope for this change but `Display::displayPopup` should probably take a `string_view` directly.

stellar-aria: Yeah I was planning on doing that as part of continuing to refactor the displays now that we're post-display unification at some point.

stellar-aria:
Fixed with #553

PR #557, opened 2023-09-19
Improve internal RAM memory utilization

PaulFreund:
# Current state

* make getRegion more secure
* Output "Full internal" popup in debug builds if internal allocation fails
* Remove staticVoices array not in use freeing ~28k
* Add a .nmdump to the binary output folder containing all symbols with their sizes ordered descending
* Added new linker sections and automatic relocation with defines in definitions.h:
    * PLACE_INTERNAL_FRUNK places data and bss objects in the memory before the application, already packed full now (freeing ~144k)
    * PLACE_SDRAM_BSS places data in SDRAM, memory is nulled
    * PLACE_SDRAM_DATA places data in SDRAM, initialized memory is populated (automatic relocation)
    * PLACE_SDRAM_TEXT places functions in SDRAM (automatic relocation). Functions in SDRAM seem to be a lot slower and that might cause unexpected issues so not done
* Placed l10n string into SDRAM, ~30k savings

Total static reduction so far: **188k**

# Some additional info

* Available internal memory in debug V4.1.4-alpha3: 1384992
* Available internal memory in debug release/1.0: 1093440
* Difference to recover: 291552 bytes -> **285k**

# Out of scope 
* Reduce memory footprint of parameters (currently ~8k per output and kit row per clip)
* Size of parameter data per clip and output UnpatchedParamSet 1588 + PatchedParamSet 3084 + PatchCableSet 2592 + Destination*(kMaxNumPatchCables+1) 396
* This amounts to ~299k of memory for 5 clip variations of a kit with 8 rows


src/deluge/memory/memory_region.h
```diff
@@ -75,6 +75,8 @@ class MemoryRegion {
 
 	void writeTempHeadersBeforeASteal(uint32_t newStartAddress, uint32_t newSize);
 	void sanityCheck();
+
+public:
```
m-m-adams, line 79: If they need to be public should these be moved to the first public section instead?

src/deluge/model/settings/runtime_feature_settings.h
```diff
@@ -34,7 +34,7 @@ enum RuntimeFeatureStateToggle : uint32_t { Off = 0, On = 1 };
 // Declare additional enums for specific multi state settings (e.g. like RuntimeFeatureStateTrackLaunchStyle)
 enum RuntimeFeatureStateSyncScalingAction : uint32_t { SyncScaling = 0, Fill = 1 };
 
-/// Every setting needs to be delcared in here
+/// Every setting needs to be delcared in here - Warning - Probably don't use them in the rendering path
```
m-m-adams, line 37: The master comp is already used in the audio path, but it's just 1 check per render. I guess that's ok if we haven't noticed a performance change

m-m-adams (commented):
Left a couple questions but generally this looks ready to merge

src/deluge/memory/memory_region.h
PaulFreund: Agreed, fixed

src/deluge/model/settings/runtime_feature_settings.h
PaulFreund: I checked all the calls to runtimeFeatureSettings.get and all references to RuntimeFeatureSettingType::MasterCompressorFx to make sure and didn't find an occurence that was in the audio path. If there is I can reconsider since the savings are about 5k

PaulFreund:
Got confirmation that this works with old bootloaders from @alex-r-t and verified that in RelWithDebInfo I get the same amount of voices in @m-m-adams culling test :)

m-m-adams (approved):

PR #562, opened 2023-09-22
Convert uses of static_vector in gui to std::vector

stellar-aria:
This reduces the overall memory footprint by reducing the number of template instantiations of static_vector, while the majority of the std::vector uses are simply instances that contain string_views.


stellar-aria:
ugh I thought this was based on release/1.0, I'll resubmit once it's rebased.

PR #573, opened 2023-09-25
POC Sustain Support

tastycode:
I've finally gotten my sustain pedal to work. It's funny how it's actually a physical mechanism, the hammers are lifted off of the strings so they don't mute when the hammer settles back. This is akin to short circuiting the note-off logic. Once the sustain pedal is lifted, all the notes on the piano get muted. We are only sending note-offs for the ones we've tracked are being sustained.

Any way you could try it out and try to figure out where the hole in my logic is that's causing the occasional hung note?

TODO: 
- [ ] Figure out why some notes still hang
- [ ] Consider a max sustain length
- [ ] Figure out why note offs don't record when using sustain
- [ ] Determine highest CPU usage samples for sustain
- [ ] Implement community feature menu

### DIFF
--- a/src/deluge/io/midi/midi_engine.cpp
+++ b/src/deluge/io/midi/midi_engine.cpp
@@ -935,6 +935,8 @@ void MidiEngine::midiMessageReceived(MIDIDevice* fromDevice, uint8_t statusType,
 		// All these messages, we should only interpret if there's definitely a song loaded
 		if (currentSong) {
 
+		// If the SoundEditor is the active UI, give it first dibs on the message
+
 			switch (statusType) {
 
 			case 0x09: // Note on
--- a/src/deluge/model/instrument/melodic_instrument.cpp
+++ b/src/deluge/model/instrument/melodic_instrument.cpp
@@ -22,6 +22,7 @@
 #include "gui/ui/root_ui.h"
 #include "gui/views/instrument_clip_view.h"
 #include "gui/views/view.h"
+#include "hid/display/display.h"
 #include "io/midi/midi_device.h"
 #include "io/midi/midi_device_manager.h"
 #include "memory/general_memory_allocator.h"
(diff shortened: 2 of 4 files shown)


m-m-adams:
I've just done a quick review and I think this might need to be approached differently - there is already mechanisms within noterow to track ignored note offs, and then to release the note if a state changes. 

Look at the ONCE mode logic - basically you'll want to pass note offs as normal, ignore them if sustain is active for that instrument, (setting ignored note off to true), and then when sustain is released stop all notes that have ignored a note off

To answer your questions:

1. Unsure, but this code 'fails closed' in that any bugs will lead to stuck notes
2. I don't think that's necessary - I'd suggest adding to the cull logic so that notes sustained on are deprioritized instead
3. You've stopped the note off before the recording call - you'd need to call recordNoteOff in the sustain branch. But if you use the suggested implementation of ignoredNoteOn it doesn't matter, that will put all sustain functionality behind the seuqencer instead
4. This is an interesting idea but belongs in a cull overhaul (e.g. rank culling by how much stopping a voice will help)
5. Unnecessary IMO, there's no impact on not using sustain

m-m-adams (changes requested):
I've just done a quick review and I think this might need to be approached differently - there is already mechanisms within noterow to track ignored note offs, and then to release the note if a state changes.

Look at the ONCE mode logic - basically you'll want to pass note offs as normal, ignore them if sustain is active for that instrument, (setting ignored note off to true), and then when sustain is released stop all notes that have ignored a note off

tastycode:
> I've just done a quick review and I think this might need to be approached differently - there is already mechanisms within noterow to track ignored note offs, and then to release the note if a state changes.

First off, thank you @m-m-adams 

I'm absolutely happy to do it however people believe it should be done. The one question I have, is do people want this? Is this worth working on?
So at a high level, you are saying the note row implements some functionality around ignoring notes? I wanted to place the logic wherever it makes sense and I wouldn't have thought to look there. Let me poke around and see if I see what you see.

m-m-adams:
Yeah, it's there for one shot samples. If a note off is received in once mode then ignoredNoteOff is set to true, then if the sample mode changes and ignoredNoteOff is true the noterow immediately switches to release

In your case you'd add some logic to treat sustained notes the same way


Personally I'd love this and sostenuto! I'd say it's worth working on for sure

m-m-adams:
Just came across it again, it's actually in voice->noteOff in model/voice/voice.cpp

tastycode:
Closing this temporarily while i go back and work on it. Don't want it cluttering the space.

PR #578, opened 2023-09-28
add missing return statement

m-m-adams:
Missed return statement, worked in debug but not in release


m-m-adams:
Small change - just going to merge myself

Issue #555, opened 2023-09-18
MPE instrument sending to other learned channels while recording 


alifeinbinary:
### Please describe the problem:

While recording, the Deluge is sending MIDI notes and CC data received from a MPE device connected to the USB host port to other MIDI devices connected via DIN MIDI that have been learned. It's like `doingMidiThru = true` while recording. I'm not familiar enough yet with the code base but I think https://github.com/SynthstromAudible/DelugeFirmware/blob/680762bd681c58ad7dcf4acbf45b54394ba980c3/src/deluge/model/instrument/midi_instrument.cpp#L553 may be partially to blame. 

Here's how I have things configured, just to give you a quick idea how I have things set up.

### MIDI devices connected to DIN MIDI via a MIDI hub
Nord - Channels 1 through 4
Elektron Keys - Channels 5 through 8 
Elektron Rytm - Channels 13, 14
Oh-coast - Channel 16

### MPE device connected to USB host MIDI
Osmose > Roland UM-One > USB host

### Deluge settings
Differentiate inputs > Enabled
MIDI Thru > Disabled
MIDI > Devices > DIN = (in: upper > 0, lower > 0 | out: upper > 0, lower > 0)
MIDI > Devices > UM-ONE = (in: upper > 14, lower > 0 | out: upper > 14, lower > 0)

### What is the expected behavior?

It's expected that the MIDI recorded from a MPE device on the USB host port isn't sent to other MIDI devices connected to the Deluge's DIN ports even when there's channel overlap. 

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

1.0.0-DEV-664C410

### If possible provide the steps to reproduce the issue and upload additional media:

1. Turn on
2. Apply the following settings to the Deluge 

- Differentiate inputs > Enabled
- MIDI > MIDI Thru > Disabled
- MIDI > Devices > DIN = (in: upper > 0, lower > 0 | out: upper > 0, lower > 0)
- MIDI > Devices > UM-ONE = (in: upper > 14, lower > 0 | out: upper > 14, lower > 0) or whatever your MPE synth requires

3. Learn MIDI synth connected to DIN ports
4. Create new track, learn MPE synth to it
5. Enable record, press play
6. Play the MPE synth long enough to hear notes being played on the synth connected over DIN MIDI when there's channel overlap between what the MIDI synth receives on and the MPE synth is sending

[...]

PR #584, opened 2023-09-30
Bugfix/midi mpe zone matching

m-m-adams:
Fixes issues I found while attempting to reproduce #555 

Breaks out the device/midi channel/MPE matching code to its own function, notes/ccs/aftertouch/pitch bend now use that and a switch on the result instead of nested if/else/gotos


m-m-adams:
Updated to treat mod wheel as unipolar

PaulFreund:
Missing some MPE understanding but looks legitimate. overall

PaulFreund (approved):

PR #585, opened 2023-09-30
Add cpputest for portable tests

m-m-adams:
Test by:
```
cd tests
cmake .
make
./RunAllTests
```
This is only tested on x86_64 linux, and it builds a 32 bit x86 test binary to avoid 32 bit addressing issues

Currently works but the only tests are that small allocation succeeds and large allocation fails

All source files are included, but only memory and util are currently built. This way mocks can be added to over ride hardware dependent behaviour in the deluge implementations. I've written simple mocks to compile in most cases, and made display->freezewitherror and debug::println work with terminal std out. Mocks are located in /tests/mocks

The main source is modified slightly for easier mocking - in particular there's an ifdef around fixedpoint.h to use standard implementations instead of the asm calls if compiling for x86. I also removed some unnecessary imports, and switched the resizeable array to use deluge dealloc since the GMA singleton is non portable.

This uses a seperate cmakelists file within tests. It can probably be added as a target under the main one later but I was getting pretty confused in the paths and globs


.vscode/launch.json
```diff
@@ -2,40 +2,13 @@
...
-        "/interface/cmsis-dap.cfg",
-        "${workspaceRoot}/contrib/renesas-rz-a1lu.cfg",
-      ],
-      "device": "R7S721020",
-      "runToEntryPoint": "main",
-      "svdFile": "${workspaceRoot}/contrib/rza1.svd",
-      "rttConfig": {
-        "enabled": true,
-      },
+      "miDebuggerPath": "gdb",
```
bobtwinkles, line 11: Did you mean to fully replace this launch config with the run tests one?

src/deluge/util/fixedpoint.h
```diff
@@ -24,6 +25,7 @@ typedef int32_t q31_t;
 #define ONE_Q15 65536
 #define NEGATIVE_ONE_Q31 -2147483648
 #define ONE_OVER_SQRT2_Q31 1518500250
+#if !defined(i386)
```
bobtwinkles, line 28: at some point it would be good to break these out to separate header probably, i can see us wanting to support e.g. aarch64 and x86_64 as well. that can be followup work though.

tests/memory_tests.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+class StealableTest : public Stealable {
+public:
+	void steal(char const* errorCode) { mock().actualCall("steal"); }
+	bool mayBeStolen(void* thingNotToStealFrom) { return true; }
+	int32_t getAppropriateQueue() { return 0; }
+	int32_t testIndex;
+};
+
+bool testReadingMemory(void* address, uint32_t size) {
```
bobtwinkles, line 20: these utilities should be `static` functions to minimize the risk of mystery ODR violations in the future I think.

bobtwinkles: Or if you want to be C++, put them in an anonymous namespace.

tests/mocks/mock_display.cpp
```diff
@@ -0,0 +1,58 @@
+#include "hid/display/display.h"
+#include "hid/display/oled.h"
+#include <iostream>
+
+class mockDisplay : public deluge::hid::Display {
```
bobtwinkles, line 5: nit: `mockDisplay` -> `MockDisplay`

.vscode/launch.json
m-m-adams: Oops, deleted them locally forever ago and forgot that would get checked in. I'll restore them

src/deluge/util/fixedpoint.h
m-m-adams: Makes sense - changing to check for 32 bit arm instead for future flexibility

m-m-adams:
@bobtwinkles I've addressed your changes

PaulFreund (approved):

PR #588, opened 2023-10-03
Bugfix/synth kit e073 s001

### LINKED ISSUE #582: Synth preset drum rows cause duplications when switching between Instrument types

When creating a kit using a synth preset and switching to another instrument type and back, the kit rows are duplicates and mangled, resulting in multiple rows where there should be less.

No new rows are created, as per standard behavior, and the only row that exists is the synth preset.

Also, that E073/S001 does not popup, and that the correct initial Synth Preset number (i.e. 0) is displayed.

Compare this to creating a new empty Kit (using SHIFT + [KIT]) and switching to [SYNTH] and back.

#395

7SEG

nightly

1.0.0-nightly-C56CDB5

- Boot Deluge
- Press SHIFT + [KIT] to create a new kit
- Press BACK to back out to empty kit
- Press the sole Kit row's [AUDITION] + [SYNTH] to load a synth preset (desired preset 0)
  - At this point E073 pops up onto the display, and when it disappears there's no preset number (and the sound generated is an empty synth patch) until scrolled at least once, at which point "1" will show up instead of the actual first preset, "0".
  - Occasionally, S001 will show up and freeze the Deluge
- Press [SELECT] to select a synth preset
- Verify that the Synth-preset Kit row's [AUDITION] pad generates the correct sound
- Press [SYNTH]
- Press [KIT]
  - E073 pops up again
- There are now three Kit rows:
  - The topmost is SYNT000.
  - The second has no name when pressed but still creates a sound (empty synth patch)
  - The bottom-most row (the one that should have been set to a synth preset) is now empty once again

m-m-adams:
Fixes a null pointer de reference if not replacing an instrument

Fixes the first load getting over written by default drum

Fixes lingering drums in memory without note rows

Fix #582 

I believe there is a memory leak in the sound->readFromFile method and memory gets full after ~300 preset loads. That's not fixed by this PR but I'll hopefully get it with expanded cpputest coverage

### DIFF
--- a/src/deluge/storage/storage_manager.cpp
+++ b/src/deluge/storage/storage_manager.cpp
@@ -1406,7 +1406,7 @@ int32_t StorageManager::loadInstrumentFromFile(Song* song, InstrumentClip* clip,
 
 deleteInstrumentAndGetOut:
 		newInstrument->deleteBackedUpParamManagers(song);
-		void* toDealloc = dynamic_cast<void*>(newInstrument);
+		void* toDealloc = static_cast<void*>(newInstrument);
 		newInstrument->~Instrument();
 		GeneralMemoryAllocator::get().dealloc(toDealloc);
 
@@ -1488,7 +1488,7 @@ int32_t StorageManager::loadSynthToDrum(Song* song, InstrumentClip* clip, bool m
 	// If that somehow didn't work...
 	if (error || !fileSuccess) {
 
-		void* toDealloc = dynamic_cast<void*>(newDrum);
+		void* toDealloc = static_cast<void*>(newDrum);
 		newDrum->~Drum();
 		GeneralMemoryAllocator::get().dealloc(toDealloc);
 		return error;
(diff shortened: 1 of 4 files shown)


src/deluge/storage/storage_manager.cpp
```diff
@@ -1488,13 +1488,25 @@ int32_t StorageManager::loadSynthToDrum(Song* song, InstrumentClip* clip, bool m
...
 			return error;
 		}
 	}
-	song->deleteBackedUpParamManagersForModControllable(*getInstrument);
-	(*getInstrument)->wontBeRenderedForAWhile();
+	//these have to get cleared, otherwise we keep creating drums that aren't attached to note rows
+	if (*getInstrument) {
+		song->deleteBackedUpParamManagersForModControllable(*getInstrument);
+		(*getInstrument)->wontBeRenderedForAWhile();
+		void* toDealloc = dynamic_cast<void*>(*getInstrument);
```
stellar-aria, line 1505: should be a static cast

stellar-aria (changes requested):
LGTM pending that one dynamic cast

PR #598, opened 2023-10-07
Bugfix/transpose decimal

### LINKED ISSUE #326: Disabling 'Mod. Depth Decimals' on 7SEG doesn't disable the feature.
This seems to work as intended on OLED, however on 7SEG decimals still remain.

Bin I'm using is: deluge7SEG-v1.0.0-dev+2023.08.09-72926a5 (7SEG release)

m-m-adams:
Fix #326 (just the transpose part)

I don't know that it's worth fixing the decimals always displaying, there's no real downside and it will complicate the code significantly

### DIFF
--- a/src/deluge/gui/menu_item/transpose.h
+++ b/src/deluge/gui/menu_item/transpose.h
@@ -35,7 +35,7 @@ public:
 	uint8_t getPatchedParamIndex() final { return PatchedParam::getPatchedParamIndex(); }
 	uint8_t shouldDrawDotOnName() final { return PatchedParam::shouldDrawDotOnName(); }
 
-	void drawValue() override { display->setTextAsNumber(this->getValue(), shouldDrawDotOnName()); }
+	void drawValue() final { Decimal::drawValue(); }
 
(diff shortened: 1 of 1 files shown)


src/deluge/gui/l10n/seven_segment.cpp
```diff
@@ -380,6 +373,16 @@ PLACE_SDRAM_DATA Language seven_segment{
         {STRING_FOR_NOTES_CLEARED, "CLEAR"},
         {STRING_FOR_AUTOMATION_CLEARED, "CLEAR"},
 
+        /* Strings Specifically for Automation Instrument Clip View
+        * automation_instrument_clip_view.cpp
+        */
+        {STRING_FOR_AUTOMATION, "AUTO"},
+        {STRING_FOR_COMING_SOON, "SOON"},
```
PaulFreund, line 380: Is it intentional that some strings like STRING_FOR_AUTOMATION_OVERVIEW are not translated in 7SEG?

m-m-adams: oops, this is targetting the wrong branch

PaulFreund (approved):

PR #620, opened 2023-10-13
Bugfix/duplicate midi clocks

### LINKED ISSUE #490: USB Midi clock running too fast and inconsistent (double clock)

Attached Deluge to Macbook via USB that is running Protokol (https://hexler.net/protokol). 

Deluge appears to be sending Midi clock from two of the three USB Midi devices it provides, resulting in a double clock that is instable and fluctuating.

grafik

Cross check with Release 4.1.3 firmware shows correct behaviour:
grafik

In reverse when receiving Midi clock something similar happans, clock is roughly doubled and fluctuating.

Discussed with @ilovesteak in #nightly-testing on Discord.
Bug seems to be a regression from splitting the USB-Midi device into three devices. The behaviour can be desirable when using UDB-Midi functionality with multiple receiving devices (MPE and non-MPE).
Proposed solution is a configurable parameter in the Midi Clock Out section with parameters port 1/2/both/off.

The expected behaviour is to send Midi clock only once.
As described, sending clock on both USB-Midi devices can be desirable, thus a user selectable configuration has been proposed.

_No response_

OLED

nightly

deluge-v1_0_0-nightly+2023_08_29-dbaa842.bin

See description.

m-m-adams:
Fix #490 

Creates a per midi device setting to choose whether to send clocks on that device

### DIFF
--- a/src/deluge/gui/menu_item/midi/device_send_clock.h
+++ b/src/deluge/gui/menu_item/midi/device_send_clock.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2014-2023 Synthstrom Audible Limited
+ *
+ * This file is part of The Synthstrom Audible Deluge Firmware.
+ *
+ * The Synthstrom Audible Deluge Firmware is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License as published by the Free Software Foundation,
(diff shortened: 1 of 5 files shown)


stellar-aria (approved):
Standard addition of a runtime variable and relevant MenuItem, LGTM

PR #622, opened 2023-10-14
Minor bugfixes

### LINKED ISSUE #581: SVF Morph can overflow

With high modulation and a high value, SVF morph can overflow to become negative

It saturates or is in a more limited range that can only reach max value with modulation? I'm not really sure yet

_No response_

OLED, 7SEG

nightly, community

_No response_

Set morph to 50, modulate with something

m-m-adams:
Slot browser expects numberEditPos to be -1 as default, on oled it's not initialized so is 0. This caused the code to always take the branch at line 932 assuming the user was editing a numeric song file using the pre-folder slot system

The mod encoder action used to not work if shift was held, making sticky shift more annoying than it should be. I've just commented that out so it's easy to find if I've overlooked a conflicting function. This is the root modencoder action that eventually gets called from all other UIs and views if nothing else has claimed the encoder

Fix #581 by saturating morph

Inline lpf scale input, missed when inline-functions was disabled

Define memset to be faster than the nano.specs version

Move y axis to a new enum to stop MPE<->non MPE crossover

### DIFF
--- a/src/definitions_cxx.hpp
+++ b/src/definitions_cxx.hpp
@@ -818,9 +818,10 @@ enum class ExistenceChangeType {
 enum CCNumber {
 	CC_NUMBER_PITCH_BEND = 120,
 	CC_NUMBER_AFTERTOUCH = 121,
-	CC_NUMBER_NONE = 122,
+	CC_NUMBER_Y_AXIS = 122,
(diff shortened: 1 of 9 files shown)


stellar-aria (approved):
LGTM.

PR #627, opened 2023-10-16
Added optimized low level functions and optimization of VoiceSample::…

PaulFreund:
# Changes

* Added (more) optimized versions of: memset, memcpy, strcmp and memmove, all faster than V4.1.4-alpha3
* GCC automagically replaced the memory moving loop in SampleLowLevelReader::bufferIndividualSampleForInterpolation with memmove which was slower, prevented that (was not happening in V4.1.4-alpha3)

# Performance numbers
All tests conducted with TEST song from @sichtbeton (5 tracks, sample heavy)

## VoiceSample::render performance
Measured with `numSamples = NUM_SAMPLES_FOR_CPU_USAGE_REPORT;` in Ozone tracing only the function call itself for the first 10 frames
* Every rendering pass is now faster than with V4.1.4-alpha3 in the ballpark of 100 or more cycles saved

## Voice count with increasing tempo
Benchmarked by increasing tempo until Deluge is at the edge of dropping the chord voices, higher is better
* 127 - Deluge+firmware+V4p1p3+OLED.bin (Official)
* 132 - feature/microOptimization with smoothing -> can lead to clicks
* 144 - feature/microOptimization with without smoothing

## Sampled performance increase per function compared to V4.1.4-alpha3
* memset: 1.86 times faster
* memcpy: 12.84 times faster


soymonitus:
This is so great

m-m-adams (approved):
Up to you if you want to remove the smoothing in this pr

Issue #583, opened 2023-09-29
Master Compressor render time is slow


m-m-adams:
### Please describe the problem:

The master compressor takes about 100 uS to render, approximately on par with the reverb, 10 synth voices, or 2 sample voices. This leads to earlier culling than on official firmware

<img width="313" alt="image" src="https://github.com/SynthstromAudible/DelugeFirmware/assets/52469396/64d27556-45cb-4e4e-84e3-af28f8426ef9">

### What is the expected behavior?

Master compressor is a simple effect that should be extremely fast

Should be re written with integer math - I think the FPU latency plus cache loss is causing the issues

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from release/1.0

PR #630, opened 2023-10-17
Master compressor

m-m-adams:
RMS calculation is done in floats due to the number of sqrt/logs required, rest of the compressor is integer based on extending the existing sidechain envelopes

UI:
in song or arranger view, affect entire sidechain is a one knob compressor, lowering the threshold and auto increasing make up gain. Pressing sidechain will flash COMP, and switch to showing gain reduction on the top LED meter and input volume on the bottom. They are not identically scaled in order to make better use of the lights. While in this mode the bottom knob (reverb) adjusts the ratio, ranging from about 1/8 to infinity/brick wall limiting

Attack and release can be adjusted on the affect entire attack/release knobs since they're otherwise unused.

I have removed the community settings as there is no longer a downside to the compressor, it's approximately 10x faster than in #583 
image

Todo:

- [x] speed up compressor
- [x] overhaul UI 
- [x] save/restore settings

In a future PR I'll move this into global effectable proper so that it can properly leverage the parameter system and kits can have their own compressor as well. Currently it relies on the fallthrough method for gold knobs without params attached


baymud:
Thanks for digging in to this! As per the video I posted on Discord, I've noticed a few things on 7SEG when testing this compressor: 

- It seems to default to maximum attack/release times when starting a new track, which seems counter intuitive
- It also seems to default to maximum compressor amount (sidechain LEDs are off, but when I go to turn it up it jumps to fully lit and I have to reduce it)
- Numbers on the display for the various parameters would help, especially if you ever want to recall/recreate certain settings, as right now you'd have to match the LEDs as best you could
- The input meter is useful, but the moving LEDs make it hard to tell where you're at when adjusting the ratio (maybe a brief interruption in the input meter or just values on the display)
- When editing ratio you can still press the encoder to change reverb size (not sure if this is a problem, or people may want it, but I found it confusing at first)

m-m-adams:
Thanks for the feedback! Some of those I'll handle in a different PR to switch to the proper deluge parameter management system, but I'd like to get this merged to solidify the audio side before that's done

>     * It seems to default to maximum attack/release times when starting a new track, which seems counter intuitive
> 
>     * It also seems to default to maximum compressor amount (sidechain LEDs are off, but when I go to turn it up it jumps to fully lit and I have to reduce it)

Both are bugs in the interaction between the older compressor code and this one. I'll fix in this PR

>     * LEDs in general seem to reset to off for amount, attack and release once you navigate away and back again (but it doesn't sound like it's resetting the parameters, so maybe just an LED issue)

Tracking but out of scope for this PR, it'll fix itself in the main deluge parameter system

>     * Numbers on the display for the various parameters would help, especially if you ever want to recall/recreate certain settings, as right now you'd have to match the LEDs as best you could
> 
>     * The input meter is useful, but the moving LEDs make it hard to tell where you're at when adjusting the ratio (maybe a brief interruption in the input meter or just values on the display)

Will fix in this PR

>     * When editing ratio you can still press the encoder to change reverb size (not sure if this is a problem, or people may want it, but I found it confusing at first)

I'll fix that too, not sure if it's good or not but it seems counter intuitive

PaulFreund:
Code looks good but need to get rid with that interaction with sidechain :)

m-m-adams:
Gonna see if I can integrate the master comp with the 32->24 bit conversion to optimize the headroom

PaulFreund:
Forgot to press approve but did review it ;)

PR #635, opened 2023-10-19
General CMake cleanup

stellar-aria:
This does some cleanup of the CMake build a bit and makes the built-in firmware semver depend on CMake project version for a single consistent source.

There is now a `MAINTAINER_MODE` option to enable the absolute glut of warnings we actually have lurking about. `-fanalyzer` in particular has quite a bit to say about some uninitialized variables that get used in places...

There should be _no_ changes to the actual outputted firmware.

Please double check that firmware size is consistent and has not changed, if it has I'll revert the built-in CMake LTO removal (which in theory was getting overridden anyways)


PaulFreund:
I started testing and right now for the Release build

```
C:\Data\prj\DelugeFirmware\Official\toolchain\win32-x86_64\arm-none-eabi-gcc\bin\arm-none-eabi-nm.exe: C:/Data/prj/DelugeFirmware/Official/build/Release/deluge.elf: no symbols
```
It seems we have less symbols in our elf file now, not yet sure which change is the reason

PaulFreund:
Unfortunately right now this change costs me 3-5 Voices in the pure synth voice test

PaulFreund:
At least 3 of them are down to 
```C++
    LINKER:--sort-common
    LINKER:--sort-section=name
```

PaulFreund:
Still 1-2 Voices I'd like to account for, need more investigation. Raw rendering seems to be the same speed

PaulFreund (approved):

PR #651, opened 2023-10-27
Bugfix/sidechain stuck

m-m-adams:
Envelope offset wasn't initialized, leading to a bug where if an envelope was used but never had a sidechain hit sent to it its value would be random


m-m-adams:
Fix works, I just want to make sure there aren't any side effects before merging

PR #653, opened 2023-10-27
Enable Stutter Automation

seangoodvibes:
Removed code that disabled ability to record stutter with mod encoder.

Note: we do not know what impact enabling this will have. There must have been a reason it was disabled previously. But let's test and find out!


Icoustic:
So do we know more about what impact it has now? 😃

PR #658, opened 2023-10-29
Added Stutter Rate to Automation View

seangoodvibes:
Stutter Rate parameter can be accessed using the select encoder. It was added as the last parameter in the select encoder list. So you can access it by scrolling select 1 left of Master Volume.


m-m-adams (approved):

seangoodvibes:
> You have a conflict in community features but otherwise good

Resolved thanks!

PR #660, opened 2023-10-30
Add "black box" to hard crash handler

PaulFreund:
In case of a CPU fault the handler will print the last link register address and as many pointers to the code section it can find in the stack data. 

PXL_20231030_151257117

Red is decoration, pink means LR came from ARM USR mode, blue would be SYS mode, stack pointers will alternate green and light blue. 

Every color block is one address, every column is one byte, from left to right highest to lowest byte and from top to bottom lowest to highest bit, so rotating 90 degree will make it easy to read and enter in winodws calculator.

Release ELF will now contain debug symbols (they are stripped in the bin) and it can be disassembled for example with GDB:

```
arm-none-eabi-gdb.exe -batch -ex 'set print asm-demangle' -ex 'file build/Release/deluge.elf' -ex "disassemble 0x200B951C-32,+64"
```


jamiefaye:
I put in a known illegal instruction as a test. Here is what I saw. Looks like it works to me.

jamiefaye:
[](url)

PR #671, opened 2023-11-01
feature: virtual 7SEG on physical OLED screen

bfredl:
Added community feature where this is is one of
- entirely disabled
- activated by toggle shortcut (shift + affect-entire)
- on at boot time (still toggleable)

This setting also "works" on 7SEG hardware, but you need to display the emulated display externally using e.g. the webmidi software.

The exact shortcut can of course be bikeshedded, I just picked something which seemed free (affect-entire doesn't care about shift).

Note: I haven't spent much time on making the display look "nice" but it should be readable enough. IIUC upcoming GFX work by @stellar-aria will support things like font rendering of arbitrary size, where a more stylized display could be rendered as a font later on.


ok-reza:
this seems to be working really well, so awesome !

i made a custom font for the OLED display which i just updated since some of my dimensions were wrong. i would love to see it implemented eventually as it's meant to recreate the 7SEG font as close as possible within the OLED display specs.

regular 7SEG font:
7-SEG_for_OLED_Regular_Screen_Guide_v2

bold 7SEG font:
7-SEG_for_OLED_Bold_Screen_Guide_v2

i also have a slanted version of the regular and bold font as additional options, but can share that at a later time.

bfredl:
Added documentation.

src/deluge/hid/buttons.cpp
```diff
@@ -97,6 +97,16 @@ ActionResult buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) {
 		}
 	}
 
+	if (b == AFFECT_ENTIRE) {
```
m-m-adams, line 100: I'm not sure about making a shortcut for this, I think it would be better triggered by the menu. I imagine most people won't change often and there's a limited number of 2 button shortcuts

bfredl: Could it be a three-button shortcut? Or perhaps designate one 2 button shortcut as the "dev feature" shortcut which could be reused by multiple dev options, but only one at a time?

I like the idea of being able to verify the OLED and 7SEG state of the same sub-menu/sub-state whithout needing to lose all context by exiting the current view; messing with the settings menu system; and then hopefully recreating the exact same sub-state but completely losing the flow of what I was working on.

m-m-adams: That's true and I can definitely see the appeal. A 3 button shortcut seems appropriate to me!

bobtwinkles: bikeshed proposal: shift+learn+affect-entire

bfredl: updated with this combination (shift+learn+affect-entire)

bobtwinkles:
I can get this to crash by trying to switch display mode on the version menu, and there's a few weird things with the combo (it generally seems to misbehave when deep in menus) but I think this is OK to merge for now -- the live switching is mostly a dev feature anyway, and it seems stable enough with the existing emulation.

m-m-adams:
Agreed, odd crashes with this feature shouldn't affect end users (and if it does the answer is "don't do it")

m-m-adams (approved):

ok-reza:
here is an additional guide on implementing the 7seg font
image

i'm not sure how relevant either the horizontal or vertical example will be, but just trying to assist ! i can edit/adjust as requested

PR #677, opened 2023-11-03
fix memory leak in cache manager

### LINKED ISSUE #643: Make sure stealable memory is deallocated at the right time
One way or another stealable sample memory is not deallocated efficiently at the moment. Secondary mechanisms are needed to reclaim memory where it coul already have been freed or queued to be freed. A possible cause is that samples have to many reasons attached to them (usage counter).

What can often be seen before secondary reclaim begins is:
```
stole and grabbed neighbouring stuff too...........
```

This is a reminder to find a solution

m-m-adams:
When stealing exactly one stealable, if that was the first stealable run looked at, the longest runs for the queue would be set to 0. This would last until the next time a cluster was marked as stealable in that queue, or a stealable from that queue was reclaimed as neighboring memory from another queue.

Instead set the longest run to 0xFFFFFFFF to indicate that queue needs to be re checked

Fix #643


m-m-adams:
Also changes to call a correct destructor, and correctly mark wontBeRenderedForAWhile as virtual

PaulFreund (approved):

PR #678, opened 2023-11-03
Fixed Midi Mod Encoder bugs

seangoodvibes:
Fixed a couple bugs with MIDI Clips and the use of Mod Encoders.

1) Fixed mod encoder setting value to max when assigning midi cc to mod encoder

2) Fixed inability to change midi cc value after it was set to max


stellar-aria (approved):

src/deluge/gui/views/view.cpp
```diff
@@ -874,10 +874,8 @@ void View::modEncoderAction(int32_t whichModEncoder, int32_t offset) {
 				//if current knobPos exceeds 127, e.g. it's 128, then it needs to drop to 126 before a value change gets recorded
 				//if newKnobPos exceeds 127, then it means current knobPos was 127 and it was increased to 128. In which case, ignore value change
 				if ((getRootUI() == &instrumentClipView) || (getRootUI() == &automationInstrumentClipView)) {
-					if (clip->output->type == InstrumentType::MIDI_OUT) {
-						if ((knobPos == 64) || (newKnobPos == 64)) {
-							return;
-						}
+					if ((clip->output->type == InstrumentType::MIDI_OUT) && (newKnobPos == 64)) {
```
m-m-adams, line 877: clamp the value instead, this will cause bugs if the offset is greater than 1

std::clamp(newKnobPos, -64, 63)

PR #679, opened 2023-11-03
respect song volume in metronome

m-m-adams:
Modifies metronome to respect song volume

At max volume metronome level is unchanged


stellar-aria:
Closing as included/superseded by #683 so we don't need to rebase that one

PR #682, opened 2023-11-04
adjust buffer sizes and positions to remove clicking

m-m-adams:
Sets num samples to be between the numSamplesBeforeLappingPlayhead, an adjusted number with added samples in dire situations, and the time to the next click

Fixes the latency correction code to correct multiple tx buffers of latency. I'm not clear why this happens but it occurs periodically while songs are playing

### DIFF
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -595,7 +595,8 @@ startAgain:
 	//the max error is 0.3ms. At 100bpm 24ppq it is 25ms per pulse
 	//this works out to a 1% error in the absolute worse case of alternating
 	//no extension and max extension, approximately 10x better than average usb midi accuracy.
-	numSamples = std::max<int32_t>(numSamples, MINSAMPLES);
+	int32_t minSamples = std::min<int32_t>(unadjustedNumSamplesBeforeLappingPlayHead, MINSAMPLES);
+	numSamples = std::max<int32_t>(numSamples, minSamples);
 	numSamplesLastTime = numSamples;
 	memset(&renderingBuffer, 0, numSamples * sizeof(StereoSample));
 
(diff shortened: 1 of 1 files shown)


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -1411,11 +1412,15 @@ void slowRoutine() {
...
-	if (latencyWithinAppropriateWindow >= SSI_TX_BUFFER_NUM_SAMPLES) {
+	while (latencyWithinAppropriateWindow >= SSI_TX_BUFFER_NUM_SAMPLES) {
 		i2sRXBufferPos += (SSI_TX_BUFFER_NUM_SAMPLES << (2 + NUM_MONO_INPUT_CHANNELS_MAGNITUDE));
 		if (i2sRXBufferPos >= (uint32_t)getRxBufferEnd()) {
 			i2sRXBufferPos -= (SSI_RX_BUFFER_NUM_SAMPLES << (2 + NUM_MONO_INPUT_CHANNELS_MAGNITUDE));
 		}
+		latencyWithinAppropriateWindow =
+		    (((rxBufferWriteAddr - (uint32_t)i2sRXBufferPos) >> (2 + NUM_MONO_INPUT_CHANNELS_MAGNITUDE))
+		     - SSI_TX_BUFFER_NUM_SAMPLES)
+		    & (SSI_RX_BUFFER_NUM_SAMPLES - 1);
```
bobtwinkles, line 1423: Am I understanding correctly that this is a busy wait until the I2S ADC gives us enough samples to slam in to the record buffer? The loop could probably benefit from an explanatory comment.

m-m-adams: It's just a bug fix in the existing correction code. If the rx buffer was in the 256-384 range it would correct it one receive buffer of samples, but if it was outside that range to start it would still be wrong.  

Replacing the if with a while loop adds 128 until it's in the 256-384 range relative to the tx buffer since it occasionally lands in the 2000ish range. I think it's caused by rendering getting ahead of the desired (per Rohan's comment) 128 sample latency

stellar-aria (approved):

PR #690, opened 2023-11-06
Add black box pointer output to freeze handler

PaulFreund:
* Fixes _SEGGER_RTT address not being visible in ELF
* Blackbox
    * Fixed stack pointer extraction
    * Added first 4 chars of commit ID in sidebar
    * Added blackbox to freeze handler (yellow sidebar instead of red)
    * Added address output via RTT in RelWithDebInfo build
* Fault handler: Switch to user mode before capture loop to help debugger see stack frames

Implements #688

### DIFF
--- a/src/RZA1/diskio.c
+++ b/src/RZA1/diskio.c
@@ -202,12 +202,14 @@ DRESULT disk_read_without_streaming_first(BYTE pdrv, /* Physical drive nmuber to
 
     BYTE err;
 
-    if (currentlyAccessingCard) {
-		if (ALPHA_OR_BETA_VERSION) {
-			// Operatricks got! But I think I fixed.
-        	FREEZE_WITH_ERROR("E259");
-		}
-	}
+    if (currentlyAccessingCard)
+    {
+        if (ALPHA_OR_BETA_VERSION)
+        {
+            // Operatricks got! But I think I fixed.
+            FREEZE_WITH_ERROR("E259");
+        }
+    }
 
     //uint16_t startTime = MTU2.TCNT_0;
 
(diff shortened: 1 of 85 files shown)


src/RZA1/diskio.c
```diff
@@ -203,7 +203,13 @@ DRESULT disk_read_without_streaming_first(BYTE pdrv, /* Physical drive nmuber to
     BYTE err;
 
     if (currentlyAccessingCard)
-        freezeWithError("E259"); // Operatricks got! But I think I fixed.
+    {
+        if (ALPHA_OR_BETA_VERSION)
+        {
+            // Operatricks got! But I think I fixed.
+            FREEZE_WITH_ERROR("E259");
```
m-m-adams, line 210: I think this should be unguarded? looks like it could lead to file corruption if ignored

src/RZA1/diskio.c
```diff
@@ -247,7 +253,12 @@ DRESULT disk_write(BYTE pdrv, /* Physical drive nmuber to identify the drive */
     BYTE err;
 
     if (currentlyAccessingCard)
-        freezeWithError("E258");
+    {
+        if (ALPHA_OR_BETA_VERSION)
+        {
+            FREEZE_WITH_ERROR("E258");
```
m-m-adams, line 259: same here

src/RZA1/diskio.c
PaulFreund: That check was previously here https://github.com/SynthstromAudible/DelugeFirmware/pull/690/files#diff-130d77e2bdfd7b590ceb981f0e747fabfa079b3ec5b547e8117108db13dee193L83 which is repurposed so the logic didn't change

m-m-adams (approved):

Issue #693, opened 2023-11-06
Master compressor basically just distorting


tobiasphilippen:
### Please describe the problem:

I find it almost impossible to make the master compressor sound good. It starts to distort very quickly, and changing the attack and release settings does not change that significantly. Also, even without any LPF engaged, volume jumps can be very abrupt. I find the compressor basically unusable as it is. Settings also used to be not saved properly, so that when opening a song, compressor setting was 50, and the sound was completely distorted. This seems to have been fixed in the current firmware version.

Also, settings from 0 to 50 for ratio do not make sense.

The compressor implemented by AlterAlter seemed to work fine and was well done from an engineering point of view. Maybe this one could be incorporated again?

### What is the expected behavior?

Compressor not distorting at all if set to less than 50 with a reasonably long release time.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

community

### What is the firmware name:

1.0.0-BETA.5A9C21D

m-m-adams:
It was not well done from an engineering point of view, it used approximately 15 voices worth of CPU and did not integrate to any of the deluge systems. Try turning down the song level and compressor threshold, both will lead to less distortion.

The actual ratio is from 2 to 20 it's just displayed as 0-50 like all other deluge parameters 

What is the issue with the LPF?

tobiasphilippen:
What I meant by engineering is *audio* engineering. Of course, I would not know about how it was done in terms of programming. And I have never used it, just watched the video. So, no, not saved anything with settings from the old compressor.
As to ratio: Well, pitch for example is displayed in semitones and cents – so that the numbers have a musically relevant meaning ... might make sense to have the usual ratios and milliseconds for attack and release, too.
Audio levels are already quite low, but I can try that. Why would lowering the threshold lead to less distortion? When I turn up the upper dial, will this lower the threshold and increase the make up gain at the same time?
Concerning LPF: I just read that there seemed to have been a distortion issue if compressor and LPF was engaged at the same time ...
Thank you.

m-m-adams:
Yes, there is auto makeup gain tied to the threshold and ratio. Due to popular demand to be able to clip, if the song level is above 30 or so it can clip the output. I'm open to tweaking that

The lpf and compressor interaction is fixed, it was a bug in the previous compressor and one of the reasons for the change

m-m-adams:
The song level is also post compressor, lowering it should get rid of any clipping to

[...]

PR #695, opened 2023-11-07
tweak master comp behaviour and parameters

m-m-adams:
Tweak params in response to #693, fix some issues caused by tiny sample windows

Threshold range lowered to be more usable

Maximum attack/release times increased

DC offset removal removed, it no longer works as sample windows can be single samples post #682 

Extra gain reduction while clipping removed, it no longer works without DC offset removal

### DIFF
--- a/src/deluge/model/global_effectable/global_effectable.cpp
+++ b/src/deluge/model/global_effectable/global_effectable.cpp
@@ -311,12 +311,12 @@ int32_t GlobalEffectable::getKnobPosForNonExistentParam(int32_t whichModEncoder,
 				break;
 
 			case CompParam::ATTACK:
-				current = AudioEngine::mastercompressor.getAttack();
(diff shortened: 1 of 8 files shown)


src/deluge/model/global_effectable/global_effectable.cpp
```diff
@@ -295,36 +295,34 @@ bool GlobalEffectable::modEncoderButtonAction(uint8_t whichModEncoder, bool on,
 }
 
 int32_t GlobalEffectable::getKnobPosForNonExistentParam(int32_t whichModEncoder, ModelStackWithAutoParam* modelStack) {
-	int displayLevel = -64;
+	int current;
```
PaulFreund, line 298: Current should be initialized (for example to zero) since there are paths that don't set it

PaulFreund (approved):

PR #701, opened 2023-11-09
Feature/optimize division

PaulFreund:
Updated the PR, should be throughly reviewed to check if I overlooked some loss of accuracy in those conversions


src/deluge/dsp/delay/delay_buffer.cpp
```diff
@@ -133,8 +133,10 @@ void DelayBuffer::setupForRender(int32_t userDelayRate, DelayBufferSetup* setup)
 
 	if (isResampling) {
 
-		setup->actualSpinRate = ((uint64_t)userDelayRate << 24) / nativeRate; // 1 is represented as 16777216
-		setup->divideByRate = 0xFFFFFFFF / (setup->actualSpinRate >> 8);      // 1 is represented as 65536
+		setup->actualSpinRate =
+		    (uint64_t)((float)(userDelayRate << 24) / (float)nativeRate); // 1 is represented as 16777216
```
m-m-adams, line 137: should be a double since it's going into a uint64, this is only 24 bits of accuracy

src/deluge/dsp/delay/delay_buffer.cpp
```diff
@@ -133,8 +133,10 @@ void DelayBuffer::setupForRender(int32_t userDelayRate, DelayBufferSetup* setup)
...
+		setup->divideByRate =
+		    (uint32_t)((float)0xFFFFFFFF / (float)(setup->actualSpinRate >> 8)); // 1 is represented as 65536
```
m-m-adams, line 139: this could use an intermediate variable to avoid converting float to int to float

src/deluge/dsp/filter/filter.h
```diff
@@ -81,7 +81,7 @@ class Filter {
 
 		//this is 1q31*1q16/(1q16+tan(f)/2)
 		//tan(f) is q17
-		divideBy1PlusTannedFrequency = (int64_t)ONE_Q31U * ONE_Q16 / (ONE_Q16 + (tannedFrequency >> 1));
+		divideBy1PlusTannedFrequency = (q31_t)(288230376151711744.0f / (float)(ONE_Q16 + (tannedFrequency >> 1)));
```
m-m-adams, line 84: I'm going to look at this all holistically, I suspect the whole setup routine can be simplified to use float math to calculate the coefficients

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -533,7 +533,8 @@ void AudioClip::render(ModelStackWithTimelineCounter* modelStack, int32_t* outpu
 	// To stop things getting insane, limit to 32x speed
 	//if ((sampleLengthInSamples >> 5) > clipLengthInSamples) return false;
 
-	uint64_t requiredSpeedAdjustment = ((uint64_t)sampleLengthInSamples << 24) / clipLengthInSamples;
+	uint64_t requiredSpeedAdjustment =
+	    (uint64_t)((float)((uint64_t)sampleLengthInSamples << 24)) / (float)clipLengthInSamples;
```
m-m-adams, line 537: double

src/deluge/model/sample/sample_playback_guide.cpp
```diff
@@ -102,10 +102,10 @@ uint64_t SamplePlaybackGuide::getSyncedNumSamplesIn() {
 	}
 	// If following external clock, that could happen
 	// TODO: is that still necessary?
-	return (uint64_t)(lengthInSamples * currentTickWithinSample
-	                  + (uint64_t)timeSinceLastInternalTick * lengthInSamples / timePerInternalTick
-	                  + (sequenceSyncLengthTicks >> 1)) // Rounding
-	       / sequenceSyncLengthTicks;
+	return (uint64_t)(((float)(lengthInSamples * currentTickWithinSample)
```
m-m-adams, line 105: I think this might rely on the integer division rounding but I'm not sure

m-m-adams (commented):
Left a few comments, I'll look into the filter ones

stellar-aria (approved):

PR #711, opened 2023-11-11
Feature: Performance View

seangoodvibes:
# Performance View

20907a86d7c1dbc258b98cb31aced7b2fd7a8055_2_1332x1000

## Description:

Polyend Play inspired Performance View for the Deluge.

Each column represents a different "FX" and each pad/row in the column corresponds to a different FX value.

Specifications:
  - Perform FX menu to edit Song level parameters and enable/disable editing mode.
  - 16 FX columns
    - 1 Param assigned to each column
  - 8 FX values per column
    - Long press pads in a column to change value momentarily and reset it (to the value before the pad was pressed) upon pad release
    - Short press pads in a column to the change value until you press the pad again (resetting it to the value before the pad was pressed)
  - Editing mode to edit the FX values assigned to each pad and the parameter assigned to each FX column
  - Save defaults as PerformanceView.xml file
    - Adjustable default Values assigned to each FX column via "Value" editing mode or PerformanceView.xml
    - Adjustable default Param assigned to each FX column via "Param" editing mode or PerformanceView.xml
    - Adjustable default "held pad" settings for each FX column via Performance View or PerformanceView.xml (simply change a held pad in Performance View and save the layout to save the layout with the held pads).
  - Load defaults from PerformanceView.xml file

#### 2) Short press a pad in the columns to hold that pad's value (parameter and held value is displayed on the screen)

- Note: If you save the layout while pad's are in a "held" state, the next time you re-load that layout it will re-load the held pad's and change the associated parameter's value based on the previous held state.

#### 4) You can edit the default values for each pad in the Performance View by entering a "Value Editing Mode." Here are the instructions for Value Editing mode:

> Enter using Shift + Keyboard button

or 

> Enter using the menu
> 
> - Enter/Exit Editing Mode by pressing down on the Select Encoder to open up the Perform FX Menu
> - In the Perform FX menu, select the "Editing Mode" sub menu and press down on Select Encoder to enter the menu
> - Change Editing Mode to "Value"

Value Editing Mode:
- While in the Value Editor, pressing any pad on the grid will open up the menu corresponding to the parameter for that pad. It will display the current value assigned to that pad when you pressed it.
- You can edit that pad's value by turning the select encoder while selecting a single pad (so it's highlighted white) or holding down on that pad. The updated value will be reflected on the display.
- After you have edited a value, the Save button will start flashing to indicate that you have "Unsaved" changes. Press the save button to save your changes. Once saved, the Save button will stop blinking. To avoid being distracting, the save button will only blink when you are in editing mode. If you exit editing mode it will stop blinking and if you re-enter editing mode it will blink again to remind you that you have unsaved changes.
- You can re-load saved changes using the Load button. This will cause you to lose unsaved changes and the Save button will stop blinking.
- Defaults are saved in an XML file on your SD card called "PerformanceView.XML" - deleting this file will cause the PerformanceView to revert back to its regular default values for each pad.
- Once you are done with editing mode, repeat the steps above to set editing mode back to "Disabled" and exit out of the menu to use Performance View in its regular state

#### 5) You can edit the Parameter assigned to each FX column by entering a "Param Editing Mode." Here are the instructions for Param Editing mode:

> Enter using Shift + Keyboard button. You will need to cycle passed the Value Editing Mode to enter the Param Editing Mode.

or 

> Enter using the menu
> 
> - Enter/Exit Editing Mode by pressing down on the Select Encoder to open up the Perform FX Menu
> - In the Perform FX menu, select the "Editing Mode" sub menu and press down on Select Encoder to enter the menu
> - Change Editing Mode to "Param"

Param Editing Mode:
- Param Editing will show you an overview of the parameters that are assignable to the FX columns in Performance View by lighting dimly the shortcut pads for these parameters.
- Parameters that have been assigned to one or more columns will have their shortcut pads lit up bright white.
- Press on a shortcut pad to illuminate the FX columns that that parameter has been assigned to.
- While holding a shortcut pad, press on the FX columns to assign or unassign a parameter to/from a column.
- Press <> + back to clear all existing Parameter assignments.
- When a Parameter has not been assigned to a column, that column will be lit grey and be unusable in Performance View until you assign a Parameter. This applies to editing the values for that FX column as well (assign a Parameter first, then you can edit the values).
- Parameters are saved to PerformanceView.xml. You can manually edit the Parameters in the xml as well, but you must use the exact Parameter names. It is recommended to save a fresh PerformanceView.xml and back it up so you have a record of the Parameter Names.

## Default FX and Colour Assignments
* The default Param and Colour Assignments for each FX column in Performance View are as follows
* Note: if you re-arrange the parameters to different FX columns, they will bring the colours noted below with them.

Columns:

Red:
1 = LPF Cutoff
2 = LPF Resonance

Orange:
3 = HPF Cutoff
4 = HPF Resonance

Yellow:
5 = Bass EQ
6 = Treble EQ

Light Green:
7 = Reverb

Light Blue:
8 = Delay Amount
9 = Delay Rate

Light Pink:
10 = MOD-FX Rate
11 = MOD-FX Depth
12 = MOD-FX Feedback
13 = MOD-FX Offset

Dark Pink:
14 = Decimation
15 = Bitcrush

Dark Blue:
16 = Stutter Rate + Stutter Trigger

## PR Summary

Summary of additions provided as part of this PR:
- New performance view (accessible using the Keyboard button from Song View and Arranger View)
- Render "Song View" on the display in Song View
- Render "Arranger View" on the display in Arranger View
- Add new Song FX menu to Song View and Arranger View (accessed by pressing Select Encoder)
- Adjusted Quantized Stutter implementation from PR #357 so that it does not need to change the value of the stutter rate parameter in order to do it's quantized stutter calculation


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -256,7 +270,8 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 
 	// Back button
 	else if (b == BACK) {
-		if (currentUIMode == UI_MODE_NONE || currentUIMode == UI_MODE_AUDITIONING) {
+		if (currentUIMode == UI_MODE_NONE || currentUIMode == UI_MODE_AUDITIONING
+		    || getRootUI() == &performanceSessionView) {
```
seangoodvibes, line 274: This lets you use back button while stuttering

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -274,23 +289,37 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 
 	// Save button
 	else if (b == SAVE) {
-		if (on && currentUIMode == UI_MODE_NONE && !inSettingsMenu() && !editingCVOrMIDIClip()
-		    && currentSong->currentClip->type != CLIP_TYPE_AUDIO) {
+		if (on && (currentUIMode == UI_MODE_NONE || getRootUI() == &performanceSessionView) && !inSettingsMenu()
```
seangoodvibes, line 292: this lets you use save button while stuttering (in performance view only)

src/deluge/gui/views/view.cpp
```diff
@@ -1164,7 +1165,7 @@ void View::modButtonAction(uint8_t whichButton, bool on) {
 	if (activeModControllableModelStack.modControllable) {
 		if (on) {
 
-			if (isUIModeWithinRange(modButtonUIModes)) {
+			if (isUIModeWithinRange(modButtonUIModes) || (getRootUI() == &performanceSessionView)) {
```
seangoodvibes, line 1168: this lets you use mod buttons while stuttering (only in performance view)

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1139,6 +1139,12 @@ int32_t ModControllableAudio::getStutterRate(ParamManager* paramManager) {
 	UnpatchedParamSet* unpatchedParams = paramManager->getUnpatchedParamSet();
 	int32_t paramValue = unpatchedParams->getValue(Param::Unpatched::STUTTER_RATE);
```
seangoodvibes, line 1141: the changes below adjust the quantized stutter implementation so that it does not need to change the parameter value in order to do it's calculations. When you change the parameter value and re-center it to 0, it changes what is displayed in the menu and on the mod encoder led's, which is not accurate. This caused problems when wanting to trigger stutter and adjust the value in the menu in performance view.

seangoodvibes:
Marked PR as ready for review. Further development of this view will occur as part of subsequent PR's.

bobtwinkles (commented):
I haven't reviewed the main performance view code yet, but here's some minor stuff in the mean time

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+	backupPerformanceLayout();
+	resetPerformanceView(modelStack);
+	readDefaultsFromFile();
+	logPerformanceLayoutChange();
+	updateLayoutChangeStatus();
+}
+
+//read defaults from XML
+void PerformanceSessionView::readDefaultsFromFile() {
```
bobtwinkles, line 1854: I'm a little nervous about this inline sdcard work, but setting up the machinery to defer it seems like something that should be done in followup work if it turns out to be a problem -- since we're just loading 1 cluster from the filesystem root *hopefully* it's fairly unlikely to click.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+	}
+
+	//disable button presses for Vertical encoder
+	//disable back button press since undo doesn't work well in this view atm.
+	else if (b == Y_ENC) { //|| (b == BACK)) {
+		goto doNothing;
+	}
+
+	else {
+notDealtWith:
```
bobtwinkles, line 1023: no need for this label, I think

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+		//if you're already in soundEditor, check if you're in the right menu
+		else if (soundEditor.getCurrentMenuItem()
+		         != paramShortcutsForSongView[lastSelectedParamShortcutX][lastSelectedParamShortcutY]) {
+			goto potentialShortcutPadAction;
+		}
+		//otherwise no need to do anything as you're already displaying the menu for the parameter
+		else {
+			return;
+		}
+potentialShortcutPadAction:
```
bobtwinkles, line 1144: Should use `||` in the `if` statements above and remove the gotos.

src/deluge/gui/views/performance_session_view.h
```diff
@@ -0,0 +1,202 @@
...
+	void savePerformanceViewLayout();
+	void loadPerformanceViewLayout();
+	bool defaultEditingMode;
+	bool editingParam; //if you're not editing a param, you're editing a value
+	bool justExitedSoundEditor;
+
+	//public so Action Logger can access it
+	void updateLayoutChangeStatus();
+	PadPress lastPadPress;
+	FXColumnPress FXPress[kDisplayWidth];
```
bobtwinkles, line 116: nit: `fxPress` instead of `FXPress`

src/deluge/model/consequence/consequence_performance_layout_change.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#pragma once
+#include "gui/views/performance_session_view.h"
+#include "model/consequence/consequence.h"
+#include <cstdint>
+
+class ConsequencePerformanceLayoutChange final : public Consequence {
+public:
+	ConsequencePerformanceLayoutChange(PadPress(*padPressBefore), PadPress(*padPressAfter),
```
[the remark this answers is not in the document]
bobtwinkles: Oh the parenthesis are fine syntactically, though stylistically I'd still remove them.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+				if (!isParamStutter(layoutForPerformance[xDisplay].paramKind, layoutForPerformance[xDisplay].paramID)) {
+					FXPress[xDisplay].padPressHeld = true;
+					FXPress[xDisplay].timeLastPadPress = AudioEngine::audioSampleTimer;
+
+					backupXMLDefaultFXPress[xDisplay].padPressHeld = FXPress[xDisplay].padPressHeld;
+					backupXMLDefaultFXPress[xDisplay].timeLastPadPress = FXPress[xDisplay].timeLastPadPress;
+				}
+			}
+		}
+		if (FXPress[xDisplay].padPressHeld) {
```
bobtwinkles, line 2008: This can break if the XML tags are out of order which is probably not what we want -- instead, we should automatically set `padPressHeld` if any of the relevant tags appear.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+		storageManager.exitTag();
+	}
+}
+
+void PerformanceSessionView::readDefaultFXHoldStatusFromFile(int32_t xDisplay) {
+	char const* tagName;
+	//loop through the hold tags
+	while (*(tagName = storageManager.readNextTagOrAttributeName())) {
+		//<status>
+		if (!strcmp(tagName, STRING_FOR_PERFORM_DEFAULTS_HOLD_STATUS_TAG)) {
```
bobtwinkles, line 1996: I don't think we want to use the `STRING_FOR_*` constants here, instead the prevailing style seems to be just having the tag name directly. This would matter especially if we release translations.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+		memcpy(&backupXMLDefaultLayoutForPerformance[xDisplay], &defaultLayoutForPerformance[xDisplay],
+		       sizeParamsForPerformance);
+	}
+}
+
+void PerformanceSessionView::readDefaultFXValuesFromFile() {
+	char const* tagName;
+	char tagNameFX[5];
+	tagNameFX[0] = 'F';
+	tagNameFX[1] = 'X';
```
bobtwinkles, line 1904: Should also have `tagNameFX[2] = 0` here, arrays are not auto-initialized to 0.

soymonitus:
My 2 cents regarding performance mode. I would have expected that if I hold a LPF pad to close the filter, that as I slide my finger through the pads up to down, I could close the filter gradually till I release and let go to come back to fully open filter, but I can't. It seems that you can only hold a LPF value, then release, then hold the next one, etc. Same with quick tap behavior. I white pad is shown with the value you just set but if I click the pad below it, it doesn't change the filter value (and white pad) to the pad below, I have first to unset the white one to set a new one below. It doesn't allow for a intuitive and quick performance.

soymonitus:
> Can you check again please? You can absolutely set another white pad without unsetting the previous white pad. Remember: quick tap to set white pad is truly a quick tap.
> 
> Also you can definitely move from pad to pad to do exactly what you said: move from top pad to bottom pad to gradually close filter and then let go to go back to full filter.

I was not able to do that, if moving my finger vertically on the pads this happens:
Activate pad 1 in white, finger on second pad deactivates first white pad, finger on third pad enables that pad in white, forth pad deactivates the third white. In sum, it jumps every two pads, activate, deactivate, activate, deactivate

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: > Should also have `tagNameFX[2] = 0` here, arrays are not auto-initialized to 0.

Is it needed if below I'm writing to &tagNameFX[2]? See line 1911.

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: > This can break if the XML tags are out of order which is probably not what we want -- instead, we should automatically set `padPressHeld` if any of the relevant tags appear.

Not sure I understand. The relevant tags can be there but the status can be set to "Off" - in this case we would not want to set it to held.

If you're referring to line 1976, if (FXPress[xDisplay].padPressHeld), padPressHeld is initialized to false, so this condition would only ever be true if the <status> tag is found. 

is your suggestion simply to just remove the if condition and move the code within that if to below line 1972?

src/deluge/gui/views/performance_session_view.h
seangoodvibes: Fixed thanks!

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: Fixed thanks!

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+ActionResult PerformanceSessionView::padAction(int32_t xDisplay, int32_t yDisplay, int32_t on) {
+	if (!justExitedSoundEditor) {
+		//if pad was pressed in main deluge grid (not sidebar)
+		if (xDisplay < kDisplayWidth) {
+			if (on) {
+				//if it's a shortcut press, enter soundEditor menu for that parameter
+				if (Buttons::isShiftButtonPressed()) {
+					ActionResult soundEditorResult = soundEditor.potentialShortcutPadAction(xDisplay, yDisplay, on);
+					return soundEditorResult;
```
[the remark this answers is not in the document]
seangoodvibes: Fixed thanks!

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+					changeRootUI(&sessionView);
+				}
+			}
+		}
+	}
+
+	//disable button presses for Vertical encoder
+	//disable back button press since undo doesn't work well in this view atm.
+	else if (b == Y_ENC) { //|| (b == BACK)) {
+		goto doNothing;
```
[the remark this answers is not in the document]
seangoodvibes: Fixed thanks!

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: Fixed thanks!

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: What do you mean by "inline sd card work"?

Do you mean like this should be moved to a save/load preset UI?

Yea if you we could defer it to follow up work that would be great.

I have on my action item list a request to load preset from a folder and to be able to save presets with different names using the keyboard grid UI.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+			}
+			else if (Buttons::isButtonPressed(deluge::hid::button::MIDI)) {
+				layoutVariant = 4;
+			}
+			else if (Buttons::isButtonPressed(deluge::hid::button::CV)) {
+				layoutVariant = 5;
+			}
+			else {
+				layoutVariant = 1;
+			}*/
```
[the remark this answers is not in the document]
seangoodvibes: Removed it thanks! Will work on this as part of a follow-up PR.

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+		uint8_t* occupancyMaskOfRow = occupancyMask[yDisplay];
+
+		renderRow(image + (yDisplay * imageWidth * 3), occupancyMaskOfRow, yDisplay);
+	}
+}
+
+//this function started off as a copy of the renderRow function from the NoteRow class - I replaced "notes" with "nodes"
+//it worked for the most part, but there was bugs so I removed the buggy code and inserted my alternative rendering method
+//which always works. hoping to bring back the other code once I've worked out the bugs.
```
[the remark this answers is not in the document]
seangoodvibes: Removed stale comments and added a more relevant one thanks!

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+     Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE},
+    {Param::Kind::UNPATCHED_GLOBAL, Param::Kind::NONE, Param::Kind::NONE, Param::Kind::UNPATCHED_GLOBAL,
+     Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE},
+    {Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE, Param::Kind::NONE,
+     Param::Kind::NONE, Param::Kind::NONE},
+};
+
+//mapping shortcuts to paramID
+const uint32_t paramIDShortcutsForPerformanceView[kDisplayWidth][kDisplayHeight] = {
+    {0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF},
```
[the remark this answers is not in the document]
seangoodvibes: Done thanks!

src/deluge/hid/buttons.cpp
```diff
@@ -159,7 +159,7 @@ ActionResult buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) {
 				}
 				// We got a short press, maybe enable sticky keys
 				//5th of a second
-				if (delta < kSampleRate / 5) {
+				if (delta < kHoldTime) {
```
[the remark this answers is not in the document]
seangoodvibes: Understood thanks!

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -491,8 +522,26 @@ bool SoundEditor::beginScreen(MenuItem* oldMenuItem) {
...
+			for (x = 0; x < 15; x++) {
+				for (y = 0; y < kDisplayHeight; y++) {
+					if (paramShortcutsForSongView[x][y] == currentItem) {
+						goto doSetupBlinkingForSessionView;
+					}
+				}
+			}
+
+			if (false) {
+doSetupBlinkingForSessionView:
```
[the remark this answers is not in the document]
seangoodvibes: I see this code logic also being used for the audio clips below.

A single break; will get you out of the second for loop. what about exiting the first for loop?

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -900,6 +978,16 @@ ActionResult SoundEditor::padAction(int32_t x, int32_t y, int32_t on) {
 		return ActionResult::REMIND_ME_OUTSIDE_CARD_ROUTINE;
 	}
 
+	//used to convert column press to a shortcut to change Perform FX menu displayed
+	if (((getRootUI() == &performanceSessionView) || (getCurrentUI() == &performanceSessionView))
+	    && !Buttons::isShiftButtonPressed() && performanceSessionView.defaultEditingMode
```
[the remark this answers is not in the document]
seangoodvibes: No I don't think so. These checks are only relevant when you're trying to call performanceSessionView::padAction from soundEditor. I only want to interact with the grid in performance view when you're in soundEditor if you're not holding shift and you're in defaultEditingMode and you're not editing parameter assignments. Otherwise I want to exit soundEditor completely if you press the grid in regular performance view or param editing mode (e.g. reach line 1053: exitCompletely();)

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+		int32_t modKnobMode = -1;
+		bool editingComp = false;
+		if (view.activeModControllableModelStack.modControllable) {
+			uint8_t* modKnobModePointer = view.activeModControllableModelStack.modControllable->getModKnobMode();
+			if (modKnobModePointer) {
+				modKnobMode = *modKnobModePointer;
+				editingComp = view.activeModControllableModelStack.modControllable->isEditingComp();
+			}
+		}
+		if (modKnobMode == 4 && editingComp) { //upper
```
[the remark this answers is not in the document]
seangoodvibes: per @m-m-adams:

> It’s the number of the currently lit mod knob light
> We probably should make it an enum

I would suggest we resolve this in a follow-up PR as this is used in many places

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+	layoutVariant = 1;
+
+	onFXDisplay = false;
+
+	performanceLayoutBackedUp = false;
+
+	justExitedSoundEditor = false;
+
+	initPadPress(&firstPadPress);
```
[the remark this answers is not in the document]
seangoodvibes: Resolved thanks!

src/deluge/gui/views/performance_session_view.cpp
bobtwinkles: Right, I missed that before sorry.

src/deluge/gui/views/performance_session_view.cpp
bobtwinkles: The `storageManager` calls end up doing I/O on the SDCard. It should be fine for an initial pass.

src/deluge/gui/views/performance_session_view.cpp
bobtwinkles: ah I see. Maybe prefix them with `TAG_FOR_` instead to make it clearer that these aren't localization strings?

src/deluge/gui/views/performance_session_view.cpp
bobtwinkles: My suggestion would be to remove the `if` condition and unconditionally parse these tags. If that would cause some problems, once the full fxpress is parsed (i.e. after the loop) we can apply some fixup logic (e.g. resetting these back to their defaults if `!fxPress[xDisplay].padPressHeld`

seangoodvibes: Ah I understand what you mean - you meant that potentially <row> and <resetValue> wouldn't be parsed. Ok I know what to do, thanks!

seangoodvibes: @bobtwinkles just made the changes, please let me know if I understood correctly! Thanks :)

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: > ah I see. Maybe prefix them with `TAG_FOR_` instead to make it clearer that these aren't localization strings?

I just removed the "STRING_FOR" prefix altogether. Should be good now :)!

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+	PadLEDs::renderingLock = true;
+
+	// erase current image as it will be refreshed
+	memset(image, 0, sizeof(uint8_t) * kDisplayHeight * (kDisplayWidth + kSideBarWidth) * 3);
+
+	// erase current occupancy mask as it will be refreshed
+	memset(occupancyMask, 0, sizeof(uint8_t) * kDisplayHeight * (kDisplayWidth + kSideBarWidth));
+
+	performActualRender(whichRows, &image[0][0][0], occupancyMask, currentSong->xScroll[NAVIGATION_CLIP],
```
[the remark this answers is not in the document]
seangoodvibes: Done! Inlined it in renderMainPads

src/deluge/gui/menu_item/performance_session_view/editing_mode.h
```diff
@@ -0,0 +1,73 @@
...
+	void readCurrentValue() override {
+		int32_t currentValue;
+		if (!performanceSessionView.defaultEditingMode) {
+			currentValue = 0;
+		}
+		else if (!performanceSessionView.editingParam) {
+			currentValue = 1;
+		}
+		else {
+			currentValue = 2;
```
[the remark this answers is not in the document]
seangoodvibes: Done. Created enum class:

enum class PerformanceEditingMode : int32_t {
	DISABLED,
	VALUE,
	PARAM,
};

src/deluge/model/consequence/consequence_performance_layout_change.cpp
```diff
@@ -0,0 +1,57 @@
...
+ConsequencePerformanceLayoutChange::ConsequencePerformanceLayoutChange(
+    PadPress(*padPressBefore), PadPress(*padPressAfter), FXColumnPress(*FXPressBefore), FXColumnPress(*FXPressAfter),
+    ParamsForPerformance(*layoutBefore), ParamsForPerformance(*layoutAfter),
+    int32_t valuesBefore[kDisplayWidth][kDisplayHeight], int32_t valuesAfter[kDisplayWidth][kDisplayHeight]) {
+	for (int32_t xDisplay = 0; xDisplay < kDisplayWidth; xDisplay++) {
+		memcpy(&FXPress[xDisplay][BEFORE], &FXPressBefore[xDisplay], sizeFXPress);
+		memcpy(&layoutForPerformance[xDisplay][BEFORE], &layoutBefore[xDisplay], sizeParamsForPerformance);
+
+		memcpy(&FXPress[xDisplay][AFTER], &FXPressAfter[xDisplay], sizeFXPress);
+		memcpy(&layoutForPerformance[xDisplay][AFTER], &layoutAfter[xDisplay], sizeParamsForPerformance);
```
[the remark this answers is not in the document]
seangoodvibes: Resolved this!

src/deluge/model/consequence/consequence_performance_layout_change.h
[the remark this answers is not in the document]
seangoodvibes: Resolved :)!

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,2045 @@
...
+
+//list of parameters available for assignment to FX columns in performance view
+
+const ParamsForPerformance songParamsForPerformance[kNumParamsForPerformance] = {
+    {Param::Kind::UNPATCHED_GLOBAL,                                    //paramKind
+     Param::Unpatched::GlobalEffectable::LPF_FREQ,                     //paramID
+     8,                                                                //xDisplay
+     7,                                                                //yDisplay
+     {rowColourRed[0], rowColourRed[1], rowColourRed[2]},              //rowColour[3]
+     {rowTailColourRed[0], rowTailColourRed[1], rowTailColourRed[2]}}, //rowTailColour[3]
```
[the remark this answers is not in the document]
seangoodvibes: Done!

src/deluge/model/action/action_logger.cpp
```diff
@@ -298,6 +300,35 @@ void ActionLogger::recordTempoChange(uint64_t timePerBigBefore, uint64_t timePer
 	}
 }
 
+//record changes to the layout in Performance View so that you can undo changes
+//changes that can be undone:
+//1) loading a new layout
+//2) setting a pad to "hold"
+//3) value editor: changing pad values
+//4) param editor: changing FX assignments to pads
```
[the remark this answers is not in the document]
seangoodvibes: > Since these are 4 separate actions, maybe we should store them as 4 separate types of consequences? It looks like `ConsequeencePerformanceLayoutChange` is quite large and breaking it in to the 4 separate actions would be both clearer semantically in the code and lighter weight in memory.

They're not entirely separate as they're kind intermingled and work together. Even if I did split them, you'd still need to execute each consequence together in a lot of cases.

So in theory yes it would make sense to split out the consequences. But it would also imply checking exactly what changed or did not change prior to logging a change which is extra cycles and checks every time you're making a change.

I thought it would be lighter to just check if there are any changes instead of specific types of changes and backup everything altogether.

Example 1) Loading a Layout - affects padPress, fxPress, ParamsForPerformance, values
- Loads hold statuses and current hold values
- Loads mappings of values to pads
- Loads mappings of params to columns
- Loads everything associated with each param mapping (paramKind, x, y coordinate, colour)

Example 2) Resetting the view - depending on the context, changes ParamsForPerformance, padPress, fxPress 
- Clears hold statuses and resets values
- Clears param assignments and layout

Example 3) Removing a param from a single column - affects ParamsForPerformance, fxPress, padPress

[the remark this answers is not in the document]
bobtwinkles: It feels like there should be a more efficient way to do this but I don't have a concrete design in my head at the moment. We can improve this later if it turns out to be trouble.

bobtwinkles (approved):
I'm somewhat concerned about the memory use of this feature but it's probably worth the squeeze. For posterity after this change, we're 23712 bytes larger than the nightly from yesterday.

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1160,6 +1160,12 @@ int32_t ModControllableAudio::getStutterRate(ParamManager* paramManager) {
...
 
+	// When stuttering, we center the value at 0, so the center is the reference for the stutter rate that we selected just before pressing the knob
+	// and we use the lastQuantizedKnobDiff value to calculate the relative (real) value
+	if ((stutterer.lastQuantizedKnobDiff != 0) && (paramValue == stutterer.valueBeforeStuttering)) {
+		paramValue = 0;
```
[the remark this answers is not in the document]
seangoodvibes: Hi @m-m-adams please try it. There is zero jumpiness in the LED's whatsoever. Whatever the LED's are at before you start stutter is what the LED's remain at when you start stutter. While stutter is engaged if you turn the knob to adjust stutter rate the LED's adjust from that starting position. After you are done stutter, it resets the led's and the stutter rate value back to the stutter rate before you began stuttering.

src/deluge/gui/views/performance_session_view.h
```diff
@@ -0,0 +1,215 @@
...
+};
+
+struct ParamsForPerformance {
+	Param::Kind paramKind;
+	ParamType paramID;
+	int32_t xDisplay;
+	int32_t yDisplay;
+	uint8_t rowColour[3];
+	uint8_t rowTailColour[3];
+	ParamsForPerformance(){};
```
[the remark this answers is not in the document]
seangoodvibes: The ParamsForPerformance(){};?

Yes, the code wouldn't compile without it because there are uses of this struct where zero parameters are passed.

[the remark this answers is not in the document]
m-m-adams: Ah - then it should initialize all members to zero. As is this doesn't initialize any members which is a future footgun

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+
+extern "C" {
+#include "RZA1/uart/sio_char.h"
+#include "util/cfunctions.h"
+}
+
+using namespace deluge;
+using namespace gui;
+
+const char* PERFORM_DEFAULTS_XML = "PerformanceView.XML";
```
m-m-adams, line 67: these should be defines or constexpr, not const

src/deluge/model/mod_controllable/mod_controllable_audio.h
```diff
@@ -147,6 +148,7 @@ class ModControllableAudio : public ModControllable {
 	bool grainInitialized;
 
 	Stutterer stutterer;
+	bool beginQuantizedStutter;
```
seangoodvibes, line 151: orphaned, remove it

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+			if ((paramKind == Param::Kind::UNPATCHED_SOUND) && (paramID == Param::Unpatched::STUTTER_RATE)
+			    && (isUIModeActive(UI_MODE_STUTTERING))) {
+				((ModControllableAudio*)view.activeModControllableModelStack.modControllable)
+				    ->endStutter((ParamManagerForTimeline*)view.activeModControllableModelStack.paramManager);
+			}
+
+			if (fxPress[xDisplay].previousKnobPosition == kNoSelection) {
+				int32_t oldParameterValue =
+				    modelStackWithParam->autoParam->getValuePossiblyAtPos(view.modPos, modelStackWithParam);
+				fxPress[xDisplay].previousKnobPosition =
```
m-m-adams, line 1239: I think you should be storing the param value and not the knob pos - params have higher accuracy than knob position allows and this could cause issues in the future

Also you're converting the knob pos back to a param value before using it anyway

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+	}
+}
+
+/// this will set a new value for a parameter
+/// if we're dealing with stutter, it will check if stutter is active and end the stutter first
+/// if we're dealing with stutter, it will change the stutter rate value and then begin stutter
+/// if you're in the value editor, pressing a column and changing the value will also open the sound editor
+/// menu for the parameter to show you the current value in the menu
+/// in regular performance view, this function will also update the parameter value shown on the display
+bool PerformanceSessionView::setParameterValue(ModelStackWithThreeMainThings* modelStack, Param::Kind paramKind,
```
m-m-adams, line 1219: This function should work on param values instead of knob pos - will allow greater accuracy in the future (particularly for things like stutter where the knob pos is very coarse)

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+                                               int32_t paramID, int32_t xDisplay, int32_t knobPos, bool renderDisplay) {
+	ModelStackWithAutoParam* modelStackWithParam = getModelStackWithParam(modelStack, paramID);
+
+	if (modelStackWithParam && modelStackWithParam->autoParam) {
+
+		if (modelStackWithParam->getTimelineCounter()
```
m-m-adams, line 1225: This is a sanity check to make sure the timeline counter is the song level global effectable?

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+			modelStackWithParam = modelStack->addParam(paramSet, summary, paramID, &paramSet->params[paramID]);
+		}
+	}
+
+	return modelStackWithParam;
+}
+
+/// converts grid pad press yDisplay into a knobPosition value default
+/// this will likely need to be customized based on the parameter to create some more param appropriate ranges
+int32_t PerformanceSessionView::calculateKnobPosForSinglePadPress(int32_t xDisplay, int32_t yDisplay) {
```
m-m-adams, line 1325: I think this should use the param value range instead of knobpos

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -0,0 +1,1931 @@
...
+		logPerformanceLayoutChange();
+	}
+
+	successfullyReadDefaultsFromFile = true;
+}
+
+/// if no XML file exists, load default layout (paramKind, paramID, xDisplay, yDisplay, rowColour, rowTailColour)
+void PerformanceSessionView::loadDefaultLayout() {
+	for (int32_t xDisplay = 0; xDisplay < kDisplayWidth; xDisplay++) {
+		memcpy(&layoutForPerformance[xDisplay], &defaultLayoutForPerformance[xDisplay], sizeParamsForPerformance);
```
m-m-adams, line 1757: You should implement a copy method instead of using memcpy - memcpy is brittle and any speed differences won't matter here

m-m-adams (commented):
I think this is looking pretty good! I'd like to see the change from using knob pos to directly using param values instead for future compatibility, no reason to limit the internal resolution of performance view to 7 bits

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: Does this change need to be a part of this PR? As much as I want to make it, I don't know if I have the energy atm to go through and change everything from knobPos to value. I know it's not a complex change, but I do use the knobPos for the value displayed on the screen so I would want to make sure I'm not breaking things.

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: yea it's a sanity check to make sure you're dealing with the same model stack as the active mod controllable stack.

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: you mean like this?

`copyLayout(layoutForPerformance[xDisplay], defaultLayoutForPerformance[xdisplay]);`

```
PerformanceSessionView::copyLayout(ParamForPerformance& layoutTo, ParamForPerformance& layoutFrom) {

layoutTo.paramKind = layoutFrom.paramKind;
layoutTo.paramID = layoutFrom.paramID;
layoutTo.xDisplay = layoutFrom.xDisplay;
layoutTo.yDisplay = layoutFrom.yDisplay;
layoutTo.rowColour[0] = layoutFrom.rowColour[0];
layoutTo.rowColour[1] = layoutFrom.rowColour[1];
layoutTo.rowColour[2] = layoutFrom.rowColour[2];
layoutTo.rowTailColour[0] = layoutFrom.rowTailColour[0];
layoutTo.rowTailColour[1] = layoutFrom.rowTailColour[1];
layoutTo.rowTailColour[2] = layoutFrom.rowTailColour[2];

}
```

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: Will make this change thanks!

src/deluge/gui/views/performance_session_view.cpp
PaulFreund: I need to look in more detail but this would generate a lot more code compared to memcpy and is more error prone, any particular reason?

seangoodvibes: > I need to look in more detail but this would generate a lot more code compared to memcpy and is more error prone, any particular reason?

Yea cause I'd need to create copy functions for each type of struct I'm trying to copy (so 3 copy functions). I initially thought of this but thought memcpy was cleaner so I went with memcpy.

src/deluge/model/mod_controllable/mod_controllable_audio.h
seangoodvibes: Done

src/deluge/gui/views/performance_session_view.cpp
seangoodvibes: Done

src/deluge/gui/views/performance_session_view.h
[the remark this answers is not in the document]
seangoodvibes: Done

PaulFreund (approved):

PR #719, opened 2023-11-13
Synth presets sometimes loading incorrectly

m-m-adams:
Adds additional logging to synth loads

In some situations a file name/synth preset can become associated with an incorrect file pointer (the underlying location on the card)

image

When this happens the sound loads incorrectly

Fixed by leveraging instrument file path to set fp in file item

Fix 715, 718

### DIFF
--- a/src/deluge/storage/file_item.cpp
+++ b/src/deluge/storage/file_item.cpp
@@ -16,6 +16,7 @@
 */
 #include "storage/file_item.h"
 #include "hid/display/display.h"
+#include "io/debug/print.h"
 #include "model/instrument/instrument.h"
 #include <string.h>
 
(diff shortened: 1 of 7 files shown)


m-m-adams:
So basically if you change scroll directions immediately after hitting "reloading" in the browser, or possibly after any bidirectional folder load, you get an incorrect file pointer

src/deluge/storage/file_item.cpp
```diff
@@ -36,6 +38,16 @@ int32_t FileItem::setupWithInstrument(Instrument* newInstrument, bool hibernatin
...
 	displayName = filename.get();
+	String tempFilePath;
+	tempFilePath.set(newInstrument->dirPath.get());
+	tempFilePath.concatenate("/");
+	tempFilePath.concatenate(filename.get());
+	bool fileExists = storageManager.fileExists(tempFilePath.get(), &filePointer);
+	if (!fileExists) {
+		Debug::print("couldn't get filepath for file");
+		Debug::println(filename.get());
+		return false;
```
PaulFreund, line 49: Shouldn't this rather be:

```suggestion
		return ERROR_FILE_NOT_FOUND;
```

PR #738, opened 2023-11-21
Bugfix/volume change clicks

m-m-adams:
Filter volume changes so that clicks are less noticeable

Stops all crackling while adjusting compressor or volume knob

Reduces clicks from adjusting modfx or filter

Song filter can still click when first engaged

### DIFF
--- a/src/deluge/dsp/master_compressor/master_compressor.cpp
+++ b/src/deluge/dsp/master_compressor/master_compressor.cpp
@@ -33,13 +33,12 @@ MasterCompressor::MasterCompressor() {
 
 	currentVolumeL = 0;
 	currentVolumeR = 0;
-	//auto make up gain
-	updateER();
+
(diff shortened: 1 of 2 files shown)


src/deluge/dsp/master_compressor/master_compressor.cpp
```diff
@@ -100,14 +103,15 @@ void MasterCompressor::render(StereoSample* buffer, uint16_t numSamples, q31_t v
 	rms = calc_rms(buffer, numSamples);
 }
 
-float MasterCompressor::runEnvelope(float in, float numSamples) {
+float MasterCompressor::runEnvelope(float state, float in, float numSamples) {
```
PaulFreund, line 106: I'd probably rename state here to prevent accidential shadowing

PaulFreund (approved):

Issue #743, opened 2023-11-22
Zoom status message (ex. 16th notes) stuck in grid mode on oled


rpc9943:
### Please describe the problem:

Popup Message in rectangle on oled screen sticks.
Hold leftright mnob down, press grid pad.
Message example "16th notes" stays on screen.
Fixed by pressing leftright knob again,

### What is the expected behavior?

Not sure why its even showing the zoom from grid view, but if it shows it, shouldnt stay on screen until i press the knob again

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

1.0.0 betaD5331ED

### If possible provide the steps to reproduce the issue and upload additional media:

1. Grid mode
2. Hold leftright knob down and press a grid pad
3. message stays on screen

PaulFreund:
Basically same bug as #741, will be fixed

PR #748, opened 2023-11-22
Fix #741 and #743 - Horizontal encoder button in Grid mode

### LINKED ISSUE #741: Solo function not present in grid mode

Solo is not working on grid mode. Works on rows.

For grid to be able to use solo function as it works in rows

_No response_

OLED

Built from community

1.0.0 betaD5331ED

Hold leftright knob down press the clip on grid mode. It does not solo. It works this way in rows view where you press the mute launch pad.

PaulFreund:


src/deluge/gui/views/session_view.cpp
```diff
@@ -3525,15 +3548,18 @@ ActionResult SessionView::gridHandlePadsLaunchWithSelection(int32_t x, int32_t y
 
 void SessionView::gridHandlePadsLaunchToggleArming(Clip* clip, bool immediate) {
 	if (immediate) {
-		if (currentUIMode == UI_MODE_HOLDING_HORIZONTAL_ENCODER_BUTTON) {
+		if (horizontalEncoderPressed) {
```
bobtwinkles, line 3551: shouldn't this be `isUIModeActive(UI_MODE_HOLDING_HORIZONTAL_ENCODER_BUTTON)`, rather than recreating that tracking with `horiznotalEncoderPressed`?

PaulFreund: Normally yes, that's how it was before but it got overwritten by clip being held state. Unfortunately there are a lot of places where ui state is written exclusively and sometimes the behavior relies on previous state to be overwritten and sometimes on an added state in some paths and overwritten in others, etc. The whole session view needs to be overhauled to be consistent again and to make future changes easier.

I hope this little hack to not break anything before Release is justified :D.

bobtwinkles:
```
$ rg 'currentUIMode = ' | wc -l
112
```
yeah that's probably too much to deal with for now

m-m-adams (approved):

Issue #730, opened 2023-11-19
Put Grain FX behind Community setting


PaulFreund:

m-m-adams:
To provide context for casual browsers this is because it uses large amounts of memory and leads to general instability since the cull system can't handle it

PR #752, opened 2023-11-24
Default disable Grain FX

PaulFreund:
Make users enable Grain FX in the community settings first to make sure everyone understands the CPU impact. Only removes menu option, does not remove GrainFX from tracks that already have it

implements #730


src/definitions_cxx.hpp
```diff
@@ -623,9 +623,10 @@ enum class ModFXType {
 	CHORUS,
 	PHASER,
 	CHORUS_STEREO,
-	GRAIN,
+	GRAIN, // Look below if you want to add another one
 };
 
+// Warning: Currently GRAIN can be disabled and kNumModFXTypes might need to be used - 1
```
m-m-adams, line 629: this seems like a footgun - maybe add a getNumModFx function that returns the current value based on whether grain is on or off?

PaulFreund: Sorry accidentially resolved. I'd agree but accessing runtimeFeatureSettings from definitions_cxx.hpp isn't trivial and I didn't have a better idea to get this done quickly :D If you have a better idea lets change it! Just removing is also not the best way since it's non obvious changing the list will result in the access chagned in global_effectable.

m-m-adams: yeah, just tried and it breaks the compilation. This seems good since it's hopefully temporary anyway

PR #757, opened 2023-11-25
Turn haveOLED and have7SEG into non-virtual calls.

topisani:
This saves roughly 1.2KB of binary size, and simplifies each check to a third of the instructions. Its not big, but for the tiny amount of impact on the code, i think it makes sense.

There could be other places where it might make sense to look at optimizing virtual calls that are called often, even outside of hot loops. `haveOLED` is called more than 200 places across the code, so it adds up


src/deluge/storage/file_item.cpp
```diff
@@ -72,7 +72,7 @@ int32_t FileItem::getDisplayNameWithoutExtension(String* displayNameWithoutExten
 	}
 
 	// 7SEG...
-	if (displayName != filename.get()) {
+	if (true) {
```
m-m-adams, line 75: This is pulled out in the other PR, so just remove this file from the commit

m-m-adams (commented):
Otherwise looks good

stellar-aria:
For context: they were virtual primarily because there were cases proposed where _both_ screens modes would be active, (such as debugging OLED over SYSEX on a dev machine while also running 7SEG physically). However that's never materialized and the hope to make the OLED UI further distanced from the 7SEG one means we should probably do this cleanup sooner rather than later.

PR #773, opened 2023-11-30
Fix auto-merge error in displayModEncoderValuePopup

topisani:
Not sure how this happened, may have been me a bit late at night or an auto merge, but that else branch is not supposed to be there. Results in any parameter with no known display name (such as midi CCs) being shown twice

### DIFF
--- a/src/deluge/gui/views/view.cpp
+++ b/src/deluge/gui/views/view.cpp
@@ -940,10 +940,6 @@ void View::displayModEncoderValuePopup(Param::Kind kind, int32_t paramID, int32_
 			popupMsg.append(name);
 			popupMsg.append(": ");
 		}
-		else {
-			int valueForDisplay = calculateKnobPosForDisplay(kind, paramID, newKnobPos + kKnobPosOffset);
-			popupMsg.appendInt(valueForDisplay);
-		}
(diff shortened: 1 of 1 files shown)


seangoodvibes (approved):
Looks good

src/deluge/gui/views/view.cpp
```diff
@@ -940,10 +940,6 @@ void View::displayModEncoderValuePopup(Param::Kind kind, int32_t paramID, int32_
 			popupMsg.append(name);
 			popupMsg.append(": ");
 		}
-		else {
```
seangoodvibes, line 943: Instead of removing the append for midi, can we replace it with an append for Midi CC #?

For Midi instruments the cc # = modelStackWithParam->paramId

So:

popupMsg.append("CC ";
popupMsg.appendInt(modelStackWithParam->paramId);
popupMsg.append(": ");

topisani: Yes, but you'd have to check the Param kind == MIDI, but I'd really rather do it in getDisplayName ForParam

seangoodvibes:
Tested, works. Code change impact negligible. Merging.

PR #774, opened 2023-12-01
Show human note name instead of integer note number, refactor

tastycode:
We had a function that converted integers into human readable notes e.g. "C#5". We were duplicating that logic in several places where we hadn't leveraged the function in util/functions. The original intent of this was to avoid using integer note values when modifying what note a kit's midi track is linked to. There was discussion around appropriate UX for this and it was discovered that displaying channel and note name on two lines is unideal because the realized vertical hierarchy is incongruent with the physical layout. The gold knobs are farther apart vertically than they are spaced apart horizontally. As a result, the display feels "flipped" because you twist the bottom knob to modify the channel and the top knob to modify the note. 

On the 7seg, the channel and note are horizontal with channel being first then note. This only makes sense because the bottom knob is slightly to the left of the top knob. 

Since compatibility with the paradigms used in the 7seg is a priority, it makes most sense to get the channel/note to fit on the same line – thus closing the gap in presentational consistency across device models. 

To do this, we reasoned that "MIDI " in `"MIDI CHANNEL 5, " is unnecessary. Additionally, "CHANNEL" can be shortened to "CHAN" and could be argued would be worth introducing inconsistency because the alternative is to break the display in two linesl

Discussion: 
https://discord.com/channels/608916579421257728/1107026299945496577/1179913677818839051

QA:

**Before**

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/5d9b3911-e5cd-4446-83ce-033f1f9a6252

**After**

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/05a9a003-de65-4289-91fb-e60dcbdeb784

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -3543,10 +3543,11 @@ void InstrumentClipView::drawNoteCode(uint8_t yDisplay) {
 }
 
 void InstrumentClipView::drawDrumName(Drum* drum, bool justPopUp) {
+
 	char const* newText;
 
 	if (display->haveOLED()) {
-		char buffer[30];
+		char buffer[50];
 
 		if (!drum) {
 			newText = "No sound";
@@ -3558,22 +3559,22 @@ void InstrumentClipView::drawDrumName(Drum* drum, bool justPopUp) {
 			newText = buffer;
 
 			if (drum->type == DrumType::GATE) {
+
 				strcpy(buffer, "Gate channel ");
 				intToString(((GateDrum*)drum)->channel + 1, &buffer[13]);
 				indicator_leds::blinkLed(IndicatorLED::CV, 1, 1);
 			}
 			else { // MIDI
-				uint8_t drumNote = ((MIDIDrum*)drum)->note;
-				char channelString[10];
-				char noteString[5];
+				char topLine[30];
+				char noteLabel[5];
 
-				strcpy(buffer, "CHAN ");
-				intToString(((MIDIDrum*)drum)->channel + 1, channelString);
-				strcat(buffer, channelString);
-				strcat(buffer, ", NOTE ");
+				sprintf(topLine, "CH: %d  N#: %d", ((MIDIDrum*)drum)->channel + 1, ((MIDIDrum*)drum)->note);
+				noteCodeToString(((MIDIDrum*)drum)->note, noteLabel);
 
-				noteCodeToString(drumNote, noteString, nullptr);
-				strcat(buffer, noteString);
+				const char* lines[] = {topLine, noteLabel};
+
+				char bufferLines[50];
+				concatenateLines(lines, sizeof(lines) / sizeof(lines[0]), buffer);
 				indicator_leds::blinkLed(IndicatorLED::MIDI, 1, 1);
 			}
 		}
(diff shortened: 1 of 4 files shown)


m-m-adams (changes requested):
Making the channel and note horizontal is nice but displaying the note code in this context is a mistake

Maybe stick the note in brackets? Like 48 (c2)

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3559,38 +3558,33 @@ void InstrumentClipView::drawDrumName(Drum* drum, bool justPopUp) {
...
+				uint8_t drumNote = ((MIDIDrum*)drum)->note;
+				char channelString[10];
+				char noteString[5];
+
+				strcpy(buffer, "CHAN ");
+				intToString(((MIDIDrum*)drum)->channel + 1, channelString);
+				strcat(buffer, channelString);
+				strcat(buffer, ", NOTE ");
 
+				noteCodeToString(drumNote, noteString, nullptr);
```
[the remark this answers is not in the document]
tastycode: Discussed at length in discord. We realized there are many use-cases to support for how people use channels and notes and their representations. So we decided to support both note number and note name in this view. This is also the only view where we show the note number. So there's no where else to change.

tastycode:
Here is the change in context, exactly as I encounter needing it. I'm playing and i have the notes whose sounds work with the melody and i need that same note in a kit and for it to be convenient to find the same note. Because 808s are tuned and it's really nice to see them along side the drum rhythm.

Also, it gave me an excuse to take the new meta glasses for a spin. Those are nifty. Now you see my life is a mess and music is the only thing holding anything together :D 

https://www.dropbox.com/scl/fi/z78v9et8m8g4pclnzy5j6/2023-11-30-ddn.mov?rlkey=a0ege5l26so2j6a8ymr37auon&dl=0

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3558,33 +3559,42 @@ void InstrumentClipView::drawDrumName(Drum* drum, bool justPopUp) {
...
 	}
 	else {
+
 		char buffer[7];
 
 		if (!drum) {
 			newText = "NONE";
+
+basicDisplay:
+			if (justPopUp && currentUIMode != UI_MODE_AUDITIONING) {
```
m-m-adams, line 3592: Can you make this a function instead of continuing to use a goto?

m-m-adams (commented):
Generally good - waiting on a video of the new popup since I don't have my OLED around to test it this week

src/deluge/gui/views/instrument_clip_view.cpp
tastycode: This was pre-existing code, the order was slightly modified, I think? I'll take a look again, I just remember this was a side effect and when i looked closely at the diff i remember realizing it was moved not added.

tastycode:
> Generally good - waiting on a video of the new popup since I don't have my OLED around to test it this week
@m-m-adams  The video was in the comment i posted when I resolved changes from your first round of feedback. See the link at the bottom of this comment https://github.com/SynthstromAudible/DelugeFirmware/pull/774#issuecomment-1835810557. Because I go through my entire workflow, it might not look at first like i'm going to be demoing this feature, but who starts up the deluge and immediately goes and maps a kit row to midi?

src/deluge/gui/views/instrument_clip_view.cpp
m-m-adams: Ya but that's always a good opportunity for some cleanup!

m-m-adams (approved):

PR #776, opened 2023-12-01
Proposal for release page

PaulFreund:


jamiefaye:
Can we include (if it isn't there already) an example of the address dump page so people can confirm they got one?

PaulFreund:
Added that 👍

PR #780, opened 2023-12-02
rescale metronome volume

### LINKED ISSUE #777: Default Metronome setting should be a sensible range instead of 22-27

Current values are 22-27

Values should start at 1, in this case that would shift to 1-6. Possibly reduce to 1-5 for a nice 'clean' number?

Possibly for a future release (OLED): I think having a bar graph underneath the value will be helpful for users to understand the metronome range at a glance. However, currently I don't think this is widely seen occurrence on the Deluge for non 0-50 values.

_No response_

OLED

Built from release/1.0

RC

_No response_

m-m-adams:
Fix #777 

A bug in reading default metronome volume in #683 caused the metronome volume to be squared, making the scaling opposite of what's desired for audio and giving it a small usable range

Range is now 1-50 exponentially scaled


seangoodvibes (approved):
Code looks good

Tested and it sounds great too

thanks for fixing my mistake

PR #783, opened 2023-12-03
Add lowercase rendering for 7SEG

stellar-aria:
Add lowercase 'c' for firmware version


src/deluge/hid/display/seven_segment.cpp
```diff
@@ -42,21 +42,41 @@ extern "C" {
...
+    0x55,
+    0x15, //N
+    0x1D, 0x67,
+    0x73, //Q
+    0x05, 0x5B,
+    0x0F, //T
+    0x3E, 0x27, 0x5C, 0x49, 0x3B, 0x6D,
+};
+
+uint8_t lowerLetterSegments[26] = {
```
m-m-adams, line 63: Could we append these + symbols for [\]^_` (probably just blank) to the end of uppercase, so that lower case letters can just be typed normally?

m-m-adams (commented):
This looks good to me - I'm not sure if fixing it in the current release is worth the (very small but non zero) chance for bugs though

src/deluge/hid/display/seven_segment.cpp
stellar-aria: Fixed

m-m-adams (approved):

PR #801, opened 2023-12-09
Support MIDI commands on many controller pedals

tastycode:
I went through 3 pedals when I first got my deluge trying to get live looping working. I eventually got it working by opening up a keyboard app on my phone, and tapping one of the piano keys with my foot. One pedal cost me $140 and i couldn't believe it wouldn't work with the deluge. I found out soon, that the deluge only supports note events for midi commands. 

So I built a solution using an RK002 cable that converted the messages to note messages. However, this fix only helps me and all the other people like me who are really excited to live loop with the deluge are stuck with pretty few options... or flumbing around with an ipad under our toes.

This PR enables the ability for program change messages to map to midi commands. 

Enabling support for the following pedals.
- Behringer FCB1010:
- Boss ES-5/ES-8
- Voodoo Labs Ground Control Pro
- Yamaha MFC1 
- Boss FC50

and many more... because apparently people really like sending program changes with pedals for some crazy reason.

**PS**: In my debugging madness i added some rudimentary reflection `UI->getName()` for a specific scenario where it was unclear which subclass was handling a particular code path. 

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/764a4d66-7603-4ab7-85e0-bb43798d1416


src/deluge/playback/playback_handler.cpp
```diff
@@ -2521,18 +2521,18 @@ bool PlaybackHandler::tryGlobalMIDICommandsOff(MIDIDevice* device, int32_t chann
...
-    	return;
-    }
-
-
-    tryGlobalMIDICommands(channel + 16, program);
-    */
+void PlaybackHandler::programChangeReceived(MIDIDevice* fromDevice, int32_t channel, int32_t program) {
+	// If user assigning MIDI commands, do that
+	if (currentUIMode == UI_MODE_MIDI_LEARN) {
+		if (getCurrentUI()->pcReceivedForMidiLearn(fromDevice, channel, program)) {}
```
m-m-adams, line 2527: I suggest we just don't do this, and only allow PC learning for non-momentary global commands

m-m-adams (changes requested):

src/definitions_cxx.hpp
```diff
@@ -1052,6 +1052,7 @@ constexpr int32_t NUM_CHANNELS = 18;
 constexpr int32_t MIDI_CHANNEL_NONE = 255;
 
 constexpr int32_t IS_A_CC = NUM_CHANNELS;
+constexpr int32_t IS_A_PC = 128; // CC128 is max.
```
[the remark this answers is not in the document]
tastycode: That makes sense. I'm silly, i did 128 because that's the maximum CC number because initially... I wanted the hack for using out of bounds for PC to be separate.. or unconflatable from the out of bounds channel strategy used for CC where we call note with a channel > 16. I'm actually on the fence somewhat between the two approaches because the approach i did choose, even if i fix the IS_A_PC values... still feels like we could approach it differently and help downstream consumers of this weird way of doing things account for less complexity? 

Thanks for pointing it out, will fix.

src/deluge/gui/views/view.cpp
```diff
@@ -751,8 +752,16 @@ void View::clearMelodicInstrumentMonoExpressionIfPossible() {
 	}
 }
 
+/* adapt program change messages to cc*/
+void View::pcReceivedForMIDILearn(MIDIDevice* fromDevice, int32_t channel, int32_t pc) {
```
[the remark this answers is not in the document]
tastycode: I will remove this, thanks for pointing it out. 

and had this been the actual code path that i needed to implement, i might have argued that using note on for CC doesn't really work either, its admittedly a bit of a hack. 

An additional reason that I will remove this, is because old midi controllers have noise/mechanism issues which cause intermittent stray CC's to be sent. Sometimes constantly. Program changes less so.

src/deluge/playback/playback_handler.cpp
tastycode: This is the real code path being used, and it seems to be the most straightforward path to implement thi  functionality while minimizing the cognitive load required to reason about signaling paths for other  devs in the future. 

Its the precise same path used for learning notes for commands. Right here: https://github.com/SynthstromAudible/DelugeFirmware/pull/801/files/2786b7007060d69fd8ca9672b4ef9b99c9033df3#diff-17a7ec38a9275b782f61ac4d14a3d043beedc0eb78fab024ea73802059ba76e9L2538


I did my best to make it as consistent as possible with the existing logic for notes. Not because i felt that there would be other places i could imagine where we use program changes for some other purpose, but just because it felt like the most maintainable / symetrical path. 

This approach _should_ not imply for example:
- Anything and everything can be learned by PC. PC is only learnable by views that  override `pcReceivedForMidiLearn` method. Which in this case, is only `SoundEditor`? 

Since its line 2527 that passes the control of the implemneted functionality into the soundeditor, line 2529 is therefore not used and only exists to not deviate from the pre-established pattern already used for note on messages further down.

m-m-adams: Global commands are learnt exclusively in the soundEditor - just call that using the same logic as the midiccReceived function

m-m-adams (changes requested):

src/deluge/gui/menu_item/midi/command.cpp
```diff
@@ -128,6 +138,10 @@ void Command::unlearnAction() {
 	}
 }
 
+void Command::learnProgramChange(MIDIDevice* device, int32_t channel, int32_t programNumber) {
+	learnCC(device, channel + IS_A_PC, programNumber, 1);
```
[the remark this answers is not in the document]
tastycode: With PCs I could imagine a toggle on / toggle off behavior. To be quite honest i have no idea what `FILL` is . I am going to QA to make sure you can't learn fill specifically until I better understand the semantics.

src/deluge/gui/ui/ui.h
```diff
@@ -109,6 +109,9 @@ class UI {
 	virtual bool canSeeViewUnderneath() { return false; }
 	virtual ClipMinder* toClipMinder() { return NULL; }
 	virtual void scrollFinished() {}
+	virtual const char* getName() { return "UI"; }
```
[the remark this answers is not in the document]
tastycode: Great suggestion. Thank you.

src/deluge/playback/playback_handler.cpp
```diff
@@ -2521,18 +2521,18 @@ bool PlaybackHandler::tryGlobalMIDICommandsOff(MIDIDevice* device, int32_t chann
...
+		else {
+			view.pcReceivedForMIDILearn(fromDevice, channel, program);
+		}
+	}
+	else {
+		// we build ontop of the CC hack
+		tryGlobalMIDICommands(fromDevice, channel + IS_A_CC + IS_A_PC, program);
```
m-m-adams, line 2534: should just be channel + IS_A_PC

src/deluge/gui/views/view.cpp
```diff
@@ -754,7 +754,7 @@ void View::clearMelodicInstrumentMonoExpressionIfPossible() {
 
 /* adapt program change messages to cc*/
 void View::pcReceivedForMIDILearn(MIDIDevice* fromDevice, int32_t channel, int32_t pc) {
-	ccReceivedForMIDILearn(fromDevice, channel + IS_A_PC, pc, 1);
+	// to be overridden
```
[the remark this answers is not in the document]
m-m-adams: Or comment as currently not used anywhere

src/deluge/gui/menu_item/midi/command.cpp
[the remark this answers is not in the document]
tastycode: Blocked from learning fill from PCs

m-m-adams:
Try formatting with ./dbt format, it'll run the exact clang tidy/format config that we check against in CI

src/definitions_cxx.hpp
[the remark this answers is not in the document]
tastycode: we should write an issue for this, with my novice level of experience everything takes me 10x longer and i'm 10x more likely to make a silly mistake and i've already bitten off more than i can chew in other areas of the codebase. However, I'm working extremely heavily with midi messages and have observed some consistency in terms of common APIs for dealing with midi data across 3-4 different codebases and libraries/languages.

src/deluge/gui/views/view.cpp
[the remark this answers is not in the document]
tastycode: we went over this and re-iterated why we are removing this and removed it.

src/deluge/playback/playback_handler.cpp
tastycode: we've eliminated most of the unnecessary code paths here but this logic remains necessary to maintain consistency and to actual trigger the sound editor's midi learn stack which arrives by the same branching logic for notes, but carry a far more specific use case. Still, the consistency in order of objects having the responsibility for this class of message has deviated as  little as possible, sparing only code paths that were not going to be used for this feature.

src/deluge/deluge.cpp
```diff
@@ -119,6 +119,9 @@ Song* currentSong = NULL;
 Song* preLoadedSong = NULL;
 
 bool sdRoutineLock = true;
+#if !defined(NDEBUG)
+String* debugPath = new String();
```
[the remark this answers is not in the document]
m-m-adams: This is still not used anywhere

src/deluge/gui/views/view.cpp
```diff
@@ -756,8 +757,16 @@ void View::clearMelodicInstrumentMonoExpressionIfPossible() {
...
 
+/* adapt program change messages to cc*/
+void View::pcReceivedForMIDILearn(MIDIDevice* fromDevice, int32_t channel, int32_t pc) {
+	// to be overridden
+}
+
 void View::ccReceivedForMIDILearn(MIDIDevice* fromDevice, int32_t channel, int32_t cc, int32_t value) {
 	if (thingPressedForMidiLearn != MidiLearn::NONE) {
+		char popupText[100];
+		sprintf(popupText, "CRFML 0 %d", thingPressedForMidiLearn);
```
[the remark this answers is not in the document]
m-m-adams: If not then this string needs to change to something user readable

src/definitions_cxx.hpp
```diff
@@ -1084,6 +1084,7 @@ constexpr int32_t MIDI_CHANNEL_NONE = 255;
 constexpr int32_t MIDI_CC_NONE = 255;
 
 constexpr int32_t IS_A_CC = NUM_CHANNELS;
+constexpr int32_t IS_A_PC = IS_A_CC + NUM_CHANNELS; // CC128 is max.
```
m-m-adams, line 1087: In retrospect from the CC hack this doesn't scale well - it'd be better to make it 2 << 24 and then just & it with the int32 (but really 5 bit) channel

That way checking if a saved midi command is a PC is note & 0xFF000000 == IS_A_PC and we don't have to worry about what happens if we need more types in the future

tastycode: I believe that we shouldn't break consistency with the CC hack, because then we'll have two different hacks. We should fix the CC and this at the same time using the approach you are advocating for here. That way we eliminate all the hacks and don't introduce inconsistency through this PR.

src/deluge/gui/menu_item/menu_item.cpp
```diff
@@ -30,6 +30,9 @@ void MenuItem::learnCC(MIDIDevice* fromDevice, int32_t channel, int32_t ccNumber
 	learnKnob(fromDevice, ccNumber, 0, channel);
 }
 
+void MenuItem::learnProgramChange(MIDIDevice* fromDevice, int32_t channel, int32_t programNumber) {
```
m-m-adams, line 33: as a stub this should be in the header

src/deluge/gui/views/view.cpp
```diff
@@ -756,6 +757,11 @@ void View::clearMelodicInstrumentMonoExpressionIfPossible() {
 	}
 }
 
+/* adapt program change messages to cc*/
+void View::pcReceivedForMIDILearn(MIDIDevice* fromDevice, int32_t channel, int32_t pc) {
```
m-m-adams, line 761: nit: move to header

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -1111,10 +1111,21 @@ ActionResult SoundEditor::verticalEncoderAction(int32_t offset, bool inCardRouti
...
+		getCurrentMenuItem()->learnProgramChange(fromDevice, channel, program);
+		return true;
+	}
+	return false;
+}
 bool SoundEditor::noteOnReceivedForMidiLearn(MIDIDevice* fromDevice, int32_t channel, int32_t note, int32_t velocity) {
 	return getCurrentMenuItem()->learnNoteOn(fromDevice, channel, note);
 }
 
+bool SoundEditor::programChangeReceived(MIDIDevice* fromDevice, uint8_t channel, uint8_t program) {
```
m-m-adams, line 1125: nit: probably should be in the header too

src/definitions_cxx.hpp
m-m-adams: yep reasonable

m-m-adams (approved):
GTG when format is updated

PR #805, opened 2023-12-11
Allow single row copy. Allow gentle paste

tastycode:
All that was changed
- Added a flag to `pasteNotes` to avoid clearing the current clip before pasting
- When copying notes, check to see if any of the audition pads are pressed, if so... only copy those rows
- Appropriate proposed shortcuts

The ability to copy/paste between kits/synths and the scaling logic as well as what happens when you paste into a clip that doesn't have scale on.. all this logic existed already. 

I realize it might be intuitive to paste into a specific location, but this was a bit too hairy for me to handle, so pasting semantics haven't changed. It's literally as if the clip you copied from only had one row in it. 

[Video Demo / QA](https://www.dropbox.com/scl/fi/6rfouwmlf353l4zt2vm7d/2023-12-11-copy-notes-del.mp4?rlkey=ukf4mr364gz5llm4ocy99pv9i&dl=0)


m-m-adams (commented):
The code is good but I'm not sure on the UI for pasting

Here's my suggestion:
If you hold an audition row while pasting, do the soft paste as follows:

    i=0, j=0
   while (i < num pressed rows):
      while (j < num copied rows):
        if copied row has notes:
          copy notes to held row
          j+=1
          i+=1
         continue
        else:
          j+=1

So that n copied rows get copied into m held rows, filling from the bottom

bobtwinkles:
Please also add this to `community_features.md` so it doesn't get lost, and whatever user interface we come up with is documented

tastycode:
> The code is good but I'm not sure on the UI for pasting

I'm very open to other ideas in terms of which shortcut to use for the behavior. Functionally, pasting actually hasn't changed semantics whatsoever other than its not removing the notes of the clip before doing so. Outside of this PR, if you copy from anywhere, and paste. You'll see the notes in the same locations as in this PR. The only difference with the new command, is that the existing notes won't be wiped out. 
I mean this in every facet and variant of behavior, for example:



If you have an instrument clip that only has a single rhythmic row of one note, and you are viewing the grid at 1/16/pad and copy notes.
Then you go to another instrument, no change in zoom, scroll up the grid just a bit so it starts on D3 instead of C3. 

Paste the notes.

The notes will end up transposed by two semitones. 

This is the case before this PR and after. 

If you then create a kit track with exactly one sample/pad. Hit paste, regardless of scroll, that pad will be filled with the on notes rhythm.

Then try... going to a kit track and zooming into 1/32nd note. 

Then add a second pad to the kit, and press paste. The rows will fill from the bottom, from the source clip, and they'll be 2x as fast as they originally were. 




If you try this whole exercise with multiple notes instetad, you'll discover very unique semantics in specific situations.

Due to the number of logical branches to test here, I was shy to implement a specific "Paste in this position" functionality, because the current paste functionality already does so, its just in a way that's relative to the pasting viewport. 

I wanted to be through this PR quickly , so the compromise was to retain the complexity of the existing logic. To simplify QA. Issues could arise simplky out of the way notes are pasted over existing notes and altering the quite complex semantics of what you can do with paste (e.g. copy from a scale, and turn on scale then paste) 

I considered a lot of different

tastycode:
> I considered a lot of different

Sorry, github cut me off there. The existing paste logic is crazy complex and just to test all the scenarios that already exist just took me 45 minutes of fiddling around. This is why i was shy to introduce any new semantics WRT note positioning.

tastycode:
@m-m-adams ⚠️  DO NOT MERGE. 

- [ ] Select menu no longer works outside of instrument clip

m-m-adams:
It's fine, your branch was just out of date

PR #810, opened 2023-12-12
Kit Affect Entire/Global FX Menu

seangoodvibes:
Added new menu that opens when when pressing select in a Kit clip with affect entire enabled in order to control kit affect entire / global FX.

With affect entire enabled in a kit, after you've entered the Kit FX menu, you can use shortcuts to open up the various menu's for editing the kit affect entire params

Also took the opportunity to clean up the code and remove some excess / duplicate audio clip menu definitions that were for the same parameters defined as part of the new global FX menu's for song/arranger/performance view/kit affect entire


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -816,6 +838,16 @@ ActionResult SoundEditor::potentialShortcutPadAction(int32_t x, int32_t y, bool
 			goto doSetup;
 		}
 
+		//For Kit Instrument Clip with Affect Entire Enabled
```
m-m-adams, line 841: hmm I'm a bit less sure about this, I wonder if this will lead to more confusion than it helps. Maybe these should only work if you've already opened the kit fx menu?

seangoodvibes: Agreed. Made the change as suggested :)

m-m-adams (approved):

PR #814, opened 2023-12-12
Prep for dbt-toolchain relocation

litui:
Change to the dbt-toolchain paths (not the version -- still at v10) to support the move of the dbt-toolchain repo to [SynthstromAudible](https://github.com/SynthstromAudible/dbt-toolchain).


litui:
Tested on my end and no issues downloading the toolchain fresh. Merging to get it over with.

PR #826, opened 2023-12-14
Added new scales

soymonitus:
Most of the scales present in the Novation products plus a few more, so to more or less represent a good number of variations for both major and minor scales, western and eastern scales.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -458,76 +458,13 @@ traverseClips:
 		goto traverseClips;
 	}
 
-	// Determine the majorness or minorness of the scale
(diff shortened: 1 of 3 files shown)


soymonitus:
There was a code trying to figure out the majorness or minorness of a set of notes, but since adding a lot of new scales, the code was not working correctly for those. I found out that just removing the code made easier for the system to find out the new scales. And for the pre-existing scales it was detecting the scale perfectly given a set of notes in the clip, so no idea why that code was needed.

m-m-adams (approved):

m-m-adams:
This is now reverted due to the scale bug

PR #833, opened 2023-12-16
move user friendly MPE match function from melodic instrument to learned midi

m-m-adams:
Refactor this to make non melodic instrument uses of MPE easier

@seangoodvibes should help with your follow channel


bobtwinkles:
Can this break any existing mappings, and if so can we correct for that somehow? If not, this looks good to merge.

m-m-adams:
No, this just moves the existing functions to a better abstraction level but they're still only used within melodic instruments

At some point in the future I want to change the MPE matching in kits and learned params to use these functions, which might have mapping issues but that will be down the road

seangoodvibes (approved):

PR #834, opened 2023-12-16
[DRAFT] Add sysex load, and SD load commands to DBT. Add sysex logging comman…

tastycode:
- [x] Sysex Logging Works
- [x] Get_started.md updated
- [x] `./dbt loadfw sysex -K <somekey>`
- [x] `./dbt loadfw sd` 

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/d0928a3e-1ca3-40d5-8798-df331154beca

https://github.com/SynthstromAudible/DelugeFirmware/assets/809953/59424c75-cc7c-405a-b1db-808cd09a5a9d

⚠️ There is still a crash if you ever used logging then tried to send firmware via loadfw/sysex

### DIFF
--- a/scripts/tasks/task-loadfw.py
+++ b/scripts/tasks/task-loadfw.py
@@ -1,14 +1,16 @@
 #! /usr/bin/env python3
-import argparse
-from pathlib import Path
-import subprocess
-import os
-import datetime
-from pathlib import Path
 from iterfzf import iterfzf
+from pathlib import Path
+from pathlib import Path
+from tqdm import tqdm
+import argparse
+import datetime
 import mido
+import os
 import psutil
 import shutil
+import subprocess
+import time
 
 def argparser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(
@@ -143,8 +145,16 @@ def load_sysex(args) -> None:
     if args.verbose:
         print(f"Sending syx file to {device}")
 
+    MAX_KB_SEC = 300
+    bytes_sent_sec = 0
     with mido.open_output(device) as output:
-        for msg in mido.read_syx_file("output.syx"):
+        for msg in tqdm(mido.read_syx_file("output.syx")):
+            bytes_sent_sec +=len(msg.bin())
+            if bytes_sent_sec > MAX_KB_SEC*1024:
+                time.sleep(1)
+                bytes_sent_sec = 0
+            else:
+                time.sleep(0.001)
             output.send(msg)
     print(f"Sent syx file: {args.build} to device {args.device}")
 
--- a/scripts/tasks/task-loadfw.py
+++ b/scripts/tasks/task-loadfw.py
@@ -1,0 +1,115 @@
+#! /usr/bin/env python3
+import argparse
+import subprocess
+import os
+from pathlib import Path
+from iterfzf import iterfzf
+import mido
+import psutil
+import shutil
+
+def argparser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="loadfw",
+        description="Loads the deluge firmware via SD or sysex (see docs/dev/getting_started.md)"
+    )
+    parser.group = "Development"
+    parser.add_argument(
+        "-B", "--build", help="The firmware image to be sent", default="build/Debug/deluge.bin"
+    )
+    parser.add_argument(
+        "-V", "--verbose", help="Output additional information", default=is_verbose_mode()
+    )
+    subparsers = parser.add_subparsers()
+    subparsers.required = True
+    parser_sd = subparsers.add_parser('sd', help="Loads the deluge firmware via SD")
+    parser_sd.add_argument(
+        "-N", "--name", help="The volume name of the SD card"
+    )
+    parser_sd.set_defaults(func=load_sd)
+
+    parser_sysex = subparsers.add_parser('sysex', help="Loads the deluge firmware via sysex")
+    parser_sysex.add_argument(
+        "-D", "--device", help="The device the firmware image is being sent to", default="Deluge OUT"
+    )
+    parser_sysex.add_argument(
+        "-K", "--key", help="Your deluge insecure developer access key (See settings/community features/allow insecure)", required=True
+    )
+    parser_sysex.set_defaults(func=load_sysex)
+
+    return parser
+
+def is_verbose_mode() -> bool:
+    return os.environ.get('LOADFW_VERBOSE', False) or os.environ.get('DBT_VERBOSE', False)
+
+def run_command(command: list, verbose: bool) -> None:
+    if verbose:
+        print(f"Executing command: {' '.join(command)}")
+
+    result = subprocess.run(command, capture_output=True, text=True)
+    
+    if verbose:
+        print(f"Exit code: {result.returncode}")
+        print(f"Output: {result.stdout}")
+        if result.returncode != 0:
+            print(f"Error: {result.stderr}")
+
+    result.check_returncode()
+
+
+def load_sd(args) -> None:
+    # only list fstype=msdos
+    device_name = args.name
+    if device_name is None:
+        disk_candidates = [part for part in psutil.disk_partitions() if part.fstype == 'msdos']
+        if len(disk_candidates) == 0:
+            raise RuntimeError("No deluge-compatible partitions found")
+        print(f"Available disks: {disk_candidates}")
+        def iter_disks():
+            for disk in disk_candidates:
+                yield disk.mountpoint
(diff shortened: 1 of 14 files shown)


scripts/tasks/task-loadfw.py
```diff
@@ -0,0 +1,157 @@
...
+        source_ts = str(datetime.datetime.fromtimestamp(os.path.getmtime(source_build)))
+        target_ts = str(datetime.datetime.fromtimestamp(os.path.getmtime(dest)))
+            
+        print(f"Copied {source_build} [{source_ts}] to {device_name} {dest} [{target_ts}]")
+    except:
+        print(f"Failed to copy {source_build} to {device_name}")
+        raise
+
+def load_sysex(args) -> None:
+    loadfw_path = Path("./toolchain/darwin-arm64/loadfw")
```
bfredl, line 123: we don't want to hardcode `darwin-arm64` as it might not exist, maybe a generic `toolchain/host` folder for things that are compiled on the host (or get the right toolchain name from some ENV var, I guess)

tastycode: THANK YOU! omg.... i'm so silly

soymonitus:
Are you saying that we are going to be able to debug with just our Deluge connected with USB to the computer?? Does the command flash the last compiled firmware automatically?

tastycode:
> Are you saying that we are going to be able to debug with just our Deluge connected with USB to the computer??

Yes. I've been debugging that way all week and don't have any special hardware. FYI, to make it faster I recommend using USBmidi instead of DIN midi. 

>  Does the command flash the last compiled firmware automatically?

Does "the command"? I created multiple commands in this PR, each with their own job. It's difficult for me to parse this question because it presumes a single command where there are multiple mentioned in the PR description.

- `./dbt loadfw sd`
   
    A script to mitigate common issues with loading firmware to SD
    
     - Accidentally loading a stale image
     - Not noticing that you've inadvertently toggled write protection on the SD
     - Errors that happen here are more noticeable than when doing the same process manually
     
- `./dbt loadfw sysex -K 102120` 

    - Loads firmware over sysex
    - will prompt for confirmation of build image if there is ambiguity or possible staleness
    - If deluge midi ports have changed, allows you to select MIDI ports
    - This will crash the deluge if the below command has been ran even once since last power on
- `./dbt sysex-console` 
   
    Start streaming logs from the deluge. Side effect, deluge will crash if you try to send firmware over sysex while this has ever ran on the deluge. I have to power cycle my deluge IF i was using logging, and i want to flash over sysex. I got used to it, its still faster than SD.
    
    
@m-m-adams Hey there, was wondering if you had any feedback on this. This hasn't been in draft for a bit and just wanted to check in. The only thing that could make me feel better about it (other than fixing that crash issue, but i already spent 8+ hours on it) is ensuring the scripts work as well in windows – i don't have a windows machine.

m-m-adams:
Hey - I'll take a look at it at some point but I'm not really familiar with the sysex firmware loading (beyond the protocol) or dbt. I'm kind of hoping one of the other reviewers gets to it

scripts/tasks/task-sysex-console.py
```diff
@@ -0,0 +1,141 @@
...
+
+    result = subprocess.run(command, capture_output=True, text=True)
+    
+    if verbose:
+        print(f"Exit code: {result.returncode}")
+        print(f"Output: {result.stdout}")
+        if result.returncode != 0:
+            print(f"Error: {result.stderr}")
+
+    result.check_returncode()
```
bobtwinkles, line 49: Since this logic is shared with `task-loadfw`, maybe it could be moved to `util.py` instead?

scripts/toolchain/requirements.txt
```diff
@@ -7,3 +7,8 @@ GitPython==3.1.37
 pyserial==3.5
 ansi==0.3.6
 kconfiglib==14.1.0
+mido==1.3
+python-rtmidi
+iterfzf
+psutil
+tqdm
```
bobtwinkles, line 14: Should lock versions of these dependencies. It might also be wise to make sure this PR branch doesn't break the CI image build. There's a script to simulate what the GH runner will do, here: https://github.com/SynthstromAudible/dbt-toolchain/blob/218b2352b3449045c620de6e306435c372c79ddc/actions-support/simulate_ci_image_build.sh and you'll need to hack the docker manifest here: https://github.com/SynthstromAudible/dbt-toolchain/blob/218b2352b3449045c620de6e306435c372c79ddc/actions-support/ci-image/Dockerfile#L48-L49.

This will require a few gigs of storage space for all the toolchains, and might take 30 minutes or so to run. I'll try to run this myself in the next few days or so if you can't get to it.

src/deluge/io/debug/sysex.cpp
```diff
@@ -80,8 +80,11 @@ void Debug::sysexDebugPrint(MIDIDevice* device, const char* msg, bool nl) {
 	}
 
 	reply[5 + len] = 0xf7;
+	//0xf0, 0x7d, 0x03, 0x40, 0x00, 'm', 's', 'g', 'e', 't', ' ', 't', 'o', ' ', 'd', 'e', 'b', 'u', 'g', 'g', 0xf7
```
bobtwinkles, line 83: Should probably put this before the header, and explain what the components are.

soymonitus:
I always get CHECKSUM FAIL on the Deluge. I did the commands:
 ./dbt build debug -c -DENABLE_SYSEX_LOAD=on
 flashed it from the SD card the first time, then enabled and checked the password in Allow Insecure
 then
 ./dbt loadfw sysex -K 6BB32EB9
 
 But i get the checksum fail always

tastycode:
Okay, thats definitely related to transfer speed. I’ve got a full rewrite
of the entire stack for sysex communication that i accidentally undertook
in the works.i had to do it to enable more agility and less brittleness due
to all the message types i needed for the browser-deluge link.  It’s become
my obsession. It’s a super finicky beast, but i am almost done. I am
finally to the point where my rewrite i actually doing the same thing. So,
this is somehow great to hear that it was iffy on the old sysex stack.

tastycode:
Oh btw. Are you using din midi or UsB for the sysex transfer?

soymonitus:
> Oh btw. Are you using din midi or UsB for the sysex transfer?
I am using USB (adn selecting Deluge Port 1)

m-m-adams:
> > Oh btw. Are you using din midi or UsB for the sysex transfer?
> 
> 
> 
> I am using USB (and selecting Deluge Port 1)

Should be port 3 for sysex

M0R172:
PLEASE everybody help get this into the community FW ASAP, this would make developing so much faster and less painful! I think it would massively increase the speed (and quality, due to easier debugging) of the Deluge Community effort!

Works for me under Linux (Kubuntu 22.04), with an OLED Deluge over USB-MIDI Some Observations:

- I do not get any crashes or other problems using loadfw sysex after sysex-console, I can use both back and forth without problems, using the same midi port or a different one. I was not crazy enough to try them simultaneously from different terminals though...
- "Should be port 3 for sysex" - for me it works with whichever deluge port I select. Is there some reason to prefer 3 for sysex?
- I *sometimes* get the OLED "screen-wrapping" issue after loadfw sysex. I suspect that this has more to do with the chainloading *per se* than with the preceding sysex transfer. A Power cycle always fixes the screen, so maybe there chainload could do some additional "oled reset" to prevent this issue?

m-m-adams:
If deluge isn't sending MIDI to/from anything else then any of the three ports should work, if it is then there are rules for how sysex and other MIDI data should interleave that we don't follow. It's a deliberate choice to use the third port for separation instead

jamiefaye:
I am about to install the Synthstrom SysEx code. We might as well get it over with.

tastycode:
About two weeks ago, i merged upstream changes from my local sysex rewrite
into this branch. I hadnt pushed up my changes because i had as of yet
failed to verify sysex upload. I got sidetracked with some drama with my
roommates and making rent, i’m back but the diff between this pr’s code and
my local is still quite large. Can you just give me the day to at least
force push up my changes? Everything in the code on this PR is nearly two
months oht of date.

soymonitus:
@m-m-adams the latest merged PR about sysex-logging command I think makes this PR unneeded

tastycode:
Can you reference that PR?

soymonitus:
https://github.com/SynthstromAudible/DelugeFirmware/pull/1199

m-m-adams:
Thanks for the contributions, but I believe all of these functions are covered by #1199 

If there's something else added in this PR that I missed please open another pull request!

PR #837, opened 2023-12-16
In key transpose via Global MIDI and sequenced via MIDI clips

0beron:
Addition of transposition performance and sequencing controls to change scale / key / mode etc

Transpose events will affect all synth and MIDI clips that are set to scale mode, and will stay within the scale note of the starting point, eg if the scale is currently C major, then notes CDEFGAB will change the root note and scale to reach D Dorian, E Phyrgian, F Lydian, G mixolydian, A Aeolian and B Locrian. 
Octave will be chosen the be closest to current song root note for the first transpose event after each song load, and subsequent octaves will be obeyed. Eg, song root note == 12 (C, one octave up), any incoming 'D' will move up a second. Subsequent notes will change the octave relative to this starting point. This is to avoid having to guess the song root note when sending the first transpose control note and risking shifting multiple octaves away.

Changing to modes of a different major scale deliberately omitted for now - other notes will be ignored.
Some scheme to control this could be added but this is not fully thought out yet.

- [x] Global MIDI command that maps a whole channel to transpose behaviour
- [x] Flash storage for global MIDI command
- [x] Basic implementation of transpose behaviour
- [x] Add option to MIDI clips alongside Channels 1-16 and MPE, so that MIDI clips can sequence transpose events internally allowing recording, probability, access via MIDI learn etc.
- [x] Menu to configure control scheme. Allows chromatic or in-key transpose.
- ~~Add some kind of chord mode that allows in-key movements but also modulation to another key.~~
- [x] MIDI clips routed to transpose must not themselves transpose to avoid recursion!


soymonitus:
Hmmm I think this can be tied to the other PR from Sean in which a global midi command / channel can be used to transpose the whole song, or it could be also a shorcut using deluge encoders. The thing is that you could set in settings if the command does chromatic transposition by semitones, or it does this behavior of transposition within scale. That's my opinion. It could make sense to join both features into one, and the user can then decide if the transposition should be done by semitone steps or by scale note jumps

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -851,7 +866,10 @@ void MIDIInstrument::polyphonicExpressionEventPostArpeggiator(int32_t value32, i
                                                               int32_t whichExpressionDimension, ArpNote* arpNote) {
 
 	// If we don't have MPE output...
```
m-m-adams, line 868: comment in the wrong spot now

m-m-adams (commented):
lgtm - one comment that needs to be moved down a few lines now

m-m-adams (approved):

lobit12:
how do you enable this feature currently?  is it working well enough to try out?

0beron:
The feature is in the nightly builds of the community branch and release 1.1

It is described in the documentation in https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/community_features.md (Section 3.3. MIDI, then under #837 MIDI control over transpose / scale.)

There's an issue where audition notes getstuck on over a transpose event, which I've fixed but the new PR hasn't been merged yet, it's at https://github.com/SynthstromAudible/DelugeFirmware/pull/1319

0beron:
I'm on the synthstrom discord also if you have any other questions

lobit12:
okay thanks man! do I need an invite to get on the discord ?

sapphire-arches:
[This invite link](https://discord.com/invite/s2MnkFqZgj) should work!

PR #841, opened 2023-12-16
c1.0.1 Bugfix / Invalid Param Collection in getModelStackWithParam for Automation View

seangoodvibes:
We have received a few bug reports of crashes when entering automation view.

This may have been due to a possible scenario where getModelStackWithParam was returned with an invalid Param Collection.

To resolve this, I have modified the code to getModelStackWithParam to first get the modelStackWithParamCollection and check that it is valid before getting the modelStackWithParamId from the Param Collection and then finally getting the modelStackWithParam from the modelStackWithParamId.

### DIFF
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -2939,9 +2939,14 @@ ModelStackWithAutoParam* AutomationInstrumentClipView::getModelStackWithParam(Mo
 			}
 
 			if (summary) {
-				ParamSet* paramSet = (ParamSet*)summary->paramCollection;
-				modelStackWithParam =
-				    modelStackWithThreeMainThings->addParam(paramSet, summary, paramID, &paramSet->params[paramID]);
+				ModelStackWithParamCollection* modelStackWithParamCollection =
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2939,9 +2939,14 @@ ModelStackWithAutoParam* AutomationInstrumentClipView::getModelStackWithParam(Mo
 			}
 
 			if (summary) {
-				ParamSet* paramSet = (ParamSet*)summary->paramCollection;
-				modelStackWithParam =
-				    modelStackWithThreeMainThings->addParam(paramSet, summary, paramID, &paramSet->params[paramID]);
+				ModelStackWithParamCollection* modelStackWithParamCollection =
+				    modelStackWithThreeMainThings->addParamCollectionSummary(summary);
+				if (modelStackWithParamCollection) {
+					ModelStackWithParamId* ModelStackWithParamId = modelStackWithParamCollection->addParamId(paramID);
```
bobtwinkles, line 2945: nit: local variables should start with lower case letters

bobtwinkles (commented):
style nit aside i think this fix should work

seangoodvibes:
> style nit aside i think this fix should work

Fixed, thanks for catching that

PR #858, opened 2023-12-19
Fix upstream usb midi port reversal

m-m-adams:
Explicitly initialize port 0 for clarity

Fix switch so that port 0 and 1 aren't swapped on save


seangoodvibes (approved):
Tested, seems to work, fix is simple enough, approved.

PR #861, opened 2023-12-20
Safety checks to avoid getting crashes when trying to randomize incompatible rows

soymonitus:
Incompatible rows are:
- Uninitialized rows (they show name U1, U2, U3..)
- Empty rows (they show name NO SOUND)
- Synths preset rows


m-m-adams:
Seems to fix both issues mentioned in 786!

m-m-adams (approved):

PR #862, opened 2023-12-20
Seperate receiving midi from matching in kits and synths

m-m-adams:
Refactor to allow skipping the match section when the ultimate destination for incoming midi is already known

### DIFF
--- a/src/deluge/model/instrument/kit.cpp
+++ b/src/deluge/model/instrument/kit.cpp
@@ -658,6 +658,7 @@ void Kit::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel
 	}
 }
 
+//not updated for midi follow, this seems dumb and is just left for backwards compatibility
 bool Kit::offerReceivedPitchBendToLearnedParams(MIDIDevice* fromDevice, uint8_t channel, uint8_t data1, uint8_t data2,
                                                 ModelStackWithTimelineCounter* modelStack) {
 
@@ -1195,6 +1196,13 @@ void Kit::offerReceivedNote(ModelStackWithTimelineCounter* modelStack, MIDIDevic
 	}
 }
 
+void Kit::offerReceivedPitchBendToDrum(ModelStackWithTimelineCounter* modelStackWithTimelineCounter, Drum* thisDrum,
+                                       MIDIDevice* fromDevice, uint8_t channel, uint8_t data1, uint8_t data2,
+                                       int32_t level, bool* doingMidiThru) {
+	int16_t value16 = (((uint32_t)data1 | ((uint32_t)data2 << 7)) - 8192) << 2;
+	thisDrum->expressionEventPossiblyToRecord(modelStackWithTimelineCounter, value16, 0, level);
+}
+
 void Kit::offerReceivedPitchBend(ModelStackWithTimelineCounter* modelStackWithTimelineCounter, MIDIDevice* fromDevice,
                                  uint8_t channel, uint8_t data1, uint8_t data2, bool* doingMidiThru) {
 
@@ -1216,8 +1224,8 @@ void Kit::offerReceivedPitchBend(ModelStackWithTimelineCounter* modelStackWithTi
 			}
 			else { // Or, if Drum does not have MPE input, then this is a channel-level message.
 yesThisDrum:
-				int16_t value16 = (((uint32_t)data1 | ((uint32_t)data2 << 7)) - 8192) << 2;
-				thisDrum->expressionEventPossiblyToRecord(modelStackWithTimelineCounter, value16, 0, level);
+				offerReceivedPitchBendToDrum(modelStackWithTimelineCounter, thisDrum, fromDevice, channel, data1, data2,
+				                             level, doingMidiThru);
 			}
 		}
 	}
--- a/src/deluge/model/instrument/kit.cpp
+++ b/src/deluge/model/instrument/kit.cpp
@@ -628,13 +628,19 @@ void Kit::renderOutput(ModelStack* modelStack, StereoSample* outputBuffer, Stere
 	                                      shouldLimitDelayFeedback, isClipActive, InstrumentType::KIT, 8);
 }
 
+//offer the CC to kit gold knobs without also offering to all drums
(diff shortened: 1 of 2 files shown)


src/deluge/model/drum/kit.cpp
```diff
@@ -628,13 +628,19 @@ void Kit::renderOutput(ModelStack* modelStack, StereoSample* outputBuffer, Stere
 	                                      shouldLimitDelayFeedback, isClipActive, InstrumentType::KIT, 8);
 }
 
+//offer the CC to kit gold knobs without also offering to all drums
+void Kit::offerReceivedCCToModControllable(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber, uint8_t value,
+                                           ModelStackWithTimelineCounter* modelStack) {
+	ModControllableAudio::offerReceivedCCToLearnedParams(
+	    fromDevice, channel, ccNumber, value,
+	    modelStack); // NOTE: this call may change modelStack->timelineCounter etc!
```
bobtwinkles, line 636: style nit: move the comment so this wraps.

src/deluge/model/drum/kit.h
```diff
@@ -50,6 +50,8 @@ class Kit final : public Instrument, public GlobalEffectableForClip {
 	                  int32_t* reverbBuffer, int32_t reverbAmountAdjust, int32_t sideChainHitPending,
 	                  bool shouldLimitDelayFeedback, bool isClipActive);
 	void notifySamplesInterruptsSuspended();
+	inline void offerReceivedCCToModControllable(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber,
+	                                             uint8_t value, ModelStackWithTimelineCounter* modelStack);
```
bobtwinkles, line 54: `inline` functions require the definition to be available at point of use, not just declaration. I think that probably happens in practice because this only gets used in `kit.cpp`, but using it like this in the header feels a bit odd.

src/deluge/model/drum/kit.cpp
```diff
@@ -994,177 +994,182 @@ void Kit::getThingWithMostReverb(Sound** soundWithMostReverb, ParamManager** par
...
+	else {
+		if (thisNoteRow) {
+			if (shouldRecordNotes && thisDrum->auditioned
+			    && ((playbackHandler.recording == RECORDING_ARRANGEMENT && instrumentClip->isArrangementOnlyClip())
+			        || currentSong->isClipActive(instrumentClip))) {
 
-					if (thisNoteRow && thisDrum->type == DrumType::SOUND
-					    && !thisNoteRow->paramManager.containsAnyMainParamCollections()) {
-						FREEZE_WITH_ERROR("E326"); // Trying to catch an E313 that Vinz got
+				if (playbackHandler.recording == RECORDING_ARRANGEMENT && !instrumentClip->isArrangementOnlyClip()) {}
```
bobtwinkles, line 1139: I see this is inherited but this construct is super weird.

src/deluge/model/drum/kit.cpp
```diff
@@ -994,177 +994,182 @@ void Kit::getThingWithMostReverb(Sound** soundWithMostReverb, ParamManager** par
...
+
+	for (Drum* thisDrum = firstDrum; thisDrum; thisDrum = thisDrum->next) {
+
+		// If this is the "input" command, to sound / audition the Drum...
+		// Returns true if midi channel and note match the learned midi note
+		// We don't need the MPE match because all types of matches should sound the drum
+		if (thisDrum->midiInput.equalsNoteOrCCAllowMPE(fromDevice, channel, note)) {
+			receivedNoteForDrum(modelStack, fromDevice, on, channel, note, velocity, shouldRecordNotes, doingMidiThru,
+			                    thisDrum);
+		}
```
bobtwinkles, line 1175: I think there's a behavior change here -- in the original version of the loop, only the first drum that claims the note when `on == true` will get that event. This version doesn't do that, I think.

bobtwinkles: `lookingForFirstDrumForNoteOn` in the original code.

m-m-adams: Yeah it's intentional - removes a micro optimization to simplify the code. It only saves a check if two drums in the same kit are mapped to the same midi note

src/deluge/model/drum/kit.cpp
m-m-adams: Yes, as far as I can tell it is a check in case of a niche situation in arranger view where the clip might not actually exist 

I couldn't think of an appropriate way to simplify it

src/deluge/model/drum/kit.h
m-m-adams: Yeah fair point, and that won't apply once it gets called by midi follow. I'll remove the inline since this is non critical code

src/deluge/model/drum/kit.cpp
```diff
@@ -652,6 +658,7 @@ void Kit::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel
 	}
 }
 
+//not updated for midi follow, this seems dumb and is just left for backwards compatibility
```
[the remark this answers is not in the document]
m-m-adams: Pitch bend is a mod source in the synth engine post 4.0, so there's no reason to learn it to a parameter anymore. We could keep it so that it doesn't change the sound of old presets in case someone did use it

src/deluge/model/drum/kit.cpp
```diff
@@ -1223,27 +1246,14 @@ void Kit::offerReceivedCC(ModelStackWithTimelineCounter* modelStackWithTimelineC
...
-					int16_t value16 = (value - 64) << 9;
-					thisDrum->expressionEventPossiblyToRecord(modelStackWithTimelineCounter, value16, 1, level);
-				}
-			}
-
-			// If not an MPE input, we don't want to respond to this CC74 at all (for this Drum).
+		int32_t bend_range = BEND_RANGE_FINGER_LEVEL;
+		switch (thisDrum->midiInput.checkMatch(fromDevice, channel)) {
+		case MPE_MASTER:
+			bend_range = BEND_RANGE_MAIN;
```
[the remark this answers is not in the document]
m-m-adams: Cool thanks!

bobtwinkles (approved):

m-m-adams:
Force push is a rebase to incorporate the kit files moving into the instruments folder

PR #864, opened 2023-12-21
MPE fix - allow note offs on channel 16

### LINKED ISSUE #863: Stuck notes on MPE when Lower is configured to use all 16 channels and Upper is disabled

I get stuck notes when learning a midi MPE keyboard (Lower) and then playing it, with channels setup to take the whole 16 channels (Upper zone disabled).

I don't get stuck notes

_No response_

OLED

Nightly

deluge-v1_0_0-nightly+2023_12_21-58b0d04

Boot Deluge, connected to an MPE keyboard in the midi input and a midi monitor in the output
Set Upstream device 1 -> MPE -> IN -> Lower num of channels -> 15
Set Upstream device 1 -> MPE -> IN -> Upper num of channels -> 0
Set Upstream device 1 -> MPE -> OUT -> Lower num of channels -> 15
Set Upstream device 1 -> MPE -> OUT -> Upper num of channels -> 0
Create Midi clip and set output channel to MPE Lower
learn the keyboard with Learn + hold audition + press a key
Play any notes in the keyboard
See the midi monitor getting notes but never get the note offs, stuck notes

Repeat but setting MPE IN Lower NumChannels to 14 and MPE OUT Lower NumChannels to 14 and there won't be any stuck notes

m-m-adams:
Fix #863

### DIFF
--- a/src/deluge/model/instrument/midi_instrument.cpp
+++ b/src/deluge/model/instrument/midi_instrument.cpp
@@ -723,7 +723,7 @@ void MIDIInstrument::noteOffPostArp(int32_t noteCodePostArp, int32_t oldOutputMe
 
 	// Or, MPE
 	else {
-		if (oldOutputMemberChannel <= 14) {
+		if (oldOutputMemberChannel <= 15) {
 			mpeOutputMemberChannels[oldOutputMemberChannel].lastNoteCode = noteCodePostArp;
 			mpeOutputMemberChannels[oldOutputMemberChannel].noteOffOrder = lastNoteOffOrder++;
 
--- a/src/deluge/model/instrument/midi_instrument.cpp
+++ b/src/deluge/model/instrument/midi_instrument.cpp
@@ -49,7 +49,7 @@ MIDIInstrument::MIDIInstrument() : NonAudioInstrument(InstrumentType::MIDI_OUT) 
 	modKnobMode = 0;
 	memset(modKnobCCAssignments, CC_NUMBER_NONE, sizeof(modKnobCCAssignments));
 
-	for (int32_t c = 1; c <= 14; c++) {
+	for (int32_t c = 0; c <= 15; c++) {
 		mpeOutputMemberChannels[c].lastNoteCode = 32767;
 		mpeOutputMemberChannels[c].noteOffOrder = 0;
 	}
@@ -593,7 +593,7 @@ void MIDIInstrument::noteOnPostArp(int32_t noteCodePostArp, ArpNote* arpNote) {
 		                                   ? MIDIDeviceManager::lowestLastMemberChannelOfLowerZoneOnConnectedOutput
 		                                   : 14;
 
-		uint8_t numNotesPreviouslyActiveOnMemberChannel[15];
+		uint8_t numNotesPreviouslyActiveOnMemberChannel[16];
 		memset(numNotesPreviouslyActiveOnMemberChannel, 0, sizeof(numNotesPreviouslyActiveOnMemberChannel));
 
 		int32_t outputMemberChannelWithNoteSharingInputMemberChannel = 16; // 16 means "none", I think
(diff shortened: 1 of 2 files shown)


src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -723,7 +723,7 @@ void MIDIInstrument::noteOffPostArp(int32_t noteCodePostArp, int32_t oldOutputMe
 
 	// Or, MPE
 	else {
-		if (oldOutputMemberChannel <= 14) {
```
soymonitus, line 726: Have you checked if the change would be needed in lines 52 and 606?

soymonitus (commented):
Just one comment I was wondering

src/deluge/model/instrument/midi_instrument.cpp
m-m-adams: Yeah, and looks like there's a couple 1s that might need to be zero as well

Thanks for pointing it out!

soymonitus: I tested it and it worked but there was just one small moment where one of the keys of the keyboard didn't trigger sound for a while. Maybe due to these extra changes needed, or maybe not. I will re-test again with your latest changes to see if I encounter any glitch in playing

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -602,7 +602,8 @@ void MIDIInstrument::noteOnPostArp(int32_t noteCodePostArp, ArpNote* arpNote) {
 			// Count up notes per member channel. This traversal will *not* find the new note that we're switching on, which will have had its toMIDIChannel set to 16 by Arpeggiator (we'll decide and set it below).
 			for (int32_t n = 0; n < arpeggiator.notes.getNumElements(); n++) {
 				ArpNote* thisArpNote = (ArpNote*)arpeggiator.notes.getElementAddress(n);
-				if (thisArpNote->outputMemberChannel >= 1 && thisArpNote->outputMemberChannel <= 14) {
```
soymonitus, line 605: This seems super right, in fact, the calculation of lowestMemberChannel and highestMemberChannel will give you the REAL member channels configured for each zone, to use them in the comparisons in the code (if blocks and whatnot)

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -723,39 +724,37 @@ void MIDIInstrument::noteOffPostArp(int32_t noteCodePostArp, int32_t oldOutputMe
 
 	// Or, MPE
 	else {
-		if (oldOutputMemberChannel <= 15) {
```
soymonitus: Good to completely remove this check since it becomes useless

src/deluge/model/instrument/midi_instrument.h
```diff
@@ -73,7 +73,8 @@ class MIDIInstrument final : public NonAudioInstrument {
 
 	int8_t modKnobCCAssignments[kNumModButtons * kNumPhysicalModKnobs];
 
-	MPEOutputMemberChannel mpeOutputMemberChannels[15]; // Numbers 1 to 14. 0 is bogus
```
soymonitus, line 76: The comment that was there before is bugging me ( // Numbers 1 to 14, 0 is bogus).
Does that mean that the code in .cpp file should be checking always from index 1 of the array and ignoring mpeOutputMemberChannels[0] ? it should ignore it for notes if MPE Lower Zone but should be taken into account if MPE UPPER ZONE. All code in the cpp file is ignoring the last member channel of either zone, first because they set the length of this array to 15 so last channel was always impossible to select, and by adding that comment they are trying to tell us not to use index 0 of the array, which we can see the like in line 52:

```
    for (int32_t c = 1; c <= 14; c++) {
		mpeOutputMemberChannels[c].lastNoteCode = 32767;
		mpeOutputMemberChannels[c].noteOffOrder = 0;
	}
```

soymonitus: I think we are doing it right now with your modifications because we take into account channel 15 for lower zone, and channel 0 for upper zone.

soymonitus (approved):
I was seeing some things I was not so sure at the beginning of my review, but after re-checking I edited my comments.
In sum, LGTM

src/deluge/model/instrument/midi_instrument.h
m-m-adams: Ya, it's correct in official only because of a different bug which causes channel 1 and 16 to always be treated as MPE

PR #865, opened 2023-12-21
MIDI Loopback

colinmeyer:
This PR introduces a virtual midi loopback, having the same effect as connecting DIN MIDI Out to DIN MIDI In with a physical cable.

There are many uses for this functionality, including

- additive synthesis: have one midi clip trigger the sounds of two or more synth clips
- generative melodies: have two or more midi clips (e.g. 1 and 1a) both controlling the same synth clip
- macros: have several alternative midi clips with different CC modulations to be sent to a synth clip


m-m-adams (commented):
Downside here is it means you can't learn midi to an external controller while a midi sequence is playing. I'd like to see a dedicated way to learn to the loopback device, maybe shift+learn?

colinmeyer:
@m-m-adams not quite understanding what you mean here. This patch does not block midi from going in or out of the delly; If you have midi from an external controller coming in, you can also learn that to delly synths. Or you can learn the midi from a delly clip to both internal and external synths. The patch should have the same effect as plugging a DIN cable from MIDI out to MIDI in (or from MIDI out to external MIDI in on another device, that device's MIDI out to delly MIDI in). Have done basic testing to confirm this

m-m-adams:
Yeah the second one - I get what you mean but as a result there's a current workflow that breaks. A common workaround for the lack of a master control channel while live looping is to unlearn midi from one track and then learn to a new one, while the track is still playing

Trying that on this PR, when I attempt to learn a clip to my controller it instead learns to one of the midi clips that's currently being output

External synths would typically be setup ahead of time, there's not really a reason to change that live. I feel strongly that this shouldn't be merged as is because it is a significant change in how the midi learning system works. Once midi follow is merged and stable that could be revisited but I still think ideally loopback midi would be learned separately from other midi

colinmeyer:
@m-m-adams this PR is a draft, because it is not complete. The intention is to add a per-song setting (off by default) for midi loopback

m-m-adams:
I think it would make more sense to have it always enabled but selectable

1. While learning
2. From midi clips (press midi again to send to loopback and it says channel 5 (loopback) on the screen

colinmeyer:
Now with Song level parameter, adjustable in Song's sound menu, and stored/read from Song.XML file

cowboy:
Instead of looping all channels back, can there be an option to loop back all with a way to select individual channels? I love the feature, but I'm concerned about it being all-or-nothing.

colinmeyer:
@cowboy May I suggest that you try it first? It works quite decently, as-is. You can try my patch, or you can simply use a physical loopback cable, i.e. plug din midi out -> din midi in

Without understanding your use case, I'm hesitant to make the feature more complicated. Currently it is easy to understand and use

colinmeyer:
@m-m-adams have a look now, maybe try it out, rebased cleanly over midi-follow, and works very nicely with that. Feels pretty intuitive, to me, your controller controls and records to whichever clip (midi or synth instrument) is currently active, and you can learn midi clips to instrument clips as you like

m-m-adams:
It is indeed much better with midi follow and I think it's workable as is

For future work I think it would be great to have a non-learn way to select the loopback input since the interaction with learn is definitely still sub optimal

src/deluge/io/midi/midi_device.cpp
```diff
@@ -555,3 +555,17 @@ void MIDIDeviceDINPorts::sendSysex(uint8_t* data, int32_t len) {
 		bufferMIDIUart(data[i]);
 	}
 }
+
+void MIDIDeviceLoopback::writeReferenceAttributesToFile() {}
+
+void MIDIDeviceLoopback::writeToFlash(uint8_t* memory) {}
```
m-m-adams, line 561: Needs to be implemented to learn to commands or midi follow

m-m-adams: I'm not sure that's actually a useful thing to do, so maybe a popup in midi follow to say that it doesn't work is a better idea

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -61,6 +61,7 @@ MIDIDeviceUSBUpstream upstreamUSBMIDIDevice_port1{0};
 MIDIDeviceUSBUpstream upstreamUSBMIDIDevice_port2{1};
 MIDIDeviceUSBUpstream upstreamUSBMIDIDevice_port3{2};
 MIDIDeviceDINPorts dinMIDIPorts{};
+MIDIDeviceLoopback loopbackMidi{};
```
m-m-adams, line 64: Add to config menu to allow MPE settings to be made - https://github.com/SynthstromAudible/DelugeFirmware/blob/4f9dfcf9f5b44c95fb412da716dd0ce65ba18630/src/deluge/gui/menu_item/midi/devices.cpp#L112

src/deluge/io/midi/midi_engine.cpp
```diff
@@ -911,7 +917,7 @@ bool lastWasNoteOn = false;
 #endif
 
 void MidiEngine::midiMessageReceived(MIDIDevice* fromDevice, uint8_t statusType, uint8_t channel, uint8_t data1,
-                                     uint8_t data2, uint32_t* timer) {
+                                     uint8_t data2, uint32_t* timer, bool notDoingLoopback) {
```
m-m-adams, line 920: This isn't necessary, the check can be on received device != loopback device

m-m-adams (approved):

m-m-adams (changes requested):
Sorry - approved accidentally and can't seem to change it

colinmeyer:
I changed it back to a draft PR, so it won't merge

src/deluge/io/midi/midi_device.cpp
```diff
@@ -555,3 +555,17 @@ void MIDIDeviceDINPorts::sendSysex(uint8_t* data, int32_t len) {
 		bufferMIDIUart(data[i]);
 	}
 }
+
+void MIDIDeviceLoopback::writeReferenceAttributesToFile() {}
```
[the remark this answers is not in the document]
colinmeyer: Any guidance for what should be here? The other device definitions seem mysterious to me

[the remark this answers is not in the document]
m-m-adams: Oh you just need to make a port attribute called "loopback", then add it here to match the reading - https://github.com/SynthstromAudible/DelugeFirmware/blob/1bcd598573be85bf0fe19937dd3b9bc9786caa7c/src/deluge/io/midi/midi_device_manager.cpp#L342

src/deluge/io/midi/midi_engine.cpp
```diff
@@ -495,6 +496,11 @@ void MidiEngine::sendMidi(uint8_t statusType, uint8_t channel, uint8_t data1, ui
 	if (statusType == 0x0F || MIDIDeviceManager::dinMIDIPorts.wantsToOutputMIDIOnChannel(channel, filter)) {
 		sendSerialMidi(statusType, channel, data1, data2);
 	}
+
+	// Send loopback (other than clock) to delly
+	if (currentSong->midiLoopback && channel != 0x08) { // channel 0x08 is clock
```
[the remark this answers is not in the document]
colinmeyer: addressed in latest code

src/deluge/io/midi/midi_device.cpp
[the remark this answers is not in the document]
colinmeyer: saving MPE settings for loopback now works

src/deluge/io/midi/midi_device.cpp
colinmeyer: Implemented these, I think at least the first one is necessary for getting loopback in MidiDevices.XML for saving MPE settings? Er, still haven't got this stuff mapped in my head enough

src/deluge/io/midi/midi_device_manager.cpp
colinmeyer: Loopback present in midi devices menu now

src/deluge/io/midi/midi_engine.cpp
colinmeyer: Thanks for pointing that out, felt clumsy while typing it in; now cleaner

src/deluge/io/midi/midi_device.cpp
m-m-adams: Yeah MPE settings are saved in a file, global midi learning is saved in flash

m-m-adams (approved):
Looks good!

lobit12:
sorry for the dumb question but how do you actually enable this feature?

also how do you transpose one track using another?   so far if I physically plug in a midi cable and to a learn between a synth and a midi clip I can get the midi clip to trigger the synth but I cant transpose the original sequence

PR #868, opened 2023-12-22
Bugfix/flash validation

0beron:
Validation for all enum classes read in from flash storage to avoid out of range values.

### DIFF
--- a/src/deluge/storage/flash_storage.cpp
+++ b/src/deluge/storage/flash_storage.cpp
@@ -117,25 +117,9 @@ namespace FlashStorage {
 120: gridAllowGreenSelection
 121: defaultGridActiveMode
 122: defaultMetronomeVolume
--- Future reservation for follow mode: --
-123: midiFollow enable/disable
-124: midiFollow set follow channel synth
-125: midiFollow set follow channel kit
-126: midiFollow set follow channel param
-127: midiFollow set kit root note
-128: midiFollow display param pop up
-129: midiFollow feedback
-130: midiFollow feedback automation mode
-131: midiFollow feedback filter to handle feedback loops
---
-132: defaultSessionLayout
-133: defaultKeyboardLayout
-134: gridUnarmEmptyPads
--- Future reservation for transpose global MIDI command: --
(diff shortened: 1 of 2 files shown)


src/deluge/storage/flash_storage.cpp
```diff
@@ -420,11 +414,23 @@ void readSettings() {
 			defaultBendRange[BEND_RANGE_MAIN] = 12;
 		}
 	}
-	midiEngine.midiTakeover = static_cast<MIDITakeoverMode>(buffer[113]);
+
+	if (util::to_underlying(MIDITakeoverMode::SCALE) < buffer[113]) {
```
bobtwinkles, line 418: Here as well

src/deluge/storage/flash_storage.cpp
```diff
@@ -260,7 +244,12 @@ void readSettings() {
 	cvEngine.setCVTranspose(1, buffer[15], buffer[19]);
 
 	for (int32_t i = 0; i < NUM_GATE_CHANNELS; i++) {
-		cvEngine.setGateType(i, static_cast<GateType>(buffer[22 + i]));
+		if (util::to_underlying(GateType::SPECIAL) < buffer[22 + i]) {
```
bobtwinkles, line 247: Please add a sentinel like `GridDefaultActiveModeMaxElement`

src/deluge/storage/flash_storage.cpp
```diff
@@ -359,7 +348,12 @@ void readSettings() {
 	else {
 		audioClipRecordMargins = buffer[61];
 		playbackHandler.countInEnabled = buffer[62];
-		keyboardLayout = static_cast<KeyboardLayout>(buffer[69]);
+		if (util::to_underlying(KeyboardLayout::QWERTZ) < buffer[69]) {
```
bobtwinkles, line 351: Similarly should have a sentinal

bobtwinkles (commented):
One more small change, then LGTM. Thanks for the change!

bobtwinkles (approved):

PR #869, opened 2023-12-22
Toggle bool class refactoring

seangoodvibes:
Consolidated all Toggle menu's into one Toggle Bool class and pass bool variable as pointer in menu definition

This PR does not handle pointer based bool toggles (e.g. delay ping pong)


src/deluge/gui/menu_item/toggle.h
```diff
@@ -14,4 +14,17 @@ class Toggle : public Value<bool> {
 	virtual void drawValue();
 	void drawPixelsForOled();
 };
+
+class ToggleBool : public Toggle {
+public:
+	using Toggle::Toggle;
+
+	ToggleBool(l10n::String newName, l10n::String title, bool& newToggle) : Toggle(newName, title), t(newToggle) {}
```
m-m-adams, line 22: This should have a comment on the bool pointers lifetime - in this case it must last at least as long as the menu exists, and so can't be valid for song specific toggles

seangoodvibes: Added thanks!

Issue #872, opened 2023-12-23
Quantized stutter feature causes midi CC0 and midi CC Bend to be quantized to the same scale


entzmingerc:
### Please describe the problem:

Turn the quantized stutter feature on, create a midi clip, then change a gold knob to midi CC0. If you turn the knob, the value doesn't smoothly change values but instead displays: (4th, 8th, 16th, 32nd, 64th). Similarly, if you change the gold knob to "Bend" by turning select 4 ticks to the left, this parameter also is quantized to the same set of values. I did not test every CC value possible, but a random selection didn't appear to show any other CCs were affected. 

### What is the expected behavior?

CC0 and Bend should both smoothly change from 1 to 127 like the other CCs, similar to how the parameters perform when quantized stutter feature is off. 

### Is there a relevant Pull request?

no, but a search of quantized stutter brings up other PRs related to the feature

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Release 1.0/Amadeus

### What is the firmware name:

deluge-c1_0_0.bin

### If possible provide the steps to reproduce the issue and upload additional media:

Quantized stutter on
midi clip
press and hold top mod encoder and turn Select knob to select CC0
turn the mod encoder and observe the quantization of CC0
similarly, set the mod encoder to Bend which is 4 ticks of Select knob to the left from CC0
turn the mod encoder and observe the quantization of Bend

soymonitus:
I just checked and could not replicate....hmmm but maybe because I am in the community build. Seems already fixed there

PR #874, opened 2023-12-23
Fix quantized stutter LEDs to only apply to sounds params and not midi params

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/872

### DIFF
--- a/src/deluge/gui/views/view.cpp
+++ b/src/deluge/gui/views/view.cpp
@@ -1121,12 +1121,8 @@ void View::setKnobIndicatorLevel(uint8_t whichModEncoder) {
 	}
 
 	// Quantized Stutter FX
-	if (modelStackWithParam->paramId == Param::Unpatched::STUTTER_RATE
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/view.cpp
```diff
@@ -1122,6 +1122,8 @@ void View::setKnobIndicatorLevel(uint8_t whichModEncoder) {
 
 	// Quantized Stutter FX
 	if (modelStackWithParam->paramId == Param::Unpatched::STUTTER_RATE
```
seangoodvibes, line 1124: I would suggest simplifying this as follows:

```
Param::Kind kind = modelStackWithParam->paramCollection->getParamKind();

if (isParamQuantizedStutter(kind, modelStackWithParam->paramId) && !isUIModeActive(UI_MODE_STUTTERING)) {
```

seangoodvibes (changes requested):

seangoodvibes (approved):
Looks good!

PR #877, opened 2023-12-24
Enhancement / Grid View ~ Add pink mode to grid view for performance view

seangoodvibes:
Pink Mode: Grid View Performance View Launcher

- From grid view, a new pink mode has been created in the sidebar in the very bottom right hand corner.
- When you press the pink pad, performance view will appear. 
  - if you tap the pink pad (momentary press), it will switch from grid view to performance view
  - if you hold the pink pad (long press), it will switch to performance view only while you are holding the pad and when you let go it will go back to grid view
- If you have entered performance view from grid mode, you can exit back out by pressing the pink pad or pressing one of the other mode pads (e.g. green, blue). Optionally you can still exit the old way as well by pressing the keyboard button.
- Note: When in grid mode, the keyboard button will no longer enter into performance view. Only the pink pad will get you into performance view.

To do:

- [x] Disable keyboard entry to performance view while in grid mode
- [x] Add short/long press entry to performance view (short = enter performance view, long = peak performance view)
- [x] Render last column as per grid view (e.g. show all the modes) - factor out the renderSideBar code in grid view so performance view can call it
- [x] Add bool in performance view for grid view to write to to indicate that you've entered performance view from grid mode

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -3332,11 +3332,6 @@ ActionResult SessionView::gridHandlePads(int32_t x, int32_t y, int32_t on) {
 			modeHandleResult = gridHandlePadsLaunch(x, y, on, clip);
 			break;
 		}
-		case SessionGridModePerformanceView: {
-			changeRootUI(&performanceSessionView);
-			uiNeedsRendering(&performanceSessionView);
-			return ActionResult::DEALT_WITH;
-		}
 		}
 
 		if (modeHandleResult == ActionResult::DEALT_WITH) {
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -2835,17 +2835,17 @@ void SessionView::gridRenderActionModes(int32_t y, RGB image[][kDisplayWidth + k
 	bool modeActive = false;
 	RGB modeColour = colours::black;
 	switch (y) {
-	case 7: {
+	case GridMode::GREEN: {
 		modeActive = (gridModeActive == SessionGridModeLaunch);
 		modeColour = colours::green; // Green
 		break;
 	}
-	case 6: {
+	case GridMode::BLUE: {
 		modeActive = (gridModeActive == SessionGridModeEdit);
 		modeColour = colours::blue; // Blue
 		break;
 	}
-	case 0: {
+	case GridMode::PINK: {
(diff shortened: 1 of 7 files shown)


seangoodvibes:
@PaulFreund 

I was chatting with @sichtbeton on Discord about the button combo to enter performance view. He was suggesting not introducing an "alternate" mechanism for entering performance view. In arranger view we also use the Keyboard button to enter Performance View so it would be a bit confusing I think to use Keyboard in Arranger and Regular Song View but not in Grid View.

Can we have a discussion about it?

What if we did a short press to get into performance view with the keyboard button and a long press with the keyboard button to pull up the keyboard?

PaulFreund:
For completeness, we had this discussion on discord and didn't find a solution that solves all intended interaction patterns. If we want to have a split keyboard in grid the keyboard button can not be shared there (it is also needed as modifier for selection and settings). An alternative would be the triplets button for performance view but that would not work in clip view.

PaulFreund:
Proposal for Release 1.1. Remove Keyboard button in Grid view and replace it with Pink button so users don't get the expectation that keyboard opens performance mode in Grid view.

src/deluge/gui/views/session_view.cpp
```diff
@@ -2820,35 +2821,44 @@ bool SessionView::gridRenderSidebar(uint32_t whichRows, RGB image[][kDisplayWidt
...
 
+void SessionView::gridRenderActionModes(int32_t y, RGB image[][kDisplayWidth + kSideBarWidth],
+                                        uint8_t occupancyMask[][kDisplayWidth + kSideBarWidth]) {
+	// Action modes column
+	uint32_t actionModeColumnIndex = kDisplayWidth + 1;
+	bool modeExists = true;
+	bool modeActive = false;
+	RGB modeColour = colours::black;
+	switch (y) {
+	case 7: {
```
m-m-adams, line 2838: use an enum please, switching on raw ints is hard to understand

src/deluge/gui/views/session_view.cpp
```diff
@@ -3286,6 +3296,13 @@ ActionResult SessionView::gridHandlePads(int32_t x, int32_t y, int32_t on) {
 				gridModeActive = SessionGridModeEdit;
 				break;
 			}
+			case 0: {
```
m-m-adams, line 3299: same here. I know this isn't your code but it needs to be cleaned up

m-m-adams (approved):

PR #883, opened 2023-12-24
Bugfix / Invalid Param Collection in getModelStackWithParam for Automation/Performance View

seangoodvibes:
Expanding on the fix that was applied to automation view in #841 

I have factored out the code logic for getting modelStackWithAutoParam to two functions in the ModelStackWithThreeMainThings class:

1) getUnpatchedAutoParamFromId
2) getPatchedAutoParamFromId

This reduces the foot print of the code for getting the modelStackWithAutoParam for both Automation View and Performance View


bobtwinkles:
If we're applying the same fix in 2 places, could this be refactored to a function on the appropriate `ModelStack`subclass?

seangoodvibes:
> If we're applying the same fix in 2 places, could this be refactored to a function on the appropriate `ModelStack`subclass?

@bobtwinkles the reason for the bug fix is due to my improper usage of the functions to get the model stack with param.

the automation view fix dealt with getting the model stack with param for instrument clip params.

this fix for performance view deals with getting the model stack with param for the song params.

bobtwinkles:
Right, my suggestion is is adding a function like `addParamCollectionAndId` but taking a `ParamCollectionSummary` and `paramID` rather than duplicating this chunk of logic.

seangoodvibes:
> Right, my suggestion is is adding a function like `addParamCollectionAndId` but taking a `ParamCollectionSummary` and `paramID` rather than duplicating this chunk of logic.

Ah good point! I didn't realize addParamCollectionAndId was a function.

```
		ParamCollectionSummary* summary = modelStack->paramManager->getUnpatchedParamSetSummary();
		if (summary) {
			ModelStackWithParamId* modelStackWithParamId = modelStack->addParamCollectionAndId(summary->paramCollection, summary, paramID);
			if (modelStackWithParamId) {
				modelStackWithParam = summary->paramCollection->getAutoParamFromId(modelStackWithParamId, true);
			}
		}
```

Would this be sufficient?

bobtwinkles:
Sorta, but I think that last `if` statement can also be folded in to a function. Something like `ParamCollectionSummary::getAutoParamFromId` which internally dispatches to its `ParamCollection` after doing `addParamCollectionAndId` to a provided `ModelStack`. Then everywhere this chain of `if` statements is currently used could become a single line, and that API could be used instead so if the logic is needed in the future it's easy to reach for a function that does the right thing automatically.

seangoodvibes:
> Sorta, but I think that last `if` statement can also be folded in to a function. Something like `ParamCollectionSummary::getAutoParamFromId` which internally dispatches to its `ParamCollection` after doing `addParamCollectionAndId` to a provided `ModelStack`. Then everywhere this chain of `if` statements is currently used could become a single line, and that API could be used instead so if the logic is needed in the future it's easy to reach for a function that does the right thing automatically.

Ah yes

What about two functions that are a subset of ModelStackWithThreeMainThings?

`ModelStackWithThreeMainThings::getUnpatchedAutoParamFromId`
`ModelStackWithThreeMainThings::getPatchedAutoParamFromId`

Then within each you would have:

```
			ParamCollectionSummary* summary = nullptr;

			summary = modelStackWithThreeMainThings->paramManager->getUnpatchedParamSetSummary();

			if (summary) {
				ModelStackWithParamId* modelStackWithParamId =
				    modelStackWithThreeMainThings->addParamCollectionAndId(summary->paramCollection, summary, paramID);
				if (modelStackWithParamId) {
					modelStackWithParam = summary->paramCollection->getAutoParamFromId(modelStackWithParamId, true);
				}
			}
		}
```

```
			ParamCollectionSummary* summary = nullptr;

			summary = modelStackWithThreeMainThings->paramManager->getPatchedParamSetSummary();

			if (summary) {
				ModelStackWithParamId* modelStackWithParamId =
				    modelStackWithThreeMainThings->addParamCollectionAndId(summary->paramCollection, summary, paramID);
				if (modelStackWithParamId) {
					modelStackWithParam = summary->paramCollection->getAutoParamFromId(modelStackWithParamId, true);
				}
			}
		}
```

bobtwinkles:
That looks like it should work :smiley_cat:

seangoodvibes:
> That looks like it should work 😺

Made the change :)

PR #885, opened 2023-12-25
Fix quantized stutter LEDs to only apply to sounds params and not midi params

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/872
But this time using a code that is protected against NULL pointer exception with the paramCollection being checked before

### DIFF
--- a/src/deluge/gui/views/view.cpp
+++ b/src/deluge/gui/views/view.cpp
@@ -1122,6 +1122,8 @@ void View::setKnobIndicatorLevel(uint8_t whichModEncoder) {
 
 	// Quantized Stutter FX
 	if (modelStackWithParam->paramId == Param::Unpatched::STUTTER_RATE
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/view.cpp
```diff
@@ -1122,6 +1122,8 @@ void View::setKnobIndicatorLevel(uint8_t whichModEncoder) {
 
 	// Quantized Stutter FX
 	if (modelStackWithParam->paramId == Param::Unpatched::STUTTER_RATE
```
m-m-adams, line 1124: Redundant - have the next checks do modelStackWithParam->isParam(unpatched sound/global, Param::Unpatched::STUTTER_RATE) instead

soymonitus:
Now similar to how it was before, but using Sean suggestion, checking paramCollection for nullity

seangoodvibes (approved):

Issue #875, opened 2023-12-23
Arpeggiator should not be an option of the Kit/Affect Entire FX menu


soymonitus:
### Please describe the problem:

There is an option to change Arp settings from the new added Affect-Entire-Kit menu. When changing mode Deluge freezes. Besides, why arp settings for affect entire? Shouldn't the menu be completely removed? Also, maybe even Stutter should not be there as an option because it is available through the knobs

### What is the expected behavior?

There is no arpegiator menu in Affect Entire Kit menu

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

soymonitus:
@seangoodvibes You developed this menu I think so asking you if there is a reason to have arp settings available

seangoodvibes:
> @seangoodvibes You developed this menu I think so asking you if there is a reason to have arp settings available

I added it because Arp Gate is available as a Kit Affect Entire param, but I'm not sure if it does anything. If it doesn't it can be removed.

Probably same for stutter rate.

soymonitus:
Why would arp gate affect a bunch of kit rows which could even not have arp ON? It doesn't make much sense to me, even if it works somehow because the code paths allow it, it is not intuitive at all and should be removed

PR #886, opened 2023-12-25
Remove Some Params from Global FX Menu's and from Automation View

seangoodvibes:
Addressing issue #875 reported by @soymonitus 

- Removed Stutter Rate from menu's (Song, Performance View, Kit Affect Entire)
- Removed Arp and Portamento from Kit Affect Entire menu
- Removed Arp and Portamento from Kit Affect Entire in Automation View
- Removed Portamento from Kit Rows in Automation View

### DIFF
--- a/src/deluge/gui/views/automation_instrument_clip_view.cpp
+++ b/src/deluge/gui/views/automation_instrument_clip_view.cpp
@@ -2746,7 +2746,7 @@ void AutomationInstrumentClipView::selectEncoderAction(int8_t offset) {
 						offset += 1;
 					}
 					idx = getNextSelectedParamArrayPosition(offset, clip->lastSelectedParamArrayPosition,
-															kNumNonKitAffectEntireParamsForAutomation);
+					                                        kNumNonKitAffectEntireParamsForAutomation);
 				}
 			}
 			auto [kind, id] = nonKitAffectEntireParamsForAutomation[idx];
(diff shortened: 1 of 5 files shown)


src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2785,6 +2798,17 @@ void AutomationInstrumentClipView::selectEncoderAction(int8_t offset) {
...
+			//don't make portamento available for automation in kit rows
+			if ((instrument->type == InstrumentType::KIT)
+			    && (clip->lastSelectedParamID == Param::Unpatched::Sound::PORTAMENTO)) {
+				if (offset < 0) {
+					offset -= 1;
+				}
+				else if (offset > 0) {
+					offset += 1;
+				}
+				goto obtainNextParam;
```
m-m-adams, line 2810: can you make this a proper loop?

seangoodvibes: > can you make this a proper loop?

Check now thanks! :)

src/deluge/gui/views/automation_instrument_clip_view.cpp
```diff
@@ -2710,81 +2722,69 @@ void AutomationInstrumentClipView::selectEncoderAction(int8_t offset) {
...
+						offset += 1;
+					}
+				}
+				else {
+					clip->lastSelectedParamID = id;
+					clip->lastSelectedParamKind = kind;
+					clip->lastSelectedParamArrayPosition = idx;
+					break;
+				}
+			} while (true);
```
m-m-adams, line 2787: A while true loop with a break is worse than a goto - you need to refactor so that the break condition is directly used as the loop condition

It might be easier to change the block above (getting the next param) into a separate function that can be called a second time when required

m-m-adams (changes requested):

src/deluge/gui/views/automation_instrument_clip_view.cpp
seangoodvibes: I refactored into a separate function as suggested. Works nicely thanks!

m-m-adams (approved):

PR #888, opened 2023-12-26
Enhancement / Mod Encoders ~ Add Mod Button Gold Knob Contextual Popups

seangoodvibes:
When you press on the gold knob mod buttons to switch what the gold knobs control, it now displays a pop-up to tell you what the current settings are for the current Gold Knob parameters selected. 

For example, if you press the LPF/HPF/EQ button it will tell you if you have currently selected LPF, HPF, or EQ. If you have selected LPF or HPF, it will also tell you the mode - e.g. LPF: 24DB

On 7SEG the pop-up is split into two actions (press and release): e.g. for LPF, when you press down on the button it will display LPF, and when you let go of the button it will display 24DB.

- [x] Get it working for OLED
- [x] Get it working for 7SEG


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -2327,77 +2326,77 @@ void ModControllableAudio::switchDelaySyncType() {
...
+	display->displayPopup(getDelaySyncTypeDisplayName());
+}
 
-	char const* displayText;
+char const* ModControllableAudio::getDelaySyncTypeDisplayName() {
 	switch (delay.syncType) {
 	case SYNC_TYPE_TRIPLET:
-		displayText = "Triplet";
-		break;
+		return "Triplet";
```
m-m-adams, line 2335: not safe - these will be on the stack and not valid after the return. Have them take a char* as input and write the result into it, so the caller manages the memory

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -2327,77 +2326,77 @@ void ModControllableAudio::switchDelaySyncType() {
...
+}
 
+char const* ModControllableAudio::getDelaySyncLevelDisplayName() {
+	// Note: SYNC_LEVEL_NONE (value 0) can't be selected
+	delay.syncLevel = (SyncLevel)(delay.syncLevel % SyncLevel::SYNC_LEVEL_256TH); // cycle from 1 to 9 (omit 0)
 	StringBuf buffer{shortStringBuffer, kShortStringBufferSize};
-	currentSong->getNoteLengthName(buffer, (uint32_t)3 << (SYNC_LEVEL_256TH - delay.syncLevel), "");
-	display->displayPopup(buffer.data());
+	currentSong->getNoteLengthName(buffer, (uint32_t)3 << (SYNC_LEVEL_256TH - delay.syncLevel));
+	return buffer.data();
```
m-m-adams, line 2354: Not safe, returning pointer to stack

m-m-adams (changes requested):

m-m-adams (approved):

PR #892, opened 2023-12-27
Refactor/seperate midi matching from receiving in melodic instruments

m-m-adams:
Breaks out matching and receiving midi to allow midi follow to integrate more easily

Renames drum received functions from offerXtodrum to receivedXForDrum to make it more clear that it must already have been matched


seangoodvibes (approved):
Looks good thanks again for doing all this :)!

seangoodvibes (approved):

PR #898, opened 2023-12-28
Rename compressor to sidechain and update imports

m-m-adams:
To avoid future confusion - otherwise each clip will have a "compressor" that is a triggered envelope along with a compressor that is a compressor

Sidechain envelope is moved to sidechain folder and named sidechain

Master compressor is moved to compressor folder and named RMSFeedbackCompressor in case other types are added in future


seangoodvibes (approved):
Straightforward changes, makes sense :)!

PR #901, opened 2023-12-30
Add velocity on sidebar pads in isomorphic keyboard view

robmccoll:
Inspired by the velocity drums keyboard, this change adds a way to adjust note velocity in the isometric layout. 

- Sidebar pads set velocity from 7 (bottom left) to 127 (top right)
- Moving right increases by 8, moving up increases by 16
- The selected velocity is used for all new notes until another velocity key is pressed
- Selected velocity is flashed as a popup on selection
- Pads range in color from dim red to bright red
- Selected pad has other channels set to 0xFF creating a range of cyan to white

Thoughts for improvements:

- Make min / max adjustable and use a linear range between current min / max (press top right or bottom left and use the encoder?)
- Allow changing what sidebar pads set / modulate (maybe a keyboard shortcut of holding any sidebar pad and then pressing another pad)

### DIFF
--- a/src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
+++ b/src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
@@ -32,9 +32,17 @@ void KeyboardLayoutIsomorphic::evaluatePads(PressedPad presses[kMaxNumKeyboardPa
 	currentNotesState = NotesState{}; // Erase active notes
 
 	for (int32_t idxPress = 0; idxPress < kMaxNumKeyboardPadPresses; ++idxPress) {
-		if (presses[idxPress].active && presses[idxPress].x < kDisplayWidth) {
-			currentNotesState.enableNote(noteFromCoords(presses[idxPress].x, presses[idxPress].y),
-			                             getDefaultVelocity());
+		if (presses[idxPress].active) {
+			// in note columns
+			if (presses[idxPress].x < kDisplayWidth) {
+				currentNotesState.enableNote(noteFromCoords(presses[idxPress].x, presses[idxPress].y), velocity);
+			}
+			else { // in velocity columns (audition pads)
+				velocity = 7 + (presses[idxPress].x - kDisplayWidth) * 8 + presses[idxPress].y * 16;
+				char velocityStr[4] = {0};
+				intToString(velocity, velocityStr, 1);
+				display->displayPopup(velocityStr);
+			}
 		}
 	}
 }
@@ -137,6 +145,22 @@ void KeyboardLayoutIsomorphic::renderPads(uint8_t image[][kDisplayWidth + kSideB
 			++noteCode;
 			++normalizedPadOffset;
 			noteWithinOctave = (noteWithinOctave + 1) % kOctaveSize;
+		}
+	}
+}
+
+void KeyboardLayoutIsomorphic::renderSidebarPads(uint8_t image[][kDisplayWidth + kSideBarWidth][3]) {
+	// Iterate over velocity pads in sidebar
+	uint8_t brightness = 1;
+	uint8_t velocity_val = 7;
+	for (int32_t y = 0; y < kDisplayHeight; ++y) {
+		for (int32_t x = kDisplayWidth; x < kDisplayWidth + kSideBarWidth; ++x) {
+			uint8_t selection_brightness = (velocity >= velocity_val - 7 && velocity <= velocity_val) ? 0xf0 : 0;
+			image[y][x][0] = brightness + 0x04;
+			image[y][x][1] = selection_brightness;
+			image[y][x][2] = selection_brightness;
+			brightness += 10;
+			velocity_val += 8;
 		}
 	}
 }
--- a/src/deluge/gui/ui/keyboard/layout/isomorphic.h
+++ b/src/deluge/gui/ui/keyboard/layout/isomorphic.h
@@ -35,6 +35,7 @@ public:
 	void precalculate() override;
 
 	void renderPads(uint8_t image[][kDisplayWidth + kSideBarWidth][3]) override;
+	void renderSidebarPads(uint8_t image[][kDisplayWidth + kSideBarWidth][3]) override;
 
 	char const* name() override { return "Isomorphic"; }
 	bool supportsInstrument() override { return true; }
@@ -46,6 +47,7 @@ private:
 	}
 
 	uint8_t noteColours[kDisplayHeight * kMaxIsomorphicRowInterval + kDisplayWidth][3];
+	uint8_t velocity = 64;
 };
 
 }; // namespace deluge::gui::ui::keyboard::layout


m-m-adams:
Awesome thanks! We were thinking of using one bar for velocity and the other for mod wheel (y in the mod matrix), I haven't reviewed your code yet but would that be a feasible option?

robmccoll:
@m-m-adams That's much better! Can do - this commit changes to that format and makes the min and max values adjustable (with pads corresponding to a linear range between the current min and max).

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
```diff
@@ -20,27 +20,100 @@
...
+					velocityMaxHeld = true;
+				}
+			}
+			else if (pressed.x == MOD_COL) { // in mod columns (audition pads)
+				mod32 = modMin + pressed.y * modStep;
+				char modelStackMemory[MODEL_STACK_MAX_SIZE];
+				ModelStack* modelStack = setupModelStackWithSong(modelStackMemory, currentSong);
+				ModelStackWithTimelineCounter* modelStackWithTimelineCounter =
+				    modelStack->addTimelineCounter(currentSong->currentClip);
+				((MelodicInstrument*)currentInstrument())
```
robmccoll, line 75: Is this cast actually safe? I know isometric explicitly doesn't support Kit, but is there some way to verify that the current instrument is a melodic instrument?

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
```diff
@@ -20,27 +20,100 @@
...
+				else {
+					velocityMaxHeld = true;
+				}
+			}
+			else if (pressed.x == MOD_COL) { // in mod columns (audition pads)
+				mod32 = modMin + pressed.y * modStep;
+				char modelStackMemory[MODEL_STACK_MAX_SIZE];
+				ModelStack* modelStack = setupModelStackWithSong(modelStackMemory, currentSong);
+				ModelStackWithTimelineCounter* modelStackWithTimelineCounter =
+				    modelStack->addTimelineCounter(currentSong->currentClip);
```
robmccoll, line 74: Is there already model stack somewhere in the class hierarchy above here? I couldn't find one.

PaulFreund:
Hey there :) Great initative! I'm sure you already saw https://github.com/SynthstromAudible/DelugeFirmware/discussions/255. Regarding implementation it would be great if most stuff is higher up than in isomorphic (for example in the layout interface 
 / keyboard_screen) because In-Key will benefit the same as Isomorphic and in keyboard_screen. If I remember correctly the default velocity is a system setting so it would be good to use that. From usability I could imagine it would be cool to have a short press to constantly set velocity and long press combined with note input it switches back to the default velocity but that is just an idea.

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
m-m-adams: It's not great since that could change in the future - the overarching class for all clips is "output" - you can check the type by accessing the type field here https://github.com/SynthstromAudible/DelugeFirmware/blob/f7aaa2c1f13e397982d5190a84d62c1479bba211/src/deluge/model/output.h#L53 

Another solution is just making an empty processParamFromInputMIDIChannel part of the instrument class, that would let it be implemented in future if mod wheel support is added to kits

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
```diff
@@ -20,35 +20,100 @@
...
 #include "model/settings/runtime_feature_settings.h"
 #include "util/functions.h"
 #include <limits>
 
+#define VEL_COL kDisplayWidth
+#define MOD_COL (kDisplayWidth + 1)
+
 namespace deluge::gui::ui::keyboard::layout {
 
+static inline void popupUint8(deluge::hid::Display* display, uint8_t val) {
```
m-m-adams, line 34: This function is useful enough that I think it should live in the display class for convenience

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
m-m-adams: I don't think so, this is probably the best way to get it

soymonitus:
Please, implement it as an "addon" so we later can decide from a menu, which "action" we want for the left column of the keyboard view and which one for the right column. I would love this to be tweakable from the main menu so when more things are implemented (pitch bend, mod wheel, beat repeat), I can, as a user, decide to put Velocity on the right column for example, and Beat Repeat on the left column, and then I have my keyboard views the way I like

m-m-adams:
That's easy enough to do in a follow on PR. I think once this is done on the KeyBoardLayout class instead of specifically isomorphic it's good to merge

src/deluge/gui/ui/keyboard/layout/isomorphic.cpp
robmccoll: Added processParamFromInputMIDIChannel to instrument and kit

robmccoll:
[ ] initialize to instrument default velocity
[ ] short press to change, long press => return to default on release
[ ] support for alternative control types

robmccoll:
> Hey there :) Great initative! I'm sure you already saw #255. Regarding implementation it would be great if most stuff is higher up than in isomorphic (for example in the layout interface / keyboard_screen) because In-Key will benefit the same as Isomorphic and in keyboard_screen. If I remember correctly the default velocity is a system setting so it would be good to use that. From usability I could imagine it would be cool to have a short press to constantly set velocity and long press combined with note input it switches back to the default velocity but that is just an idea.

I actually missed #255 somehow 🤦 but I'll try to bring this more in line with the ideas there. I moved the code to an intermediate class ColumnControlsKeyboard instead of the KeyboardLayout (assuming other layouts may be developed that want to use the audition pads for something else) and added support to the In-Key keyboard. Modified the pad press logic to track held keys, but short press / long press is still TODO

robmccoll:
- Fixed some fuzzy math problems, un/signed clamping problems
- Use instrument default velocity
- Long presses will change a value temporarily and return to the set value on release
- Short presses change the set value
- Pressing and holding the top pad in a column lets you use the horizontal scroll to change the function of that column
- Functions now include velocity (VEL) mod (MOD) chord mode (CHRD) and beat repeat - unfinished (BEAT)
- Chord mode allows you to toggle back to no chord by short pressing the set value. Chords include 5th, Sus2, Minor, Major, Sus4, Minor7, Dom7, Major7.
- Beat repeat will need some thought to make sure the timing is correct and that the implementation isn't crazy inefficient. If anyone has suggestions, I'd appreciate hearing them. I looked at the arpeggiator and I'm not sure this can / should be implemented the same way.

soymonitus:
I gave this a test. Super cool feature. After testing it, i found the UI feels a bit unresponsive and buggy, when you press the pads like crazy, at some point if you press one pad from one column it also changes the value of the other colum, or for example pressing the first pad makes the 4th pad go white, like if there were some references being messed up

m-m-adams:
Hey sorry - I totally missed that this was ready for review again. I'll take a good look through it this weekend, but giving it a try right now and I like it!

m-m-adams:
Resolved conflicts and updated to head. Using your colour functions still but now using the new RGB class

m-m-adams (commented):
Good to go codewise. For the responsiveness I think they should be rendering the white square as soon as the touch happens, and just jump back to the stuck one afterwards

I pushed changes that I think help the responsiveness. If you're good with this I'll merge it!

m-m-adams (approved):

m-m-adams:
Before merging can you actually add an entry to community_features.md and the changelog?

robmccoll:
Fantastic - thanks for taking the time to go through it and update everything @m-m-adams! And thanks @soymonitus for the testing and feedback! I'm able to reproduce some of the bugs that @soymonitus described, so I'm going to try to fix those too before this gets merged.

- [ ] Fix transfer of chords in CMEM
- [ ] Fix column interaction bug
- [ ] Do another pass to check for immediate UI responsiveness to pad touches
- [ ] Allow for multiple simultaneous chords in CHRD and CMEM
- [ ] Update community_features.md

soymonitus:
https://github.com/SynthstromAudible/DelugeFirmware/assets/6472762/9e5b4eb0-b148-4a90-9159-516555b4ac07

I have tested you latest commit from your branch. Something funky is still going on with stuck velocity which cannot be changed, only under certain random taps

soymonitus:
Same happens with modwheel. I seem to trigger the bug (both for velocity and for modwheel) when I slide my finger over the column, the indexes then get crazy. There is a bug over there

https://github.com/SynthstromAudible/DelugeFirmware/assets/6472762/7c961b3e-9e6b-420e-8a9f-ee13690a3fac

soymonitus:
Another modwheel bug is that it only affects the notes that are already ON, but if you change the modwheel value before pressing a note and then press a note, it doesn't take the currently set modwheel value

https://github.com/SynthstromAudible/DelugeFirmware/assets/6472762/3785fe29-b298-46e5-b2d3-f1a3b417bc37

Before merging I think all these bugs need to be addressed. The feature is not usable right now. Would you move it to Draft so it doesn't get accidentally merged?

soymonitus:
@m-m-adams Don't you think that modwheel and velocity should set the value to whatever you press without mattering if it is short or long press? It is super weird and non usable that the values jump back to previous value depending on the amount of time you press the pad. Why? It would make this much more usable if no matter the type of click, just in the button press (not release) you set the new value but it doesn't jump back when released.
Regarding default velocity, I imagine this has been programmed so that the one set in settings would be the one used when entering new notes in the sequencer manually, but with the keyboard open, it should take the value set in the slidebar.

soymonitus:
The UI is inconsistent. When holding a pad in modwheel or velocity and moving the up/down encoder, you edit the value of that pad. On the other hand, when you hold a scale pad and turn the up/down encoder it changes the mode of the column to another, i expect the scale assigned to that pad to cycle till I find the scale I want (now that we have 16 scales and all cannot fit in the 8 pads) so it is saved to that pad.

soymonitus:
Testing Scales: Is the double click need as it was designed to change value? If not, then there is a bug there, if so, why not just changing scales it with a single tap? and for temporary change hold it (long press) and let it go to go back to previous value?

https://github.com/SynthstromAudible/DelugeFirmware/assets/6472762/44db7efe-ef31-43eb-b2ff-3ddab9e78cb3

soymonitus:
**CMEM BUG**: I was able doing multiple presses on the keyboard notes and on the CMEM pads to get stuck notes that then I couldn't stop anyway. Had to delete the clip so they stopped. There is something to check there.

soymonitus:
**GENERAL IMPROVEMENTS**: As an OLED user I would like to see the full name of the mode instead of the 4-letters abbreviation meant for 7seg users:
CHRD -> Chords
CMEM -> Chord memory
VEL -> Velocity
MODW -> Modwheel
SCAL -> Scales

robmccoll:
> @m-m-adams Don't you think that modwheel and velocity should set the value to whatever you press without mattering if it is short or long press? It is super weird and non usable that the values jump back to previous value depending on the amount of time you press the pad. Why? It would make this much more usable if no matter the type of click, just in the button press (not release) you set the new value but it doesn't jump back when released.

I can see the use for having things be temporary or the new normal, but I agree that the short-press / long-press seems a little unintuitive and the difference between short and long feels arbitrary. I'm fine with changing or scrapping it depending on what seems reasonable to everyone. I also thought it might be interesting to add a variable slew so that you could get smooth transitions between mod values.

> Regarding default velocity, I imagine this has been programmed so that the one set in settings would be the one used when entering new notes in the sequencer manually, but with the keyboard open, it should take the value set in the slidebar.

Yes - this is the intended behavior. The keyboard's velocity is initialized to the instrument's default velocity, but from that point onward, the keyboard's velocity (both the immediate value and the set-by-short-press value that it falls back to) do not affect the default velocity used by the sequencer and are not affected by it.

robmccoll:
@soymonitus  - thanks so much for all of this! I'm sure it's been frustrating and I'm sorry I didn't catch more of it myself, but I'm glad that you're willing to put in the effort. Keeping the quality of the community firmware up is massively important.

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
```diff
@@ -0,0 +1,461 @@
...
+		}
+		else {
+			vDisplay = velocity32;
+			velocity = (velocity32 + kHalfStep) >> kVelModShift;
+		}
+		break;
+	case MOD:
+		if (pad.active) {
+			modDisplay = modMin + pad.y * modStep;
+			getCurrentInstrument()->processParamFromInputMIDIChannel(CC_NUMBER_Y_AXIS, modDisplay,
```
m-m-adams, line 155: should be CC1 (mod wheel), this records it as MPE y to the currently sounding notes instead

soymonitus:
> I like the tap to latch/press to be momentary behaviour and it's consistent with performance mode. Found the mod wheel issue, it's the mpe y CC and not the mod wheel CC

but when you cannot operate the modwheel for example as a real touchstrip modwheel, which is what I would expect from it, I mean, sliding my finger over the column to set a value, and don't want the value to jump back arbitrarily based on how fast or slow I did the slide (because some pad detected a short press vs other that detected a long press).

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
m-m-adams: Actually looking again I think this is right as is, incoming CC1 is converted to this value. This should be fine

robmccoll:
Ok - this is probably ready for review again.
- Almost all of the major bugs were related to not checking and filtering out key presses in the dead state that I added 🤦 . Any press on another column pad or the keyboard could register a previous release a second time any time in the future.
- Chord memory is changed to capture on press only when a pad was empty to make playing easier. You must shift+pad to clear. This also fixed the stuck note bug.
- Scale modes now support all modes. The pads default to the first 7, but press and hold any pad and use the vertical encoder to change the mode associated with that pad.
- Popup labels have full length strings for OLED (I don't have an OLED, so please test that I didn't mess this up).
- Added text to the community features markdown about this.

Hopefully the formatting is ok now - I'm having problems with .so resolution using the clang-format in the prepackaged toolchain so I used my own.

m-m-adams:
Awesome I'll take another look! 


Regarding short/long press time this matches the rest of Deluge (performance view/sticky shift), so while there's some usability issues they should be addressed as a whole and it's not a problem specific to this PR

m-m-adams:
You’ve got a conflict, can you merge in community? I believe you just have to keep both

 I think it will make the checks run too since it looks like this pr has an old version of the CI config in it

m-m-adams (approved):
Awesome work, this is a ton of fun and the performance issues seem resolved. Thanks for the contribution!!

soymonitus:
This works much better than when i first tested. However, i still think that velocity and modwherl should work differently, by not having a different behavior for long and short press, only suirt press, so you can slide your finger to change values like a touch strip

lobit12:
Is it possible to record external midi input as a chord memory?

sichtbeton:
The normal mute pads allow you to exit out of menus immediately, even a couple of layers deep. could this functionality be added to the selectable sidebar functions or could the mute buttons be one of the selectable items?

PR #904, opened 2024-01-01
[DRAFT] Add tests for the voice code

sapphire-arches:
This is looking like it's going to pull in a whole bunch of nonsense along with the voice code, so that might get submitted as separate PRs first.


m-m-adams (commented):
Good to go when the error in the test build is fixed!

m-m-adams:
Maybe we should try making the core render functions constexpr and then testing at compile time?

sapphire-arches:
A more gradual approach is likely more suitable here.

PR #905, opened 2024-01-02
Remove call to newlibc sprintf due to malloc, replace with malloc-free variant

tastycode:
I rely on sprintf heavily for an upcoming PR, so removing it wasn't much of an option. I was able to find a dependency-free variant tailored for embedded systems. It does not (seem) to dynamically allocate memory. I placed it in src/lib because i didn't understand the purpose of `./lib`. 

```
zsh 10929  (git)-[20240102-sprintf-sadness]-% toolchain/darwin-arm64/arm-none-eabi-gcc/bin/arm-none-eabi-objdump -t /Users/tastycode/workspace/DelugeFirmware/build/Debug/deluge.elf | grep  malloc
201419b8 g     F .reset	0000000c __malloc_unlock
201419ac g     F .reset	0000000c __malloc_lock
2004800c g     O .bss	00000004 __malloc_free_list
20048150 g     O .bss	00000001 __lock___malloc_recursive_mutex
```


m-m-adams (commented):
At the moment this adds a bunch of bin size - could you look at replacing some of the existing formatting code with this lib and see if we can recover it? Things like the existing float/int/slot to string, probably some of the existing code around prints can use it too

src/deluge/util/functions.cpp
```diff
@@ -2690,28 +2690,32 @@ bool shouldAbortLoading() {
 
 // Must supply a char[5] buffer. Or char[30] for OLED.
 void getNoteLengthNameFromMagnitude(char* text, int32_t magnitude, bool clarifyPerColumn) {
+	char noteLengthName[64];
+	StringBuf noteLengthBuf(noteLengthName, 64);
```
m-m-adams, line 2694: I think this can allocate

tastycode: StringBuf's entire point of existence is to provide a nice interface for high level interaction with a manually managed string. 

The entire implementation is 20 lines. where would it be allocating?

```
class StringBuf {
	// Not templated to optimize binary size.
public:
	StringBuf(char* buf, size_t capacity) : capacity_(capacity), buf_(buf) {}

	void append(const char* str) { ::strncat(buf_, str, capacity_ - size() - 1); }
	void append(char c) { ::strncat(buf_, &c, 1); }
	void clear() { buf_[0] = 0; }

	// TODO: Validate buffer size. This will overflow
	void appendInt(int i, int minChars = 1) { intToString(i, buf_ + size(), minChars); }
	void appendHex(int i, int minChars = 1) { intToHex(i, buf_ + size(), minChars); }

	[[nodiscard]] char* data() { return buf_; }
	[[nodiscard]] const char* data() const { return buf_; }
	[[nodiscard]] const char* c_str() const { return buf_; }

	[[nodiscard]] std::size_t capacity() const { return capacity_; }
	[[nodiscard]] std::size_t size() const { return ::strlen(buf_); }

	[[nodiscard]] bool empty() const { return buf_[0] == 0; }

	bool operator==(const char* rhs) const { return strcmp(buf_, rhs) == 0; }
	bool operator==(StringBuf const& rhs) const { return strcmp(buf_, rhs.c_str()) == 0; }

private:
	size_t capacity_;
	char* buf_;
};
```

m-m-adams: Oh my bad, thought it was the std string buf

m-m-adams: Could you add a comment mentioning that it's totally different?

m-m-adams: Disregard, I'm gonna just rename that class later instead

src/definitions.h
```diff
@@ -70,3 +70,21 @@ extern void freezeWithError(char const* errmsg);
 #define PLACE_SDRAM_DATA __attribute__((__section__(".sdram_data")))
 
 // #define PLACE_SDRAM_TEXT __attribute__((__section__(".sdram_text"))) // Paul: I had problems with execution from SDRAM, maybe timing?
+#ifndef D_PRINTLN
+
+#if ENABLE_TEXT_OUTPUT
+
+#include "deluge/deluge.h"
+#include "lib/printf.h"
```
m-m-adams, line 78: Just move this to deluge.h, not needed here and it breaks inclusions of stdio

tastycode:
Closed in favor of https://github.com/SynthstromAudible/DelugeFirmware/pull/950

PR #907, opened 2024-01-03
Update gcc args and configuration, fix build warnings

m-m-adams:
Saves a few kb of binsize, more importantly removes some future footguns

1. Without -freestanding GCC expects syscalls
2. Replace rdimon with sys_stubs.c. Without rdimon libc expects syscalls provided by semihosting. This provides technically functional implementations for all syscalls except _sbrk, which is implemented via the GMA
3. Move always-inlined functions to headers, as is they were inlined post LTO in release but not debug. Stops the warnings
4. Consolidates the linker args in CMakeLists.txt instead of spread between that and scripts/cmake
5. Explicitly remove unreferenced lib exports from the linkerscript instead of solely relying on GC


src/resetprg.c
```diff
@@ -100,15 +100,15 @@ void _fini(void) {
...
+void emptySection(uint32_t* start, uint32_t* end) {
 	uint32_t* dst = start;
 	while (dst < end) {
 		*dst = 0;
 		++dst;
 	}
 }
 
-inline void relocateSDRAMSection(uint32_t* start, uint32_t* end) {
+void relocateSDRAMSection(uint32_t* start, uint32_t* end) {
```
bobtwinkles, line 111: Should make these `static` instead I think.

CMakeLists.txt
```diff
@@ -218,6 +224,12 @@ set(NMDUMP_FILE $<CONFIG>/deluge${EXECUTABLE_VERSION_SUFFIX}.nmdump)
 
 # add link options to generate memory map
 target_link_options(deluge PUBLIC
+
+    #use newlib nano
+    --specs=nano.specs 
+    #replace syscalls with not implemented stubs
```
bobtwinkles, line 230: Style nit: add spaces after the `#`s

src/sys_stubs.c
```diff
@@ -0,0 +1,64 @@
...
+
+void* delugeAlloc(unsigned int requiredSize, bool mayUseOnChipRam);
+
+void* _sbrk(int incr) {
+	static unsigned char* heap = NULL;
+	static unsigned char* end_heap = NULL;
+	unsigned char* prev_heap;
+
+	if (heap == NULL) {
+		heap = delugeAlloc(10000, false);
```
bobtwinkles, line 14: Can we do this with a linker section instead?

m-m-adams: Then it'll always be present, currently this is never called and only exists to do something better than corrupt memory in case of accidents

PR #911, opened 2024-01-04
Bugfix / MIDI Follow ~ Add control of Performance View Song params  

seangoodvibes:
Summary of changes:

- [x] Added ability for midi follow cc's to control song level param's while in performance view
- [x] Remove mod encoder popup when editing the same param as displayed on the screen
- [x] Update performance view display with midi follow cc value, don't display popup if editing same param as shown on display
- [x] Clean up the code by refactoring re-rendering code for automation view and performance view to separate functions

### DIFF
--- a/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
+++ b/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
@@ -1710,10 +1710,21 @@ bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice
 				//if you're in automation view and editing the same parameter that was just updated
 				//by a learned midi knob, then re-render the pads on the automation editor grid
 				if (getRootUI() == &automationInstrumentClipView) {
-					int32_t id = modelStackWithParam->paramId;
-					Param::Kind kind = modelStackWithParam->paramCollection->getParamKind();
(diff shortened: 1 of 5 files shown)


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1709,13 +1710,21 @@ bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice
...
+						//for the same clip active in automation view
+						int32_t id = modelStackWithParam->paramId;
+						Param::Kind kind = modelStackWithParam->paramCollection->getParamKind();
+						possiblyRefreshAutomationEditorGrid(clip, kind, id);
 					}
 				}
+				//placeholder: if support is added for midi learning to song params, then you will need
+				//to possibly refresh the performance view display (see midi follow code below for doing this)
+				//you will need to check that you're dealing with a song param and not a clip param otherwise
+				//you may incorrectly refresh the performance view display
```
m-m-adams, line 1727: nit: this implies that song params should get their own param::kind

m-m-adams, line 1727: nit: this implies they should have different param kinds

m-m-adams (approved):

PR #925, opened 2024-01-07
menus: clamp to avoid overflow on shift+scroll

### LINKED ISSUE #871: hard crashes with shift + turn select scrolling in small menus

Shift + turn select should increment parameter values by 5, or increment your index in a menu by 5 for fast scrolling. In menus with less than five items, it seems to crash. Crashes reproducible with the following parameters:

sound parameters: 
Delay Type (digital, analog)
FM Destination (carrier, mod1)
synth mode (subtractive, FM, ring)
sequencer direction (forward, reverse, ping pong)
interpolation (sync, linear)
pitch/speed (link, independent)
sample mode (cut, once, loop, stretch)
HPF type (HP_L, sv_n, sv_b)
Priority (low, med, high)
polyphony (poly, mono, legato, auto)

settings menu:
CV (out1, out2)
Gate > out1 (vtrig, strig)
midi > tovr (jump, pick, scal)
midi > device (din) this doesn't show the same crash error, but it just freezes
midi > device > MPE (in, out)
pads > shor (1.0, 3.0)
pads > keyb (qwer, a2er, qrt2)
prev (on, off, cond)
curs (slow, fast, off)
reco > moni (on, off, cond)

out of range error: 
default > swing > 1 to 99 range, when on 1 and shiftscrolling it doesn't change value, if you scroll to 2 then shiftscroll left it goes to -3 and goes even more negative. Similarly you can go far above 99.
default > tempo > does +5 / -5, and seems to have a lower limit 60 and upper limit 240, but with shift scroll I can scroll past this, I don't think we should be able to set the defaults, this value also goes negative so I'm not sure what happens if we save a project default tempo with -50. 

please no crash, thanks
1 item, don't crash
2 items, switch between both items for each select encoder tick, which is what happens if you add 5 with wrap around for 2 items
3 items, step 2 ahead
4 items, step 1 ahead
5 items, step 0 ahead
6 items, step 5 ahead

https://github.com/SynthstromAudible/DelugeFirmware/issues/664 looks similar to this to me

7SEG

Release 1.0/Amadeus

deluge-c1_0_0.bin

Go to the sound parameter or settings menus listed above.
Hold shift, then turn select left or right.
Many of the menus listed above should immediately crash, or should crash after 2-3 ticks of the select knob. 

For the default > tempo > 60 to 240, you can shiftscroll down to 60, turn select once to 61, then shiftscroll to the left to 56 and down to negative tempo values. Similarly, you can do this to scroll past 240. Same for the swing menu with the expected range 1 to 99. 

IMG_20231222_172122027

bobtwinkles:
this change is the result of auditing all menu items that override selectEncoderAction to ensure none of them can go out-of-bounds if abs(offset) > 1

Fixes #871

### DIFF
--- a/src/deluge/gui/menu_item/enumeration.cpp
+++ b/src/deluge/gui/menu_item/enumeration.cpp
@@ -14,6 +14,28 @@ void Enumeration::beginSession(MenuItem* navigatedBackwardFrom) {
 void Enumeration::selectEncoderAction(int32_t offset) {
 	this->setValue(this->getValue() + offset);
 	int32_t numOptions = size();
+	int32_t sign = (offset < 0) ? -1 : ((offset > 0) ? 1 : 0);
+
+	switch (numOptions) {
+	case 0:
+		[[fallthrough]];
+	case 1:
+		[[fallthrough]];
+	case 2:
+		offset = 1 * sign;
+		break;
+	case 3:
+		offset = std::min<int32_t>(offset * sign, 2) * sign;
+		break;
+	case 4:
+		offset = 1 * sign;
+		break;
+	case 5:
+		offset = std::min<int32_t>(offset * sign, 4) * sign;
+		break;
+	default:
+		break;
+	}
 
 	if (display->haveOLED()) {
 		if (this->getValue() >= numOptions) {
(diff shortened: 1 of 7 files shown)


src/deluge/gui/menu_item/enumeration.cpp
```diff
@@ -14,6 +14,28 @@ void Enumeration::beginSession(MenuItem* navigatedBackwardFrom) {
...
+	case 1:
+		[[fallthrough]];
+	case 2:
+		offset = 1 * sign;
+		break;
+	case 3:
+		offset = std::min<int32_t>(offset * sign, 2) * sign;
+		break;
+	case 4:
+		offset = 1 * sign;
```
m-m-adams, line 31: Should this be +- 3 to act like the others?

m-m-adams (approved):

PR #927, opened 2024-01-07
Bugfix / MIDI Follow ~ Update Automation MIDI Clip View for MIDI Follow CC's

seangoodvibes:
When in automation view and using MIDI Follow, matched CC's update their corresponding CC parameter values in a MIDI clip

This fixes a bug where the automation view editor grid was not updating to reflect the updated values.

It also removes the "can't control popup" message that would appear to say that the parameter that CC is mapped to can't be controlled. That popup is not relevant for MIDI clips. Also removed this pop-up for CV clips as it's not relevant there either.


src/deluge/io/midi/midi_follow.cpp
```diff
@@ -480,6 +480,14 @@ void MidiFollow::offerReceivedCCToMelodicInstrument(ModelStackWithTimelineCounte
 	MelodicInstrument* melodicInstrument = (MelodicInstrument*)clip->output;
 	melodicInstrument->receivedCC(modelStackWithTimelineCounter, fromDevice, match, channel, ccNumber, value,
 	                              doingMidiThru);
+
+	//if you're in automation midi clip view and editing the same CC that was just updated
+	//by a learned midi knob, then re-render the pads on the automation editor grid
+	if (clip->output->type == InstrumentType::MIDI_OUT && getRootUI() == &automationInstrumentClipView) {
+		if (((InstrumentClip*)clip)->lastSelectedParamID == ccNumber) {
+			uiNeedsRendering(&automationInstrumentClipView);
```
m-m-adams, line 488: I think you should look for a cleaner way to do this, this is just going to lead to whack a mole as ways to change a parameter are added. It probably needs to be in the instrument itself

seangoodvibes: > I think you should look for a cleaner way to do this, this is just going to lead to whack a mole as ways to change a parameter are added. It probably needs to be in the instrument itself

This makes sense.

I've moved it to melodic instrument now. Can you take a look if you think this implementation is better?

I don't have a midi controller with me atm so can't test if this works, but it should.

m-m-adams (approved):

PR #928, opened 2024-01-07
view: only check param kind when autoparam valid

bobtwinkles:
... in View::setKnobIndicatorLevel

The paramCollection and autoParam members are implcitly related -- the paramCollection is valid iff the autoParam is valid.

Fixes the crash found by triff on Discord

0x201199ef
0x20119b13
0x200fcdac
0x200759ad

on commit f7aaa2c1f13e397982d5190a84d62c1479bba211


m-m-adams:
There's a more core problem - modelStackWithAutoParam is larger than ModelStackWithThreeMainThings, but the mod_controllable default implementation casts between them and then sets autoparam to 0, which will clobber a memory location unless it's a modelStackWithAutoParam to start with

m-m-adams:
Satisfied after today's discussion about the extra reserved space in view.h

Sketchy but a future problem!

m-m-adams (approved):

PR #934, opened 2024-01-08
Feature/option to combine mpe to monophonic midi output

m-m-adams:
Convenience settings for MPE and poly aftertouch users

Adds two settings to midi outputs to combine poly expression to mono. The aftertouch option converts MPE and poly aftertouch to mono aftertouch (highest pressure wins), the MPE option converts y-axis to mod wheel (highest wins) and poly pitch bend to average mono pitch bend.

This makes using an MPE controller with midi follow much more convenient


src/deluge/model/instrument/non_audio_instrument.h
```diff
@@ -53,6 +57,8 @@ class NonAudioInstrument : public MelodicInstrument, public ModControllable {
 	ModControllable* toModControllable() { return this; }
 
 	int32_t channel = 0;
+	uint8_t cachedBendRanges
+	    [2]; // Cache these here just in case there's no ParamManager - because CVInstruments don't do backedUpParamManagers.
```
bobtwinkles, line 61: style nit: move comment above for wrapping

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -863,3 +933,55 @@ void MIDIInstrument::polyphonicExpressionEventPostArpeggiator(int32_t value32, i
...
+	}
+
+	// Are multiple notes sharing the same output member channel?
+	ArpeggiatorSettings* settings = getArpSettings();
+	if (settings == nullptr || settings->mode == ArpMode::OFF) { // Only if not arpeggiating...
+		int32_t numNotesFound = 0;
+		int32_t mpeValuesSum = 0; // We'll be summing 16-bit values into this 32-bit container, so no overflowing
+		int32_t mpeValuesMax = -ONE_Q31;
+		for (int32_t n = 0; n < arpeggiator.notes.getNumElements();
+		     n++) { // This traversal will include the original note, which will get counted up too
```
bobtwinkles, line 956: style nit: move comment down a line for wrapping

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -212,25 +220,39 @@ int32_t MIDIInstrument::getOutputMasterChannel() {
...
 		int32_t valueSmall = (newValue >> 18) + 8192;
 		midiEngine.sendPitchBend(masterChannel, valueSmall & 127, valueSmall >> 7, channel);
 		break;
 	}
 
-	case 1:
+	case 1: {
+		//mono y expression is limited to positive values, double it to match range
+		int32_t polyPart = (lastCombinedPolyExpression[whichExpressionDimension] >> 25) + 64;
+		int32_t monoPart = lastMonoExpression[whichExpressionDimension] >> 24;
```
bobtwinkles, line 243: Should we sum these in a higher precision fixed-point space (i.e. prior to the shifts) to get better rounding? I doubt it matters much in practice.

bobtwinkles (commented):
mostly stupid style nitpicks, functionally i don't see any obvious problems

src/deluge/model/instrument/midi_instrument.cpp
m-m-adams: They both fully max out the int32 range so they need to shift down first anyway, this way it's easier to handle the fact that one is bipolar and the other is unipolar

PR #953, opened 2024-01-11
[DRAFT] Script to error upon _malloc_r

tastycode:


m-m-adams:
No longer needed - with some other build changes I was able to remove the _sbrk stub, and any builds with malloc will now fail to link

PR #955, opened 2024-01-12
Unlimited Drum Randomizer

M0R172:
Please test!

Drum randomizer now does not have a limit on number of files to randomize, and is more efficient by pulling code outside of loops.

Also, SHIFT+(random pad) now randomizes ALL non-muted kit rows.

It uses "Reservoir Sampling with k=1" for randomly choosing from a previously unknown number of files in the directory without a second pass. This gets rid of the need to allocate a list of N filenames for choosing from N files.


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1282,93 +1282,112 @@ const uint32_t auditionPadActionUIModes[] = {UI_MODE_AUDITIONING,
...
-					if (result != FR_OK) {
-						display->displayError(ERROR_SD_CARD);
-						return ActionResult::DEALT_WITH;
-					}
-					while (true) {
-						result = f_readdir_get_filepointer(&staticDIR, &staticFNO,
-						                                   &thisFilePointer); /* Read a directory item */
-						if (result != FR_OK || staticFNO.fname[0] == 0) {
-							break; // Break on error or end of dir
+		bool randomize_all = false;
```
bobtwinkles, line 1296: nit: variables should use `camelCase` (applies to all variables introduced here, including `randomize_all`, `maxrow`, and `chosen_fn`)

m-m-adams:
Hey thanks for the PR! I'm slightly concerned about performance impacts if someone tries to use this while a song is playing, since the 25 item limit exists to prevent hiccups in audio rendering. 

While this PR is better designed and removes the need for the file name array, I think the worst case execution could slow down audio rendering. Some of our users seem to keep multiple hundreds of samples in the same folder so that case needs to be handled

Before merging I'd like sanity checks on the max number of files to read. One approach would be a max of say 50 files, and then cheating a bit by starting the next randomization reads from the current sample instead of the first file in the dir

alternatively you might be able to call the audio routine from here in the case of a long loop, but it's better to avoid that if possible since it makes audio execution more difficult to trace

Thanks!

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1282,93 +1282,112 @@ const uint32_t auditionPadActionUIModes[] = {UI_MODE_AUDITIONING,
...
+		{
+			maxrow = getCurrentInstrumentClip()->noteRows.getNumElements();
+			randomize_all = true;
+		}
+
+
+		String chosen_fn;
+		chosen_fn.set("Nothing to randomize.");
+
+		for (int32_t i = 0; i < maxrow; i++) {
```
bobtwinkles, line 1308: this loops should probably go from `firstRandomRow` to `lastRandomRow`, where those two variables are set to either `getCurrentInstrumentClip()->getNoteRowOnScreen(lastAuditionedYDisplay, currentSong)->y` (for the single random case, you'll need to check if the returned note row is null as well and bail in that case) or `firstRandomRow=0`/`lastRandomRow=getCurrentInstrumentClip()->noteRows.getNumElements()` for the randomize all case.

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1282,93 +1282,112 @@ const uint32_t auditionPadActionUIModes[] = {UI_MODE_AUDITIONING,
...
-					continue;
-				}
-				AudioFileHolder* afh = r->getAudioFileHolder();
-				if (afh == NULL) {
-					continue;
-				}
+		if (sdRoutineLock) {
+			return ActionResult::REMIND_ME_OUTSIDE_CARD_ROUTINE;
+		}
+		if (currentSong->currentClip->output->type != InstrumentType::KIT) {
```
bobtwinkles, line 1291: Can use `getCurrentInstrumentType()` here.

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1282,93 +1282,112 @@ const uint32_t auditionPadActionUIModes[] = {UI_MODE_AUDITIONING,
...
-						getCurrentInstrument()->beenEdited();
 					}
 				}
-			}
-		}
-		if (numRandomized > 0) {
-			display->displayPopup(deluge::l10n::get(deluge::l10n::String::STRING_FOR_RANDOMIZED));
-			return ActionResult::DEALT_WITH;
 		}
+		display->displayPopup(chosen_fn.get());
```
bobtwinkles, line 1389: I like the idea of showing the file name here, but have a few UI concerns:
- Does this work with filenames longer than what can be shown across the OLED screen? I'm not sure we handle wide popups like that gracefully
- This is mostly meaningless for the randomize all case

bobtwinkles (commented):
Thanks for the contribution! A few style nits, but this looks like a nice improvement overall.

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: a) Fair point, but the file name is currently already shown when pressing the audition pad, so if it works there...
And if it does not, the problem should be addressed in the popup function...

b) Agreed, but IMO still more informative than "Randomized". But if you prefer "Randomized all" or something else, I am fine with it.

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: OK. Does that have further advantages apart from avoiding the train wreck style of programming?
But I agree that the optical improvement would be already be worth the change...

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Sorry, guilty of not reading the styleguide.

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Single random:
OK on the null check.
But why is: getCurrentInstrumentClip()->getNoteRowOnScreen(lastAuditionedYDisplay, currentSong)->y
better than: getCurrentInstrumentClip()->getNoteRowOnScreen(i, currentSong)

All Random:
I think what you describe is exactly what is happening, but maybe I am overlooking something?

M0R172:
> Hey thanks for the PR! I'm slightly concerned about performance impacts if someone tries to use this while a song is playing, since the 25 item limit exists to prevent hiccups in audio rendering.
> While this PR is better designed and removes the need for the file name array, I think the worst case execution could slow down audio rendering. Some of our users seem to keep multiple hundreds of samples in the same folder so that case needs to be handled

I tried it with a few hundred sample files, and a number of tracks playing. I turned the volume of the drumset down to only hear the other tracks, and randomized the complete kit. Although it took about a second to randomize, I could not hear any dropouts of the other tracks in this time (TBH, I have no Idea how this can work, but it does). What happens if you have a few thousand files in each samples directory? Good question!

> Before merging I'd like sanity checks on the max number of files to read. One approach would be a max of say 50 files, and then cheating a bit by starting the next randomization reads from the current sample instead of the first file in the dir

This is not trivial... because if one were to use only the filename (that is currently stored with the sample), one would have to search the directory for that file entry next time round, which is exactly the operation making the whole thing slow. So one would have to remember the directory offset somewhere (where?) and then go to that offset directly using the low-level fs functions... I will look into it. 

> alternatively you might be able to call the audio routine from here in the case of a long loop, but it's better to avoid that if possible since it makes audio execution more difficult to trace

I agree that the code is confusing enough as it is...

m-m-adams:
Ya store the pointer to the SD location, not finding the name again

m-m-adams:
Disregard my performance comments - the sd dma complete interrupt calls the audio rendering so this shouldn't interfere

Sample streaming should be fine too with the calls to load enqueued clusters

M0R172:
Ok, that explains some things (including why the randomizing gets slower when tracks are playing).

I will look into your and boib's suggestions during the weekend, and mark the PR as draft until then.

Cheers,
M

src/deluge/gui/views/instrument_clip_view.cpp
bobtwinkles: > But why is: getCurrentInstrumentClip()->getNoteRowOnScreen(lastAuditionedYDisplay, currentSong)->y
better than: getCurrentInstrumentClip()->getNoteRowOnScreen(i, currentSong)

It avoids running the loop for more iterations than necessary. The idea is to compute the exact range of rows, rather than needlessly iterating over all visible rows. It also allows the `if (randomize_all){}` check in the loop to go away, since both the `randomize_all` and single row cases will treat `i` as the row index.

>   I think what you describe is exactly what is happening, but maybe I am overlooking something?

You're correct, I missed the line that updates `maxRows`.

src/deluge/gui/views/instrument_clip_view.cpp
bobtwinkles: Style is probably the primary motivator, but it also compiles to slightly smaller code since a function call is a single instruction (plus maybe one register transfer to set up the argument) versus several chained load instructions.

src/deluge/gui/views/instrument_clip_view.cpp
bobtwinkles: Let's use "randomized all" for the all case.

src/deluge/gui/views/instrument_clip_view.cpp
bobtwinkles: This won't fix variable names, but you can use `dbt format` to make the automatic format check happy as well

src/deluge/gui/views/instrument_clip_view.cpp
bobtwinkles: as a heads up, I just renamed the recommend function to `getCurrentOutputType` (and the enum is `OutputType` rather than `InstrumentType`)

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Ah, but

GetCurrentInstrumentClip()->getNoteRowOnScreen(lastAuditionedYDisplay, currentSong)->y

Will only give me the LAST auditioned row - the current code allows (as did the OG drum randomizer) holding several audition pads and randomizing all those rows at once, which is a functionality I would not like to drop.

With the current code

`for (int32_t i = 0; i < maxrow; i++) {
				if (randomize_all || auditionPadIsPressed[i])`

there is very little overhead for the non-pressed pads anyway. I do not see how this could be any more efficient while keeping the multiple press functionality.

bobtwinkles: ah I see, in that case the current logic is fine.

M0R172:
OK:

- I changed the variable names,
- changed the popups to discern between multiple pressed, all active, and single row randomized (in which case the new filename is displayed)
- Added a null check (which I am not sure is strictly necessary, but better safe than sorry, I guess)
- Fixed a logic bug: Now, when neither shift is pressed, nor an audition pad is held, we correctly fall through to the normal edit action so the note can be edited
- I tested it with large folders: randomizing from 250 sample is basically instant, 7200(!) samples can take a second, but still feels usable to me and no crashes or dropouts result (the ui seems to be blocked in this time, though)

From my standpoint this seems suitable for release. 

I would be willing to revisit this at a later stage if people discover bugs or lots of people think the performance is too bad 
(I do have an idea how to make it faster, but this would be low-level fat hacking that would require some time to write and test - it also would make the randomness a bit less random, which I would like to avoid)

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1302,69 +1332,71 @@ ActionResult InstrumentClipView::padAction(int32_t x, int32_t y, int32_t velocit
 				if (afh == NULL) {
 					continue;
 				}
-
-				static int32_t MaxFiles = 25;
-				String fnArray[MaxFiles];
-				char const* currentPathChars = afh->filePath.get();
-				if (currentPathChars == NULL) {
+				char const* path = afh->filePath.get();
+				if (path == NULL) {
```
bobtwinkles, line 1336: `path` will never be null, but it might be `&nothing`. Can probably skip this null check entirely, if it crashes something the blackbox will tell us pretty immediately.

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1302,69 +1332,71 @@ ActionResult InstrumentClipView::padAction(int32_t x, int32_t y, int32_t velocit
...
+				if (slashAddress == NULL) {
+					continue;
+				}
+
+				// Open directory of current audio file
+				*slashAddress = 0;
+				FRESULT result = f_opendir(&staticDIR, path);
+				if (result != FR_OK) {
+					display->displayError(ERROR_SD_CARD);
+					return ActionResult::DEALT_WITH;
```
bobtwinkles, line 1349: I think this will leave the `AudioFileHolder` in a bad state, since `slashAddress` writes directly to the underlying memory. Instead, we should copy the directory name in to `chosenFilename`.

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1302,69 +1332,71 @@ ActionResult InstrumentClipView::padAction(int32_t x, int32_t y, int32_t velocit
...
-						soundDrum->name.set(fn);
-						numRandomized++;
-						getCurrentInstrument()->beenEdited();
-					}
+				// Assign new audio file
+				if (fileCount) {
+					String filePath;
+					filePath.set(path);
+					filePath.concatenate("/");
+					filePath.concatenate(chosenFilename);
```
bobtwinkles, line 1369: Rather than allocating a new string, I think we can re-use the one at `afh->filePath`

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1302,69 +1332,71 @@ ActionResult InstrumentClipView::padAction(int32_t x, int32_t y, int32_t velocit
...
-							continue; // Ignore dot entry
-						}
-						audioFileManager.loadAnyEnqueuedClusters();
-						fnArray[numSamples].set(staticFNO.fname);
-						numSamples++;
-						if (numSamples >= MaxFiles) {
-							break;
-						}
+					if (random(fileCount++) == 0) { // Algorithm: Reservoir Sampling with k=1
+						strncpy(chosenFilename, staticFNO.fname, 256);
```
bobtwinkles, line 1360: Use `sizeof(chosenFilename)` instead of 256, so they can't get out of sync

bobtwinkles (commented):
A couple of other small tweaks and this is probably good to merge. Thanks again!

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Changed to check for &nothing. 

Leaving the other null check in as it is not expensive, not in the hot loop, and the whole function is only triggered by user interaction, i.e. will not run very often in the grand scheme of things.

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Done!

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Done!

src/deluge/gui/views/instrument_clip_view.cpp
M0R172: Done!

Oops, I assumed that `get()` retuned a copy!

I opted for just writing the slash back into `slashAddress` right after using the shortened path to open the dir. 
As paths can theoretically be up to 65k chars long, I want to avoid copying or making `chosenFilename` that big.

Admittedly, this leaves the "bad state" for the duration of the `f_opendir` call, but I think the afh thrying to reload the file at that exact instant is, if not plain impossible, *extremely* unlikely...

bobtwinkles:
LGTM now, nice work!

PR #957, opened 2024-01-12
Bugfix/sync levels

m-m-adams:
Fix for sync levels from #950 

@tastycode


src/deluge/util/functions.cpp
```diff
@@ -2690,8 +2690,8 @@ bool shouldAbortLoading() {
 
 // Must supply a char[5] buffer. Or char[30] for OLED.
 void getNoteLengthNameFromMagnitude(char* text, int32_t magnitude, bool clarifyPerColumn) {
-	char noteLengthName[64];
-	StringBuf noteLengthBuf(noteLengthName, 64);
+	//char noteLengthName[64];
```
bobtwinkles, line 2693: Delete this buffer since we don't need it?

PR #963, opened 2024-01-13
Enhancement / MIDI Defaults ~ Add Select Kit Row Default Menu

seangoodvibes:
Added new Select Kit Row menu to the Midi Defaults menu in `Settings > MIDI > Select Kit Row`

When this setting is enabled, midi notes received for learned kit row's will update the kit row selection in the kit clip.

This updates the mod controllable stack that the mod encoders control as well so that you can update the parameters for the drum being auditioned without needing to manually select the same drum as the drum corresponding to the learned note received.

This is useful for midi follow as well because the updated drum selection will not only allow you to use your midi follow controller to control the parameters for that drum, but it will also send updated midi feedback as the drum selection changes.

Refactored InstrumentClipView::SetSelectedDrum to improve it and enable its use with midi.

Improvements made:

1) it would set the selected drum even if it's the same drum that's currently selected. This happens when you audition a kit row for instance. So I added a check to make sure we're not re-selecting the same drum. In that instance the only thing that would happen is a re-rendering of the audition pad (so it lights up when you press it).

2) it would set the mod controllable when changing the selected drum, even if affect entire is enabled. This did not make sense because with affect entire enabled you cannot control the mod controllable stack of the selected drum. So there is no need to redraw mod led's or set the mod controllable. So I added a check to see if affect entire is enabled. Without check, this call to set the mod controllable would also unnecessarily send midi follow feedback when nothing has changed.

3) Only redraw clip if the drum being selected is in the currently opened kit clip (for drum selections received through midi, you may be changing the drum selection of a different clip than the current clip selected).

4) Added check to only set the mod controllable or redraw automation view when you actually changed the selected drum.

These changes should improve performance.

### DIFF
--- a/src/deluge/model/instrument/kit.cpp
+++ b/src/deluge/model/instrument/kit.cpp
@@ -25,6 +25,8 @@
 #include "io/debug/print.h"
 #include "io/midi/midi_device.h"
 #include "io/midi/midi_device_manager.h"
+#include "io/midi/midi_engine.h"
+#include "io/midi/midi_follow.h"
 #include "memory/general_memory_allocator.h"
 #include "model/clip/instrument_clip.h"
 #include "model/clip/instrument_clip_minder.h"
@@ -1018,6 +1020,9 @@ void Kit::receivedNoteForDrum(ModelStackWithTimelineCounter* modelStack, MIDIDev
                               bool* doingMidiThru, Drum* thisDrum) {
 	InstrumentClip* instrumentClip = (InstrumentClip*)modelStack->getTimelineCounterAllowNull(); // Yup it might be NULL
 
+	//do we need to update the selectedDrum?
+	possiblySetSelectedDrumAndRefreshUI(thisDrum);
+
 	bool recordingNoteOnEarly = false;
 
 	bool shouldRecordNoteOn =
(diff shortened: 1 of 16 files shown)


src/deluge/model/instrument/kit.cpp
```diff
@@ -1172,6 +1177,52 @@ void Kit::receivedNoteForDrum(ModelStackWithTimelineCounter* modelStack, MIDIDev
...
+		//if no, update selected drum
+		if (selectedDrum != thisDrum) {
+			selectedDrum = thisDrum;
+
+			//is the instrument clip for this kit the currently selected clip? (could be selected by holding clip in song)
+			//only need to potentially refresh the kit clip if we're actively in that clip
+			//we use this function from midi follow because it was designed to detect clip selections in song / arranger view
+			//and when a clip is selected in song / arranger view, you will need to refresh the mod led indicators
+			//and if using midi follow you may need to send updated midi feedback (because parameter context has changed)
+			Clip* clip = midiFollow.getClipForMidiFollow();
```
m-m-adams, line 1192: refactor - if this is being used outside midi follow, and doesn't need the midi follow object, then it shouldn't belong to the midi follow class. If it does rely on the midi follow object then it shouldn't be used here where there's no control on the midifollow internal variables.

Using it like this is likely to cause future issues since any changes to how midi follow clips are selected will also impact kit clip selection, and that won't be obvious to look at when reviewing changes to midi follow

src/deluge/model/instrument/kit.cpp
```diff
@@ -1172,6 +1177,52 @@ void Kit::receivedNoteForDrum(ModelStackWithTimelineCounter* modelStack, MIDIDev
...
+			if (clip && clip == activeClip) {
+				//are we currently in the instrument clip UI?
+				//if yes, we may need to refresh it (main pads and / or sidebar)
+				UI* currentUI = getCurrentUI();
+				if (currentUI == &instrumentClipView || currentUI == &automationInstrumentClipView) {
```
m-m-adams, line 1197: Can you call instrumentClipView::setSelectedDrum instead? There's a bunch of duplicated logic here

seangoodvibes: > Can you call instrumentClipView::setSelectedDrum instead? There's a bunch of duplicated logic here

the problem with using InstrumentClipView::setSelectedDrum is that it uses getCurrentKit() and getCurrentClip(), which means it won't work with pressing a clip in Song or Arranger because currentSong->currentClip isn't updated.

I did however make two improvements to setSelectedDrum because I noticed a couple things:

1) it would set the selected drum even if it's the same drum that's currently selected. This happens when you audition a kit row for instance. So I added a check to make sure we're not re-selecting the same drum. In that instance the only thing that would happen is a re-rendering of the audition pad (so it lights up when you press it).

2) it would set the mod controllable when changing the selected drum, even if affect entire is enabled. This did not make sense because with affect entire enabled you cannot control the mod controllable stack of the selected drum. So there is no need to redraw mod led's or set the mod controllable. So I added a check to see if affect entire is enabled

seangoodvibes: > Can you call instrumentClipView::setSelectedDrum instead? There's a bunch of duplicated logic here

Just as a further follow up. I have further refactored setSelectedDrum which enables me to now call that function instead of adding duplicate code to the Kit class.

m-m-adams (approved):

PR #969, opened 2024-01-13
[DRAFT] feature request: remove default cable: aftertouch -> volume

hexagon5un:
Since 4.0, the default has been to link aftertouch in to master level, with a strength of 33.  Unfortunately, this overdrives / clips on a lot of patches, and volume on aftertouch is a strange choice on its face.  

(Should pressure affect pitch bend?  Vibrato?  Filter cutoff?  IMO, it shouldn't be a global default.)

And for all the times that you're not actively using aftertouch, it just eats up a patch cable slot.

This is a super simple change, and will make a number of people happy to get rid of it without having to explicitly un-set the volume-pressure cable in every patch.

https://forums.synthstrom.com/discussion/comment/20999
https://forums.synthstrom.com/discussion/comment/23574/#Comment_23574

### DIFF
--- a/src/deluge/processing/sound/sound.cpp
+++ b/src/deluge/processing/sound/sound.cpp
@@ -307,12 +307,6 @@ void Sound::setupDefaultExpressionPatching(ParamManager* paramManager) {
 	if (patchCableSet->numPatchCables >= kMaxNumPatchCables) {
 		return;
 	}
-	patchCableSet->patchCables[patchCableSet->numPatchCables++].setup(
-	    PatchSource::AFTERTOUCH, Param::Local::VOLUME, getParamFromUserValue(Param::Static::PATCH_CABLE, 33));
-
-	if (patchCableSet->numPatchCables >= kMaxNumPatchCables) {
-		return;
-	}
 
 	if (synthMode == SynthMode::FM) {
 		patchCableSet->patchCables[patchCableSet->numPatchCables++].setup(


m-m-adams:
This exists for MPE where it makes as much sense as mapping velocity to volume, and is recommended in the MPE spec. I personally don't use it but it's simple enough to make yourself a template patch and name it so it is at the top of your list

If this is a feature you are keen on then adding the ability for a user to define a patch to be loaded for new synths would be helpful. Otherwise we lean heavily towards keeping existing behaviour in place

hexagon5un:
Patch compatibility with mainline is certainly desirable.  But I thought this was a simple fix to a problem that many people were encoutering, so I thought I'd push it. 

The workaround doesn't work for me, though, or maybe I don't understand it?  It seems to only affect that _one_ first patch -- 000.rich.saw.bass -- and none of the others on my Deluge. It also doesn't remove the effect from the init patch when I create a new one.  

(And it doesn't show up in the XML file?!  How does the Deluge know that the aftertouch dependency has been zeroed?  This is _very_ surprising behavior, and implies that if I send someone this same patch, it won't behave the same on their Deluge with respect to aftertouch?)

Was the workaround that I start every new patch by loading a modified init instead of hitting "new"?  That'll work, but it doesn't do anything for the 400+ other patches on the Deluge, many of which distort on an aftertouch keybed because of the extra volume.

I feel like I should take this discussion elsewehere, and return later, because there's a bunch I don't understand yet.

m-m-adams:
These are all great reasons, and if you can make it configurable or allow choosing a different synth template to be created with shift+synth I'd be happy to include it. While you personally may prefer this change, it is just as likely that we'll receive more complaints from people who find that their patches have all changed without their input, or start receiving bug reports about aftertouch not doing anything. 

The default expression patching only occurs when a pre 4.0 firmware file is read on a post 4.0 deluge (it's the function above the one you modified). If you then save the synth, it is no longer a pre 4.0 patch and will not have any default patching applied. You can bump the version of your existing patches if you do not like the default conversion behaviour.

The workaround I suggested for new synths is the same one Reza mentioned in the forum post you linked, create a new template patch setup however you'd like and begin your sound design there.

Queroland:
Hey, 
I once had a similar problem with the volumes default and patches on my presets.. I wrote a windows-program which allows you to set all at once for all synth programs in your library.. you can find it there:
[https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1](url)

Queroland:
[https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1](url)

Queroland:
Hey, I once had a similar problem with volumes and I worte a windows-program which allows you to set all patches and volumes at once for all your presets.. have a look here:
[https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1]

Queroland:
Hey, I once had a similar problem with volumes and I worte a windows-program which allows you to set all patches and volumes at once for all your presets.. have a look here:
https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1
or here:
https://github.com/Queroland/DelLeague/releases/tag/V2.0

PR #971, opened 2024-01-14
Fix offline build for test binaries

bobtwinkles:


bobtwinkles:
I'm not sure `task-test` is actually in the right place because it probably won't work in most dbt environments (it needs a host compiler, which we don't ship). Maybe it should go in `contrib` or something instead.

m-m-adams:
Seems fine to me, we don't need any special compiler or libraries and I think having a host compiler to build the test binary is a pretty small requirement

scripts/dbt/project.py
```diff
@@ -1,47 +0,0 @@
-import os
-from SCons.Script import GetOption
-from dbt.util import vcheck, vprint
-
-
-def compose_valid_targets(cmd_env=None):
```
m-m-adams, line 6: Is this file just unused? It doesn't seem related to tests

bobtwinkles: yeah, this is for the old SCons build. I just noticed it while adding the build task, it should probably technically be a separate PR...

m-m-adams (approved):

PR #970, opened 2024-01-14
Streamline recording new clips while Deluge is playing

cowboy:
This assumes Deluge is in grid mode, Deluge is playing, and Rec is enabled.

The following steps assume default active is set to green, but should work regardless of how default active is set. The updated behavior works in green mode.

Instead of:

1. Hold blue mode button + long-press empty clip pad(s) to create clip(s)
2. Short-press clip(s) created in step 1 to queue them up for recording at the beginning of the next bar
3. Record clip(s)

This allows one to:

1. Short-press empty clip pad(s) to create clip(s) and queue them up for recording at the beginning of the next bar
2. Record clip(s)

In my testing, this makes recording new clips significantly easier.

PR #973, opened 2024-01-14
[DRAFT] Set currentClip on clip create, even if clip is not entered

cowboy:
Per discussion in discord and comment https://github.com/SynthstromAudible/DelugeFirmware/pull/970#pullrequestreview-1820269660

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -3408,9 +3408,12 @@ ActionResult SessionView::gridHandlePadsEdit(int32_t x, int32_t y, int32_t on, C
 				if ((x + currentSong->songGridScrollX) <= trackCount) {
 					Output* track = gridTrackFromX(x, trackCount);
 					clip = gridCreateClip(gridSectionFromY(y), track, nullptr);
-					// Immediately start playing it for new tracks
-					if (clip != nullptr && track == nullptr) {
-						gridToggleClipPlay(clip, true);
+					if (clip != nullptr) {
+						currentSong->currentClip = clip;
+						// Immediately start playing it for new tracks
+						if (track == nullptr) {
+							gridToggleClipPlay(clip, true);
+						}
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


Queroland:
Hey,
I once had a similar problem with velocities and patch cables and I wrote a C# program to set patches and volume for all my synth programs at once.. I only runs on windows..but you can set all patches at once..check it out here:
[(https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1)]

cowboy:
> Hey, I once had a similar problem with velocities and patch cables and I wrote a C# program to set patches and volume for all my synth programs at once.. I only runs on windows..but you can set all patches at once..check it out here: [https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1](url)

I'm not sure I understand, how is that relevant to this PR? Thanks!

m-m-adams:
> > Hey, I once had a similar problem with velocities and patch cables and I wrote a C# program to set patches and volume for all my synth programs at once.. I only runs on windows..but you can set all patches at once..check it out here: [https://forums.synthstrom.com/discussion/4247/deluge-xml-fixer-and-with-set-volumes/p1?new=1](url)
> 
> I'm not sure I understand, how is that relevant to this PR? Thanks!

I think that was intended for the PR about changing the default modulation settings

m-m-adams:
2 things:

1. Rows mode is not handled but should have the same behaviour
2. I think it's cleaner to move the currentClip change to press as well, as is there is different behaviour between first creating a clip and pressing on an existing clip. The addition would be setting currentClip here https://github.com/SynthstromAudible/DelugeFirmware/blob/a89de9a577dfcd0d038b16961198620a8ddc8e10/src/deluge/gui/views/session_view.cpp#L3316  and here https://github.com/SynthstromAudible/DelugeFirmware/blob/a89de9a577dfcd0d038b16961198620a8ddc8e10/src/deluge/gui/views/session_view.cpp#L615 if a clip is returned

cowboy:
> * Rows mode is not handled but should have the same behaviour

I don't use Rows mode, can you give me an idea of what needs to change?

m-m-adams:
I just mean the same addition but in the equivalent spot in the non-grid session code, which I think is the default path of the handle button function

m-m-adams:
Handled by #1178

PR #976, opened 2024-01-14
Added Loopy Pro template for Midi Follow

soymonitus:
It is setup to send and receive on channel 15, when Deluge is connected via USB, so it is detected as “Deluge Port 1”


bobtwinkles:
It might be good to reduce the size of the image a bit, as it's currently larger on disk than the actual template zip (I assume the .zip format is what's required to actually feed it to loopy pro?) Otherwise looks fine to me

seangoodvibes:
> It might be good to reduce the size of the image a bit, as it's currently larger on disk than the actual template zip (I assume the .zip format is what's required to actually feed it to loopy pro?) Otherwise looks fine to me

It doesn't need to be a zip no. The actual loopy pro template is compressed inside the zip. I think it should be uploaded without the zip.

soymonitus:
If you unzip it, it is like a container folder with several other files. And the size is quite bigger. Are you ok with that? The image snapshot is for reference purposes, to include it in the docs or webpages. If it is not the place to put images let me know and I can delete it. Or do you want me to make the size smaller?

bobtwinkles:
Given the size and how people will probably interact with this template (downloading just the file from GH), it seems like we should keep the project compressed. I was hoping the zip had only 1 file or stuff we could rip out but that doesn't seem to be the case.

PR #991, opened 2024-01-15
Feature: Added new scales

soymonitus:
- Make scale detector give preference to any previously set scale if the notes contained in the sequencer clips properly match the scale (usefull for when you disable scale mode and then enable it again, you don’t want your scale to change because Deluge detected the majorness or minorness of it) => this is useful to protect 5-notes scales to be undesirably upgraded to a 7-note scale.
- Modified Roli Lumi code to also cycle through transpositions of its supported scales, because some Deluge scales are equivalent to Lumi scales if you transpose them. So now we cycle through the 12 possible transpositions of a scale, updating root note in Lumi in case it is different from Deluge

TODO:
- [x] Fix bug when creating a new song and is initialized with a scale with less than 7 notes, root notes appear repeated till they fill 7 rows
- [x] Move scales with less than 7 notes to a community feature flag, so the transposing behavior is only removed if you include those => changed to not need feature flag as now all scales are transposed between each other


soymonitus:
Ready for review!

M0R172:
> Removed the behavior of transposing all notes when you cycle through scales. As we now have scales with less than 7 notes, and could even have scales with 8 notes in the future, it is unfeasible to keep this "transposing" feature. Now, when you cycle through scales, the notes already entered remain in the same position (not transposed), and get hidden from the sequencer if they don't belong to the newly selected scale.

Hmm, I think having invisible notes (which, I presume would still get played) in the sequencer could be quite confusing...

Just as a thought - How about coloring out-of-scale notes differently (conceptually like notes that are not on the current zoom level's timing grid, but of course not using the same color as for time-shifted notes)?

soymonitus:
I need to fix a bug when creating a new song initialized with a less-than-7-notes scale.

soymonitus:
> > Removed the behavior of transposing all notes when you cycle through scales. As we now have scales with less than 7 notes, and could even have scales with 8 notes in the future, it is unfeasible to keep this "transposing" feature. Now, when you cycle through scales, the notes already entered remain in the same position (not transposed), and get hidden from the sequencer if they don't belong to the newly selected scale.
> 
> Hmm, I think having invisible notes (which, I presume would still get played) in the sequencer could be quite confusing...
> 
> Just as a thought - How about coloring out-of-scale notes differently? Conceptually like notes that are not on the current zoom level's timing grid, but of course not using the same color as for time-shifted notes - the obvious choice would be the middle color between the adjacent notes, but the color difference might be too subtle to see.

I checked with other members and they found that possibility to have notes not be transposed when changing scale to be a plus.

If I keep the old transposing behavior, it would be totally impossible to change from a scale to another scale with fewer notes because then it is not clear which rows in the octave are transposed to which rows in the new octave set. Imagine you have notes in all the notes within an octave of a 7-note scale, and then you change to a blues scale (6-notes), what do I do with the extra row? I cannot delete it, and I cannot quantize and merge with another row, that would be worse. Hence the option to not transpose at all.

When you create a song you would normally choose your scale at the beginning. If you ever want to change your scale, it would be up to you to edit your melody to the new notes. It happens that if you go out of scale mode, then move some notes up or down, when you press the scale button again, the Deluge will guess the new scale based on the notes present in the sequencer so it is not a big deal.

M0R172:
Sorry, my post was unclear, I not referrin to

> the notes already entered remain in the same position (not transposed) 

For which I understand the motivation when there are scales with different number of notes.

But I was referring to:

> get hidden from the sequencer if they don't belong to the newly selected scale.

Which I understand as: The notes are still there, are not in the current scale, are not visible in the sequencer, but are audible. This, I would find very confusing. I hear notes playing, but I can not see them - unless I change out of scale mode, or into a scale that contains them, and then they suddenly pop up out of nowhere?

Thus, the idea to *still display* the out-of-scale notes, but mark them by a *different color* (just as out-of-timing notes). This way, they could still be seen, and edited, even when not belonging to the scale. I think this would be a valuable addition, even with the existing scales, as it would allow staying in scale mode, while occasionally "borrowing" a note from out-of scale (which is how a lot of music is composed).

M0R172:
To be even more annoying (Sorry!):

>  As we now have scales with less than 7 notes, and could even have scales with 8 notes in the future, it is unfeasible to keep this "transposing" feature. Now, when you cycle through scales, the notes already entered remain in the same position (not transposed)

What if there were two options available to the user (mabye discerned by holding shift): 
a) Only allow changing to scales that have the same number of notes as the current one, and transposing notes accordingly
b) Allow changing to any scale, but not transposing

I think both have their uses, but a) is one of those happy-accident machines that I would strongly  prefer not to lose! Writing a melody or harmony and then changing scales is just so much fun!

Important point:

This is definitely easier to implement than my idea with the colored off-scale notes! 
b) is what you just implemented, and 
a) is the old Deluge behaviour plus categorizing scales by number of notes...

soymonitus:
I have implemented the runtime flag! Ready for review

soymonitus:
@M0R172 Paul Freund suggested me in the Discord channel to put the new uneven-length scales under a Community Feature toggle, so all the new added 7-note scales are just added without Feature flags, with the official transposing behavior. And for users who also want uneven-length scales, they can turn ON the feature, which apart from including those new scales, gets rid of the transposing

M0R172:
OK, I consider that much better than just removing some potentially useful functionality, thank you!

soymonitus:
I move this back to Draft because I want to investigate the possibility to do "conditional" transposing based on the notes present in the clips, so to avoid needing a feature flag for the scales with less than 7 notes.

soymonitus:
@PaulFreund @m-m-adams @M0R172 I have revamped the implementation to allow for intelligent/conditional transposition. No more need for community feature setting.
All scales are ordered in descending order, first the 7-note scales, then the 6-note scales, then the 5-note scales. The default scale in settings is saved with offsets for these 3 groups so we can add more 7 and 6 note scales in between in the future with backwards compatibility.
As you cycle through scales, the notes present in clips are taken into account to allow to go from a scale to another one with fewer notes, and you can only go to that one if there are not more notes than it can fit. If you for example are in the BLUES scale (6 notes) and click the Shift+ scale button to go to PENTATONIC, you will go there if you have entered notes in just 5 of the 6 rows (among all clips). If you are occupying all rows, you will omit all 5-note scales and clicking Shift+ scale will take you to the list start (Major).

src/deluge/gui/menu_item/defaults/scale.h
```diff
@@ -17,17 +17,67 @@
...
+				// apply offset to 5 note scales
+				v = OFFSET_5_NOTE_SCALE + v - FIRST_5_NOTE_SCALE_INDEX;
+			}
+			else if (v >= FIRST_6_NOTE_SCALE_INDEX) {
+				// apply offset to 6 note scales
+				v = OFFSET_6_NOTE_SCALE + v - FIRST_6_NOTE_SCALE_INDEX;
+			}
+			FlashStorage::defaultScale = v;
+		}
+	}
```
sapphire-arches, line 74: style nit: newline here

src/deluge/storage/flash_storage.cpp
```diff
@@ -367,7 +368,7 @@ void readSettings() {
 		defaultKeyMenu.lower = buffer[57];
 		defaultKeyMenu.upper = buffer[58];
 
-		defaultScale = buffer[59];
+		defaultScale = buffer[148];
```
sapphire-arches, line 371: we should read the old location and leave a tombstone value (e.g. 0xff), using the new location if the old location contains the tombstone.

sapphire-arches (commented):
1 minor style nit and maybe something we can improve on the flash storage, otherwise this looks OK architecturally.

src/deluge/storage/flash_storage.cpp
sapphire-arches: Also, I'm assuming you're moving locations to have a larger datatype? But that isn't necessary for now so maybe finding a way to re-use the existing location would be ideal

soymonitus: What about checking for firmware version? The thing is that RANDOM and NONE are adjacent to the existing scales so that doesn't allow us to insert more scales in the middle. Now that I moved Random and None to the end (254 and 255) to have free space for new scales, I need to use a different FlashStorage slot as if we use the existing one, and a user loads the official firmware on top of the community one, he will get rubbish from the presetScaleNotes array, pointing to bad memory places. I added different locations for 7, 6 and 5 note scales as then we can add more scales in the future, have them in order, and have the saved flashstorage be still valid when you upgrade firmware. I will add extra code to check for out of bounds indexes to reset it back to 0, in case the user downgrades to a previous community version with less scales.

soymonitus: To avoid crashing the official firmware we could consider the selection "NONE" as a tombstone value for the old slot. So if the user goes back to official firmware they will have None as selected default scale, if they change it to something else (like Minor or Dorian or whatever), and then later install the community firmware again, we recover that value again to the new slot, and write None again in the old slot

soymonitus:
@sapphire-arches  I have added code to import the old value into the new slot. I decided to use the scale NONE as the detected tombstone value, to avoid crashing official firmware if you downgrade.

sapphire-arches (approved):

Issue #471, opened 2023-08-29
Saving settings to SD card can silently fail if card is in RO mode


sapphire-arches:
### Please describe the problem:

When writing out settings (MIDI device config, community features) the save will silently fail if the SD card is in RO mode.

### What is the expected behavior?

A warning should be presented indicating that saving settings didn't work. 

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

nightly, community, Built from community, Built from release/1.0

### What is the firmware name:

dbaa84209d7edf7079366871e8933a98635805d1

M0R172:
This should be closed, I guess.

sapphire-arches:
Fixed by #1001

PR #1001, opened 2024-01-16
Issue #471 Saving settings to SD card can silently fail if card is in RO mode 

M0R172:
Displays SD  Write Error in these cases.

I changed function that is called in all the cases causing this issue, adding a second boolean argument:

`int32_t createXMLFile(char const* pathName, bool mayOverwrite = false, bool displayErrors = true);`

Note that the error code is always returned, as before, but the default is now to also display the error.

Motivation:
- There were currently two calls to this where some error codes are to be handled silently by the caller, versus four calls that just failed silently, as mentioned in issue #471. The latter cases containing more recent additions to the codebase.
- I thus think it is better to prevent any such errors from passing silently *by accident* if any more calls to  `createXMLFile` get added in the future - if you *want* errors to pass silently, you now have to consciously make that decision.


M0R172:
How to test:

- Switch your SD card to read-only (little plastic switch on the side)
- Insert into Deluge
- Enter Menu with "shift->push select"
- Exit Menu with "back"
- "SD Write Error" is displayed after "saving settings"

M0R172:
Please note that issue #1002 mentioned above is NOT caused by this PR!

src/deluge/storage/storage_manager.cpp
```diff
@@ -1026,10 +1026,14 @@ int32_t StorageManager::createFile(FIL* file, char const* filePath, bool mayOver
 	return NO_ERROR;
 }
 
-int32_t StorageManager::createXMLFile(char const* filePath, bool mayOverwrite) {
+int32_t StorageManager::createXMLFile(char const* filePath, bool mayOverwrite, bool displayErrors) {
 
 	int32_t error = createFile(&fileSystemStuff.currentFile, filePath, mayOverwrite);
 	if (error) {
+		if (displayErrors) {
```
m-m-adams, line 1033: nit: can you just && these instead of nesting?

m-m-adams (commented):
Just a minor style nit, otherwise good to go

As a follow up, would you be interested in adding a check on SD insertion to see if the SD is writeable? (The sdCardInserted function will get called on boot too)

src/deluge/storage/storage_manager.cpp
M0R172: Actually it seems I can't, because if the error should not be displayed, the function still needs to return the error...

M0R172:
> As a follow up, would you be interested in adding a check on SD insertion to see if the SD is writeable? (The sdCardInserted function will get called on boot too)

That is indeed on my agenda.

m-m-adams (approved):

Issue #998, opened 2024-01-16
Enhancement / MIDI Follow Channel UX


seangoodvibes:
### Please describe the problem:

Right now we have 3 channel's which are a bit confusing:

1. Synth
2. Kit
3. Param

The names mean nothing at the moment.

The Param channel is used separately only for Sending Midi Feedback

### What is the expected behavior?

I think we should have one channel called the feedback channel

the other channels would be Instrument A/B/C channels

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

PR #1003, opened 2024-01-16
Enhancement / MIDI Follow ~ Rename follow channel's and change feedback toggle to channel setting

seangoodvibes:
Resolves issue #998 

The current channel names of Synth/Kit/Param are confusing because you can learn to any of these channels and your "learning" works for Synths, Kits and Params.

To avoid user confusion, I have renamed the channels to Channels A, B, C.

I have also replaced the MIDI Follow Feedback toggle of Enabled/Disabled with a MIDI Follow Feedback Channel menu so you can learn the channel you want to use specifically for MIDI Follow Feedback. If you unlearn the Feedback channel, then MIDI Follow Feedback will be disabled.


m-m-adams (approved):
Concerns addressed - device is not displayed to user, so while learning is possible they hopefully won't be mislead by it

PR #1010, opened 2024-01-17
fix: blatantly illegal destructor calls for Drum on things constructed as SoundDrums

sapphire-arches:


sapphire-arches:
i feel like i tried this before and it caused some Problems, but surely this is better than the status quo (which doesn't destruct the Sound parts of the SoundDrum at all, and might even run the Drum destructor on some random Sound memory

sapphire-arches:
this was not illegal. `~Drum` is virtual, and dispatches correctly:

https://gcc.godbolt.org/z/GPsqc4xPb

PR #1011, opened 2024-01-17
Refactor Params (again)

sapphire-arches:
This refactors how the Param enum works, again, to try and reduce redundancy and recover... rorage? (storage)

I think we might want to merge these 3 changes as separate commits, so once this PR is approved I'll create separate PRs for each commit and merge them that way. There's a nonzero risk that this breaks compatibility with existing performance view saved layouts (low impact, I think -- only if you have ARP_GATE or PORTAMENTO mapped to perf view and why would you do that). There's also a risk it could break instrument/drum preset loading.

Through some arcane compiler magic I do not fully understand this patch series actually reduces the total image size by like 5k bytes:

```
   text	   data	    bss	    dec	    hex	filename
1295764	  20992	 319436	1636192	 18f760	with_changes.elf
1297324   20992  322924 1641240  190b18 baseline.elf
```


seangoodvibes (approved):

sapphire-arches:
Latest size, now that I deleted an accidental function and compressed some stuff to array lookups instead of branch hell:

```
1297324   20992  322924 1641240  190b18 baseline.elf
1294444	  20992	 319436	1634872	 18f238	b303e1c.elf
```

for a total of 6368 bytes saved.

sapphire-arches:
This still needs manual testing to make sure load/save is still working, but otherwise should be ready for review

m-m-adams (approved):
I think this is good to go - not noticing any issues loading presets 

For merging it @PaulFreund is able to bypass the queue and do a merge commit

soymonitus:
This PR makes the deluge crash on tweaking affect entire parameters

PR #1013, opened 2024-01-18
Bugfix / update global param kind definitions in performance view / midi follow / automation view

seangoodvibes:
I discovered that I was not properly assigning the right param kind in global param contexts (song, arranger, audio clips, kit affect entire).

I also noticed that midi follow feedback for global effectible param's mod fx depth and mod fx rate were incorrectly getting sent for the CC mapped to arp gate and portamento because those params share the same ID.

I fixed all these things by assigning the right param kind's and excluding arp gate and portamento from the global param context logic.


seangoodvibes:
Closing. Addressing this in the automation view for arranger PR. https://github.com/SynthstromAudible/DelugeFirmware/pull/1039

PR #1017, opened 2024-01-19
Fix `dbt format` speed

stellar-aria:
Fixes speed issue where each child process would be spawning a child process of its own to find a valid `clang-format` executable, instead of being passed the information from the parent


m-m-adams (approved):

cowboy:
"whoops"

PR #1018, opened 2024-01-19
Feature: ONCE launch mode for clips

soymonitus:
Similar to FILL launch mode but starts at beat 1 and auto-mutes after playing once.

Tasks:
- [x] Implement the basics
- [x] Write documentation in the community features document
- [x] Test and test to find edge cases


soymonitus:
@m-m-adams @entzmingerc Ready for testing and review

soymonitus:
Had to solve merge conflicts due to latest change of launchstyle to enum.
The feature is complete and tested. Could someone review the PR?

src/deluge/model/song/song.cpp
```diff
@@ -2925,7 +2925,9 @@ void Song::turnSoloingIntoJustPlaying(bool getRidOfArmingToo) {
 			// Just get rid of arming
 			for (int32_t l = 0; l < sessionClips.getNumElements(); l++) {
 				Clip* loopable = sessionClips.getClipAtIndex(l);
-				loopable->armState = ArmState::OFF;
+				if (loopable->launchStyle != LaunchStyle::FILL && loopable->launchStyle != LaunchStyle::ONCE) {
```
m-m-adams, line 2928: Should this check that the launch style is default instead?

src/deluge/model/song/song.cpp
```diff
@@ -2940,7 +2942,9 @@ void Song::turnSoloingIntoJustPlaying(bool getRidOfArmingToo) {
 		clip->soloingInSessionMode = false;
 
 		if (getRidOfArmingToo) {
-			clip->armState = ArmState::OFF;
+			if (clip->launchStyle != LaunchStyle::FILL && clip->launchStyle != LaunchStyle::ONCE) {
```
m-m-adams, line 2945: Same here

src/deluge/playback/mode/session.cpp
```diff
@@ -1515,7 +1522,8 @@ void Session::armClipsToStartOrSoloWithQuantization(uint32_t pos, uint32_t quant
 
 	// If we were doing this just for one Clip (so a late-start might be allowed too)...
 	if (clip) {
-		if (clip->launchStyle == LaunchStyle::DEFAULT) {
+		if (clip->launchStyle == LaunchStyle::DEFAULT || (clip->launchStyle == LaunchStyle::FILL && !doLateStart)
```
m-m-adams, line 1525: Why aren't once and fill allowed to do late starts? Also this expression can just be DEFAULT||!doLateStart

src/deluge/playback/mode/session.cpp
```diff
@@ -1700,7 +1708,8 @@ void Session::armClipToStartOrSoloUsingQuantization(Clip* thisClip, bool doLateS
 		// If late start...
 		if (doLateStart) {
 
-			if (thisClip->armState != ArmState::OFF) { // In case also already armed
+			if (thisClip->armState != ArmState::OFF && thisClip->launchStyle != LaunchStyle::FILL
```
m-m-adams, line 1711: Check for default instead

src/deluge/gui/context_menu/launch_style.cpp
```diff
@@ -14,8 +14,9 @@ namespace deluge::gui::context_menu {
 enum class LaunchStyle::Value {
```
m-m-adams, line 14: Can this reuse the launch style enum?

src/deluge/gui/context_menu/launch_style.cpp
```diff
@@ -43,6 +45,9 @@ bool LaunchStyle::setupAndCheckAvailability() {
 	case ::LaunchStyle::FILL:
 		valueOption = Value::FILL;
 		break;
+	case ::LaunchStyle::ONCE:
+		valueOption = Value::ONCE;
```
m-m-adams, line 49: If you use the same enum this whole switch is avoidable

src/deluge/gui/context_menu/launch_style.cpp
```diff
@@ -71,6 +76,9 @@ void LaunchStyle::selectEncoderAction(int8_t offset) {
 	case Value::FILL:
 		clip->launchStyle = ::LaunchStyle::FILL;
 		break;
+	case Value::ONCE:
+		clip->launchStyle = ::LaunchStyle::ONCE;
```
m-m-adams, line 80: Same with this one

src/deluge/playback/mode/session.cpp
```diff
@@ -560,6 +560,13 @@ void Session::doLaunch(bool isFillLaunch) {
 			}
 		}
 
+		// Arm it again if a ONCE clip, so it stops at the launchEvent
+		if (!isFillLaunch && (clip->activeIfNoSolo || clip->soloingInSessionMode)
```
m-m-adams, line 564: Can it be a fill launch if it's a once clip?

src/deluge/gui/context_menu/launch_style.cpp
stellar-aria: yeah it should be, that's on me. I can make that change real fast in the community trunk.

src/deluge/model/song/song.cpp
soymonitus: If we ever want to add more launch modes it is better to check for these two specific modes which do something different to normal, this line makes sure they are not unarmed to stop. They are armed before in the code but for Default clips they will get unarmed once they have changed state at the end of the loop. In sum, if we add later another launch mode this code will still work and won't execute unexpected behavior for the new mode which doesn't belong to it. The same applies to other changes in this PR of the 'if' conditions

src/deluge/model/song/song.cpp
soymonitus: Same as before, this avoids those two type of clips to be unarmed. If not, the Fill and Once clips will stay playing forever. This make sure that those two launchmode clips will get stopped after one play, and not other launchmodes that we might add in the future

src/deluge/playback/mode/session.cpp
soymonitus: The late starts don't trigger launch events for later. This was a bug from previous Fill mode, if you do Shift + audition, they will start playing and never be stopped after that.

src/deluge/playback/mode/session.cpp
soymonitus: Same as before

src/deluge/gui/context_menu/launch_style.cpp
soymonitus: Same

src/deluge/playback/mode/session.cpp
soymonitus: This is some hack implemented by 0beron, which calls doLaunch() method outside the regular end-of-loop moment. That time when doLaunch is called with isFillLaunch=true, it is meant to only evaluate fill clips and nothing more because it was scheduled to happen in the middle of a clips loop. So not wanting to start Once clips in the start of measure 4 in a 4 measures clips loop because a 1-measure Fill clip was scheduled to launch

soymonitus:
@m-m-adams Answered all questions and applied patch from stellar-aria to improve menu items

src/deluge/model/song/song.cpp
m-m-adams: Right but that's the default behaviour, so a new mode wouldn't want to inherit it anyway

src/deluge/playback/mode/session.cpp
m-m-adams: simplifying still applies

soymonitus: Ok I'll modify it

src/deluge/model/song/song.cpp
soymonitus: Oh yea, you are right. Modified it!

src/deluge/playback/mode/session.cpp
m-m-adams: I think this should be if launchstyle == default as well. Reason being that positive conditions are simpler to add to later

soymonitus: Done

m-m-adams (approved):

PR #1022, opened 2024-01-20
build: use a separate static library for GUI files

sapphire-arches:
This saves 162,552 bytes and fixes the ODR warnings

```
1296836   21136  318532 1636504  18f898 lto errors
1134184   21104  318664 1473952  167da0 separate GUI lib
```


sapphire-arches:
This regresses voice count by 5 (73 -> 68) currently.

m-m-adams (commented):
GTG - this slightly improves performance with multisamples (e.g. when we're memory limited) and slightly lowers performance with raw rendering. On the whole samples are a larger issue for us currently so this is a good change, and fixes linker behaviour which could break at any moment

m-m-adams (approved):

PR #1027, opened 2024-01-21
General cleanup while bughunting

stellar-aria:
Just some little QoL things:
- override in places
- not storing a `sizeof()` result (for compiler optimizations)
- returning a pair using std::pair
- simplified zeroing of sidebar data using the new colours instead of per-channel

### DIFF
--- a/src/deluge/gui/ui/ui.cpp
+++ b/src/deluge/gui/ui/ui.cpp
@@ -63,6 +63,11 @@ int32_t numUIsOpen = 0; // Will be 0 again during song load / swap
 
 UI* lastUIBeforeNullifying = nullptr;
 
+/**
+ * @brief Get the greyout rows and columns for the current UI
+ *
+ * @return std::pair<uint32_t, uint32_t> a pair with [rows, columns]
+ */
 std::pair<uint32_t, uint32_t> getUIGreyoutRowsAndCols() {
 	uint32_t cols = 0;
 	uint32_t rows = 0;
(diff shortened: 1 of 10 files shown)


src/deluge/gui/ui/ui.cpp
```diff
@@ -55,23 +56,23 @@ void UI::close() {
...
+int32_t numUIsOpen = 0; // Will be 0 again during song load / swap
 
-UI* lastUIBeforeNullifying = NULL;
-
-void getUIGreyoutRowsAndCols(uint32_t* cols, uint32_t* rows) {
-	*cols = 0;
-	*rows = 0;
+UI* lastUIBeforeNullifying = nullptr;
 
+std::pair<uint32_t, uint32_t> getUIGreyoutRowsAndCols() {
```
m-m-adams, line 66: semantically I feel like this should be a struct with rows and columns, I'm not super set on that though if there's a reason not to

m-m-adams (commented):
One small nit but good to go otherwise

src/deluge/gui/ui/ui.cpp
stellar-aria: yeah either one is probably fine, I just feel like a struct is probably overkill here since it won't get used anywhere else.
The details of the return type should be documented in doxygen and that would take care of any confusion, ideally.
...Maybe I should just do that >.>

stellar-aria: Documented.

PR #1029, opened 2024-01-21
menu: fix highlights for range entries

### LINKED ISSUE #1023: OLED inversion region in Default Swing menu when range is used is offset
The inversion region appears to be at the y level of the separator which means it's a few pixels too high to actually highlight the number correctly.

sapphire-arches:
Fixes #1023

### DIFF
--- a/src/deluge/gui/menu_item/range.cpp
+++ b/src/deluge/gui/menu_item/range.cpp
@@ -195,7 +195,8 @@ void Range::drawPixelsForOled() {
 		hilightStartX = stringStartX;
 		hilightWidth = digitWidth * leftLength;
 doHilightJustOneEdge:
-		baseY += 4;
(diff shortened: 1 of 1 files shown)


src/deluge/gui/menu_item/range.cpp
```diff
@@ -195,7 +195,8 @@ void Range::drawPixelsForOled() {
 		hilightStartX = stringStartX;
 		hilightWidth = digitWidth * leftLength;
 doHilightJustOneEdge:
-		deluge::hid::display::OLED::invertArea(hilightStartX, hilightWidth, baseY - 1, baseY + digitHeight + 1,
+		baseY += 4;
```
m-m-adams, line 198: nit: it's not clear to me why this is happening so it probably needs a comment

m-m-adams (approved):

Issue #1032, opened 2024-01-22
MIDI Follow XML file load fails when parameter name is "none"


seangoodvibes:
### Please describe the problem:

If you open MIDIFollow.XML and rename one of the parameters to <none></none>, the deluge crashes on load with error "BBBB"

### What is the expected behavior?

No crash, parameter setting is ignored if parameter name is invalid

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1034, opened 2024-01-22
BugFix / MIDI Follow ~ Fix "BBBB" crash in MIDIFollow.XML load parsing routine

seangoodvibes:
This fixes: https://github.com/SynthstromAudible/DelugeFirmware/issues/1032

It was reported that MIDI Follow could cause Deluge to crash with error "BBBB" on load whenever the param "none" was entered as a tag in the MIDIFollow.XML file.

This was a symptom of a bigger problem with the XML parsing function in MIDIFollow.

Basically, you cannot parse a tag more than once. This was happening when "none" was entered as a param because the grid loop would loop through multiple params which would return a param name of "none" which would cause a match with the XML tag and cause a re-read of the value within that XML tag.

the Function "readTagOrAttributeValueInt()" does not allow you to re-read the value within that tag and crashes with "BBBB" on the second read.

To fix this, we do a couple things:

1) we check that the x, y position in the loop actually corresponds to a param

2) a param name corresponds to one grid x, y shortcut. So after you've found that param and the corresponding x, y shortcut and after you've loaded the cc number from the XML tag into the paramToCC array, you are now done with the loop. You should exit both for loops and proceed to processing the next XML tag if there is one.

### DIFF
--- a/src/deluge/io/midi/midi_follow.cpp
+++ b/src/deluge/io/midi/midi_follow.cpp
@@ -773,28 +773,27 @@ void MidiFollow::readDefaultMappingsFromFile() {
 		for (int32_t xDisplay = 0; xDisplay < kDisplayWidth; xDisplay++) {
 			for (int32_t yDisplay = 0; yDisplay < kDisplayHeight; yDisplay++) {
 				//let's see if this x, y corresponds to a valid param shortcut
-				if ((patchedParamShortcuts[xDisplay][yDisplay] != kNoParamID)
-				    || (unpatchedNonGlobalParamShortcuts[xDisplay][yDisplay] != kNoParamID)
-				    || (unpatchedGlobalParamShortcuts[xDisplay][yDisplay] != kNoParamID)) {
-					//if we have a valid param shortcut, let's confirm the tag name corresponds to that shortcut
-					if ((!strcmp(tagName, params::paramNameForFile(params::Kind::PATCHED,
-					                                               patchedParamShortcuts[xDisplay][yDisplay])))
-					    || (!strcmp(tagName, params::paramNameForFile(
-					                             params::Kind::UNPATCHED_SOUND,
-					                             params::UNPATCHED_START
-					                                 + unpatchedNonGlobalParamShortcuts[xDisplay][yDisplay])))
-					    || (!strcmp(tagName, params::paramNameForFile(
-					                             params::Kind::UNPATCHED_GLOBAL,
-					                             params::UNPATCHED_START
-					                                 + unpatchedGlobalParamShortcuts[xDisplay][yDisplay])))) {
-						//tag name matches the param shortcut, so we can load the cc mapping for that param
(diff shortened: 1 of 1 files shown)


src/deluge/io/midi/midi_follow.cpp
```diff
@@ -766,29 +766,46 @@ void MidiFollow::readDefaultsFromFile() {
...
 	char const* tagName;
+	bool foundParam;
 	while (*(tagName = storageManager.readNextTagOrAttributeName())) {
-		for (int32_t xDisplay = 0; xDisplay < kDisplayWidth; xDisplay++) {
-			for (int32_t yDisplay = 0; yDisplay < kDisplayHeight; yDisplay++) {
-				if (!strcmp(tagName, params::paramNameForFile(params::Kind::PATCHED,
-				                                              patchedParamShortcuts[xDisplay][yDisplay]))) {
-					paramToCC[xDisplay][yDisplay] = storageManager.readTagOrAttributeValueInt();
+		//don't even process this tag if the param name is "none" as you can't find a matching param
+		if (strcmp(tagName, "none")) {
```
sapphire-arches, line 773: I think this should be made more general -- the checks below against `kNoParamID` should prevent us from ever generating a `"none"` from `paramNameForFile`, so there's no foot-gun where the compare will succeed despite the param name being invalid (baring more bugs in `paramNameFromFile`).

seangoodvibes: Removed the strcmp for "none"

src/deluge/io/midi/midi_follow.cpp
```diff
@@ -766,29 +766,43 @@ void MidiFollow::readDefaultsFromFile() {
...
-				                 params::paramNameForFile(params::Kind::UNPATCHED_GLOBAL,
-				                                          params::UNPATCHED_START
-				                                              + unpatchedGlobalParamShortcuts[xDisplay][yDisplay]))) {
-					paramToCC[xDisplay][yDisplay] = storageManager.readTagOrAttributeValueInt();
+				//let's see if this x, y corresponds to a valid param shortcut
+				if ((patchedParamShortcuts[xDisplay][yDisplay] != kNoParamID)
+				    || (unpatchedNonGlobalParamShortcuts[xDisplay][yDisplay] != kNoParamID)
+				    || (unpatchedGlobalParamShortcuts[xDisplay][yDisplay] != kNoParamID)) {
+					//if we have a valid param shortcut, let's confirm the tag name corresponds to that shortcut
+					if ((!strcmp(tagName, params::paramNameForFile(params::Kind::PATCHED,
```
sapphire-arches, line 780: We should check if the arrays do not have `kNoParamID` case by case. So something like 

`(patchedParamShortcuts[xDisplay][yDisplay] != kNoParamID && !strcmp(tagName, params::paramNameForFile(params::Kind::PATCHED, patachedParamShortcuts[xDisplay][yDisplay])) || unpatchedNonGlobal || unpatchedGlobal`

PR #1035, opened 2024-01-22
thumb by default for clarity

m-m-adams:
Thumb GUI lib was building whole binary with thumb

This PR makes that explicit, further reducing code size. Followups will seperate the DSP folder from the main code to be built with ARM but currently that leads to ODR violations with the GMA and ParamManagers

There is no net performance effect of this PR vs. current head but it is -100k bin size. Compared to the arm build it's -5 pure synth arp voices, +2-3 multi sample voices.


src/deluge/CMakeLists.txt
```diff
@@ -28,14 +20,14 @@ target_link_libraries(deluge PUBLIC
...
     eyalroz_printf
-    deluge_gui
 )
 
 target_compile_definitions(deluge PUBLIC
     $<$<CONFIG:DEBUG,RELWITHDEBINFO>:IN_HARDWARE_DEBUG=1>
     $<$<CONFIG:DEBUG,RELWITHDEBINFO>:ENABLE_TEXT_OUTPUT=1>
 )
 
+
```
sapphire-arches, line 30: Extra newline

src/deluge/dsp/util.h
```diff
@@ -15,15 +15,6 @@
...
 
-#include "definitions_cxx.hpp"
-#include "dsp/filter/filter.h"
-#include "dsp/filter/filter_set.h"
-#include "dsp/filter/lpladder.h"
-#include "dsp/filter/svf.h"
-#include "dsp/timestretch/time_stretcher.h"
-#include "processing/sound/sound.h"
-#include "storage/storage_manager.h"
-#include "util/functions.h"
```
sapphire-arches, line 26: Did you mean to include these changes?

m-m-adams: Yes, just a bunch of un needed includes I came across

PR #1036, opened 2024-01-22
Build/dsp in arm

m-m-adams:
Ready to go - builds DSP code in ARM, using a unity build in place of LTO. Significant bin size reduction, down to 1.1mb

+10 voices compared to pure thumb, returning to the arm performance level


src/deluge/dsp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+    NE10
+    eyalroz_printf
+)
+target_compile_options(deluge_dsp PRIVATE
+    ${DELUGE_COMMON_COMPILE_OPTIONS}
+    # use a unity build instead (combine all cpp to one before compiling), which has similar gains without the arm/thumb issues
+    "-fno-lto"
+    # we're not using lto so go ahead and inline
+    "-finline-functions"
+    # Build dsp code with thumb mode
```
sapphire-arches, line 39:
```suggestion
    # Build dsp code with arm mode
```

src/deluge/dsp/delay/delay.cpp
```diff
@@ -32,17 +29,8 @@ Delay::Delay() {
...
-	if (!song) {
-		song = currentSong;
-	}
-	if (song) {
-		syncLevel = (SyncLevel)(8 - (song->insideWorldTickMagnitude + song->insideWorldTickMagnitudeOffsetFromBPM));
-	}
-	else {
-		syncLevel = (SyncLevel)(8 - FlashStorage::defaultMagnitude);
-	}
+	syncLevel = (SyncLevel)5;
```
sapphire-arches, line 32: Does this get overwritten later?

src/deluge/CMakeLists.txt
```diff
@@ -7,6 +8,9 @@ file(GLOB_RECURSE deluge_SOURCES CONFIGURE_DEPENDS
   ${CMAKE_CURRENT_LIST_DIR}/*.cpp
 )
 
+# Filter out GUI sources as those are built as part of the deluge_gui library
```
sapphire-arches, line 11:
```suggestion
# Filter out DSP sources as those are built as part of the deluge_dsp library
```

src/deluge/dsp/delay/delay.cpp
m-m-adams: Yeah it's set by mod controllable audio/global effectable now in order to strip out all the dependencies on song and param

PR #1053, opened 2024-01-24
Documentation / Add Touch OSC Midi Follow Template

seangoodvibes:
Adding Touch OSC template created by @s.e.s on Discord 

- It is setup to send and receive on Channel 1 when the Deluge is setup in the Touch OSC Connections menu

touch_osc


soymonitus:
I found from the template that it is setup to send/receive on channel 1. Maybe good to note that in the documentation

PR #1054, opened 2024-01-24
Chore/clangfmt update

m-m-adams:
Run format with new clang format version

Bump clang format version in workflow

Ignore format in blame


sapphire-arches (approved):

m-m-adams:
Needs to be a merge commit, squash will break the blame ignore

PR #1077, opened 2024-01-26
Add clangd completion engine to VSCode

stellar-aria:
This also sets clangd as the default engine.


Makefile
```diff
@@ -1,16 +0,0 @@
-ifeq ($(OS),Windows_NT)
```
m-m-adams, line 1: I like the make file, don't kill please!

m-m-adams (approved):

PR #1078, opened 2024-01-26
Send Firmware via Sysex. Cross-Platform, Single python file, not using old loadfw, only one python dependency.

M0R172:
Explicitly acknowledging the prior art of https://github.com/SynthstromAudible/DelugeFirmware/pull/834,
this is a slightly more minimalistic version of only one of the provided features. 
I hope this will make it easier to evaluate and test, and thus get it into community soon.

This aims to solve two problems:
1. currently loadfw must be compiled from c
2. currently loadfw uses alsa, which is linux-only
3. I could not make it work on my machine

`dbt loadfw 3 abcd1243 /path/deluge.bin`
loads the firmware /path/deluge.bin to the deluge with key abcd1243, connected on MIDI port 3

`dbt loadfw -h`
Not only displays help, but also lists available midi ports, output on my system:

> mw@mw-G5-Laptop:~/DelugeFirmware$ ./dbt loadfw -h
> usage: loadfw [-h] [-d D] port_number hex_key firmware_file
> 
> Send firmware to Deluge via MIDI
> 
> positional arguments:
>   port_number    MIDI port number (example: 0)
>   hex_key        8-digit Deluge Hex Key (example 1234abcd)
>   firmware_file  Path to firmware binary file (example ./build/Debug/deluge.bin)
> 
> options:
>   -h, --help     show this help message and exit
>   -d D           Delay in milliseconds between SysEx packets. Default is 1. Increase in case of checksum
>                  errors.
> 
> usage example: dbt loadfw 3 abcd1243 /path/deluge.bin
> 
> loads /path/deluge.bin to the deluge with key abcd1243, connected on MIDI port 3
> 
> Available MIDI ports:
> 
> Port # 0 : Midi Through:Midi Through Port-0 14:0
> Port # 1 : Deluge:Deluge MIDI 1 24:0
> Port # 2 : Deluge:Deluge MIDI 2 24:1
> Port # 3 : Deluge:Deluge MIDI 3 24:2


sapphire-arches:
This works on my machine (Weird Linux).

soymonitus:
Wow i want to test it with macOs!

M0R172:
IMO they should stay separate: While they of course complement each other, 
they are technically independent, 
and I think making a small amount of changes per PR makes each PR easier to test and agree on. 

I actually hope that we get this into community faster via two than via one PR...

soymonitus:
Not meaning to join both on your side but just asking if they were complementary and i could on my own merge both and test with both at the same build

soymonitus:
ModuleNotFoundError: No module named 'rtmidi'

m-m-adams:
Works on M1 mac - I suspect the intel mac issues are on the python-rtmidi side as they've had mac packaging issues in the past

M0R172:
@soymonitus 
When you force a reinstall of the toolchain (e.g. by cloning into a new directory), everything should be installed automatically, since it is in the requirements file (is there some easier way to update the toolchain? Should this be automatically happening on a pull?)

If you need to manually install the package: you have to call 
`pip install python-rtmidi` NOT rtmidi, and NOT rtmidi-python...
Don't ask me why this is so...

soymonitus:
> @soymonitus When you force a reinstall of the toolchain (e.g. by cloning into a new directory), everything should be installed automatically, since it is in the requirements file (is there some easier way to update the toolchain? Should this be automatically happening on a pull?)
> 
> If you need to manually install the package: you have to call `pip install python-rtmidi` NOT rtmidi, and NOT rtmidi-python... Don't ask me why this is so...

Do you have an M1 mac? I tried with an Intel one and the wheel it downloaded was wrong, i had to do:
dbt shell
pip install python-rtmidi==1.5.8 --no-binary python-rtmidi,rtmidi

soymonitus:
On the other hand I cannot get it to work, after reaching 100%, the Deluge says CHECKSUM FAIL. It also fails for the PR from tasycode, but sysex logging works. I have no idea if it is just my Deluge or something else

M0R172:
After reinstalling the build chain, now it also does not work here - I have NO IDEA what is going on anymore...

m-m-adams:
Same - now it's crashing mine. Not sure on the issue

soymonitus:
If it is of any help, with the speed set to -d 1 i get the checksum fail, and with speed set to -d 2 or higher I get a crash. Maybe my Deluge was needed some small delay in syx sending, and now the crashes I am getting are the same that you are getting so if that is fixed it will be fixed for me also if I use delay

M0R172:
Yes, these are definitely two separate issues. That the checksum can fail if data is sent too fast is no big surprise, and it seems that 1ms delay is about the limit of what works over USB midi - it works for me 99% of the time, but not  always... fine, so I might increase the default to 2 ms, the transfer is still pretty fast.

But this crash is really driving me mad now - I *never* had anything like this before updating the toolchain!

sapphire-arches:
A couple notes from debugging with Monitus:
 - It seems like release worked on the original (pre-toolchain update) PR, but debug did not.
 - The chainload logic currently does a whole bunch of illegal stuff including:
   - Assuming the code for the chainload function is linearly laid out in memory and < 256 bytes in size
   - Blindly copying the chainload stub to memory without really ever checking that it's going to be legal

M0R172:
Hmm, I thought I'd just try adding
`__attribute__((optimize("align-functions=256")))`

before

`static void chainloader(uint32_t code_start, uint32_t code_size, uint32_t code_exec, char* buf)`

Alas:

```
/tmp/ccuwrBDR.s: Assembler messages:
/tmp/ccuwrBDR.s:90: Fatal error: alignments greater than 64 bytes not supported in .text sections.

```
But maybe there is some way to change this limit?

bfredl:
> The chainload logic currently does a whole bunch of illegal stuff including:
>    Assuming the code for the chainload function is linearly laid out in memory and < 256 bytes in size
>    Blindly copying the chainload stub to memory without really ever checking that it's going to be legal

Ye, the current implementation was completely empirical, rather than following some kind of rulebook. There's a lot of "legal" stuff which doesn't work as the renesas chip doesn't behave as an arm-a9 is documented (for instance, any attempt of using "standard" instructions for data vs instruction cache managhement causes a hard lockup).

M0R172:
Any place where I could find more info on this topic, i.e. why the current chainloader looks like it does?

M0R172:
BTW, I just considered if this PR could go back to "ready for review" as it seems pretty clear that the python script  works as it should (maybe someone should verify on windows, but who uses that, anyway ;-) ). 

The chainloading bug related to the new toolchain seems a technically independent problem, and  might actually be easier to reproduce and diagnose with python loadfw available...

Any opinions?

sapphire-arches:
yeah, let's evolve this tool in followup PRs. Thanks!

PR #1079, opened 2024-01-26
Sysex Firmware load: Deluge pads as progress bar, Fix oled wrap bug

M0R172:
1. Cear visual feedback *on the device* that firmware is being uploaded 
2. The occasional "oled wrap" bug, corrupting the display when chainloading seems fixed (by stopping the scroll animation when uploading), tested by uploading 100 times in a row without any screen issues.
3. Sysex loading is now enabled by default, which I think makes sense for dev firmware versions.

Feedback welcom!


sapphire-arches:
1. Can we trust the `PadLEDs` code to still exist during firmware loading? IIRC from when I read it a while ago the logic does not protect any code regions except itself.
2. I think we should fix this by sending a reset sequence to the OLED on boot. This would also fix the pixel offset bug when using "normal" flashing.
3. Sysex firmware loading is intentionally disabled by default as it's a bit of a security concern. I'm not super opposed to enabling it by default, but this should be done in `CMakeLists.txt` not by manually defining it in the header.

M0R172:
> Can we trust the PadLEDs code to still exist during firmware loading? IIRC from when I read it a while ago the logic does not protect any code regions except itself.

I assume that while uploading (which is when I animate the pad leds), the new firmware is just loaded into a normal memory buffer, and only in the final step it replaces the currently running code. I might be wrong, but two things indicate to me that this is most likely true:
1. When you upload, but not "check and load" (which is triggered by an extra sysex command, see sysex.cpp) the new firmware, the Deluge keeps running normally, which it could not do if the code was already overwritten.
2. Is basically the same: When the status of the last pad is updated, 143/144 of the new firmware is already loaded - it would be a huge coincidence if the led code was in that part, so I again assume the running code is only overwritten in the end (or maybe the execution pointer is just set to the new buffer?).

> I think we should fix this by sending a reset sequence to the OLED on boot. This would also fix the pixel offset bug when using "normal" flashing.

I agree that this should be the proper and logical solution, and believe me, I tried...

Stopping the scroll animation before "reboot" was the only thing that would work for me.
It is a purely heuristic hack, but it seems to works, and I don't see it doing any harm.

If someone who knows more about the hardware knows the proper incantations to solve this, 
that line of code can be replaced anytime. But since so far no fix was offered AFAIK, this is *much* better than nothing for now...


> Sysex firmware loading is intentionally disabled by default as it's a bit of a security concern. I'm not super opposed to enabling it by default, but this should be done in CMakeLists.txt not by manually defining it in the header.

Is there a way to do this from the command line when calling dbt build??? I searched intensely and could not find it. if there is something like
`dbt build release -DENABLE_SYSEX_LOAD`
*but actually working, which this is not*, *PLEASE* tell me!

M0R172:
On the topic of fixing the OLED boot wrap issue: I tried all of the below, and combinations thereof (although certainly not all possible permutations), both before reboot, or during startup in deluge_main - nothing helped, except stopping the scroll animation before reboot. Again, if someone knows/finds out how to do it "properly", I would be happy to hear about it, but until then, I'd rather have an ugly hack than a messed-up screen...

`
deluge::hid::display::OLED::clearMainImage();
deluge::hid::display::OLED::sendMainImage();
if (deluge::hid::display::have_oled_screen) {
	oledRoutine();
}

PIC::selectOLED();
PIC::flush();
PIC::waitForFlush();
PIC::deselectOLED(); 


R_RSPI1_Stop(); // NOPE

display->freezeWithError("");

	
MainOLED_WCom(0xFD); // SET COMMAND LOCK
MainOLED_WCom(0x12);
MainOLED_WCom(0xAE); //DOT MARTIX DISPLAY OFF
MainOLED_WCom(0x81); //CONTARST CONTROL(00H-0FFH)
MainOLED_WCom(0xFF);
MainOLED_WCom(0xA4); //ENTIRE DISPLAY OFF(0A4H-0A5H)
R_RSPI_WaitEnd(SPI_CHANNEL_OLED_MAIN);
`

stellar-aria:
> Is there a way to do this from the command line when calling dbt build??? I searched intensely and could not find it. if there is something like
dbt build release -DENABLE_SYSEX_LOAD
but actually working, which this is not, PLEASE tell me!

You'll need to pass it to `dbt configure`, not `dbt build`, as the setting is stored in the CMakeCache. `dbt build` is just running ninja, not really CMake.

M0R172:
Hey, thanks a lot!

Just one thing: Do I have to redo this whenever I pull from community?

stellar-aria:
It should get saved in the CMakeCache, unless you nuke your build folder or manually reconfigure

M0R172:
OK thanks again - I removed the ENABLE_SYSEX_LOAD flag from the PR to make it less "controversial".

sapphire-arches:
I think you're the first person to find a reliable repro for the offset issue. If you use the SET COLUMN ADDRESS and SET PAGE ADDRESS commands described in https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf in `oledMainInit` , does it fix the problem?

M0R172:
> I think you're the first person to find a reliable repro for the offset issue. 

I am probably just fastest to write firmware over sysex - the screen starts to wrap pretty reliably after 3 to 5 times overwriting the firmware...

> If you use the SET COLUMN ADDRESS and SET PAGE ADDRESS commands described in https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf in oledMainInit , does it fix the problem?




SET PAGE ADDRESS is already in `oledMainInit`:
```
#if OLED_MAIN_HEIGHT_PIXELS != 64
	MainOLED_WCom(0x22); // Set page address start / end
	MainOLED_WCom((64 - OLED_MAIN_HEIGHT_PIXELS) >> 3);
	MainOLED_WCom(7);
#endif
```
BTW, the if statement could be removed. IF there ever was a deluge with a 64 pixel high screen, these lines would indeed not be necessary, but do the correct thing for that case (set start to 0 and end to 7, which the manual says is done at reset anyway)


Current code is also doing

```
	MainOLED_WCom(0x00); // SET LOW COLUMN START ADDRESS
	MainOLED_WCom(0x10); //SET HIGH COLUMN START ADDRESS

	MainOLED_WCom(0x20); //SET MEMORRY ADDRESSING MODE
	MainOLED_WCom(0x00); // Horizontal
```
Which looks wrong, because the manual says 0x0_ 0x1_ are only for page adressing mode, but we are clearly selecting horizontal mode. However, if I remove the first two of these lines and instead do

```
MainOLED_WCom(0x21); //Set Column Address
MainOLED_WCom(0x00); // Start
MainOLED_WCom(0x10); // End
```
I see absolutely no change. Display works just as before  but/and screen wrapping after reboot still happens...

src/deluge/io/debug/sysex.cpp
```diff
@@ -154,6 +173,7 @@ void Debug::loadCheckAndRun(uint8_t* data, int32_t len) {
 	unpack_7bit_to_8bit((uint8_t*)fields, sizeof(fields), data + 4, 14);
 
 	if (handshake != fields[0]) {
+		display->displayPopup("Incorrect Handshake Key!");
```
sapphire-arches, line 176: This should use `deluge::l10n::get(deluge::l10n::String::STRING_FOR_BAD_KEY)` with entries in `english.cpp` and `seven_segment.cpp`. Search for `STRING_FOR_WRONG_SIZE` in those files for an example.

sapphire-arches:
OK, the display controller on this OLED is apparently not a complete SSD1306 clone. I've filed #1084 to track this reset issue, but we'll ship the workaround for now. One minor comment about using the I10N system but otherwise this looks good to go.

PR #1095, opened 2024-01-27
chore: remove unused dependencies from toolchain

sapphire-arches:
This doesn't appear to be used anywhere, get rid of it for now.


sapphire-arches:
These were `GitPython` indirect dependencies, I think.

PR #1098, opened 2024-01-27
py: use black to format code

sapphire-arches:


sapphire-arches:
Need to check if this works with the vscode setup.

stellar-aria (approved):

Issue #1097, opened 2024-01-27
Community setting for default reverb model


sapphire-arches:
This maybe should go in the normal defaults menu.

m-m-adams:
I agree that this should just be a normal default setting

Issue #1082, opened 2024-01-26
Mutable Reverb mode persists when loading different songs


ok-reza:
### Please describe the problem:

Once Mutable is selected as a reverb mode, it persists when loading different songs until Deluge is turned off.

### What is the expected behavior?

Reverb mode should be saved per project, if no reverb mode is present in xml (eg. old project files) then freeverb should be default.

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

1/26 nightly: https://github.com/SynthstromAudible/DelugeFirmware/actions/runs/7673299358

### If possible provide the steps to reproduce the issue and upload additional media:

1. turn on deluge, switch reverb to mutable (sound settings>FX>reverb>mutable)
2. load another project that already utilizes reverb
3. note mutable reverb (tends to be longer and more present)

PR #1099, opened 2024-01-28
Enhancement / Set default reverb mode to mutable, read/write reverb mode from Song XML, set older firmware song's to Freeverb

seangoodvibes:
This PR closes https://github.com/SynthstromAudible/DelugeFirmware/issues/1097 and https://github.com/SynthstromAudible/DelugeFirmware/issues/1082

It does so by doing the following:

1. Mutable reverb is now the default for new songs
2. Song's created/saved before the current firmware version (FIRMWARE_4P1P4_ALPHA) will load the "Freeverb" reverb mode
3. New song's saved with the current firmware version (FIRMWARE_4P1P4_ALPHA) or later firmwares will save the current reverb mode to the Song XML file and will read and set that reverb mode when the song is loaded.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -187,7 +187,6 @@ Song::Song() : backedUpParamManagers(sizeof(BackedUpParamManager)) {
 	reverbCompressorSync = SYNC_LEVEL_8TH;
 	AudioEngine::reverb.setModel(deluge::dsp::Reverb::Model::MUTABLE);
 
-
 	masterCompressorAttack = 10 << 24;
 	masterCompressorRelease = 20 << 24;
 	masterCompressorThresh = 0;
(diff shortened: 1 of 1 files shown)


ok-reza:
great work!
hm i think it's best practice to still have freeverb as the default for new projects, it seems like mutable still may have a round of tuning its parameters and there are still some issues with it (reverb cuts off after 12 seconds) and users unaware may be confused why their reverb is behaving differently (also it's implemented differently in its current state which will add to confusion eg. width doesn't solely control stereo width)
eventually, a user deliberately choosing their preferred reverb model seems like the best option for users

seangoodvibes:
> great work!
> 
> hm i think it's best practice to still have freeverb as the default for new projects, it seems like mutable still may have a round of tuning its parameters and there are still some issues with it (reverb cuts off after 12 seconds) and users unaware may be confused why their reverb is behaving differently (also it's implemented differently in its current state which will add to confusion eg. width doesn't solely control stereo width)
> 
> eventually, a user deliberately choosing their preferred reverb model seems like the best option for users

Hey, thanks for the feedback. 

I still think we should change the default as I'm pretty sure the consensus is that the existing reverb isn't great and contributes to the perception that the deluge doesn't sound good. setting a better reverb as the default will help reduce the amount of complaints / comments about how the deluge sounds "out of the box."

m-m-adams:
Gonna merge this, caveat that firmware version for songs being loaded isn't set properly yet. I'll send a follow on PR for that as I've found it while troubleshooting another song xml loading issue

m-m-adams (approved):

PR #1101, opened 2024-01-28
Fix sysex loader - Version 2 (final) (Really Final) (FINAL FINAL)

sapphire-arches:
Fixes the sysex chainload logic by porting it to assembly and disabling the MMU.

### DIFF
--- a/src/deluge/util/chainload.S
+++ b/src/deluge/util/chainload.S
@@ -87,7 +87,7 @@ chainloader_copy_start:
   add       r6, r6, #4
   b         chainloader_copy_start
 chainloader_copy_done:
-  // Reload arguments and jump to the new code
+  // Reload arguments and jump to the second-stage chainloader
   ldr       r0, [sp, #0x00]
   ldr       r1, [sp, #0x04]
   ldr       r2, [sp, #0x08]
@@ -98,6 +98,7 @@ chainloader_copy_done:
   // Ensure instruction alignment
   .align 2
   // Start of the second-stage chainloder
+  // Expects r0, r1, r2, and r3 to be set as described in the arugments
 chainload_code_start:
   // Copy the new image to $r0
   // r4 is scratch storage (the data being copied)
@@ -125,7 +126,6 @@ buffer_copy_done:
   dsb
   isb
   // Jump to the new firmware image
-  ldr       r2, [sp, #0x08]
   bx        r2
 
   // Align to a full instruction so the first-stage chainloader copies enough bytes
--- a/src/deluge/util/chainload.S
+++ b/src/deluge/util/chainload.S
@@ -1,0 +1,134 @@
+/*
+ * Copyright © 2024 Synthstrom Audible Limited
+ *
+ * This file is part of The Synthstrom Audible Deluge Firmware.
+ *
+ * The Synthstrom Audible Deluge Firmware is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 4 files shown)


bfredl (approved):
This seems to work reliable in both Release and Debug mode, so a big :+1: from me.

M0R172:
Same here, 50 consecutive firmware loads, both in debug and release, without any problem! YOU ROCK!

I discovered a little aesthetics bug in my progressbar code (it occasionaly skips a pad), maybe you'd include this little fix, so the update process looks as flawless as it works? (Otherwise I'll open a seperate PR, but the change seems too small for that)

in sysex.cpp `Debug::loadPacketReceived`
replace current progress display code (lines 153-161) by
```
uint32_t pad = (18 * 8 * pos) / load_bufsize;
uint8_t col = pad % 18;
uint8_t row = pad / 18;
PadLEDs::image[row][col] = {static_cast<RGB::channel_type>(255 / 7 * row), 0,
						static_cast<RGB::channel_type>(255 - (255 / 7) * row)};
if ((pos / 512) % 16 == 0) {
PadLEDs::sendOutMainPadColours();
PadLEDs::sendOutSidebarColours();
}
```

Again: Thank you for the fantastic work - this will be a whole new era of Deluge development!

soymonitus:
It works for me! but i need to set speed to -d 2 (2ms). With 1ms it gives me checksum fail. Wouldn't it be nice to setup the default speed to 2ms, just in case other deluge users experience the same issue as me. For me 2ms works all the time

M0R172:
> Wouldn't it be nice to setup the default speed to 2ms?

I agree. I will make a small bugfix PR combining the new default with the pad-led-progressbar fix I mentioned above, and maybe a little clarification of the help text.

> Another note. The documentation needs to be changed for the loadfw commnad. The index of ports seems to be 0-indexed. So for Deluge Port 1 the comnand in reality is:
> ./dbt loadfw 0 ....
> and for Deluge Port 3 the command is:
> ./dbt loadfw 2
> ./dbt loadw 3 will fail, there is no port 3 for the command

The port numbers are indeed zero indexed.

However, these port numbers refer to the midi ports *of the computer/os*, not of the deluge...

IF you have no other MIDI ports (physical, virtual, whatever your OS decides is a MIDI port), Deluge's MIDI port 1 will be your/loadfw's midi port 0, and Deluge's port 3 will be system/loadfw port 2. On my machine, as a counterexample I have some virtual loopback port as port 0 (no idea why), and Deluge port 3 *happens to be* the computer's port 3.

When you do 

`dbt loadfw -h`

It will output a list of the actual midi ports registered on the system, with names and port numbers, so you can select the correct port.

What I might do is choose something like 99 as an example of the port number, so it is more clear that it is an example and not the port number you should actually use...

src/deluge/util/chainload.S
```diff
@@ -0,0 +1,134 @@
...
+  b         buffer_copy
+buffer_copy_done:
+  // flush L1I$
+  mov       r0, #0
+  mcr       p15, 0, r0, c7, c5, 0      @;; ICIALLU
+  // Issue memory barriers
+  dsb
+  isb
+  // Jump to the new firmware image
+  ldr       r2, [sp, #0x08]
```
m-m-adams, line 128: duplicate of line 93?

src/deluge/util/chainload.h
```diff
@@ -5,4 +5,7 @@
 #define OFF_USER_CODE_EXECUTE 0x28
 #define OFF_USER_SIGNATURE 0x2c
 
+/// Chainload from a buffer.
+///
+/// Currently only implemented for ARM.
```
m-m-adams, line 10: can we add a note about entry point location being the value of buffer+0x28 for the sake of future us?

M0R172:
Why was this closed but not merged???

sapphire-arches:
I got overeager deleting old branches, sorry. I'll restore it in a few
hours.

PR #1102, opened 2024-01-28
Add util::map and expand Mutable parameter range

stellar-aria:


src/deluge/util/misc.h
```diff
@@ -70,6 +70,11 @@ constexpr T set_top_n_bits(const T v) {
 	}
 	return output;
 }
+
+template <typename T>
+[[nodiscard]] constexpr T map(T x, T in_min, T in_max, T out_min, T out_max) {
```
m-m-adams, line 75: This would be more useful with two types (to map param manager q31_ts to floats) and a control over log-lin-exp shaping. As is this is pretty specific to the reverb 

example of other code that could use it -
https://github.com/SynthstromAudible/DelugeFirmware/blob/e23e966757dd374921be794e564706e23a695e8c/src/deluge/dsp/compressor/rms_feedback.h#L34

m-m-adams: Which is fine, but in this form I think it's better to just leave it in the reverb code instead of utils

stellar-aria: This is a really just a pretty generic parameter mapping function. It's actually just Arduino's IIRC haha.

The problem with two types is that how those two types interact can be obscured by the templated function. I had an issue with that using two different types of integers at one point I think 🤔

It's easier to just do the type conversion first, and then the scale mapping.

m-m-adams: yeah fair enough

m-m-adams (approved):

PR #1114, opened 2024-01-29
feature: DX7 synth

bfredl:
This still needs a bit more work, but I'm currently in the process of reworking my old DX7 branch to integrate better with the native UI of the deluge, like using soundEditor menus instead of the homegrown "DX EDITOR" UI I played with last summer.

What has which has been reworked so for is the importing of existing patches in .syx patch bank format (those which contain 32 dx7 patches each). This can be done as follows:

- place such files in a "DX7/" folder in the root of the sdcard
 - PRESS "STUTTER / CUSTOM 1" + "SYNTH" to create a new synth in dx7 mode (just a placeholder, need to come up with the proper shortcut grammar for things like this)
 - in the menu, select the first option
 - select a .syx file
 - patches inside the file can now be auditioned (enable keyboard view to audition while in this menu)

# pros 

- classic synth engine with a lot of patches available.
- more powerful FM engine in some aspects (separate envelopes per operator, etc)
- added codesize is reasonably small (~28 kb) for the amount of sounds being made available
- polyphony is quite high, at least with the "modern" style engine which has a SIMD implementation. (~48-56 ish voices is reasonable)

# cons

- having two separate FM engines integrated into the core might seem confusing and redundant.
- Even with the reworked UI, will likely always feel a bit like a "foreign citizen" in the deluge architecture. For example the osc1/osc2/mod1/mod2 structure is not used, but instead one will need to work with a separate set of operators (existing "inside" osc1, can be compared with "kernels" of the Waldorf Iridium I guess)

# alternatives

- implement some kind of "plugin system" for synth engines. This will take a quite some effort both in design and implementation, especially how the UI:s of such plugins should be integrated into the menu system, and how their parameters should be saved into xml/json song files.

While this path is something I might be interested on working on myself, the time scale for such a thing will be much longer ahead, compared to the effort of finalizing this rework. 

- add features from this into the deluge FM engine, which would be more in line with #227 . A related option for this could be to port the DX7-style envelope and add one each to MOD1 and MOD2. This will miss out on the ability of using existing dx7-patches however (which was one of my motivations for doing this in the first place.)


sapphire-arches:
Size smoke check:

```
1133412   21344  348924 1503680  16f1c0 DX7 baseline
1162284   21540  362724 1546548  179934 DX7 support
```

(total delta + 43k bytes). Probably reasonable for a whole synth engine

lobit12:
Im curious if this is using the DX7 oscillator type thats found in the alternative Plaits firmware?

being able to load classic 6 op banks would be amazing, I think its fine to have it as a 'separate' functionality to the current 4op like synth.

bfredl:
> Im curious if this is using the DX7 oscillator type thats found in the alternative Plaits firmware?

It is a DX7 compatible oscillator but It is based on dexed (which in turn is based on music-synthesizer-for-android), not any code from plaits. you can load DX7 .syx files in the common 32-preset bank format.

lobit12:
did you add this to the nightly build already?   how can I try it out?

bfredl:
This is starting to be ready for general review. The most "controversial" aspect to discuss might be UI shortcuts needed to operate this conveniently. All of the above behaviors are only active behind a community setting

- There needs to be a shortcut for the action "create new DX7 synth". Currently this is CUSTOM1+SYNTH, but open for suggestions. Not sure where an action like this would fit in the menu system. (This action uses different defaults to make the overall synth structure work well with the dx7 engine, in particular making the amp env1 fully open and removing the velocity to global volume mapping, as these both are controlled per operator).

- when a dx7 synth is active, the default sidebars change from velocity+modwheel to velocity + dx7 operators sidebar. This is very useful to quickly get into editing of the six operator params.

- when editing any inner DX7 parameter, a secondary "layer" of shortcuts are active to quickly jump between most common dx7 parameters

m-m-adams:
Seems to get about 50 voices so I think we just need to worry about UI stuff

m-m-adams:
I've got a couple ideas for improving how integrated this is with the deluge

First one - go the opposite way and have this entirely replace the synth engine when it's active instead

Pros:
1. Less confusing as a synth mode than an oscillator mode
2. Would let us use shortcuts directly to DX7 controls
3. Would let the synth menu get entirely replaced by the DX7 menu, so there's less confusing navigation
Cons:
1. No filters
2. Shortcut pads would be mislabelled

Second one - leave it in the oscillator, but have the DX7 presets form a macro controlled oscillator. My idea here would be to use "wavetable position" as time for the DX7 envelopes, allowing the user to "scrub through time" by modulating position with an envelope or other mod source. PW could be used to control a global DX7 mod depth or something
Pros:
1. Would integrate really nicely
2. Avoids on board editing issues
3. Sounds really cool
Cons:
1. Less powerful

lobit12:
I gotta say I love being able to access filter and envelope options on the classic 6OP sounds- its super cool,  I think the only other device that did it was the fabled Yamaha FS1R.  I bugged befredl for an early version so had a couple months on it now.

If I had to choose between the two options It would be the first (wouldn't want to lose access to editing options) but I actually really like the way its acting as a layer of the subractive engine.  Thats just my 2c.   of course would love to have gold knob access / automatable params and maybe that would be worth losing the filter for.. that would be a tough choice.  

 I definitely think as far as 6OP editing goes the grid is laid it very logically (operators on rows) and has pretty much most of the important stuff there, very intuitive editing.

bfredl:
I really like the ability to use filters as a post-processing effects. That they do not work on FM4 is a consequence of a mis-match how samples are processed, IIRC. As there is no such limitation on DX7 (it just produces sound just like a wavetable, as you mentioned), it seems like an artificial restriction to not allow this. Also I think layering a sample as a initial attack together with a DX7 tail can be a very powerful sound design tool.

>  My idea here would be to use "wavetable position" as time for the DX7 envelopes, allowing the user to "scrub through time" by modulating position with an envelope or other mod source.

Unfortunately DX7 envelopes does not work a way for this to really make sense. The sound can hugely be a affected by from where the envelope happens to be when you "key up", as the envelope keeps it state but changes the rate.. They are not a time-reversable thing you can scan through like a wavetable.

I'd rather just add operator 1-6 LEVEL as mod destinations, something I plan to do as a follow-up, as this PR is big enough as it is. Importantly, _including_ making them accessible in the automation view, to fully integrate this with other added community functionality. 

> PW could be used to control a global DX7 mod depth or something

This is already being done in this PR. Sensitivity is controlled by ampmod on each operator, consistently how it works on DX7 and many of its emulations (so it works with many existing patches).

m-m-adams:
Maybe two modes, a DX7 synth where the shortcuts are permanently the DX7 editor pads and there's no filters etc., and a DX7 oscillator that just loads a preset and then has PW and wavetable position as macros to control it?

bfredl:
> I'm good to merge if we put the oscillator type behind a community setting to signify that it's not stable 

Ye. It almost is already. I could make it impossible to select "DX7" type by accident without going through the shortcut to set up the patch properly, which is behind a community setting.

m-m-adams:
> > I'm good to merge if we put the oscillator type behind a community setting to signify that it's not stable
> 
> Ye. It almost is already. I could make it impossible to select "DX7" type by accident without going through the shortcut to set up the patch properly, which is behind a community setting.

Yeah I think that's reasonable

src/deluge/model/settings/runtime_feature_settings.cpp
```diff
@@ -157,6 +157,9 @@ void RuntimeFeatureSettings::init() {
 	// EnableGrainFX
 	SetupOnOffSetting(settings[RuntimeFeatureSettingType::EnableGrainFX], STRING_FOR_COMMUNITY_FEATURE_GRAIN_FX,
 	                  "enableGrainFX", RuntimeFeatureStateToggle::Off);
+	// EnableGrainFX
```
m-m-adams, line 160: nit: dx7

src/deluge/gui/l10n/g_seven_segment.cpp
```diff
@@ -135,6 +135,7 @@ PLACE_SDRAM_DATA Language seven_segment{
         {STRING_FOR_ANALOG_SQUARE, "ASQUARE"},
         {STRING_FOR_ANALOG_SAW, "ASAW"},
         {STRING_FOR_SAMPLE, "SAMP"},
+        {STRING_FOR_DX7, "DX7"},
```
sapphire-arches, line 138: Looks like the source `.json` for these didn't get checked in, I still need to fix the genfile check...

bfredl:
>  put the oscillator type behind a community setting to signify that it's not stable

This is done now. The only way to create a dx7 synth is via the custom shortcut (behind community setting), you cannot select "dx7" type by accident anymore (just like left/right input are not available all the time, so there is some precedent)

lobit12:
sweet!  nice one guys

PR #1119, opened 2024-01-29
loadfw: default sysex delay 2ms, clarify help message, fix pad led progress display

M0R172:


stellar-aria:
I'm going to format this so that merging isn't blocked

stellar-aria (approved):

M0R172:
Ah, thanks! I ran dbt format and it would not change anything...

PR #1125, opened 2024-01-29
Fix mod button popups

soymonitus:
- Make mod button clicks that show current mod encoder modes be an sticky popup until you release the button
- Made all labels consistent between different screens and shorter for better readability
- Fixed popup bug in 7Seg not correctly differentiating between temporary and sticky
- Fixed calculation of sidechain speed in global effectable not taking into account the song resolution


soymonitus:
Ready to review @seangoodvibes

seangoodvibes (approved):

Issue #1135, opened 2024-01-31
Midi Follow / Automation View ~ Bug fix - potential invalid modelStackWithParam pointer, refactor modelStackWithParam to output


seangoodvibes:
### Please describe the problem:

User reported crash associated with invalid modelStackWithPointer

identified "else" condition where the code would attempt to create a modelStackWithParam for an audio clip when the ClipType may not be an audio clip.

### What is the expected behavior?

Always check that the clip type matches the type of modelStackWithParam you're creating

Also, getModelStackWithParam in the automation view class should be factored out to the Output class so that it can be called as a virtual function.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1136, opened 2024-01-31
Bugfix / invalid modelStackWithParam pointer + refactor modelStackWithParam to Output and Song classes

seangoodvibes:
This PR closes issue # https://github.com/SynthstromAudible/DelugeFirmware/issues/1135

- Fixed: identified "else" condition where the code would attempt to create a modelStackWithParam for an audio clip when the ClipType may not be an audio clip.

- Refactored getModelStackWithParam for different output types from Automation View class to the various Output classes.
- Refactored getModelStackWithParam for the song from the Performance View class to the Song class.

Left to do:
- [x] Got a bug with midi follow and getting model stack with param for Synth clip / Kit row. Can't control params .. means I'm getting a null modelStackWithParam...on the hunt for it!

### DIFF
--- a/src/deluge/model/instrument/melodic_instrument.cpp
+++ b/src/deluge/model/instrument/melodic_instrument.cpp
@@ -698,3 +698,24 @@ void MelodicInstrument::polyphonicExpressionEventPossiblyToRecord(ModelStackWith
 
 	expressionValueChangesMustBeDoneSmoothly = false;
 }
+
+ModelStackWithAutoParam* MelodicInstrument::getModelStackWithParam(ModelStackWithTimelineCounter* modelStack,
+                                                                   Clip* clip, int32_t paramID,
+                                                                   deluge::modulation::params::Kind paramKind) {
+	ModelStackWithAutoParam* modelStackWithParam = nullptr;
+
+	ModelStackWithThreeMainThings* modelStackWithThreeMainThings =
+	    modelStack->addOtherTwoThingsButNoNoteRow(toModControllable(), &clip->paramManager);
+
+	if (modelStackWithThreeMainThings) {
+		if (paramKind == deluge::modulation::params::Kind::PATCHED) {
+			modelStackWithParam = modelStackWithThreeMainThings->getPatchedAutoParamFromId(paramID);
+		}
+
+		else if (paramKind == deluge::modulation::params::Kind::UNPATCHED_SOUND) {
+			modelStackWithParam = modelStackWithThreeMainThings->getUnpatchedAutoParamFromId(paramID);
+		}
+	}
+
+	return modelStackWithParam;
+}
--- a/src/deluge/model/output.h
+++ b/src/deluge/model/output.h
@@ -19,6 +19,7 @@
 
 #include "definitions_cxx.hpp"
 #include "model/clip/clip_instance_vector.h"
+#include "modulation/params/param.h"
 #include "util/d_string.h"
 #include <cstdint>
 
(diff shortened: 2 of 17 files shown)


src/deluge/model/output.h
```diff
@@ -146,6 +148,11 @@ class Output {
 	void endArrangementPlayback(Song* song, int32_t actualEndPos, uint32_t timeRemainder);
 	bool recordingInArrangement;
 
+	virtual ModelStackWithAutoParam*
+	getModelStackWithParam(ModelStackWithTimelineCounter* modelStack, Clip* clip,
+	                       int32_t paramID = deluge::modulation::params::kNoParamID,
+	                       deluge::modulation::params::Kind paramKind = deluge::modulation::params::Kind::NONE) = 0;
```
sapphire-arches, line 154:
```suggestion
	                       deluge::modulation::params::ParamType paramID,
	                       deluge::modulation::params::Kind paramKind) = 0;
```

I don't think it ever makes sense to call this without providing a `paramID` and `paramKind` so there's no need for default arguments.

src/deluge/model/song/song.cpp
```diff
@@ -5758,6 +5758,18 @@ ModelStackWithThreeMainThings* Song::addToModelStack(ModelStack* modelStack) {
 	return modelStack->addTimelineCounter(this)->addOtherTwoThingsButNoNoteRow(&globalEffectable, &paramManager);
 }
 
+// get's the model stack for Song level parameters that are being edited
+// used in performance view and in automation arranger view
```
sapphire-arches, line 5762: This documentation should move to `song.h` and use `///` comments so doxygen picks it up. I'd probably write something like this:

Gets a modelstack with the song-global unpatched param `paramID`.

That docstring tells me the most important information about the function:
  - What parameters does it accept (an `UNPATCHED_PARAM_*` paramID)
  - What, as the user of this function, can I expect to be true about its return value (or more generally, the state of the system now that I've called the function).

src/deluge/model/song/song.h
```diff
@@ -333,6 +333,7 @@ class Song final : public TimelineCounter {
 	ModelStackWithThreeMainThings* setupModelStackWithSongAsTimelineCounter(void* memory);
 	ModelStackWithTimelineCounter* setupModelStackWithCurrentClip(void* memory);
 	ModelStackWithThreeMainThings* addToModelStack(ModelStack* modelStack);
+	ModelStackWithAutoParam* getModelStackWithParam(ModelStackWithThreeMainThings* modelStack, int32_t paramID);
```
sapphire-arches, line 336:
```suggestion
	ModelStackWithAutoParam* getModelStackWithParam(ModelStackWithThreeMainThings* modelStack, deluge::modulation::params::ParamType paramID);
```

src/deluge/model/instrument/melodic_instrument.cpp
```diff
@@ -698,3 +698,24 @@ void MelodicInstrument::polyphonicExpressionEventPossiblyToRecord(ModelStackWith
 
 	expressionValueChangesMustBeDoneSmoothly = false;
 }
+
+ModelStackWithAutoParam* MelodicInstrument::getModelStackWithParam(ModelStackWithTimelineCounter* modelStack,
+                                                                   Clip* clip, int32_t paramID,
```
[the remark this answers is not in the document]
seangoodvibes: As mentioned on discord I would like to update these paramID definitions as part of a bigger paramID refactoring PR as there is inconsistency in way more places than this.

PR #1138, opened 2024-01-31
Fix Lumi scale support

soymonitus:
I found that the logic to search for a matching scale for the Lumi keyboard was wrong. If one of the first 7 scales of the Deluge is selected (from Major to Locrian), as the code is searching for transpositions first and all those scales are equivalent, the Lumi keyboard could be highlighting correctly the notes but not the root key, even though the scale is available to choose. So now to fix this:
- First search among all the Deluge scales "as is" without transpositions, if there is a match return it.
- Then if no matches, as a fallback option, search for transpositions of the scales, to at list light up the keys even though the root note is not the same. Besides, the logic was wrong when calculating the notes.

### DIFF
--- a/src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
+++ b/src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
@@ -244,15 +244,26 @@ void MIDIDeviceLumiKeys::setRootNote(int16_t rootNote) {
 // Efficient binary comparison of notes to Lumi builtin scales
 std::pair<MIDIDeviceLumiKeys::Scale, int16_t>
 MIDIDeviceLumiKeys::determineScaleAndRootNoteOffsetFromNotes(uint8_t* modeNotes, uint8_t noteCount) {
+	// First try with all scales "as is" (without transposition)
+	// Turn notes in octave into 12 bit binary
 	uint16_t noteInt = 0;
+	for (uint8_t note = 0; note < noteCount; note++) {
+		noteInt |= 1 << modeNotes[note];
+	}
+	// Compare with Lumis pre-built binary list of scales
+	for (uint8_t scale = 0; scale < MIDI_DEVICE_LUMI_KEYS_SCALE_COUNT; scale++) {
+		if (noteInt == scaleNotes[scale]) {
+			return {Scale(scale), 0};
+		}
+	}
 
-	// Try with all possible transpositions
+	// Then if no matches found, try with all possible transpositions of all the scales
 	for (int32_t i = 0; i < kOctaveSize; i++) {
 		// Turn notes in octave into 12 bit binary
+		noteInt = 0;
 		for (uint8_t note = 0; note < noteCount; note++) {
-			noteInt |= 1 << modeNotes[(note + i) % kOctaveSize];
+			noteInt |= 1 << ((modeNotes[note] + i) % kOctaveSize);
 		}
-
 		// Compare with pre-built binary list of scales
 		for (uint8_t scale = 0; scale < MIDI_DEVICE_LUMI_KEYS_SCALE_COUNT; scale++) {
 			if (noteInt == scaleNotes[scale]) {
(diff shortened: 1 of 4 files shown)


soymonitus:
@litui could you test this PR with your Lumi?

litui:
I can try this out tomorrow.

litui:
Actually, got around to this just now. It seems to work fine.

I'm not entirely confident I see the logic problem needing to be fixed however, as this never came up as an issue in my own testing (the root note doesn't suddenly shift during scale selection, and the lumi remembers the previously selected root... likewise when root is changed, it remembers the previously selected scale).

I'm not inclined to block this change, I just am looking for better understanding of an actual workflow where it appears as a problem.

litui:
I just tried the current community build as well and couldn't trigger any issues in the highlighting logic with my Lumi using the typical scale and root selection.

litui:
I think I just may not be understanding how your transposition code (and the related changes you previously made to this file) fit into the picture. As far as the Lumi highlighting support is concerned it only cares about the root and scale.

soymonitus:
Have you tried the Hirajoshi scale? Root note should be different in deluge and in lumi. Regarding the major modes, have you tried the locrian scale? Does it highlight the ckrrwct root note?

src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
```diff
@@ -244,15 +244,26 @@ void MIDIDeviceLumiKeys::setRootNote(int16_t rootNote) {
 // Efficient binary comparison of notes to Lumi builtin scales
 std::pair<MIDIDeviceLumiKeys::Scale, int16_t>
 MIDIDeviceLumiKeys::determineScaleAndRootNoteOffsetFromNotes(uint8_t* modeNotes, uint8_t noteCount) {
+	// First try with all scales "as is" (without transposition)
+	// Turn notes in octave into 12 bit binary
```
soymonitus, line 248: This piece of code is how it was before my modifications to allow for transpositions. I just recovered it because we first need to check for all scales "as is". Then, afterwards, we can check for transpositions
Example of bug: If we would first check all scales with their possible transpositions (as it was before this PR), when we take the Locrian scale from the Deluge and compare it to the Lumi scales, if searching first for transpositions, we would first find that the Major scale from the Lumi matches the one from the Deluge but with different Root key, and we would not have the opportunity to check the Lumis locrian scale, which is a better match because it matches note intervals and root key.

src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
```diff
@@ -244,15 +244,26 @@ void MIDIDeviceLumiKeys::setRootNote(int16_t rootNote) {
...
+	}
 
-	// Try with all possible transpositions
+	// Then if no matches found, try with all possible transpositions of all the scales
 	for (int32_t i = 0; i < kOctaveSize; i++) {
 		// Turn notes in octave into 12 bit binary
+		noteInt = 0;
 		for (uint8_t note = 0; note < noteCount; note++) {
-			noteInt |= 1 << modeNotes[(note + i) % kOctaveSize];
+			noteInt |= 1 << ((modeNotes[note] + i) % kOctaveSize);
```
soymonitus, line 265: This line was totally wrong. To be able to correctly transpose a scale, we have to add semitones to the modeNotes[note] value, not accessing a different index from the array as it was before. This was a bug.

soymonitus (commented):
@litui Added some explanations of my changes

src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
litui: Ah okay, I get that change now, thank you.

src/deluge/io/midi/device_specific/midi_device_lumi_keys.cpp
litui: Thanks for the explanation. I appreciate it.

litui:
I'll try to find some time to check these scales today.

litui:
Okay, so I've now tried both the pre and post-fix versions and while I see what you're trying to accomplish, I don't really agree with prioritizing the transposed scale over the correct root note.

Based on my own workflows, I would rather have the fallback of chromatic highlighting than have the incorrect root note highlighted. I don't think this decision to prioritize transposition over root note correctness should be made on our end and if it's a choice the end-user has, it should be a configurable option somewhere.

soymonitus:
@litui removed the transposition code. Please, could you review it?

soymonitus:
@litui  ready to go!

litui:
I can't test at the moment, but it looks good to me.

soymonitus:
This just sets the Lumi implementation like it was before the transposition logic I added when I added the new scales. And adds documentation on which scales are directly supported by Lumi. So it should be safe to merge.

m-m-adams (approved):

Issue #1030, opened 2024-01-21
Keyboard in file browser not showing when scrolling with horizontal encode


sichtbeton:
### Please describe the problem:

When in the file browser you scroll with the horizontal encoder, the name bar is shifting. When you then type with the pads, it types but the keyboard is not present.

### What is the expected behavior?

The keyboard should come up.

### Is there a relevant Pull request?

No

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

1.1.0-DEV-85997821

sichtbeton:
https://github.com/SynthstromAudible/DelugeFirmware/assets/62035517/86356844-1216-4b66-98ae-61225223429e

sapphire-arches:
This bug is actually quite old, potentially even going back to the initial OSS release.

sichtbeton:
> This bug is actually quite old, potentially even going back to the initial OSS release.

And obviously didnt bother anyone since :-)

PR #1145, opened 2024-02-01
FIX: Keyboard in file browser not showing when scrolling with horizontal encode #1030

M0R172:
FIX: Keyboard in file browser not showing when scrolling with horizontal encode #1030

I also took the liberty to remove the "draw keyboard on top of waveform/song preview" thing - I'd bet money that nobody will miss it - it makes the keyboard very hard to see, does not provide useful information, and IMHO does not even "look cool". It's just visual clutter and unnecessary code.

This also enabled pulling the "clear pads" and "send pads" into the draw keyboard function, saving a bunch of code duplication.

### DIFF
--- a/src/deluge/gui/ui/qwerty_ui.cpp
+++ b/src/deluge/gui/ui/qwerty_ui.cpp
@@ -53,7 +53,6 @@ bool QwertyUI::opened() {
 	return true;
 }
 
-// Won't "send"
 void QwertyUI::drawKeys() {
 	PadLEDs::clearMainPadsWithoutSending();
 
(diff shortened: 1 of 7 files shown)


src/deluge/gui/ui/qwerty_ui.cpp
```diff
@@ -55,6 +55,7 @@ bool QwertyUI::opened() {
 
 // Won't "send"
```
m-m-adams, line 56: I think this is now wrong

m-m-adams (commented):
Just update the stale comment and good to go. Thanks!

stellar-aria (approved):

Issue #1139, opened 2024-01-31
Fix MIDI Follow Channel A/B/C string for 7seg


seangoodvibes:
### Please describe the problem:

On 7seg the midi follow channels A/B/C are lablelled CHAN/CHAN/CHAN

### What is the expected behavior?

Should be labelled A/B/C instead

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Nightly

Issue #1142, opened 2024-01-31
Performance View: some stutter values are getting cut off on the 7SEG display


baymud:
### Please describe the problem:

On 7SEG, pressing the stutter pads in performance view can result in the value displaying incorrectly, as the 1st digit sometimes gets cut off.

For example, 4ths and 8ths display fine, but the following do not: 
16ths = 6ths
32nds = 2nds
64ths = 4ths

### What is the expected behavior?

For the sake of clarity, the first digit should not be cut off.  16ths could display as 16th, etc. 

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Nightly

PR #1149, opened 2024-02-01
Bugfix / update some 7 seg strings

seangoodvibes:
This closes issues https://github.com/SynthstromAudible/DelugeFirmware/issues/1139 and https://github.com/SynthstromAudible/DelugeFirmware/issues/1142

- Shortened stutter strings in performance view so they're not cut off
- Added 7seg strings for Midi Follow Channel A/B/C menu's


src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -605,13 +605,13 @@ void PerformanceSessionView::renderFXDisplay(params::Kind paramKind, int32_t par
 					buffer = "8ths";
 				}
 				else if (knobPos < 14) { // 16ths stutter: 2 leds turned on
-					buffer = "16ths";
+					buffer = "16th";
```
m-m-adams, line 608: These labels are all incorrect at non default resolutions - this bug was fixed in the main quantized stutter before Amadeus so you should be able to reuse that logic

m-m-adams (approved):

Issue #903, opened 2024-01-01
Finer resolution automation data isn't always visible at lower resolutions


pssssssst:
### Please describe the problem:

When i zoom in in automation view to get a finer resolution (eg from 16th-notes to 32nd-notes) and adjust a parameter, it shows up fine. When i zoom out again (to 16th-notes) the visual indication disappears completely. the automation is still there but you can't see it. Zooming in  makes it appear again

### What is the expected behavior?

You should be able to see the automation data at every zoom level. With note data it stays colored when its exactly on the grid and turns white when its not.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

F7AAA2C

### If possible provide the steps to reproduce the issue and upload additional media:

1. go to automation view
2. choose parameter (for example high pass cut off)
3. click and turn horizontal encoder to get to 32nd-notes resolution
4. change automation in one column
5. zoom out again to 16th-notes
6. colored column disappears

pssssssst:
oh wait. i checked again. its even weirder than that. if i change the automation in the first column (in 32nd-notes) it disappaers. If i change it in the second column it stays

PR #1155, opened 2024-02-01
Bugfix / Automation View - Calculate and Render Average Value on Automation Editor Grid

seangoodvibes:
This addresses issue https://github.com/SynthstromAudible/DelugeFirmware/issues/903
- The issue reported that if you enter automation at a lower zoom level that it does not always show up at higher zoom levels
- This is because I was getting the value from the "middle node" within a column to display automation at higher zoom levels. The middle node may not always have a value if you manually enter automation at lower zoom levels.

To resolve this: 
- I added a calculation function to calculate the average value of all nodes within a grid square position
- I then used that average to render automation on the grid

A follow PR will be opened to address this remaining issue: https://github.com/SynthstromAudible/DelugeFirmware/issues/1172

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -3412,7 +3412,6 @@ uint32_t AutomationView::getMiddlePosFromSquare(int32_t xDisplay, int32_t effect
 }
 
 // calculates value of all nodes within a square for automation editor rendering
-// and for display of square value in pad selection mode
 int32_t AutomationView::getAverageSquareKnobPosition(ModelStackWithAutoParam* modelStack, int32_t xDisplay,
                                                      int32_t effectiveLength, int32_t xScroll, int32_t xZoom) {
 	int32_t squareStart = getPosFromSquare(xDisplay, xScroll, xZoom);
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -686,8 +686,15 @@ void AutomationView::renderRow(ModelStackWithAutoParam* modelStackWithParam, RGB
 
 	for (int32_t xDisplay = 0; xDisplay < kDisplayWidth; xDisplay++) {
 
-		uint32_t squareStart = getMiddlePosFromSquare(xDisplay, lengthToDisplay, xScroll, xZoom);
-		int32_t knobPos = getParameterKnobPos(modelStackWithParam, squareStart) + kKnobPosOffset;
+		int32_t knobPos = 0;
+
+		if (isAutomated) {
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/automation_view.cpp
```diff
@@ -3404,6 +3411,27 @@ uint32_t AutomationView::getMiddlePosFromSquare(int32_t xDisplay, int32_t effect
...
+	int32_t squareStart = getPosFromSquare(xDisplay, xScroll, xZoom);
+	int32_t squareWidth = getSquareWidth(xDisplay, effectiveLength, xScroll, xZoom);
+	int32_t numNodesWithinSquare = squareWidth / kParamNodeWidth;
+
+	int32_t totalKnobPos = 0;
+
+	for (int32_t i = 0; i < numNodesWithinSquare; i++) {
+		int32_t value = modelStack->autoParam->getValuePossiblyAtPos(squareStart + (i * kParamNodeWidth), modelStack);
+		totalKnobPos =
+		    totalKnobPos + modelStack->paramCollection->paramValueToKnobPos(value, modelStack) + kKnobPosOffset;
```
sapphire-arches, line 3427:
```suggestion
		totalKnobPos +=  modelStack->paramCollection->paramValueToKnobPos(value, modelStack);
```

`kNobPosOffset` should be added after the averaging, to reduce the risk of overflow and save us an add operation per loop.

src/deluge/gui/views/automation_view.cpp
```diff
@@ -686,8 +686,15 @@ void AutomationView::renderRow(ModelStackWithAutoParam* modelStackWithParam, RGB
...
+
+		if (isAutomated) {
+			knobPos = getAverageSquareKnobPosition(modelStackWithParam, xDisplay, lengthToDisplay, xScroll, xZoom);
+		}
+		else {
+			uint32_t squareStart = getPosFromSquare(xDisplay, xScroll, xZoom);
+			knobPos = getParameterKnobPos(modelStackWithParam, squareStart) + kKnobPosOffset;
+		}
 
 		RGB& pixel = image[xDisplay];
```
sapphire-arches, line 699: Can we render the min/max knob position in the range with like the note tail color, to give some indication that there is more data lurking below the surface?

seangoodvibes: > Can we render the min/max knob position in the range with like the note tail color, to give some indication that there is more data lurking below the surface?

Can I address this in a follow up PR? I started digging into this but the way I want to show lower data is a somewhat complex undertaking and I don't want to accidentally break this PR or break automation view because it's going to require me playing around with the nodes a bit.

PR #1156, opened 2024-02-01
Enhancement / Automation View - Better Integration of Automation View with Menu System

seangoodvibes:
This PR will:

- [x] add back the settings and sound menu's to automation view
- [x] allow you to jump from the sound menu's for a specific parameter straight into the automation editor for that parameter
- [x] move the interpolation shortcut to the interpolation pad in the first column
- [x] blink interpolation shortcut when interpolation is on 
- [x] disable select parameter scrolling in automation view for non-midi clip automation
- [x] add adjustment of parameter values using select encoder
- [x] test the above!


sapphire-arches:
Maybe that submenu should move to `GLOBAL SETTINGS > DEFAULTS > UI > AUTOMATION` instead? Right now Automation View is the only place where you can't access the global settings which feels odd.

seangoodvibes:
> Maybe that submenu should move to `GLOBAL SETTINGS > DEFAULTS > UI > AUTOMATION` instead? Right now Automation View is the only place where you can't access the global settings which feels odd.

@m-m-adams said we shouldn't nest the automation menu under UI and that it should be a top level automation menu instead.

As for not being able to access global settings, I did think about this.

I think the right fix would be to re-instate shift + select to bring us into settings
and then just pressing on select encoder would open the automation menu

this would necessitate changing the interpolation shortcut which is a good idea anyway. I want to use shift + interpolation and blink the interpolation shortcut to indicate that interpolation is on.

sapphire-arches:
That sounds good to me, I think.

src/deluge/gui/menu_item/patched_param.h
```diff
@@ -31,6 +31,7 @@ class PatchedParam : public Param, public MenuItemWithCCLearning {
 	PatchedParam() = default;
 	PatchedParam(int32_t newP) : Param(newP) {}
 	MenuItem* selectButtonPress();
+	ActionResult buttonAction(deluge::hid::Button b, bool on) { return Param::buttonAction(b, on); }
```
m-m-adams, line 34: I think this should just pass through inheritance?

src/deluge/gui/menu_item/patched_param/integer.h
```diff
@@ -37,6 +37,7 @@ class Integer : public PatchedParam, public menu_item::IntegerContinuous {
 
 	uint8_t shouldDrawDotOnName() final { return PatchedParam::shouldDrawDotOnName(); }
 	MenuItem* selectButtonPress() final { return PatchedParam::selectButtonPress(); }
+	ActionResult buttonAction(deluge::hid::Button b, bool on) final { return PatchedParam::buttonAction(b, on); }
```
m-m-adams, line 40: same here

src/deluge/gui/menu_item/unpatched_param.h
```diff
@@ -38,6 +38,7 @@ class UnpatchedParam : public Param, public IntegerContinuous, public MenuItemWi
 	[[nodiscard]] int32_t getMaxValue() const override { return Param::getMaxValue(); }
 	[[nodiscard]] int32_t getMinValue() const override { return Param::getMinValue(); }
 	MenuItem* selectButtonPress() final { return Param::selectButtonPress(); }
+	ActionResult buttonAction(deluge::hid::Button b, bool on) final { return Param::buttonAction(b, on); }
```
m-m-adams, line 41: also here

src/deluge/gui/menu_item/patched_param.h
seangoodvibes: > I think this should just pass through inheritance?

I tried removing the three button action calls and it did not work anymore.

I treated it the same as the selectButtonAction function which is also added to all three of these classes which inherit param.

m-m-adams (approved):
LGTM

PR #1159, opened 2024-02-02
Feature / Transpose Chromatically All Scale Mode Instrument Clips (excl. Kits) from Song View's (Song / Arranger / Arranger Automation / Performance)

seangoodvibes:
Added ability to chromatically transpose all scale mode instrument clips (excluding kits) from all the various song views (Song Row/Grid, Arranger, Arranger Automation, Performance View)

Uses the same shortcut as when you're in a Synth/Midi/CV clip (Press and turn <>) to transpose the clips chromatically by +/- # of semitones (customizable, see below)

The display will show you the current root note and scale (before and after transposition):

image

You can adjust the number of semitones offset that the master transpose will transpose clips by, by holding shift and turning <>. The display will show you the transpose offset.

image


M0R172:
Beautiful!

Ideas for the future, based on this:

1. transpose in-scale (does it already?)
2. automizable (is it already?)
3. automizable by midi notes relative to some root note
4. Now we can instant-harmonize a song by external keyboard or midi track... (I think there is an issue/pr somewhere suggesting this, it's one of the most raved-about features of the oxi one sequencer)
5. Profit!
6. Maybe extend to audio, as the Deluge can easily do this, IIRC (I hardly use audio)

Just some brainstorming here, the new shortcut is definitely great on it's own, even more so as it is super-consistent with current behaviour!

soymonitus:
When transposing.. say 1 semitone, will this just change root note and not scale? I mean, if you have your clips in C Major scale, and you turn up 1 notch, will the scale now be C# major?

seangoodvibes:
> When transposing.. say 1 semitone, will this just change root note and not scale? I mean, if you have your clips in C Major scale, and you turn up 1 notch, will the scale now be C# major?

Correct this does not change the scale. It changes the root note (eg C major to C# major).

src/deluge/model/song/song.cpp
```diff
@@ -5732,6 +5750,64 @@ int32_t Song::addInstrumentsToFileItems(OutputType outputType) {
...
+		masterTransposeOffset = 0;
+	}
+	displayMasterTransposeOffset();
+}
+
+void Song::displayMasterTransposeOffset() {
+	DEF_STACK_STRING_BUF(popupMsg, 40);
+
+	if (display->haveOLED()) {
+		popupMsg.append("Transpose Offset: \n");
```
m-m-adams, line 5791: should be - "Transpose Interval: " rather than offset. Offset sounds like they should be added, interval is clearer that you're changing the interval that the transposer will adjust by

Same applies to the names in the code

m-m-adams (commented):
Looks good! One UI nit

m-m-adams (approved):

Issue #1061, opened 2024-01-24
Holding an audition pad and scrolling with vertical encoder will make notes hang in Automation View


ok-reza:
### Please describe the problem:

Notes play indefinitely when you are in Automation View and press and hold an audition pad and scroll with the vertical encoder. Notes can be stopped if you navigate to any hung notes and press it again.

This behavior doesn't occur in normal clip view.

### What is the expected behavior?

Notes properly cut off when they are no longer pressed in Automation View

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

1.1.0 - 1AD7933 (1/24)

PR #1160, opened 2024-02-02
FIX: Holding an audition pad and scrolling with vertical encoder will make notes hang in Automation View #1061

M0R172:
FIX: Holding an audition pad and scrolling with vertical encoder will make notes hang in Automation View #1061


seangoodvibes:
Well that was an easy fix eh? Thanks for doing this! Just tested and works perfect.

seangoodvibes (approved):

PR #1163, opened 2024-02-03
Move compressor into global effectable

m-m-adams:
First step of per clip comps - editing not yet possible but this finished the architectural changes

### DIFF
--- a/src/deluge/dsp/compressor/rms_feedback.cpp
+++ b/src/deluge/dsp/compressor/rms_feedback.cpp
@@ -29,13 +29,13 @@ RMSFeedbackCompressor::RMSFeedbackCompressor() {
 void RMSFeedbackCompressor::updateER(float numSamples, q31_t finalVolume) {
 
 	// int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
-	float songVolume = logf(finalVolume) - 2;
+	float songVolumedB = logf(finalVolume) - 2;
 
-	threshdb = songVolume * threshold;
+	threshdb = songVolumedB * threshold;
 	// this is effectively where song volume gets applied, so we'll stick an IIR filter (e.g. the envelope) here to
 	// reduce clicking
 	float lastER = er;
-	er = std::max<float>((songVolume - threshdb - 1) * ratio, 0);
+	er = std::max<float>((songVolumedB - threshdb - 1) * ratio, 0);
 	// using the envelope is convenient since it means makeup gain and compression amount change at the same rate
 	er = runEnvelope(lastER, er, numSamples);
 }
--- a/src/deluge/dsp/compressor/rms_feedback.cpp
+++ b/src/deluge/dsp/compressor/rms_feedback.cpp
@@ -18,25 +18,18 @@
 #include "dsp/compressor/rms_feedback.h"
 
 RMSFeedbackCompressor::RMSFeedbackCompressor() {
-
-	// an appropriate range is 0-50*one q 15
-
-	thresholdKnobPos = 0;
-	sideChainKnobPos = ONE_Q31 >> 1;
-	// this is 2:1
-	ratioKnobPos = 0;
-
-	currentVolumeL = 0;
-	currentVolumeR = 0;
-	er = 0;
-	setSidechain(sideChainKnobPos);
+	setAttack(0);
+	setRelease(0);
+	setThreshold(0);
+	setRatio(0);
+	setSidechain(0);
 }
 // 16 is ln(1<<24) - 1, i.e. where we start clipping
 // since this applies to output
 void RMSFeedbackCompressor::updateER(float numSamples, q31_t finalVolume) {
 
 	// int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
-	float songVolume = std::log(finalVolume) - 2;
+	float songVolume = logf(finalVolume) - 2;
 
 	threshdb = songVolume * threshold;
 	// this is effectively where song volume gets applied, so we'll stick an IIR filter (e.g. the envelope) here to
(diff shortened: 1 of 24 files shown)


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1288,11 +1288,20 @@ void ModControllableAudio::writeTagsToFile() {
 	storageManager.closeTag();
 
 	// Sidechain compressor
-	storageManager.writeOpeningTagBeginning("compressor");
-	storageManager.writeSyncTypeToFile(currentSong, "syncType", compressor.syncType);
-	storageManager.writeAbsoluteSyncLevelToFile(currentSong, "syncLevel", compressor.syncLevel);
-	storageManager.writeAttribute("attack", compressor.attack);
-	storageManager.writeAttribute("release", compressor.release);
+	storageManager.writeOpeningTagBeginning("Sidechain");
```
sapphire-arches, line 1291: Generally the tags are lower `pascalCase`

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1503,33 +1512,68 @@ int32_t ModControllableAudio::readTagFromFile(char const* tagName, ParamManagerF
...
+				storageManager.exitTag("compHPF");
+			}
+			else {
+				storageManager.exitTag(tagName);
+			}
+		}
+		storageManager.exitTag("AudioCompressor");
+	}
+	// this is actually the sidechain but pre 1.1 songs save it as compressor
+	else if (!strcmp(tagName, "compressor") || !strcmp(tagName, "Sidechain")) { // Remember, Song doesn't use this
```
sapphire-arches, line 1549: Is the  `Remember, Song doesn't use this` comment leftover from when the code was originally copied from somewhere else? It seems like song does in fact use this.

src/deluge/model/mod_controllable/mod_controllable_audio.h
```diff
@@ -159,7 +162,7 @@ class ModControllableAudio : public ModControllable {
 	StereoSample grabbedSample;
 	StereoSample lastGrabbedSample;
 
-	SideChain compressor; // Song doesn't use this, despite extending this class
+	SideChain sidechain; // Song doesn't use this, despite extending this class
```
sapphire-arches, line 165: Similarly I think `Song` does use this now?

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -18,25 +18,18 @@
...
+	setRatio(0);
+	setSidechain(ONE_Q31 >> 1);
 }
 // 16 is ln(1<<24) - 1, i.e. where we start clipping
 // since this applies to output
 void RMSFeedbackCompressor::updateER(float numSamples, q31_t finalVolume) {
 
 	// int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
-	float songVolume = std::log(finalVolume) - 2;
+	float songVolume = logf(finalVolume) - 2;
```
sapphire-arches, line 32: Should this be called `volumeDB` or something instead?

src/deluge/dsp/compressor/rms_feedback.h
```diff
@@ -20,48 +20,56 @@
...
 #include "dsp/stereo_sample.h"
+#include "util/fixedpoint.h"
 #include <cmath>
 
 class RMSFeedbackCompressor {
 public:
 	RMSFeedbackCompressor();
-	void setup(q31_t attack, q31_t release, q31_t threshold, q31_t ratio, q31_t sidechain_fc);
 
+	// takes in knob positions in the range 0-ONE_Q31
```
sapphire-arches, line 30: Which 2 parameters are the knob positions?

Also, `///` so doxygen picks it up <3

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
m-m-adams: It's an old comment but I'm still pretty sure it doesn't - it reads the tags because it inherits from mod controllable but never actually calls the sidechain comp

src/deluge/dsp/compressor/rms_feedback.h
m-m-adams: All of them, but yeah I'll fix that

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
sapphire-arches: It does use the `compressor` parameters though right?

src/deluge/model/mod_controllable/mod_controllable_audio.h
sapphire-arches: I was confusing this with the compressor, which I'm pretty sure the song does use.

src/deluge/model/song/song.cpp
```diff
@@ -1667,27 +1647,32 @@ int32_t Song::readFromFile() {
 				lastSelectedParamArrayPosition = storageManager.readTagOrAttributeValueInt();
 				storageManager.exitTag("lastSelectedParamArrayPosition");
 			}
-
+			// legacy section, read as part of global effectable (songParams tag) post 1.1
```
sapphire-arches, line 1650: Maybe use `c1.1` here to distinguish it from Official 1.1

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
m-m-adams: Yeah this is why I renamed the parameter - song uses the real compressor, but not the sidechain envelope (formerly named compressor in the code)

Issue #1147, opened 2024-02-01
Update currentSong->currentClip after Selecting a Clip


seangoodvibes:
### Please describe the problem:

After a selecting a clip in arranger or song view, the clip selection goes back to the previous clip you entered

This relates to PR #973

### What is the expected behavior?

Whether you enter a clip or select a clip in song/arranger view, the clip selection (stored in currentSong->currentClip) should reflect the last clip interacted with.

### Is there a relevant Pull request?

https://github.com/SynthstromAudible/DelugeFirmware/pull/973

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

Built from community

seangoodvibes:
FYI @cowboy

seangoodvibes:
> Does this actually relate to any code in #970 or just the comments there that led to the creation of #973?

Ah maybe doesn't relate to 970, just saw the comments there. So more related to 973.

PR #1167, opened 2024-02-03
Input Needed! Update currentSong->currentClip after Selecting a Clip #1147

M0R172:
Does this fix Issue #1147? I'm not completely sure...

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -709,7 +709,6 @@ holdingRecord:
 
 				// If Clip present here...
 				else if (clip) {
-					currentSong->currentClip = clip;
 
 					// If holding down tempo knob...
 					if (Buttons::isButtonPressed(deluge::hid::button::TEMPO_ENC)) {
@@ -729,7 +728,6 @@ removePendingOverdub:
 
 					// Or, normal action - select the pressed Clip
 					else {
-
 						selectedClipYDisplay = yDisplay;
 						// This is only interresting for changing colour
 						clipWasSelectedWithShift = Buttons::isShiftButtonPressed();
@@ -871,6 +869,8 @@ justEndClipPress:
 							                                                     // selected,
 						}
 						// and we don't want the loading animation or anything to get stuck onscreen
+						currentSong->currentClip = clip;
+						display->displayPopup("NEW CURRENT");
 						clipPressEnded();
 					}
 				}
@@ -921,7 +921,6 @@ justEndClipPress:
 
 				// If Clip is present here
 				if (clip) {
-					currentSong->currentClip = clip;
 					return view.clipStatusPadAction(clip, on, yDisplay);
 				}
 			}
@@ -937,7 +936,6 @@ justEndClipPress:
 
 				// If Clip is present here
 				if (clip) {
-					currentSong->currentClip = clip;
 					switch (currentUIMode) {
 					case UI_MODE_MIDI_LEARN:
 						if (sdRoutineLock) {
@@ -3323,9 +3321,7 @@ ActionResult SessionView::gridHandlePads(int32_t x, int32_t y, int32_t on) {
 	}
 	else {
 		gridActiveModeUsed = true;
-
 		Clip* clip = gridClipFromCoords(x, y);
-		currentSong->currentClip = clip;
 		ActionResult modeHandleResult = ActionResult::NOT_DEALT_WITH;
 		switch (gridModeActive) {
 		case SessionGridModeEdit: {
@@ -3453,21 +3449,26 @@ ActionResult SessionView::gridHandlePadsEdit(int32_t x, int32_t y, int32_t on, C
 		// First finger up
 		if (gridFirstPressedX == x && gridFirstPressedY == y) {
 			// Open clip if no other pad was previously pressed, timer has not run out and clip is pressed
-			if (isUIModeActive(UI_MODE_CLIP_PRESSED_IN_SONG_VIEW) && performActionOnPadRelease
-			    && AudioEngine::audioSampleTimer - selectedClipTimePressed < kShortPressTime) {
+			if (isUIModeActive(UI_MODE_CLIP_PRESSED_IN_SONG_VIEW) && performActionOnPadRelease) {
+				if (AudioEngine::audioSampleTimer - selectedClipTimePressed < kShortPressTime) {
 
-				// Not allowed if recording arrangement
-				if (playbackHandler.recording == RecordingMode::ARRANGEMENT) {
-					display->displayPopup(deluge::l10n::get(deluge::l10n::String::STRING_FOR_RECORDING_TO_ARRANGEMENT));
+					// Not allowed if recording arrangement
+					if (playbackHandler.recording == RecordingMode::ARRANGEMENT) {
+						display->displayPopup(
+						    deluge::l10n::get(deluge::l10n::String::STRING_FOR_RECORDING_TO_ARRANGEMENT));
+					}
+					else {
+						if (clip != nullptr) {
+							transitionToViewForClip(clip);
+						}
+						return ActionResult::ACTIONED_AND_CAUSED_CHANGE;
+					}
 				}
 				else {
-					if (clip != nullptr) {
-						transitionToViewForClip(clip);
-					}
-					return ActionResult::ACTIONED_AND_CAUSED_CHANGE;
+					currentSong->currentClip = clip;
+					display->displayPopup("NEW CURRENT");
 				}
 			}
-
 			clipPressEnded();
 		}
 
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -709,6 +709,7 @@ ActionResult SessionView::padAction(int32_t xDisplay, int32_t yDisplay, int32_t
 
 				// If Clip present here...
 				else if (clip) {
+					currentSong->currentClip = clip;
```
m-m-adams, line 712: I think this belongs around line 734 and 778, where selectedClipYDisplay is set - otherwise it could select a clip that's being deleted

src/deluge/gui/views/session_view.cpp
```diff
@@ -920,7 +921,7 @@ ActionResult SessionView::padAction(int32_t xDisplay, int32_t yDisplay, int32_t
 
 				// If Clip is present here
 				if (clip) {
-
+					currentSong->currentClip = clip;
```
m-m-adams, line 924: I don't know that this should select clips - remember the purpose of currentClip is to control the target for looper footswitches and midi follow. Muting/launching a clip might not be a great indication that the user wants to select it

src/deluge/gui/views/session_view.cpp
```diff
@@ -936,7 +937,7 @@ ActionResult SessionView::padAction(int32_t xDisplay, int32_t yDisplay, int32_t
...
```
m-m-adams, line 940: similar here - not sure this is actually going to be a clip we want to select

m-m-adams (commented):
As far as closing the issue goes - the grid one is good and the row one within the clip area is good, but I'm not sure about the mute and audition pads in rows mode. My gut feel is that they shouldn't select a clip since it's a less deliberate action than pressing the clip is? I could go either way though

Code wise just a minor fix in the row current clip addition - it's current location is above a check that could delete the clip, so it needs to be moved down to where get selectedClipYDisplay variable is set

M0R172:
Agree with your points. I admit that I was quite unclear about why and when we really want to change the currentClip. It just seemed technically easy to fix, so I thought I could shorten the backlog...

Now I discovered that there actually is a function

`setActiveModControllableTimelineCounter`

which seems to be setting midi follow (does that include footsswitches? logically it should, IMO) -
so if that is the motivation, maybe

- setting `currentClip` is not necessary (but maybe still useful)?
- `setActiveModControllableTimelineCounter` needs to be called instead/additionally?

So I guess there is some more research needed in that direction, and maybe a discussion of when the "current" clip *should* actually change - 
If I was jamming on some channel with my midi controller, I guess I would like the bar set quite high for when to change the target channel, so probably not on launching or muting some other track... 
OTOH, it should be painless if I *want* to switch channels - seems like a bit of a trade-off...

So, we basically a need list of all the things you can do to a row/clip/column in song/arranger/grid mode, and then for each item decide if that should change the "current" and/or "midi-follow" target... input welcome!

seangoodvibes:
Hey thanks for tackling this.

As far as midi follow goes, that "setactivemodcontrollabletimeline" function is for triggering the sending of midi feedback.

The current clip problem doesn't come into play with that function.

Where the current clip problem comes into play with midi follow is where right now into order to play a clip, you need either enter the clip or press and hold the clip in arranger / song.

We want to improve that by allowing you to select the clip without being able to enter it so you could play it without having to enter it.

This means, if you press and hold on a clip in arranger or song view and then let go, that you could then play that clip (midi follow would send the notes there).

For sending notes midi follow uses currentSong->currentClip, so if you want your notes to go through to a specific clip, that clip needs to be set as the current clip.

m-m-adams:
Yeah - I think the changes to set current clip when a clip pad is held are all we need

M0R172:
Here's a version that mostly does what I think is correct. To help with testing, it displays a popup when currentClip is set. 

What I am not completely happy with is the fact that it also triggers when *multiple* pads are pressed and released again, even if some action results, like cloning rows in song mode - but I don't see an easy way to block these cases. Do you think this needs to be fixed, or is this acceptable or even good?

Feedback welcome!

M0R172:
One special case that I definitely think should be handled:

I really like the ability to hold a clip and quickly change the mod button parameters without entering a clip - but I think doing this should not change the active clip... Opinions?

seangoodvibes:
> One special case that I definitely think should be handled:
> 
> I really like the ability to hold a clip and quickly change the mod button parameters without entering a clip - but I think doing this should not change the active clip... Opinions?

I don't think we have a way of distinguishing between whether you're pressing to change mod parameters or just pressing to select it.

So I think pressing it should change the active clip. Why do you not want the active clip to change?

M0R172:
Hmm, I thought about this scenario:

I am jamming on track/clip A with my midi keyboard, notice track B being a bit loud, quickly adjust B a bit down using the press+modencoder method, and then having to change active track back to A to continue jamming...

OTOH one could well argue that midi follow on the active clip is more for convenience in quick experiments in the studio and for performing one should have an actual fixed midi configuration. 

Plus, you are correct - there is no really good way to discern this, other than checking if the encoders were turned before releasing the pad, but this would add more complexity and create an inconsistency...

**Bottom line:**
If people like the current state, I am happy to leave it like this for now (it is better than before, and further tweaks could always be made later if deemed necessary).

seangoodvibes:
> Hmm, I thought about this scenario:
> 
> I am jamming on track/clip A with my midi keyboard, notice track B being a bit loud, quickly adjust B a bit down using the press+modencoder method, and then having to change active track back to A to continue jamming...
> 
> OTOH one could well argue that midi follow on the active clip is more for convenience in quick experiments in the studio and for performing one should have an actual fixed midi configuration.
> 
> Plus, you are correct - there is no really good way to discern this, other than checking if the encoders were turned before releasing the pad, but this would add more complexity and create an inconsistency...
> 
> **Bottom line:** If people like the current state, I am happy to leave it like this for now (it is better than before, and further tweaks could always be made later if deemed necessary).

Well I think midi follow is really good for live performing as well so we need to keep your use case in mind. We're a bit stuck here.

M0R172:
Just to put the thought out here, not necessarily advocating this:

What if changing the midi follow/ current clip would require a shortcut, e.g. "hold learn"+"press clip". This would still be much quicker than entering and exiting a clip , but would hardly be done by accident (I think doing it by accident would suck in a live scenario) - one could even make a user setting to decide if mere selection, or the shortcut are required, depending on personal priority: Easy change or safe against accidental change...

seangoodvibes:
> Just to put the thought out here, not necessarily advocating this:
> 
> What if changing the midi follow/ current clip would require a shortcut, e.g. "hold learn"+"press clip". This would still be much quicker than entering and exiting a clip , but would hardly be done by accident (I think doing it by accident would suck in a live scenario) - one could even make a user setting to decide if mere selection, or the shortcut are required, depending on personal priority: Easy change or safe against accidental change...

@m-m-adams I think this is a good suggestion about being intentional about wanting to change the currentClip, especially with the midi follow use case and wanting to make sure you're not accidentally sending notes to the wrong clip.

I also think it would be good to highlight if possible the active clip in grid/row view so you know exactly what clip is currently selected for midi follow and other use cases of currentClip (looping?).

M0R172:
"blurring" the whole thing might be a bit strong, but I agree that some kind of visual cue about the active track could be helpful!

soymonitus:
What about holding Learn? Right now even in Song view, the pads flash white to indicate that a clip is midi learned to an external midi controller. What about flashing the row in pale blue for the currently selected midi follow clip?

soymonitus:
One other topic:

someone mentioned the other day that he had a midi clip that played an external synth, which was connected through line in, and that he wanted to create an audii clip to record what he was playing in his external midi keyboard, using midi follow mode through that Midi clip he had created. The problem he had was that when creating an audio clip to loop the incoming external synth audio, suddenly no notes passed to the midi clip and he could not play it and at the same time record it
i found that the issue happens because beginLinearRecording method sets the currentClip, like if you had entered it:
https://github.com/SynthstromAudible/DelugeFirmware/blob/d4060c1b246537ed9902aec2fdbc0330ae660799/src/deluge/model/clip/clip.cpp#L1026

and since midi_follow.cpp asks for the getCurrentClip if no clip is pressed, then the audio clip becomes the clip focused for midi follow
i saw that when midi notes, pitchbend or aftertouch is received the getSelectedClip method is called with useActiveClip true
https://github.com/SynthstromAudible/DelugeFirmware/blob/d4060c1b246537ed9902aec2fdbc0330ae660799/src/deluge/io/midi/midi_follow.cpp#L151

what would you think if we, instead of recurring to getCurrentClip, we have an extra variable in Song next to currentClip, which is lastFocusedPlayableClip (or lastFocusedInstrumentClip) (whatever is named to say that the clip can receive midi notes: synth, midi, cv or kit)
then lastFocusedPlayableClip is always set to the same as currentClip, in the places of the code where it is set, BUT only if the clip to be set is not an audioclip

Issue #1153, opened 2024-02-01
resetting Deluge to factory default settings by holding Select and powering on doesn't work


seangoodvibes:
### Please describe the problem:

Can't reset the Deluge to factory defaults by holding down the Select encoder and powering on the Deluge

### What is the expected behavior?

You should be able to factory reset the Deluge by holding down Select and powering on the Deluge.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

m-m-adams:
https://github.com/m-m-adams/DelugeFirmware/blob/5c6e3812687db63d8d22cc360741fa9a2a85f07c/src/deluge/deluge.cpp#L631 

Problem appears to be here - I don't think it actually gets hit

M0R172:
Confirmed by print debugging: The case gets hit, it just never `looksOk`... 

Without the `looksOk` check, any combination of presses that includes the select button triggers factory reset...

PR #1169, opened 2024-02-04
FIX: resetting Deluge to factory default settings by holding Select and powering on doesn't work #1153

M0R172:
FIX: resetting Deluge to factory default settings by holding Select and powering on doesn't work #1153

NOTE: This fixes the select encoder being ignored due to another (currently unknown) PIC event message (129) happening at boot.

It still only factory resets the settings that are on the official deluge, i.e. no community settings. Changing that would be a technically unrelated issue.


src/deluge/drivers/pic/pic.h
```diff
@@ -77,6 +77,9 @@ class PIC {
 	 */
 	enum class Response : uint8_t { // Technically matches Message except in some spots????
 		NONE = 0,
+
+		WHATEVER_THIS_IS = 129,
```
sapphire-arches, line 81:
```suggestion
		UNKNOWN_BOOT_RESPONSE = 129,
```

sapphire-arches (commented):
minor quibble on the name but otherwise LGTM

PR #1171, opened 2024-02-05
toolchain: fix setup on windows

sapphire-arches:
This was broken because I assumed that the code in the `if` statement `true`/`false` blocks was variable-expanded one line at a time, but it turns out CMD expands the whole expression as it parses the if.

### DIFF
--- a/dbt.cmd
+++ b/dbt.cmd
@@ -1,19 +1,17 @@
 @echo off
-call "%~dp0scripts\toolchain\dbtenv.cmd" env
+
 set "SCRIPT_PATH=%DBT_ROOT%"
+call "%~dp0scripts\toolchain\dbtenv.cmd" env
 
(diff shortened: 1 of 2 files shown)


dbt.cmd
```diff
@@ -1,19 +1,17 @@
 @echo off
-call "%~dp0scripts\toolchain\dbtenv.cmd" env
+
```
stellar-aria, line 2: DBT_ROOT gets set from dbtenv.cmd, so all uses of SCRIPT_PATH in dbtenv.cmd should just be DBT_ROOT instead, and this won't need to be switched

m-m-adams: I think we can just leave this as future work, it's better to have something working but inefficient for anyone who tries to download and setup on windows at the moment

PR #1174, opened 2024-02-05
BugFix isRelevant flag for GlobalEffectable ModFXType params in Global Menu's

seangoodvibes:
Global effectible Mod FX menu's weren't using the isRelevant flag to make the Mod FX params listed conditional on Mod FX type selected.

Upon investigation, I found:

- [x] we were missing a Depth menu for the unpatched global effectable mod fx param. This meant that the Depth param didn't have an isRelevant check in the Global Menu's To fix this I added an unpatched menu class for the Depth param.
- [x] the isRelevant check was only working for sound params and not global effectable params. So I replaced sound with a new function I added to the sound and global effectable classes, and accessed through the currentModControllable class (ModControllableAudio) as a virtual function.
- [x] Refactored isRelevant and checkPermissionToBeginSession menu functions to receive a ModControllableAudio* instead of a Sound* and then cast the modControllable to a Sound where the sound pointer was still needed.

I think this is what Rohan originally wanted when he wrote "// TODO: really want to receive a ModControllableAudio here!"

### DIFF
--- a/src/deluge/gui/menu_item/osc/sync.h
+++ b/src/deluge/gui/menu_item/osc/sync.h
@@ -25,7 +25,8 @@ public:
 	using Toggle::Toggle;
 	void readCurrentValue() override { this->setValue(soundEditor.currentSound->oscillatorSync); }
 	void writeCurrentValue() override { soundEditor.currentSound->oscillatorSync = this->getValue(); }
-	bool isRelevant(Sound* sound, int32_t whichThing) override {
+	bool isRelevant(ModControllableAudio* modControllable, int32_t whichThing) override {
+		Sound* sound = static_cast<Sound*>(modControllable);
 		return (whichThing == 1 && sound->synthMode != SynthMode::FM && sound->sources[0].oscType != OscType::SAMPLE
 		        && sound->sources[1].oscType != OscType::SAMPLE);
 	}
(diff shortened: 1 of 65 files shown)


m-m-adams:
Can you change the function signature to receive a modcontrollable instead of a sound? Right now it's passed in a sound and pulls the modcontrollable from the global soundEditor context instead, which is not great

seangoodvibes:
> Can you change the function signature to receive a modcontrollable instead of a sound? Right now it's passed in a sound and pulls the modcontrollable from the global soundEditor context instead, which is not great

done

src/deluge/gui/menu_item/osc/sync.h
```diff
@@ -25,7 +25,8 @@ class Sync final : public Toggle {
 	using Toggle::Toggle;
 	void readCurrentValue() override { this->setValue(soundEditor.currentSound->oscillatorSync); }
 	void writeCurrentValue() override { soundEditor.currentSound->oscillatorSync = this->getValue(); }
-	bool isRelevant(Sound* sound, int32_t whichThing) override {
+	bool isRelevant(ModControllableAudio* modControllable, int32_t whichThing) override {
+		Sound* sound = static_cast<Sound*>(modControllable);
```
m-m-adams, line 29: I don't think these are valid - the mod controllable might not be a sound now, so that needs to be checked

m-m-adams: Hm actually probably fine since you can't reach these leaf menus if you're not editing a sound

m-m-adams (approved):

PR #1175, opened 2024-02-05
Refactor "Naming" of Sidechain Compressor to Sidechain

seangoodvibes:
Refactored references to the word Compressor when used in the Sidechain context to just use the word Sidechain. This will help avoid confusion with the Master Compressor.

This also includes renaming the Sidechain Compressor menu to just say Sidechain

**Only thing not covered by this PR:** there are still references to compressor in the file name strings. I didn't touch them because I didn't want to break file compatibility.


src/deluge/gui/menu_item/patch_cable_strength.h
```diff
@@ -31,7 +31,7 @@ class PatchCableStrength : public Decimal, public MenuItemWithCCLearning {
 	[[nodiscard]] int32_t getMaxValue() const final { return kMaxMenuPatchCableValue; }
 	[[nodiscard]] int32_t getNumDecimalPlaces() const final { return 2; }
 	virtual int32_t getDefaultEditPos() { return 2; }
-	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange) override;
+	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange);
 	virtual ParamDescriptor getDestinationDescriptor() = 0;
```
seangoodvibes, line 35: not sure why but it wouldn't compile unless I removed override from this function...

src/deluge/gui/menu_item/patch_cable_strength/fixed.h
```diff
@@ -25,7 +25,7 @@ class Fixed : public Regular {
 	Fixed(l10n::String newName, int32_t newP = 0, PatchSource newS = PatchSource::LFO_GLOBAL)
 	    : Regular(newName), p(newP), s(newS) {}
 
-	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange) final;
+	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange);
```
seangoodvibes, line 28: same thing here, had to remove final to get it to compile...

src/deluge/gui/menu_item/patch_cable_strength/regular.h
```diff
@@ -27,7 +27,7 @@ class Regular : public PatchCableStrength {
 	ParamDescriptor getDestinationDescriptor() final;
 	PatchSource getS() final;
 	ParamDescriptor getLearningThing() final;
-	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange) override;
+	MenuPermission checkPermissionToBeginSession(Sound* sound, int32_t whichThing, MultiRange** currentRange);
```
seangoodvibes, line 30: and had to remove override here too for same reason

m-m-adams (commented):
You need to add the overrides back in and fix the underlying problem. There's two different classes named MultiRange (one in menu item, one in storage) and the change in include order in menus.cpp changes which one gets used. I believe the one we want in checkPermissionToBeginSession is the storage version but I'm not actually sure, so you'll need to investigate that

m-m-adams (approved):

PR #1178, opened 2024-02-05
Refactor/get set api currentclip

### LINKED ISSUE #1147: Update currentSong->currentClip after Selecting a Clip

m-m-adams:
Refactor currentClip to use get/set api

Sets currentClip on following actions:
- pressing a row in rows mode
- pressing a square in selection mode in grid
- creating a clip in grid or row mode, including the create+arm for recording workflow

Existing behaviour of setting current clip on entering clip level views is redundant but retained. Otherwise a future code change allowing entering a clip without one of the above actions would break clip view logic.

Fix #1147 

I believe this makes #1167 and #989 redundant

### DIFF
--- a/src/deluge/deluge.cpp
+++ b/src/deluge/deluge.cpp
@@ -387,14 +387,14 @@ bool readButtonsAndPads() {
 void setUIForLoadedSong(Song* song) {
 
 	UI* newUI;
-
+	Clip* currentClip = currentSong->getCurrentClip();
 	// If in a Clip-minder view
-	if (getCurrentClip() && song->inClipMinderViewOnLoad) {
-		if (getCurrentClip()->onAutomationClipView) {
+	if (currentClip && song->inClipMinderViewOnLoad) {
+		if (currentClip->onAutomationClipView) {
 			newUI = &automationView;
 		}
-		else if (getCurrentClip()->type == ClipType::INSTRUMENT) {
-			if (getCurrentInstrumentClip()->onKeyboardScreen) {
+		else if (currentClip->type == ClipType::INSTRUMENT) {
+			if (((InstrumentClip*)currentClip)->onKeyboardScreen) {
 				newUI = &keyboardScreen;
 			}
 			else {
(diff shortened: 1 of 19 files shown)


soymonitus:
And also makes https://github.com/SynthstromAudible/DelugeFirmware/pull/973 redundant

soymonitus:
There is still one issue to fix. If you are in Green mode in Grid. When you hold down a clip for a second, it allows you to edit it and it is supposed to also change the current clip but it doesn't. if only does it while in Blue mode. But I have it set to always Green, and Blue mode is temporary. It should also set currentclip when the action is by holding it. 

His latest commit in hist branch achieved this by checking  AudioEngine::audioSampleTimer - pressTime) >= kShortPressTime
https://github.com/SynthstromAudible/DelugeFirmware/pull/1167/commits/eb411290b522f7dc92de6379703b6ea48560ac36

m-m-adams:
What grid settings are you using? I'm not seeing that behaviour. In green mode it just launches for me and doesn't allow me to edit the clip

M0R172:
I assume he means that when you hold a clip in green mode, you can e.g. use the mod encoders on it, without changing the clip's mute/launch status - not that you actually enter the clip. 

I would agree that in that case (long press) the current clip should change to be consistent with the other view modes.

m-m-adams:
> I assume he means that when you hold a clip in green mode, you can e.g. use the mod encoders on it, without changing the clip's mute/launch status - not that you actually enter the clip.
> 
> I would agree that in that case (long press) the current clip should change to be consistent with the other view modes.
> 
> Otherwise I agree with your PR and the reasoning behind it, and closed mine.

thanks for the initial work! It made it way quicker to find the places to add the checks

I've updated to handle holding clips in green mode now too so that should work

seangoodvibes (approved):
Code looks good to me. No concerns.

Issue #1179, opened 2024-02-06
Shift + hold and turn vertical encoder inside a clip transposes all other clips. Proposal to change it to nudge notes vertically for that clip


soymonitus:
### Please describe the problem:

When you are inside a clip and hold and turn the vertical encoder it will nudge all notes in the clip 1 octave up or down, but only for that clip. Contrarily, when you hold Shift and hold and turn the encoder it transposes not only this clip but all the other clips in the song, changing the musical mode (same scale, new root note).
This is inconsistent and should not affect other clips.

### What is the expected behavior?

I would expect notes to be nudge up or down by one position within the current sequencer grid (scale could be on or off). If scale is On, then notes will nudge up or down in tune with the scale. If scale is off, the notes will go up or down one semitone, but only for that clip, and not for all clips

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

Official

PR #1183, opened 2024-02-07
Implemented transpose in scale for Shift + hold and turn vertical encoder

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1179

Now the command "Shift + hold and turn vertical encoder" inside a clip instead of transposing all clips in the song, it will nudge notes vertically according to the mode the clip is in (scale or non-scale).
If in scale, the notes are nudged vertically respecting the scale.
If clip not in scale mode, the notes are nudged by semitones.
Colors of the audition pads and notes are updated (they were not before).

The existing command "Hold and turn vertical encoder" keeps working as always, that is, transposing an octave all notes for this clip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ In addition, a number of improvements have been made to how the OLED display is 
 - Added support for "gentle paste" of notes which pastes notes without removing old ones.
 - Fixed numerous crash bugs around parameter automation when entering and leaving clip view.
 - The default ModFX type for songs is now DISABLED rather than FLANGER.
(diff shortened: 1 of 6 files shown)


soymonitus:
@m-m-adams @seangoodvibes this is ready for testing!

soymonitus:
I made some changes after looking at the code @0beron is writing for his PR, to avoid future conflicts between the two PRs. The features are now clearly differentiated. He can still make use of the instrumentClip's transpose() method while I use a new one called nudgeNotesVertically

m-m-adams (approved):

PR #1186, opened 2024-02-07
Filter type cleanup

m-m-adams:
Fixes:

- Morph now displays properly as Drive in ladder filters and FM in the high pass ladder
- Gold knob clicks only cycle within families - 12/24/drive, band/notch

Improvements:

- Adds a template EnumStringMap that takes in an array of enum to string pairs and provides an enum to string and string to enum function. 
- Adds a functional API for splitting a filtermode enum into its family and type
- Concentrates all filter config related functionality into one header, making it much easier to add more filters in the future

### DIFF
--- a/src/deluge/model/mod_controllable/filters/filter_config.h
+++ b/src/deluge/model/mod_controllable/filters/filter_config.h
@@ -38,13 +38,12 @@ using specificFilterType = uint8_t;
 // must match order of filter family declaration for indexing to work
 constexpr uint8_t numForVariant[kNumFamilies] = {kNumLadders, kNumSVF, kNumHPLadders};
 
-struct FilterType {
-	FilterFamily family;
-	specificFilterType type;
-	FilterMode mode;
-
-	FilterType() = default;
-	FilterType(FilterMode mode) : mode(mode) {
+/// A filters family (types which can all share an implementation for toggling) and specific type within that family
+class SpecificFilter {
+public:
+	SpecificFilter() = default;
+	/// Construct a filtertype from a mode
+	SpecificFilter(FilterMode mode) {
 		switch (mode) {
 		case FilterMode::OFF:
 			family = FilterFamily::NONE;
@@ -76,6 +75,11 @@ struct FilterType {
 			break;
 		}
 	}
+
+	FilterFamily getFamily() { return family; }
+
+	specificFilterType getType() { return type; }
+
 	[[nodiscard]] l10n::String getMorphName() const {
 
 		switch (family) {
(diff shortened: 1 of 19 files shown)


M0R172:
Really like the refactorings, BUT:

- Gold knob clicks only cycle within families - 12/24/drive, band/notch

Why make changing filter types harder? I really like that I can currently try out all kinds of filters on a sound without any additional shortcut or menu diving!

m-m-adams:
That's in support of future filter additions - the click to cycle is getting unwieldy already and I want to add in some more types!

m-m-adams:
Hm, they're all there in the shortcut list still. I found this a significant improvement personally as I usually set my presets up using shortcuts and just tweak on the gold knobs. When I press the gold knob it's almost always to change band to notch or 12db to 24db, and with this change that's a lot faster to do

M0R172:
Thanks for considering it, I guess everybody has a different editing style...

m-m-adams:
Reverted it for now so it doesn't hold up the bug fixes and refactoring in this PR

src/deluge/model/mod_controllable/filters/filter_config.h
```diff
@@ -0,0 +1,145 @@
...
+			return l10n::String::STRING_FOR_MORPH;
+		default:
+			return l10n::String::STRING_FOR_NONE;
+		};
+	};
+
+	SpecificFilter& incrementMode() {
+		type = (type + 1) % numForVariant[util::to_underlying(family)];
+		return *this;
+	} // namespace deluge::dsp::filter
```
sapphire-arches, line 128: Old comment?

src/deluge/util/container/enum_to_string_map.hpp
```diff
@@ -0,0 +1,35 @@
...
+	constexpr Enum operator()(const char* str) {
+		for (int i = 0; i < N; i++) {
+			if (!strcmp(str, stringList[i])) {
+				return static_cast<Enum>(i);
+			}
+		}
+#if ALPHA_OR_BETA_VERSION
+		char popup[25];
+		sprintf(popup, "no match for:%s", str);
+		freezeWithError(popup);
```
sapphire-arches, line 31: Does this work on 7seg?

m-m-adams: It does not, I'll just log it instead. It's probably better not to freeze on any unrecognized tag later anyway

src/deluge/model/mod_controllable/filters/filter_config.h
m-m-adams: No, it's in the filter namespace intentionally. It's only in the models folder so that it compiles as thumb. It's not much code so probably wouldn't make a big difference to move back though

Issue #1189, opened 2024-02-07
decideOnCurrentPlaybackMode function doesn't identify automation arranger view or performance view accessed from arranger


seangoodvibes:
### Please describe the problem:

The function void PlaybackHandler::decideOnCurrentPlaybackMode() checks the UIType to determine if the playback mode should be arranger playback or session playback.

When in Automation Arranger View or Performance View accessed from arranger, the playback mode is incorrectly set as session

### What is the expected behavior?

Playback mode for Automation Arranger View or Performance View accessed from arranger should be Arrangement

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1191, opened 2024-02-08
Bug fix - set correct playback mode when in Alternate Arranger Views - Automation / Performance

seangoodvibes:
Bug fix to set correct playback mode to arrangement when in an arranger view or arranger white clip

This wasn't working when in automation arranger view or in performance view launched from arranger

This closes issue https://github.com/SynthstromAudible/DelugeFirmware/issues/1189


m-m-adams:
This adds bugs for non-automation view - for example, if you enter a session only clip, then switch to arranger it will play in session. If you enter a white clip in arranger, then switch to session, it will play in arranger. 

The correct fix is modifying automationview->toClipMinder() to return NULL if it's doing song automation, and leaving the launch selection intact

Issue #996, opened 2024-01-16
Enhancement / Add Song/Grid Mute Column to Sidebar to Launch Clips while in Performance View


seangoodvibes:
### Please describe the problem:

Request to see the sections from grid view / clips from song view so that you can launch sections / clips while in performance view

### What is the expected behavior?

Same interaction with mute column in sidebar in performance view as when you are in Song/Grid view

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

PR #1194, opened 2024-02-08
Add Arranger and Session Sidebar's to Performance View

seangoodvibes:
Added Arranger and Session Sidebar's to Performance View

This closes issue https://github.com/SynthstromAudible/DelugeFirmware/issues/996

### DIFF
--- a/src/deluge/gui/views/performance_session_view.cpp
+++ b/src/deluge/gui/views/performance_session_view.cpp
@@ -256,10 +256,6 @@ void PerformanceSessionView::focusRegained() {
 
 	updateLayoutChangeStatus();
 
-	if (defaultEditingMode) {
-		indicator_leds::blinkLed(IndicatorLED::KEYBOARD);
-	}
-
 	if (display->have7SEG()) {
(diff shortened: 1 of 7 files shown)


src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -881,18 +897,40 @@ ActionResult PerformanceSessionView::padAction(int32_t xDisplay, int32_t yDispla
 			}
 			uiNeedsRendering(this); // re-render pads
 		}
+		// pressing the first column in sidebar to trigger sections / clips
+		else if (xDisplay == kDisplayWidth) {
```
m-m-adams, line 901: this section would be a lot easier to follow if you inverted the order of the checks - something like this:

if (xDisplay >= kDisplayWidth) {
if (in arranger) {
    do arranger things
}
else {
    if (not in grid) {
        do session view things
    }
    else {
        do grid session things
   }
}
}

m-m-adams (approved):

PR #1198, opened 2024-02-09
Arpeggiator improvements

soymonitus:
Added new features to the arpeggiator (Note: Ratchets are currently available only for Synth clips, but expected to be
  extended to Midi clips and Kit rows in the future). The new features include:
    **1. Splitted the old `Mode` setting into separate settings** `Mode` (Off or Arpeggiator) `Octave Mode` (Up, Down,
    Up&Down, Alternate or Random) and `Note Mode` (Up, Down, Up&Down, AsPlayed or Random) settings, so you can setup
    individually how octaves are walked and how notes are walked in the sequence.
    **2. The `Mode` pad shortcut is now an `Arp preset` shortcut**, which will update the new 3 settings all at once:
          - `Off` will disable arpeggiator.
          - `Up` will setup Mode to `Arpeggiator`, Octave Mode to `Up` and Note Mode to `Up`.
          - `Down` will setup Mode to `Arpeggiator`, Octave Mode to `Down` and Note Mode to `Down`.
          - `Both` will setup Mode to `Arpeggiator`, Octave Mode to `Alternate` and Note Mode to `Up`.
          - `Random` will setup Mode to `Arpeggiator`, Octave Mode to `Random` and Note Mode to `Random`.
          - `Custom` will setup Mode to `Arpeggiator`, and enter a submenu to let you edit Octave Mode and Note Mode.
    **3. Mode (MODE):**
          - `Off` disables the arpeggiator.
          - `Arpeggiator` (ARP) enables the arpeggiator.
    **4. Octave Modes (OMOD):**
          - `Up` (UP) will walk the octaves up.
          - `Down` (DOWN) will walk the octaves down.
          - `Up & Down` (UPDN) will walk the octaves up and down, repeating the highest and lowest octaves.
          - `Alternate` (ALT)  will walk the octaves up, and then down reversing the Notes pattern (without
          repeating notes). Tip: Octave Mode set to Alternate and Note Mode set to Up is equivalent to
          the old `Both` mode.
          - `Random` (RAND) will choose a random octave every time the Notes pattern has played.
          Tip: Set also Note Mode to Random to have the equivalent to the old `Random` mode.
    **5. Note Modes (NMOD):**
          - `Up` (UP) will walk the notes up.
          - `Down` (DOWN) will walk the notes down. Tip: this mode also works in conjunction with Octave Mode
          Alternate, which will walk all the notes and octaves all the way down, and then up reversing it.
          - `Up & Down` (UPDN) will walk the notes up and down, repeating the highest and lowest notes.
          - `As played` (PLAY) will walk the notes in the same order that they were played. Tip: this mode
          also works in conjunction with Octave Mode Alternate, which will walk all the notes and octaves
          all the way up (with notes as played), and then down reversing the order of played notes.
          - `Random` (RAND) will choose a random note each time. If the Octave Mode is set to something
          different than Random, then the pattern will play, in the same octave, the same number of random
          notes as notes are in the held chord and then move to a different scale based on the Octave Mode.
          Tip: Set also Octave Mode to Random to have the equivalent to the old `Random` mode.
    **6. `Sequence Length` (LENG)** (unpatchet parameter, assignable to golden knobs):
          - If set to zero, the arpeggiator pattern will play fully.
          - If set to a value higher than zero, the pattern will play up to the set number of notes, and then
          reset itself to start from the beginning.
    **7. Ratcheting:** There are two new parameters (unpatched, assignable to golden knobs), to control how notes
      are ratcheted. A ratchet is when a note repeats itself several time in the same time interval that the
      original note has to play.
          - `Ratchet Amount` (RATC): this will set the maximum number of ratchets that an arpeggiator step
           could have (each step will randomize the number of ratchet notes between 1 and max value).
             - From values 0 to 4, no ratchet notes
             - From 5 to 19, up to 2 ratchet notes
             - From 20 to 34, up to 4 ratchet notes
             - From 35 to 50, up to 8 ratchet notes
          - `Ratchet Probability` (RPRO): this sets how likely a step is to be ratcheted
             - Being 0 (0%), no ratchets at all
             - And 50 (100%), all notes will evaluate to be ratcheted.
             
  Follow-up ideas/improvements for later (out of scope for this PR):
  - Rhythm (rests patterns), like the ones in Novation Launchkey: 0, 0x0, 0xx0, Random
  - Add "Chord" to Note Mode (to play all notes at the same time)
  - Ratchets for kit rows and midi clips
  - Sequencer (16 configurable steps with +- deviation in semitones from held note)


src/deluge/processing/sound/sound.cpp
```diff
@@ -3833,6 +3854,7 @@ void Sound::writeToFile(bool savingSong, ParamManager* paramManager, Arpeggiator
 		storageManager.writeOpeningTagBeginning("arpeggiator");
 		storageManager.writeAttribute("mode", arpModeToString(arpSettings->mode));
 		storageManager.writeAttribute("numOctaves", arpSettings->numOctaves);
+		storageManager.writeAttribute("numRatchets", arpSettings->numRatchets);
```
m-m-adams, line 3857: not used by sounds - sounds are using the unpatched param instead

src/deluge/processing/sound/sound.cpp
```diff
@@ -3690,6 +3709,8 @@ void Sound::writeParamsToFile(ParamManager* paramManager, bool writeAutomation)
 	UnpatchedParamSet* unpatchedParams = paramManager->getUnpatchedParamSet();
 
 	unpatchedParams->writeParamAsAttribute("arpeggiatorGate", params::UNPATCHED_ARP_GATE, writeAutomation);
+	unpatchedParams->writeParamAsAttribute("arpeggiatorRatchetProbability", params::UNPATCHED_ARP_RATCHET_PROBABILITY,
```
m-m-adams, line 3712: Sound checks ratchetProbability to read this parameter

You can put the UnpatchedSound enum into the template enum to string mapper class to avoid this error - https://github.com/SynthstromAudible/DelugeFirmware/blob/community/src/deluge/util/container/enum_to_string_map.hpp

Example here - https://github.com/SynthstromAudible/DelugeFirmware/blob/community/src/deluge/model/mod_controllable/filters/filter_config.cpp

soymonitus: Fixed!

src/deluge/processing/sound/sound.cpp
soymonitus: this is for the number of ratchets, which is a "menu" parameter. It is similar to octaves and cannot be automated

src/deluge/util/functions.cpp
```diff
@@ -933,43 +933,163 @@ FilterType stringToFilterType(char const* string) {
...
+		return ArpOctaveMode::ALTERNATE;
+	}
+	else if (!strcmp(string, "random")) {
+		return ArpOctaveMode::RANDOM;
+	}
+	else {
+		return ArpOctaveMode::UP;
+	}
+}
+
```
sapphire-arches, line 1092: These can/should be replaced with `EnumStringMap` I think, should provide some code size savings (see `src/deluge/model/mod_controllable/filters/filter_config.cpp`  for an example of how to use these). They can live in `arpeggiator.cpp` as static functions too, so `functions.cpp` doesn't expand forever.

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
```diff
@@ -0,0 +1,30 @@
...
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+#pragma once
+#include "gui/menu_item/unpatched_param.h"
+#include "gui/ui/sound_editor.h"
+
+namespace deluge::gui::menu_item::arpeggiator {
```
sapphire-arches, line 21: Instead of 3 separate classes, this, `RatchetProbability`, and `SequenceLength` could be a single class `deluge:gui::menu_item::unpatched_param::Arpeggiator` (in `deluge/gui/menu_item/unpatched_param`).

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
```diff
@@ -0,0 +1,30 @@
...
+class RatchetAmount : public UnpatchedParam {
+public:
+	using UnpatchedParam::UnpatchedParam;
+	bool isRelevant(ModControllableAudio* modControllable, int32_t whichThing) override {
+		return !soundEditor.editingKit() && !soundEditor.editingCVOrMIDIClip();
```
sapphire-arches, line 26:
```suggestion
		return getCurrentOutputType() != OutputType::AUDIO;
```

We do support ARP of kit rows, CV, and MIDI outputs I think? Or is there specific limitations on the ratchet implementation that prevents it working there?

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -34,6 +35,11 @@ void NonAudioInstrument::renderOutput(ModelStack* modelStack, StereoSample* star
 
 		if (activeInstrumentClip->arpSettings.mode != ArpMode::OFF) {
 			uint32_t gateThreshold = activeInstrumentClip->arpeggiatorGate + 2147483648;
+			uint32_t ratchetProbability = activeInstrumentClip->arpeggiatorRatchetProbability + 2147483648;
+			uint32_t ratchetAmount = activeInstrumentClip->arpeggiatorRatchetAmount + 2147483648;
+			uint32_t sequenceLength =
+			    (((int64_t)activeInstrumentClip->arpeggiatorSequenceLength + 2147483648) * kMaxMenuValue + 2147483648)
+			    >> 32;
```
sapphire-arches, line 42:
```suggestion
			uint32_t gateThreshold = activeInstrumentClip->arpeggiatorGate + 2147483648U;
			uint32_t ratchetProbability = activeInstrumentClip->arpeggiatorRatchetProbability + 2147483648U;
			uint32_t ratchetAmount = activeInstrumentClip->arpeggiatorRatchetAmount + 2147483648U;
			uint32_t sequenceLength =
			    (((int64_t)activeInstrumentClip->arpeggiatorSequenceLength + 2147483648) * kMaxMenuValue + 2147483648U)
			    >> 32;
```

I think these calculations can overflow otherwise? It's unclear how `gateThreshold` was working before.

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -128,47 +147,86 @@ void Arpeggiator::noteOn(ArpeggiatorSettings* settings, int32_t noteCode, int32_
...
-	    MIDI_CHANNEL_NONE; // MIDIInstrument might set this, but it needs to be MIDI_CHANNEL_NONE until then so it
-	                       // doesn't get included in the survey that will happen of existing output member channels.
+		// Save arpNote
+		arpNote = (ArpNote*)notes.getElementAddress(notesKey);
+		arpNote->inputCharacteristics[util::to_underlying(MIDICharacteristic::NOTE)] = noteCode;
+		arpNote->velocity = velocity;
+		arpNote->outputMemberChannel =
+		    MIDI_CHANNEL_NONE; // MIDIInstrument might set this, but it needs to be MIDI_CHANNEL_NONE until then so it
+		                       // doesn't get included in the survey that will happen of existing output member
+		                       // channels.
```
sapphire-arches, line 197: style nit: move comment above assignment, since you're touching this line anyway.

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -128,47 +147,86 @@ void Arpeggiator::noteOn(ArpeggiatorSettings* settings, int32_t noteCode, int32_
...
+		}
 	}
 
 noteInserted:
 	arpNote->inputCharacteristics[util::to_underlying(MIDICharacteristic::CHANNEL)] =
 	    fromMIDIChannel; // This is here so that "stealing" a note being edited can then replace its MPE data during
 	                     // editing. Kind of a hacky solution, but it works for now.
+	arpNoteAsPlayed->inputCharacteristics[util::to_underlying(MIDICharacteristic::CHANNEL)] =
+	    fromMIDIChannel; // This is here so that "stealing" a note being edited can then replace its MPE data during
+	                     // editing. Kind of a hacky solution, but it works for now.
```
sapphire-arches, line 229: style nit: move comments

m-m-adams:
The arpeggiator mode pad is a waste now - could you make it work like unison, where pressing once does note direction (starting from off) and pressing a second time does octave direction?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -241,66 +319,120 @@ void ArpeggiatorBase::switchAnyNoteOff(ArpReturnInstruction* instruction) {
...
+			}
+			else {
+				currentOctave = 0;
+			}
+		}
+	}
+	else {
+		currentOctaveDirection = (settings->octaveMode == ArpOctaveMode::DOWN)
+		                             ? -1
+		                             : 1; // Have to reset this, in case the user changed the setting.
```
sapphire-arches, line 393: Style nit: move comment above for better wrapping.

src/deluge/modulation/arpeggiator.h
```diff
@@ -120,6 +142,7 @@ class Arpeggiator final : public ArpeggiatorBase {
 	bool hasAnyInputNotesActive();
 
 	OrderedResizeableArray notes;
+	ResizeableArray notesAsPlayed;
```
sapphire-arches, line 145: If I understand correctly, `notes` tracks the currently sounding notes generated by the arpegiator while `notesAsPlayed` tracks the "live" notes from the underlying sequence? It would be good to describe this in detail in `///` doc comments here.

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
soymonitus: but they are different parameters, how are they supposed to be the same class?

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
soymonitus: Right now ratchets are limited to internal synth engine. The arp gates are calculated differently and I need to refactor it, but better in a future PR. Right now it works smoothly for synth clips only. Kits and Midi are a future improvement (maybe during the beta phase these weeks I will be able to add them, but want the PR out so people can start using/testing it)

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
sapphire-arches: The specific parameter to edit is a constructor parameter.

src/deluge/model/instrument/non_audio_instrument.cpp
soymonitus: The gateThreshold line has been like that from the beginnings so I just did the same with my new parameters. I haven't experienced any buggy behavior even setting parameter values from minimum to maximum values

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
soymonitus: I see other params inheriting from  public UnpatchedParam  and they are in different folders and classes. I think I don't understand what you mean

soymonitus:
@sapphire-arches I pushed some changes for some comments of your feedback

src/deluge/util/functions.cpp
soymonitus: I specify different names for OLED and 7SEG so probably not a good fit this EnumStringMap

src/deluge/gui/menu_item/arpeggiator/ratchet_amount.h
m-m-adams: All three of these classes are identical (you're just over riding isRelevant) and the param id is passed in the constructor

src/deluge/util/functions.cpp
m-m-adams: I'm working on converting params to use the enumStringMap as a bigger refactor, don't worry about it for now

m-m-adams:
@sapphire-arches I'm merging and opening an issue with the minor tweaks needed - I want user feedback on this

m-m-adams (approved):

PR #1199, opened 2024-02-09
Enable sysex logging to console by dbt command

soymonitus:
Command is:
`./dbt sysex-logging <port_number_index>`

To check the port number indexes execute:
`./dbt sysex-logging -h`

To enable sysex message sending from the deluge you have to enable the flag before compiling:
`./dbt configure -DENABLE_SYSEX_LOGGING=YES`
Then compile as normal:
`./dbt build`
And flash to the deluge the Debug build found in build/Debug/

To make debug log prints in the Deluge that will output to the console here is an example of how to print logs:
`D_PRINTLN("new file Index is %d", newFileIndex);`

### DIFF
--- a/scripts/tasks/task-sysex-logging.py
+++ b/scripts/tasks/task-sysex-logging.py
@@ -1,0 +1,104 @@
+#! /usr/bin/env python3
+import sys
+import time
+import argparse
+import rtmidi
+import sys
+
+
+def argparser():
+    midiout = rtmidi.MidiOut()
+    available_ports = midiout.get_ports()
+    s = "\nusage example: dbt sysex-logging 123"
+    s += "\n\nstart listening to console logs on MIDI port # 123"
+    s += "\n\nAvailable MIDI ports:\n\n"
+    for i, p in enumerate(available_ports):
+        s += "Port # " + str(i) + " : " + str(p) + "\n"
+
+    parser = argparse.ArgumentParser(
+        prog="sysex-logging",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description="Listen to console logs via MIDI",
+        epilog=s,
+        exit_on_error=True,
+    )
+    parser.group = "Development"
+    parser.add_argument(
+        "port_number",
+        help="MIDI port number (example: 123) use 'dbt sysex-logging -h' to list available ports",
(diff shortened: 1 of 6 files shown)


m-m-adams:
Can you add the info to CONTRIBUTING.md as well please?

scripts/tasks/task-sysex-logging.py
```diff
@@ -0,0 +1,104 @@
...
+def sysex_console(port):
+    midiout = rtmidi.MidiOut()
+    midiout.open_port(port)
+
+    with midiout:
+        data = bytearray(6)
+        data[0] = 0xF0  # sysex message
+        data[1] = 0x7D  # deluge (midi_engine.cpp midiSysexReceived)
+        data[2] = 0x03  # debug namespace
+        data[3] = 0x00  # sysex.cpp, sysexReceived
```
m-m-adams, line 63: 0x00 0x01 is the command to enable sysex logging - please clarify that in the comment

src/deluge/io/debug/print.cpp
```diff
@@ -110,7 +110,7 @@ void prependTimeStamp(bool isNewLine) {
 		lutHexString(Debug::readCycleCounter(), buffer);
 		buffer[8] = ' ';
 		buffer[9] = 0;
-		if (midiDebugDevice && SYSEX_LOGGING_ENABLED) {
+		if (midiDebugDevice && ENABLE_SYSEX_LOGGING) {
```
m-m-adams, line 113: this isn't needed, there's no issue always having this in the binary and we've taken advantage of it to troubleshoot issues with users before. Sysex logging enabling is handled by sending the logging enable sysex message

soymonitus: Ok I will remove that flag

soymonitus:
Not sure what I have to add to that file. I haven't seen any mention to the already existing loadfw command so not sure what to put and where in the doc

> Can you add the info to CONTRIBUTING.md as well please?

m-m-adams:
Just a mention that this exists as a way to get log messages while developing. Basically the same as what you put in the first comment, but it more properly belongs in the developer guide in contibuting.md

m-m-adams (approved):

PR #1201, opened 2024-02-09
Add new HPF to reverb, log scaling for HPF and Damping

stellar-aria:
Adds a requested feature to the Mutable reverb: HPF on the wet output to reduce the low-end rumble

Adds log-scaling controls to the Damping and HPF cutoff to match frequency scaling

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/1579


src/deluge/dsp/util.hpp
```diff
@@ -87,4 +89,22 @@ void foldBuffer(q31_t* startSample, q31_t* endSample, q31_t foldLevel) {
 		currentSample += 1;
 	} while (currentSample < endSample);
 }
+
+/* Original code for fastlog2f by Dr. Paul Beckmann from the ARM community
+forum, adapted from the CMSIS-DSP library About 25% performance increase over
+std::log10f
+*/
+inline float log2fast(float f) {
```
m-m-adams, line 97: have you benchmarked this? I would be surprised if it's actually faster. The arm built in appears to use the same number of polynomial approximation rounds, but uses direct conversion of the ieee754 float value to exponent/mantissa as it's initial estimate instead of frexp

stellar-aria: it might not be, this was originally written for Cortex-M I think. I can do a Cortex-A benchmark with a stub on a beaglebone or something if you really want

m-m-adams: I'd prefer to just use std::log. I'm open to keeping this if it benchmarks significantly better but comparing source code and assembly I believe this only saves two multiplies and adds a function call compared to std::log. That would be a large speed up without hardware floats but might even be a loss for us.

stellar-aria: yeah that's fair, I'll take it out.

seangoodvibes:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1579

m-m-adams (approved):

src/deluge/dsp/reverb/mutable/reverb.hpp
```diff
@@ -124,12 +127,27 @@ class Mutable : public Base {
...
+	[[nodiscard]] float getDamping() const override { return lp_val_; }
 
 	void setWidth(float value) override { diffusion_ = util::map(value, 0.f, 1.f, kWidthMin, kWidthMax); }
 	[[nodiscard]] float getWidth() const override { return util::map(diffusion_, kWidthMin, kWidthMax, 0.f, 1.f); };
 
+	//	void setHPF(float value) {
+	//		hp_cutoff_val_ = value;
+	//		hp_cutoff_ = value == 0.f ? 0.f : std::clamp(0.35f - std::log2(((1.f - value) * 50.f) + 1.f) / 16,
+	// 0.f, 1.f);
+	//	}
```
sapphire-arches, line 143: should be deleted

PR #1208, opened 2024-02-11
Refactor/move midi match to kit

m-m-adams:
The kit offer functions don't rely on the midi follow class and make more sense in kit. Moving them enables future work to learn a whole kit to a single input midi channel using a similar mechanism to midi follow

### DIFF
--- a/src/deluge/io/midi/midi_follow.cpp
+++ b/src/deluge/io/midi/midi_follow.cpp
@@ -401,8 +401,10 @@ void MidiFollow::sendNoteToClip(MIDIDevice* fromDevice, Clip* clip, MIDIMatchTyp
 			// Definitely don't record if muted in arrangement
 			bool shouldRecordNotes = shouldRecordNotesNowNow && currentSong->isOutputActiveInArrangement(clip->output);
 			if (clip->output->type == OutputType::KIT) {
-				offerReceivedNoteToKit(modelStackWithTimelineCounter, fromDevice, on, channel, note, velocity,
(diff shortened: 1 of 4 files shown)


src/deluge/io/midi/midi_follow.cpp
```diff
@@ -401,8 +401,10 @@ void MidiFollow::sendNoteToClip(MIDIDevice* fromDevice, Clip* clip, MIDIMatchTyp
 			// Definitely don't record if muted in arrangement
 			bool shouldRecordNotes = shouldRecordNotesNowNow && currentSong->isOutputActiveInArrangement(clip->output);
 			if (clip->output->type == OutputType::KIT) {
-				offerReceivedNoteToKit(modelStackWithTimelineCounter, fromDevice, on, channel, note, velocity,
-				                       shouldRecordNotes, doingMidiThru, clip);
+				auto kit = (Kit*)clip->output;
+				kit->receivedNoteForKit(modelStackWithTimelineCounter, fromDevice, on, channel,
```
seangoodvibes, line 405: nit: just curious, why the `auto kit` here and no auto kit in the same type of code below?

seangoodvibes (changes requested):

src/deluge/io/midi/midi_follow.cpp
m-m-adams: They mean the same thing

seangoodvibes (approved):

PR #1212, opened 2024-02-11
Move config suggestions to dir

m-m-adams:
We're having repeated issues with people's required local customizations for their editor conflicting with the checked in version. To avoid this, move .vscode to IDE_Configs/vscode and gitignore .vscode itself. I have also switched the 'recommended' intellisense from clangd back to c_cpp as people seem to have more issues setting up clangd in general, and it requires non-generic paths on macos and linux to run without errors. 

It would be nice to have dbt create .vscode when running dbt configure from vscode, however I'm not yet sure how to detect that on non-*nix systems. For now I've just made a note in CONTRIBUTING.md instead.


PaulFreund:
Fully agree it is useful. As discussed on Discord, I think we should wait with merging it until there is an automation for new users so the out of box experience is not impacted. Automates DBT install should make some initial choices

m-m-adams:
Yeah, I guess we could just always create .vscode if it doesn't exist. There's no harm even if people use a different IDE

On mac and linux vscode sets itself as $TERM_PROGRAM in its shells, do you know if there's an equivalent on windows?

m-m-adams:
Confirmed working on mac/linux/windows

PR #1213, opened 2024-02-12
[DRAFT] Enhancement / Display change in semitones in an instrument clip when nudging notes vertically

seangoodvibes:
Added a pop-up to show the number of semitones nudged vertically while you are holding down and turning the vertical encoder in an instrument clip.

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -88,6 +88,8 @@ InstrumentClip::InstrumentClip(Song* song) : Clip(ClipType::INSTRUMENT) {
 	}
 
 	outputTypeWhileLoading = OutputType::SYNTH; // NOTE: (Kate) was 0, should probably be NONE
+
+	semitonesNudged = 0;
 }
 
 // You must call prepareForDestruction() before this, preferably by calling Song::deleteClipObject()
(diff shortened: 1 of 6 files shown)


src/deluge/model/clip/instrument_clip.h
```diff
@@ -76,6 +76,8 @@ class InstrumentClip final : public Clip {
 	void seeWhatNotesWithinOctaveArePresent(bool[], int32_t, Song* song, bool deleteEmptyNoteRows = true);
 	void transpose(int32_t, ModelStackWithTimelineCounter* modelStack);
 	void nudgeNotesVertically(int32_t, ModelStackWithTimelineCounter* modelStack);
+	int32_t semitonesNudged;
```
m-m-adams, line 79: This doesn't belong here - it's almost always 0, and even when it's not it doesn't contain useful information about the clip. The value is tied to the action of turning the knob so the storage location should be too. Use a static variable in the function instead

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -1306,6 +1308,7 @@ void InstrumentClip::nudgeNotesVertically(int32_t change, ModelStackWithTimeline
 			NoteRow* thisNoteRow = noteRows.getElement(i);
 			// transpose by semitones or by octave
 			thisNoteRow->y += change;
+			semitonesNudged += change;
```
m-m-adams, line 1311: This isn't actually very useful info. Display the number of scale degrees it's been nudged instead

PR #1215, opened 2024-02-12
Improve sysex logging script

soymonitus:
Reduced the sleep wait time so it is not so CPU intensive when there are no messages to read. Free up the midiout connection when the config message was already sent, as per the docs from rtmidi


m-m-adams (approved):
Asyncio with callbacks would be a nice improvement too!

PR #1225, opened 2024-02-13
Voice efficiency

### LINKED ISSUE #804: More Dropouts Occurring with C1_0_0 Firmware

Love it all the new features. Grid mode I now can’t live without. But I’m experiencing a lot more dropouts of tracks sooner with the new version. 
min version 4.1.3 I could play a specific song with 9 tracks. Now I can’t even play 6 tracks on the same song without dropouts occurring. 

All things being equal except the new firmware update, dropouts should not be occurring with so few tracks playing. 

_No response_

OLED

Built from community

C1.0.0

I reverted back to firmware 4.1.3 and played up to 9 tracks without dropouts. I them updated to firmware C1.0.0 and played the song with the same tracks active which caused numerous dropouts. I then began removing tracks and could only get up to 6 tracks to play at a time without dropouts.

m-m-adams:
Three changes to increase voice count - 
1. <s>When a sample is no longer needed, add it to the start of the stealable queue instead of the end</s> Removed - negative side effects with songs close to max load but not over it
2. After rendering the song, always try rendering it again (exits if it's ahead or after 5 renders). Previously only done when clocks interfered with rendering
3. Re introduce the cull smoothing - in combo with 2 this no longer has the risk of over shooting our samples, as if it's not keeping up with output it'll go again and cull a second time immediately 
4. Fix soft cull logic so that more than one voice can be soft culled, avoiding the issue of "overshooting" hard culls
5. Fix voice allocation logic so that voices are solicited via culling when it's likely that the next render round will cull voices
6. Fix memory leakage between songs by immediately deallocating voices when they won't be used again
7. Soft cull more agressively, avoid going into hard culling to reduce the clicks associated with hard culls

Fix #804 (I think)

### DIFF
--- a/src/deluge/dsp/timestretch/time_stretcher.cpp
+++ b/src/deluge/dsp/timestretch/time_stretcher.cpp
@@ -168,7 +168,8 @@ void TimeStretcher::reInit(int64_t newSamplePosBig, SamplePlaybackGuide* guide, 
 void TimeStretcher::beenUnassigned() {
 	unassignAllReasonsForPercLookahead();
 	unassignAllReasonsForPercCacheClusters();
-	olderPartReader.unassignAllReasons();
+	// we might sometimes want to dealloc the clusters but I'm not sure
+	olderPartReader.unassignAllReasons(false);
 	if (buffer) {
 		delugeDealloc(buffer);
 	}
@@ -984,7 +985,7 @@ bool TimeStretcher::setupNewPlayHead(Sample* sample, VoiceSample* voiceSample, S
                                      int32_t newHeadBytePos, int32_t additionalOscPos, int32_t priorityRating,
                                      LoopType loopingType) {
 	// clear the current reasons since setting up the clusters will add new ones
-	voiceSample->unassignAllReasons();
+	voiceSample->unassignAllReasons(false);
 	bool success = voiceSample->setupClustersForPlayFromByte(guide, sample, newHeadBytePos, priorityRating);
 	if (!success) {
 		return false;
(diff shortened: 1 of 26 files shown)


m-m-adams:
image

Huge increase in synth voices. Sample voices are increased slightly but not as dramatically - the latest song in the dropouts issue goes from culling at around ~12-16 voices to ~16-20

m-m-adams:
Fixing soft culling brings the synth voice test up to 153

image

soymonitus:
Regarding my most sample heavy song, which contain pretty long rap verse audio clips, they are cut much before with this build :(
 I need to test with other songs which are more sample heavy but with no audio clips to see if they improved

m-m-adams:
Is that true with the most recent commit as well? Audio clips shouldn't be affected by the changes, they're still supposed to be culled only when there's nothing else to cull

soymonitus:
> Is that true with the most recent commit as well? Audio clips shouldn't be affected by the changes, they're still supposed to be culled only when there's nothing else to cull

I haven't tested with latest commits, i will do it today

soymonitus:
I tested with various songs and it is working really good

stellar-aria (approved):
Overall pretty simple parameter passing changes, I won't pretend to understand the internals of the culling system in audio_engine.h right now.

chewabledrapery:
Given that #804 is closed and i'm still having the same exact issues on latest build where should i submit example projects to? Appreciate the work so far but 804 is not fully resolved for me, perhaps the underlying cause is not voice culling and is something else?  I have an example that has only 4 tracks in it and it is dropping kicks in a pattern. Thanks!

m-m-adams:
Yes, please make a new issue with example songs - I can't continue debugging off that issue since all provided songs play properly

chewabledrapery:
No problem might be better to start afresh, especially as the cause may turn out to be be unrelated.

I'll record a video to accompany the example project as that may help to show the issue in better context.

The latest beta definitely improved the other example I uploaded to 804 and overall I've noticed far less "clicky" artifacts in another project so you've definitely done some sterling work, very very much appreciated.

Thanks Mark watch this space 👍😎

PR #1237, opened 2024-02-16
Fix keyboard column controls

robmccoll:
- Fix vertical encoder handlers spilling into horizontal 
- Ignore select-knob caused zero offset horizontal encoders
- Make mode highlighting following scale mode instead of selected pad
- Velocity / mod glide mode (Pad 6 + vertical to enable disable) this works, but the UI is bad, so let's leave this as an easter egg will change in the future type of thing (also the popup gets eaten by the value when the pad retriggers, so you don't actually get an on / off display)

### DIFF
--- a/src/deluge/gui/ui/keyboard/layout/column_controls.cpp
+++ b/src/deluge/gui/ui/keyboard/layout/column_controls.cpp
@@ -31,6 +31,16 @@
 #define RIGHT_COL (kDisplayWidth + 1)
 #define MAX_CHORD_NOTES 4
 
+// TODO
+// - checkout doTick* and ui timer manager for:
+//   - slew
+//   - beat repeat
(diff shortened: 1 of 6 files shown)


soymonitus:
I have fixed several bugs from the issue i created in the github issues section. Here you can check the PR:
https://github.com/SynthstromAudible/DelugeFirmware/pull/1250
I think it would be good, it my PR gets merged first, that you rebase yours cause some of your tmp fixes (like the pad6 for glide) is already set as a menu toggle in mine, and i fixed the problem with scales

m-m-adams (commented):
Agreed that the menu solution is better for glide (which should be called latching, or inverted and called momentary)

sapphire-arches:
Superseded by #1250

PR #1251, opened 2024-02-18
Feature/learn midi for kit

m-m-adams:
Allows a whole kit to be learned at once to a midi channel. It's only possible from session/arranger view as I don't have a good UI for clip view

Todo: 

- [x] Modify midi learn to work with Instrument* instead of MelodicInstrument*
- [x] Modify session view midi learning so that it can be used on all instruments
- [x] Map learned input into the kit, using the note/cc/aftertouch/pitch received for kit functions
- [x] Make it flash in session view


sichtbeton:
oh this is great.

seangoodvibes (approved):

PR #1254, opened 2024-02-18
update workflows for 1p1 release

m-m-adams:
Update the build and cherry pick workflows to target 1.1 instead of 1.0


.github/workflows/cherry-pick.yml
```diff
@@ -15,18 +15,18 @@ permissions:
 jobs:
   cherry_pick_release_v1_0:
```
sapphire-arches, line 16: Job name should be updated

PR #1272, opened 2024-02-20
#REQUEST 902 - Automatically load the last open/saved song or user-defined template at startup

joxihan:
NEW SONG : Same behavior as current firmware, boot up on blank song.
TEMPLATE SONG: The current default behavior where a new project is initialized for you.
LAST USED SONG: The last project you loaded on the Deluge.
LAST MODIFIED SONG: The last project you saved on the Deluge.


m-m-adams:
For formatting you can run ./dbt format to get the exact format config used by our CI

If you're using VSCode and install the advanced formatters extension (should be in our suggested extensions) it'll do it on save for you

joxihan:
> For formatting you can run ./dbt format to get the exact format config used by our CI
> 
> If you're using VSCode and install the advanced formatters extension (should be in our suggested extensions) it'll do it on save for you

Thanks Mark, i didn't know that, just run dbt format and checked in

src/definitions_cxx.hpp
```diff
@@ -1069,3 +1070,5 @@ enum class ClipType {
 };
 
 enum class LaunchStyle { DEFAULT, FILL, ONCE };
+
+enum StartupSongMode : uint8_t { Template, LastOpened, LastSaved, StartupSongModeMaxEvent };
```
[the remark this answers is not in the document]
joxihan: Agree as default, let me change that.

joxihan:
updated PR following shapphire-arches comments

soymonitus:
When booting the Deluge in the default template song.. when wanting to save the song it will show name "DEFAULT 2". It would be better if the Default song, when loaded, its name or file path is wiped, so it behaves like a blank song, where clicking Save will show you the next available song slot. Say your last song is named SONG34.xml, then the Deluge should suggest you the name "SONG35"

src/deluge/deluge.cpp
```diff
@@ -446,6 +449,33 @@ void setupBlankSong() {
...
+		currentSong->writeTemplateSong(templatePath);
+	}
+	else {
+		void* songMemory = GeneralMemoryAllocator::get().allocMaxSpeed(sizeof(Song));
+		currentSong = new (songMemory) Song();
+		auto filename =
+		    startupSongMode == StartupSongMode::TEMPLATE ? "DEFAULT" : runtimeFeatureSettings.getStartupSong();
+		currentSong->name.set(filename);
+		if (openUI(&loadSongUI)) {
+			loadSongUI.performLoad();
```
soymonitus, line 471: Here you can add this after the performLoad() action:
```
			if (startupSongMode == StartupSongMode::TEMPLATE) {
				// Wipe the name so the Save action asks you for a new new
				currentSong->name.clear();
			}
```

soymonitus (changes requested):
A change to make the Template feature work as intended

soymonitus:
It doesn't work with booting from songs loaded/saved inside a subfolder. Dir name or path should also be taken into account.

joxihan:
> It doesn't work with booting from songs loaded/saved inside a subfolder. Dir name or path should also be taken into account.

Good catch I'll write the subfolder support tonight hk time thx for the review

soymonitus:
Now it looks good. But i have no permissions to approve so an admin dev should take a look at it 😀 
Thanks a lot for this feature, really useful

src/deluge/deluge.cpp
```diff
@@ -446,6 +449,37 @@ void setupBlankSong() {
...
+		void* songMemory = GeneralMemoryAllocator::get().allocMaxSpeed(sizeof(Song));
+		currentSong = new (songMemory) Song();
+		auto filename =
+		    startupSongMode == StartupSongMode::TEMPLATE ? "DEFAULT" : runtimeFeatureSettings.getStartupSong();
+		currentSong->setSongFullPath(filename);
+		if (openUI(&loadSongUI)) {
+			loadSongUI.performLoad();
+			if (startupSongMode == StartupSongMode::TEMPLATE) {
+				// Wipe the name so the Save action asks you for a new new
+				currentSong->name.clear();
```
seangoodvibes, line 474: The comment says "new new" - did you mean "new name"?

seangoodvibes (changes requested):

src/deluge/gui/l10n/seven_segment.cpp
```diff
@@ -338,6 +338,11 @@ PLACE_SDRAM_DATA Language seven_segment{
         {STRING_FOR_DEFAULT_UI_KEYBOARD_LAYOUT_INKEY, "INKY"},
         {STRING_FOR_DEFAULT_UI_SONG, "SONG"},
         {STRING_FOR_DEFAULT_UI_SONG_LAYOUT_ROWS, "ROWS"},
+        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE, "STSONG"},
```
soymonitus, line 341: since screen is only 4 chars, would you say SONG instead? as we are already in the Defaults submenu, there is enough context, to suppose SONG is going to be a setting for Default Song

src/deluge/gui/l10n/seven_segment.cpp
```diff
@@ -338,6 +338,11 @@ PLACE_SDRAM_DATA Language seven_segment{
...
+        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE_BLANK, "NSONG"},
```
soymonitus, line 342: Here I would say NEW

soymonitus (commented):
Some label changes

src/deluge/deluge.cpp
joxihan: updated, thx sean

src/deluge/gui/l10n/seven_segment.cpp
joxihan: updated thx monitus, i ll play around with the 7seg emulator to have more clue on 7seg limitations

seangoodvibes (approved):

seangoodvibes:
@joxihan could you add documentation to this feature to these documents please?

1. Community Features document: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/community_features.md

I would recommend added section 3.12 to the document under Section 3. General Improvements

2. Change Log Document: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/CHANGELOG.md

Add a brief description of your change under section c1.2.0

Thanks!

joxihan:
> @joxihan could you add documentation for this feature to these documents please?
> 
> 1. Community Features document: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/community_features.md
> 
> I would recommend added section 3.12 to the document under Section 3. General Improvements
> 
> 2. Change Log Document: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/CHANGELOG.md
> 
> Add a brief description of your change under section c1.2.0
> 
> Thanks!

Documentation added :)

src/deluge/model/song/song.cpp
```diff
@@ -5781,6 +5792,29 @@ int32_t Song::convertSyncLevelFromInternalValueToFileValue(int32_t internalValue
 	return fileValue;
 }
 
+const char* Song::getSongFullPath() {
+	String fullPath;
+	fullPath.concatenate(&dirPath);
+	fullPath.concatenate("/");
+	fullPath.concatenate(&name);
+	return fullPath.get();
```
sapphire-arches, line 5800: This is unsafe, `fullPath`'s storage will be deallocated at the end of the function and the returned pointer will therefore be dangling. You should return the string object itself.

src/deluge/model/song/song.cpp
```diff
@@ -5781,6 +5792,29 @@ int32_t Song::convertSyncLevelFromInternalValueToFileValue(int32_t internalValue
...
+
+void Song::setSongFullPath(const char* fullPath) {
+	if (char* filename = strrchr((char*)fullPath, '/')) {
+		char* dir = new char[sizeof(char) * strlen(fullPath) + 1];
+
+		memset(dir, 0, sizeof(char) * strlen(fullPath) + 1);
+		strncpy(dir, fullPath, strlen(fullPath) - strlen(filename));
+
+		dirPath.set(dir);
+		name.set(++filename);
```
sapphire-arches, line 5811: memory for `dir` is leaked, should only compute the various `strlen`s once.

src/deluge/deluge.cpp
```diff
@@ -446,6 +449,37 @@ void setupBlankSong() {
...
+			loadSongUI.performLoad();
+			if (startupSongMode == StartupSongMode::TEMPLATE) {
+				// Wipe the name so the Save action asks you for a new song
+				currentSong->name.clear();
+			}
+		}
+		else {
+			setupBlankSong();
+		}
+	}
```
sapphire-arches, line 480: Should be a `switch` statement over `startupSongMode`.

src/deluge/deluge.cpp
```diff
@@ -451,17 +451,19 @@ void setupBlankSong() {
...
+	switch (startupSongMode) {
+	case StartupSongMode::TEMPLATE:
+	case StartupSongMode::LASTOPENED:
+	case StartupSongMode::LASTSAVED: {
+		if (startupSongMode == StartupSongMode::TEMPLATE) {
+			auto templatePath = "SONGS/DEFAULT.XML";
+			setupBlankSong();
+			if (!storageManager.fileExists(templatePath)) {
+				currentSong->writeTemplateSong(templatePath);
+			}
```
sapphire-arches, line 465: This code can go under the `case StartupSongMode::TEMPLATE:`

src/deluge/deluge.cpp
```diff
@@ -451,17 +451,19 @@ void setupBlankSong() {
...
-		setupBlankSong();
-	}
-	else if (startupSongMode == StartupSongMode::TEMPLATE && !storageManager.fileExists(templatePath)) {
-		setupBlankSong();
-		currentSong->writeTemplateSong(templatePath);
-	}
-	else {
+	switch (startupSongMode) {
+	case StartupSongMode::TEMPLATE:
+	case StartupSongMode::LASTOPENED:
```
sapphire-arches, line 458:
```suggestion
	case StartupSongMode::TEMPLATE:
	    [[fallthrough]];
	case StartupSongMode::LASTOPENED:
		[[fallthrough]];
```

src/deluge/deluge.cpp
```diff
@@ -477,6 +479,10 @@ void setupStartupSong() {
 		else {
 			setupBlankSong();
 		}
+	} break;
+	case StartupSongMode::BLANK:
```
sapphire-arches, line 483:
```suggestion
	case StartupSongMode::BLANK:
	    [[fallthrough]];
```

src/deluge/deluge.cpp
joxihan: hum no, because there's also the logic to load the DEFAULT.XML song if it does already exist

src/deluge/deluge.cpp
joxihan: what's that for ?

src/deluge/deluge.cpp
sapphire-arches: Right, but it can fall through that case and then execute the rest of the logic.

src/deluge/deluge.cpp
sapphire-arches: Tells the compiler that the fall through is intentional, not the result of forgetting a break;

src/deluge/deluge.cpp
joxihan: ok didn't know this trick, let me change this when i'm back

src/deluge/deluge.cpp
joxihan: interestingly the compiler does not yell at me even tho i dont set this

src/deluge/gui/l10n/seven_segment.cpp
```diff
@@ -338,6 +338,11 @@ PLACE_SDRAM_DATA Language seven_segment{
         {STRING_FOR_DEFAULT_UI_KEYBOARD_LAYOUT_INKEY, "INKY"},
         {STRING_FOR_DEFAULT_UI_SONG, "SONG"},
         {STRING_FOR_DEFAULT_UI_SONG_LAYOUT_ROWS, "ROWS"},
+        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE, "SONG"},
+        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE_BLANK, "NEW"},
+        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE_TEMPLATE, "TEMPL"},
```
sapphire-arches, line 343:
```suggestion
        {STRING_FOR_DEFAULT_UI_DEFAULT_STARTUP_SONG_MODE_TEMPLATE, "TMPL"},
```

sapphire-arches (approved):
LGTM - I think we should add some safety check so if the default song load crashes we fall back to an init song, but that can be future work. Thanks for the contribution!

joxihan:
> LGTM - I think we should add some safety check so if the default song load crashes we fall back to an init song, but that can be future work. Thanks for the contribution!

Thanks, really my pleasure. I did some sanity check it seems to fallback pretty well in case of issue loading the song, (non existing or faulty file).
What is the next step before merging it ?

seangoodvibes:
> > LGTM - I think we should add some safety check so if the default song load crashes we fall back to an init song, but that can be future work. Thanks for the contribution!
> 
> Thanks, really my pleasure. I did some sanity check it seems to fallback pretty well in case of issue loading the song, (non existing or faulty file). What is the next step before merging it ?

Merging now :)! Welcome to the team :D

PR #1280, opened 2024-02-20
Implemented ratchets on midi clips, kit rows, unsynced rate, and stopped playback

soymonitus:
Before, you were only allowed to do ratchets on internal synth clips, and only if playback was engaged, and if arp rate was synced. Now it works in all other possible situations:
- Playing live notes on the audition pads with playback stopped
- Unsynced rate
- Kit rows
- Midi clips (which will output all the individual ratchet notes to your external synth!)


soymonitus:
@seangoodvibes could you review this? Mark can't do it for a while

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -138,13 +133,13 @@ void ArpeggiatorForDrum::noteOff(ArpeggiatorSettings* settings, ArpReturnInstruc
 	}
 
 	arpNote.velocity = 0; // Means note is off
+	reset();
```
sapphire-arches, line 136: What's the purpose of this reset?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -647,57 +662,62 @@ void ArpeggiatorBase::render(ArpeggiatorSettings* settings, int32_t numSamples,
...
 	}
 
+	uint32_t gateThresholdSmall = gateThreshold >> 8;
+
 	if (isRatcheting) {
 		// shorten gate in case we are ratcheting (with the calculated number of ratchet notes)
 		gateThresholdSmall = gateThresholdSmall >> ratchetNotesMultiplier;
 	}
 
+	uint32_t maxGate = 16777216;
```
sapphire-arches, line 694:
```suggestion
	uint32_t maxGate = 1 << 25;
```

make sure I got the shift amount right, but i think that's clearer.

soymonitus: It is 1<< 24. If you can edit your proposal so i can confirm it

soymonitus: Fixed

src/deluge/modulation/arpeggiator.cpp
soymonitus: Oh i see what you mean, that function does nothing here. Removed.

soymonitus:
Changes ready @sapphire-arches

PR #1282, opened 2024-02-21
Refactor to explicitly return ErrorType

stellar-aria:
Now that we have a proper enum (ErrorType) for errors instead of macro definitions, functions that were previously returning errors typed as `int`/ `int32_t` can be properly refactored to return `ErrorType`

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -90,7 +90,7 @@ bool Browser::opened() {
 bool Browser::checkFP() {
 	FileItem* currentFileItem = getCurrentFileItem();
 	String filePath;
-	int32_t error = getCurrentFilePath(&filePath);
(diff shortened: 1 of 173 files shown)


src/deluge/gui/ui/browser/browser.cpp
```diff
@@ -826,13 +826,13 @@ int32_t Browser::arrivedInNewFolder(int32_t direction, char const* filenameToSta
...
-	return NO_ERROR;
+	return Error::NONE;
 }
 
 // You must set currentDir before calling this.
-int32_t Browser::getUnusedSlot(OutputType outputType, String* newName, char const* thingName) {
+Error Browser::getUnusedSlot(OutputType outputType, String* newName, char const* thingName) {
 
-	int32_t error = 0;
+	Error error;
```
m-m-adams, line 835: should this be initialized?

PR #1305, opened 2024-02-23
Make number of count-in bars configurable

pmatila:
First-time contributor to Deluge here (Turre on Discord), so apologies if I've missed something in the community protocol.

As for the feature: I don't know how people get by with just one bar of count-in :flushed: Except for very slow songs, I need that two bars to get my hands in position and stuff. Maybe everybody else just uses a foot-switch to control :smile: Anyway, this changes the on/off toggle into an integer-with-off, allow up to four bars of count-in.

I repurposed the count-in enabled toggle for the purpose because this should be entirely backwards compatible and so adding a separate value for the number of bars seemed silly waste of a perfectly good byte to me :smile:

I was a bit surprised IntergerWithOff renders the off-state with a zero, "Off" would seem more appropriate in at least this case. I can add that to the PR if that would be preferred.

For simplicity's sake I just made the countdown ticks a multiply of the bars, a fancier way could be counting down bars and ticks separately.


sapphire-arches:
Hi! Sorry it took so long to review this.

Looks like the format check is cranky, you can run `./dbt format` to run a clang-format pass that should match CI. Otherwise, this change LGTM.

pmatilai:
Oh, I didn't realize 'dbt format' was a thing, and a thing you're supposed to run. Maybe something to put into CONTRIBUTING.md. Anyway, formatted (one of my includes wasn't in alphabetical order) and rebased. 

I actually consider this a rather speedy review, no worries on that front. Thanks for the review!

PR #1314, opened 2024-02-23
Introduce D_TRY and D_TRY_CATCH macros

stellar-aria:
This standardizes the error handling system in a way that is both recognizeable and uses the current Deluge Error type

### DIFF
--- a/src/definitions_cxx.hpp
+++ b/src/definitions_cxx.hpp
@@ -566,16 +566,40 @@ enum class Error {
 };
 
 #define D_TRY(expr)                                                                                                    \
-	error = (expr);                                                                                                    \
-	if (error != Error::NONE) {                                                                                        \
-		return error;                                                                                                  \
-	}
+	do {                                                                                                               \
+		Error error = (expr);                                                                                          \
+		if (error != Error::NONE) {                                                                                    \
+			return error;                                                                                              \
+		}                                                                                                              \
+	} while (0)
 
-#define D_TRY_CATCH(expr, block)                                                                                       \
-	error = (expr);                                                                                                    \
-	if (error != Error::NONE) {                                                                                        \
-		block                                                                                                          \
-	}
+#define D_TRY_FINALLY(expr, finally)                                                                                   \
+	do {                                                                                                               \
+		Error error = (expr);                                                                                          \
+		(finally);                                                                                                     \
+		if (error != Error::NONE) {                                                                                    \
+			return error;                                                                                              \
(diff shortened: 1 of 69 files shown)


src/deluge/gui/context_menu/save_song_or_instrument.cpp
```diff
@@ -49,12 +49,11 @@ bool SaveSongOrInstrument::acceptCurrentOption() {
 
 	case 1: { // Create folder
 		Browser* browser = (Browser*)getUIUpOneLevel();
-		Error error = browser->createFolder();
-
-		if (error != Error::NONE) {
+		Error error;
+		D_TRY_CATCH(browser->createFolder(), {
```
robmccoll, line 53: This looks is a little strange now for two reasons:
- The macro just expects Error error to exist in scope (which could create some fun compilation errors if you don't know about that and are using it for the first time)
- If you don't know that the macro uses / assigns error, it would appear reading this that error is never assigned.

It's a subtle change, but I'd recommend changing the macro to accept a name and have it declare the Error internally like:
```
#define D_TRY_CATCH(expr, error, block)    \
	Error error = (expr);                    \
	if (error != Error::NONE) {                 \
		block                                   \
	}
```

Usage:
```
D_TRY_CATCH(browser->createFolder(), error, {
  display->displayError(error);
  close();
  return false;
});
```

Added bonus: one line of code fewer per usage.

src/definitions_cxx.hpp
```diff
@@ -565,6 +565,18 @@ enum class Error {
 	SD_CARD_NO_FILESYSTEM,
 };
 
+#define D_TRY(expr)                                                                                                    \
+	error = (expr);                                                                                                    \
+	if (error != Error::NONE) {                                                                                        \
+		return error;                                                                                                  \
+	}
```
robmccoll, line 572: For this version, again, it's a little strange to have the macro depend on Error error being in scope. What would you think of:
```
#define D_TRY(expr)                              \
do {                               \
	Error error = (expr);                                              \
	if (error != Error::NONE) {                                         \
		return error;                                                      \
	}                          \
} while(0);
```

src/deluge/gui/context_menu/save_song_or_instrument.cpp
robmccoll: Also if you do this, I'd probably wrap it in a `do { } while(0};` setup to scope it so that you can safely reuse the same error name when using the macro multiple times in the same scope.

PR #1315, opened 2024-02-23
Fix enumeration menu item value and offset wrapping bug

robmccoll:
With the current code, if you go to the Community Features => Sysex menu (so that you are looking at the on or off option), hold shift, and scroll select up and down, the firmware will crash.

This seems to be because the code below doesn't properly handle wrapping the selected value to the [0, size()) range. I'm not completely sure my reading of what the code was trying to do is accurate (W.R.T. the sign and switch) or why anything would be different on OLED vs. 7SEG, so I may be missing something here.

This does fix the bug described above (setValue is never called with an invalid value), and to me seems reasonable (recalculate the offset as the difference between the old and new values before calling into the value).

### DIFF
--- a/src/deluge/gui/menu_item/enumeration.cpp
+++ b/src/deluge/gui/menu_item/enumeration.cpp
@@ -13,48 +13,19 @@ void Enumeration::beginSession(MenuItem* navigatedBackwardFrom) {
 }
 
 void Enumeration::selectEncoderAction(int32_t offset) {
-	this->setValue(this->getValue() + offset);
 	int32_t numOptions = size();
-	int32_t sign = (offset < 0) ? -1 : ((offset > 0) ? 1 : 0);
+	int32_t startValue = getValue();
 
-	switch (numOptions) {
-	case 0:
-		[[fallthrough]];
-	case 1:
-		[[fallthrough]];
-	case 2:
-		offset = 1 * sign;
-		break;
-	case 3:
-		offset = std::min<int32_t>(offset * sign, 2) * sign;
-		break;
-	case 4:
-		offset = std::min<int32_t>(offset * sign, 3) * sign;
-		break;
-	case 5:
-		offset = std::min<int32_t>(offset * sign, 4) * sign;
-		break;
-	default:
-		break;
+	// valid values are [0, numOptions), so wrap as needed
+	int32_t nextValue = (startValue + offset) % numOptions;
+	if (nextValue < 0) {
+		nextValue += numOptions;
 	}
 
-	if (display->haveOLED()) {
-		if (this->getValue() >= numOptions) {
-			this->setValue(numOptions - 1);
-		}
-		else if (this->getValue() < 0) {
-			this->setValue(0);
-		}
-	}
-	else {
-		if (this->getValue() >= numOptions) {
-			this->setValue(this->getValue() - numOptions);
-		}
-		else if (this->getValue() < 0) {
-			this->setValue(this->getValue() + numOptions);
-		}
-	}
+	setValue(nextValue);
 
+	// reset offset to account for wrapping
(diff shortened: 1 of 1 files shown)


robmccoll:
Oh no - I got the start value - nextValue backwards 🤦‍♂️

robmccoll:
@sapphire-arches - any reason both shouldn't wrap? Wrapping seems pretty consistent with most menu systems I've ever used.

ok-reza:
i think there are two solid benefits for not wrapping:

- it's clearer on 7seg that you've reached the end of a list/folder. when you only have 4 letters displayed at a time, it's harder to know you aren't just looping over the same items without realizing it. eg. when going through every sample in a sample folder, you reach an end point and won't accidentally keep previewing already previewed sounds.
- being able to spin the encoder all the way and know you've reached the end is useful. for instance, eg. i can enter the main menu and scroll clockwise many times and know i'll land on firmware version entry. or for 2 option menu's (enabled vs disabled or on vs off) turn clockwise will ensure off and counterclockwise will always ensure on. 

that being said, wrapping is really nice is some cases.

there's a wrapping discussion i brought up here if you'd like to discuss further: https://github.com/SynthstromAudible/DelugeFirmware/discussions/515

robmccoll:
Perhaps I've ended up down a rabbit hole here - the real goal of this PR is just to fix the crash that occurs in this code. I can re-introduce the OLED doesn't wrap, 7SEG wraps if we want and then we can worry about changing the behavior later. At the very least, 7SEG should probably continue to wrap in this code since the other menu items will also wrap.

src/deluge/gui/menu_item/enumeration.cpp
```diff
@@ -13,48 +13,25 @@ void Enumeration::beginSession(MenuItem* navigatedBackwardFrom) {
...
-		break;
-	case 4:
-		offset = std::min<int32_t>(offset * sign, 3) * sign;
-		break;
-	case 5:
-		offset = std::min<int32_t>(offset * sign, 4) * sign;
-		break;
-	default:
-		break;
-	}
```
sapphire-arches, line 39: I think we still need this offset clamp logic, otherwise small menus can end up being unnavigable with the shift+select scroll shortcut.

robmccoll: Maybe we should do something alternative - shift + select offsets by 5 right? Scrolling by n-1 on a short list of n when wrapping is confusing. If I have 1,2,3,4, I'm on 2, and I shift scroll forward, the net effect is scrolling backwards by 1. How about if n <= 5, we just make shift scroll jump to the beginning or end of the list? On OLED with no wrapping, this is the behavior you'd get anyway.

ok-reza:
that may be a great solution for allowing no wrapping across the board, shift scroll to find the end or start, otherwise menus wrap with normal scroll??

sapphire-arches:
Yeah, that sounds good to me!

To be precise, the behavior we want is:
  - |offset| = 1 means scroll with wrapping is allowed
  - |offset| >= 1:
    - scrolling by offset would wrap: go to at most the end/beginning of the list
    - otherwise, scroll by the offset

Sorry to go back and forth on this one so much :sweat_smile:

robmccoll:
@sapphire-arches Glad we spent a little more time and found a good solution!

sapphire-arches:
I think this looks good. If you're feeling adventurous, it would be good to audit the other overrides of `selectEncoderAction` to make sure they behave the same. If you don't have time we can merge this as-is and file a bug to follow up on that.

m-m-adams:
Gonna merge since it fixes a crash, auditing the menus to make sure they match can be a future problem

I agree with reza that this is a good approach in general

m-m-adams (approved):

PR #1319, opened 2024-02-24
MIDI Transpose follow up.

0beron:
For cherrypick into 1.1

MIDI transpose clips get priority in playback - in session and arranger, all clips are looked at in two passes, and those for which Output::needsPriority() returns true are actioned in the first pass. This currently only applies to MIDI clips routed internally to transpose.
In session there are 4 passes, in the order:

1. session clips + priority
2. arranger only clips + priority
3. other session clips 
4. other arranger clips

I think this is required to allow a white clip in arranger that is sending MIDI -> transpose will correctly influence the other clips in either arranger or session.

Updated docs for MIDI transpose to describe playback ordering, but that notes still get cut off.

Fixed a string bug I observed in automation view on 7seg.

Fixed audition pad update in automation view when MIDI transpose changes the root note.


src/deluge/model/output.h
```diff
@@ -154,6 +154,7 @@ class Output {
 	virtual ModelStackWithAutoParam* getModelStackWithParam(ModelStackWithTimelineCounter* modelStack, Clip* clip,
 	                                                        int32_t paramID,
 	                                                        deluge::modulation::params::Kind paramKind) = 0;
+	virtual bool needsPriority() { return false; }
```
sapphire-arches, line 157:
```suggestion
    /// When true, this clip must be ticked in the early pass, as it affects other clips in the song.
	virtual bool needsEarlyPlayback() const { return false; }
```

`needsEarlyPlayback` might be a better name, and should be marked `const`.

0beron: Made that change, and now found an issue where audition notes get stuck. Converted to draft until I fix that

PR #1320, opened 2024-02-24
Fix Mutable Reverb HPF Noise (#1289)

### LINKED ISSUE #1289: Strange noise that shifts in tone while paused

**UPDATE:** it seems to be connected to the mutable reverb. changing the reverb model back to Freeverb gets rid of the noise.

After playing this song for a couple of bars and then stopping playback, wait for ca 10 seconds, then a strange noise can be heard while the Deluge is idle. the color of the noise even changes when hitting playback (you hear it when no clip is playing but the deluge itself plays). this noise doesnt go away, only reloading the song works.

https://drive.google.com/file/d/10M8WfIiOpqsLMnfVa5rqiyojEh0Uuu5S/view?usp=sharing

no noise

no

OLED

Nightly

Deluge Beta Release 20240221

load song
play for 4 bars
stop playback
listen for noise

stellar-aria:
Fixes #1289


m-m-adams (approved):
Lgtm

Issue #1307, opened 2024-02-23
Request: make Pressure on MPE keys update the velocity of the arpeggiated notes


soymonitus:
### Please describe the problem:

It would be nicer if the pressure expression parameter from an MPE controller, when holding notes with your fingers, could update the velocity sent by the post-arpeggiator notes. Right now if you initially play softly the chord to be arpeggiated, it doesn't matter if later you increase pressure on the keys, all post-arp notes will sound super quiet.

### What is the expected behavior?

Notes get higher velocity as you increase pressure on the keys of your MPE controller

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

soymonitus:
Not sure if it is considered a bug or a nice-to-have. Anyways I think it would be nicer, if your controller is MPE capable, that the arpeggiated notes generated by the deluge express better you current finger positions when a new note is sent out.

m-m-adams:
I agree, it works this way on the linnstrument arpeggiator and it's quite nice

Issue #1306, opened 2024-02-23
When live playing an arpeggiated synth clip with an MPE controller, the arp notes don't send MPE expressions


soymonitus:
### Please describe the problem:

MPE expression values are only sent when you change pressure on the keys of your controller for a single note of the arp. Subsequent notes don't get the expression of your current finger pressure. Say you add more pressure to a key, it will send aftertouch = 127 for the current note that happens to play in the arpeggiator, but after that, next notes have aftertouch=0 even if you have the key fully pressed.  UPDATE: THIS ONLY HAPPENS WHEN SELECTING NOTE MODE = AS PLAYED

### What is the expected behavior?

When a new post-arpeggiator note is sent to the synth, it should also send the "last known" expression value. Not tested other than aftertouch, but pretty sure other expression params like bend and Y are not sent either.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

soymonitus:
Since I am working on MPE arpeggiator improvements I want to auto assign this to me

PR #1326, opened 2024-02-24
Add setting to map MPE Aftertouch or Y, to velocity when doing arps. And fixed bug with MPE arp expressions not sent with "As played" mode.

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1306
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1307

Both issues were kind of a mixture of several things.

From the start, MPE pressure from an MPE keyboard was not used for setting the velocity of new notes which was a missed opportunity and made the arpeggiator experience feel like half-baked. Now you can setup in Arpeggiator -> MPE -> Pressure to Velocity if you want pressure to set velocity for each new note. Disabled by default, because players with aftertouch keyboards may not want to let aftertouch control velocity, as they would need to play super hard on the keys to get normal velocity.

On the other hand, when using the noteMode "As played", as the mpeValues were not updated when a polyphonic expression event was received (only for ascending-ordered notes) so the playing was buggy, with live values being overwriten by the initial mpe values.

NOTE: needs to be cherry picked


soymonitus:
image
Any idea why the "mpeAtToV" setting is written to the synth patch but it is not read? Always that i go to arp -> mpe -> pressure to velocity, I find it Disabled. Maybe some conversion thing from bool to int? or maybe some missed place where to read?

soymonitus:
@m-m-adams ready to review. Right now MPE->Velocity-> can be set to "Disabled", "Aftertouch" or "MPE Y". The code is prepared for the future parameters that will allow also "MPE X". In the next PR I will add support to modulate the others (as you said it was preferred to keep PRs simple)

src/definitions_cxx.hpp
```diff
@@ -620,6 +620,13 @@ enum class ArpOctaveMode {
 	RANDOM,
 };
 
+enum class ArpMpeModSource {
+	OFF,
+	AFTERTOUCH,
+	MPE_Y,
+	MPE_X,
```
m-m-adams, line 627: I'm not sure about this - I think the current behaviour of bending the note is better. If there's an edge case where it doesn't work we should fix it

soymonitus: I can remove it. Actually it would be weird to have bending and at the same time some modulation. We can leave it just as aftertouch or Y for all.params

m-m-adams: Yeah I agree

soymonitus: changes done

m-m-adams (approved):

PR #1338, opened 2024-02-25
Refactor Delay and DelayBuffer

stellar-aria:
General refactoring of dsp/delay facilities.

Including:
- Moving state classes to inner classes
- Use of references
- Increasing separation of concerns with getters/setters
- Conditional branch rearrangement for early return and reduced nesting


src/deluge/dsp/delay/delay_buffer.h
```diff
@@ -19,53 +19,65 @@
...
 
 class DelayBuffer {
 public:
-	DelayBuffer();
-	~DelayBuffer();
+	constexpr static size_t kMaxSize = 88200;
+	constexpr static size_t kMinSize = 1;
+	constexpr static size_t kNeutralSize = 16384;
+
+	struct Config {
```
m-m-adams, line 34: I think this should stay out of the class, or be set seperately and not passed as an argument to the functions that need it. It's misleading that the delaybuffer has a config but the functions need to be passed a config as well

src/deluge/dsp/delay/delay.cpp
```diff
@@ -17,190 +17,164 @@
...
+		// which is easily done now, before anything's been written
+		if (!primaryBuffer.isActive() && secondaryBuffer.isActive()
+		    && sizeLeftUntilBufferSwap == getAmountToWriteBeforeReadingBegins()) {
 
-				int32_t idealBufferSize = secondaryBuffer.getIdealBufferSizeFromRate(userDelayRate);
-				idealBufferSize = std::min(idealBufferSize, (int32_t)DELAY_BUFFER_MAX_SIZE);
-				idealBufferSize = std::max(idealBufferSize, (int32_t)DELAY_BUFFER_MIN_SIZE);
+			int32_t idealBufferSize = DelayBuffer::getIdealBufferSizeFromRate(userDelayRate);
+			idealBufferSize = std::min(idealBufferSize, (int32_t)DelayBuffer::kMaxSize);
+			idealBufferSize = std::max(idealBufferSize, (int32_t)DelayBuffer::kMinSize);
```
sapphire-arches, line 57: Is it ever correct to use the `IdealBufferSizeFromRate` without this clamping? If not we should fold it in.

src/deluge/dsp/delay/delay.h
```diff
@@ -17,28 +17,38 @@
...
-struct DelayWorkingState {
-	bool doDelay;
-	int32_t userDelayRate;
-	int32_t delayFeedbackAmount;
-};
-
 class Delay {
 public:
-	Delay();
-	void cloneFrom(Delay* other);
```
sapphire-arches, line 33: :partying_face:

src/deluge/dsp/delay/delay.cpp
```diff
@@ -17,190 +17,164 @@
...
-			workingState->userDelayRate <<= (syncLevel + 5);
+			workingState.userDelayRate <<= (syncLevel + 5);
 		}
 	}
 
 	// Tell it to allocate memory if that hasn't already happened
-	informWhetherActive(mightDoDelay, workingState->userDelayRate);
-	workingState->doDelay = isActive(); // Check that ram actually is allocated
+	informWhetherActive(mightDoDelay, workingState.userDelayRate);
+	workingState.doDelay = isActive(); // Check that ram actually is allocated
```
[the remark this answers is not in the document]
stellar-aria: Fixed!

src/deluge/dsp/delay/delay.cpp
stellar-aria: Done.

src/deluge/dsp/delay/delay_buffer.h
stellar-aria: I've refactored this so that it's clear this is a _resampling_ config. The buffer no longer has intrinsic knowledge of if it's resampling or not, instead the difference between buffer write depends on the presence of the config (as an optional)

src/deluge/dsp/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set_target_properties(deluge_dsp
         CXX_STANDARD ${DELUGE_CXX_STANDARD}
         CXX_STANDARD_REQUIRED ON
         CXX_EXTENSIONS ON
-        UNITY_BUILD true
+#        UNITY_BUILD true
```
m-m-adams, line 26: :(

src/deluge/dsp/delay/delay_buffer.h
m-m-adams: This is clearer, but as far as I can tell there's still nothing that actually uses the config inside the delay class, they all refer to a passed in config instead - I think it should just be on its own in this file

m-m-adams: Disregard - misread

m-m-adams (approved):
Excellent

src/deluge/dsp/delay/delay.cpp
```diff
@@ -221,3 +199,264 @@ void Delay::discardBuffers() {
...
+		sample.l -= postLPFL;
+
+		int32_t distanceToGoR = sample.r - postLPFR;
+		postLPFR += distanceToGoR >> 11;
+		sample.r -= postLPFR;
+	}
+
+	// Go through what we grabbed, sending it to the audio output buffer, and also preparing it to be fed back
+	// into the delay
+	for (auto [input, output] : std::views::zip(working_buffer, buffer)) {
```
m-m-adams, line 372: I'm truly impressed that this doesn't blow up our bin size

m-m-adams (approved):

PR #1340, opened 2024-02-25
Fixed Scale cycling, which was not possible from OTHER SCALE even if you emptied notes

soymonitus:
Some user in discord complained that he had added more than 7 notes, getting him into the "OTHER SCALE" type of scale. But after he removed the extra notes from the clip, he couldn't do Shift + Scale to go back to normal scales. This PR fixes this by taking into account always, even if coming from OTHER SCALES, the real number of used notes.

NOTE: needs to be cherry picked as it is a fix for the "New scales" feature in 1.1

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -3054,7 +3054,7 @@ int32_t Song::getCurrentPresetScale() {
 		// If we're here, must be this one!
 		return p;
 
-notThisOne : {}
+notThisOne: {}
 	}
 
 	return CUSTOM_SCALE_WITH_MORE_THAN_7_NOTES;
@@ -4972,7 +4972,7 @@ Instrument* Song::changeOutputType(Instrument* oldInstrument, OutputType newOutp
 			return NULL;
 		}
 
-gotAnInstrument : {}
+gotAnInstrument: {}
 	}
 
 	// Synth or Kit
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -2915,36 +2915,24 @@ const char* Song::getScaleName(int32_t scale) {
 
 int32_t Song::cycleThroughScales() {
 	int32_t currentScale = getCurrentPresetScale();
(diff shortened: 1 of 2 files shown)


src/deluge/model/song/song.cpp
```diff
@@ -2915,37 +2915,25 @@ const char* Song::getScaleName(int32_t scale) {
...
-		return 255;
-	}
 	int32_t newScale = currentScale + 1; // next scale
 	if (newScale >= NUM_PRESET_SCALES) {
 		// If past the last scale, start from the beginning
 		newScale = 0;
 	}
-	// Set it, it will
-	return setPresetScale(newScale);
+	int32_t result = setPresetScale(newScale);
```
m-m-adams, line 2923: use the error enum

soymonitus: The method returns the scale that it was able to set given the current conditions. It is not just errors, or is the error enum also used for success values?

m-m-adams: Oh ok - make a new enum class instead then. Don't just return ints with special meaning

soymonitus: But what about the scales indexes? if it returns from 0 to (NUM_SCALES - 1) as valid, that is, the method could apply a known scale, but if not, there are two possible error cases. Should I return a tuple? first value is the applied preset scale and second value the error enum?

soymonitus: The old code was returning 255 to signal a CUSTOM scale, because all the other parts of the code check for:

int32_t changedScale = setPresetScale(newScale)
**if (changedScale >= NUM_PRESET_SCALES) {
    // do sth with Custom scale case**
    
 I can do the refactor in all places but not sure if worth it

soymonitus: Found a better way to approach it, without needing that constant to represent a special error. Commit pushed.

m-m-adams: Great, I like it!

m-m-adams (approved):

PR #1349, opened 2024-02-26
only send midi if changed

### LINKED ISSUE #1333: Note Off is also sending out the current channel aftertouch/bend/modwheel values

With every note through midi out, the deluge is sending out the current channel aftertouch/bend/modwheel values at note offs.  This is just extraneous data.

Just note off, nothing else.

_No response_

OLED

Nightly

C1.1.0

_No response_

m-m-adams:
Fix #1333, only send value on note off when the note off changes it


sapphire-arches (approved):
LGTM, do we need this in 1.1?

m-m-adams:
Yeah I think so - it's harmless but at least a couple people seem bothered by it

PR #1350, opened 2024-02-26
Preserve Pad Brightness Preference Through Power Cycle

joxihan:


stellar-aria:
Instead of having an explicit default brightness, can we just have setDimmerInterval instead store the newly set value in the FlashStorage, and do something to mark the settings as unsaved, and then save at a later point when there's nothing going on to be interrupted?

That way the last set brightness will simply persist across resets, rather than defaulting to a value that may not be correct for all situations.

sapphire-arches:
I'm a little nervous about introducing implicit writes to flash from UI code, and retaining the ability to have transient as well as default brightness levels seems useful.

m-m-adams:
I agree - having a default and a temporary value seems useful to me

joxihan:
a bit confused, do i need to make a change or not at the end ? :p

stellar-aria:
😅 Looks like I'm outvoted, in which case...

LGTM :shipit:

stellar-aria (approved):

PR #1359, opened 2024-02-27
Extract kMaxSample value (2**24)

stellar-aria:
Just some magic number cleanup


src/deluge/dsp/compressor/rms_feedback.h
```diff
@@ -91,8 +91,8 @@ class RMSFeedbackCompressor {
 
 private:
 	// parameters in use
-	float a_ = (-1000.0 / 44100);
-	float r_ = (-1000.0 / 44100);
+	float a_ = (-1000.0 / kSampleRate);
```
m-m-adams, line 94: oops - these should be -1000.0f if you don't mind changing it, I think they're doubles right now

m-m-adams (commented):
LGTM but if you could fix my mistake that would be nice too!

src/deluge/dsp/compressor/rms_feedback.h
stellar-aria: Fixed!

PR #1363, opened 2024-02-27
Move to a more native C++ FatFS adapter

stellar-aria:
This introduces a new RAII-style FatFS layer, and will port existing code to use it where appropriate


src/deluge/gui/ui/save/save_song_ui.cpp
```diff
@@ -352,18 +356,15 @@ bool SaveSongUI::performSave(bool mayOverwrite) {
...
 						}
 						if (!bytesRead) {
 							break; // Stop, on rare case where file ended right at end of last cluster
 						}
 
-						UINT bytesWritten;
-						result = f_write(&recorderFileSystemStuff.currentFile, storageManager.fileClusterBuffer,
-						                 bytesRead, &bytesWritten);
-						if (result || bytesWritten != bytesRead) {
+						auto written = created.value().write({(std::byte*)storageManager.fileClusterBuffer, bytesRead});
```
m-m-adams, line 366: is there an unwrap equivalent so that this can just be created.write? We've already checked that opening succeeded in this path

src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -1465,8 +1467,9 @@ Error SampleRecorder::alterFile(MonitoringAction action, int32_t lshiftAmount, u
 
 		if (action != MonitoringAction::NONE || capturedTooMuch) {
 
-			FRESULT fres = f_open(&file, sample->filePath.get(), FA_WRITE);
-			if (fres) {
+			auto opened = this->file->open(sample->filePath.get(), FA_WRITE);
```
m-m-adams, line 1470: is this not just file->open?

src/deluge/model/sample/sample_recorder.h
```diff
@@ -118,11 +116,11 @@ class SampleRecorder {
 
 	int32_t* sourcePos;
 
-	FIL file;
+	std::optional<FatFS::File> file;
 
 private:
 	void setExtraBytesOnPreviousCluster(Cluster* currentCluster, int32_t currentClusterIndex);
-	Error writeCluster(int32_t clusterIndex, int32_t numBytes);
+	Error writeCluster(int32_t clusterIndex, size_t numBytes);
```
m-m-adams, line 123: this is a change from signed to unsigned - is that intentional?

src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -749,24 +753,22 @@ void SampleRecorder::updateDataLengthInFirstCluster(Cluster* cluster) {
 extern int32_t pendingGlobalMIDICommandNumClustersWritten;
 
 // You'll want to remove the "reason" after calling this
-Error SampleRecorder::writeCluster(int32_t clusterIndex, int32_t numBytes) {
+Error SampleRecorder::writeCluster(int32_t clusterIndex, size_t numBytes) {
```
m-m-adams, line 756: Changed signed -> unsigned

src/deluge/model/sample/sample_recorder.cpp
stellar-aria: it is but I'm having trouble distinguishing local `file` from instance `file` atm, so it's `this->file` until I can change it to `file_`

src/deluge/gui/ui/save/save_song_ui.cpp
stellar-aria: actually, yeah. the dereference operators `*` and `->` haha

src/deluge/model/sample/sample_recorder.h
stellar-aria: yes, this is safe, it's only used when the size is known to be `> 0`

src/deluge/model/sample/sample_recorder.cpp
stellar-aria: Checked to be safe

PR #1365, opened 2024-02-27
Performance/dont cull drones as often

m-m-adams:
Sets notes in "hold" when the engine detects that they're currently sustaining and receive a note on which would otherwise be ignored

Avoids culling long drones because their "time entered stage" gets increasingly long compared to newly held notes, but the sound of a drone dropping out can be much more disruptive

### DIFF
--- a/src/deluge/model/drum/drum.h
+++ b/src/deluge/model/drum/drum.h
@@ -68,7 +68,7 @@ public:
 	virtual void noteOff(ModelStackWithThreeMainThings* modelStack, int32_t velocity = kDefaultLiftValue) = 0;
 	virtual bool allowNoteTails(ModelStackWithSoundFlags* modelStack, bool disregardSampleLoop = false) = 0;
 	virtual bool anyNoteIsOn() = 0;
-	virtual bool hasAnyVoices() = 0;
+	virtual bool hasAnyVoices(bool resetTimeEntered) = 0;
 	virtual void unassignAllVoices() = 0;
 
 	virtual Error loadAllSamples(bool mayActuallyReadFiles) { return Error::NONE; }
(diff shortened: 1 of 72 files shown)


src/deluge/model/drum/drum.h
```diff
@@ -68,7 +68,7 @@ class Drum {
 	virtual void noteOff(ModelStackWithThreeMainThings* modelStack, int32_t velocity = kDefaultLiftValue) = 0;
 	virtual bool allowNoteTails(ModelStackWithSoundFlags* modelStack, bool disregardSampleLoop = false) = 0;
 	virtual bool anyNoteIsOn() = 0;
-	virtual bool hasAnyVoices() = 0;
+	virtual bool hasAnyVoices(bool resetTimeEntered) = 0;
```
sapphire-arches, line 71: The reset logic should probably go in a separate function

src/deluge/modulation/envelope.cpp
```diff
@@ -139,6 +142,12 @@ void Envelope::noteOff(uint8_t envelopeIndex, Sound* sound, ParamManagerForTimel
 		unconditionalRelease();
 	}
 }
+/// resets time to give drone notes a lower likelihood of being bulled, since they'll stay in sustain as they loop
```
sapphire-arches, line 145:
```suggestion
/// resets time to give drone notes a lower likelihood of being culled, since they'll stay in sustain as they loop
```

src/deluge/model/drum/drum.h
m-m-adams: Yeah reasonable. I copied the synth approach but the drum check for voices is a lot simpler

m-m-adams: done

PR #1366, opened 2024-02-28
Refactor/preset loading

m-m-adams:
Set preset loading with config functions instead of public member variables

Reset state when switching load types, fixes a bug where loading a synth to a kit would copy the previously loaded synth's param manager instead of the previous row

Use correct param manager for loading synth to kit, stopping a crash when no instrument has previously been loaded

Improve UX of loading synth to row:

- add title SYNTH TO ROW
- Blink synth and kit buttons
- Save the folder the drum was in so it's easier to replace


seangoodvibes:
Looks good. Just a minor bug when holding audition and hitting save, the popup gets stuck on the display. Probably just need to add a cancelPopup call.

seangoodvibes (approved):

PR #1374, opened 2024-02-29
Add Automation View for Patch Cables / Modulation Depth

seangoodvibes:
Added ability to jump from the Sound menu for Patch Cables / Modulation Depth into automation view.

For patch cables / modulation depth, the grid value ranges for each pad have been adapted to accomodate the full -50 to +50 range.

The bottom pad in the grid will set the value to -50 and the top pad in the grid will set the value to +50.

You can set the value to 0 by pressing and holding the two middle pads.

The LED indicators for the mod encoders have been updated to show the full -50 to +50 range as well.

This diagram shows the updated value ranges for the pads when automating a patch cable / modulation depth.

Screenshot 2024-02-28 at 7 43 23 PMScreenshot 2024-02-28 at 7 47 12 PM

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -71,6 +71,7 @@
 #include "model/settings/runtime_feature_settings.h"
 #include "model/song/song.h"
 #include "modulation/automation/auto_param.h"
+#include "modulation/params/param.h"
 #include "modulation/params/param_manager.h"
 #include "modulation/params/param_node.h"
 #include "modulation/params/param_set.h"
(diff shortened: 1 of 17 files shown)


src/deluge/gui/views/automation_view.cpp
```diff
@@ -3294,6 +3326,12 @@ void AutomationView::setParameterAutomationValue(ModelStackWithAutoParam* modelS
 // deleting automation, or displaying current parameter value
 // multi pad presses don't use this function
 void AutomationView::setKnobIndicatorLevels(int32_t knobPos) {
+	// if you're dealing with a patch cable which has a -128 to +128 range
+	// we'll need to convert it to a 0 - 128 range for purpose of rendering on knob indicators
+	if (!onArrangerView && (getCurrentClip()->lastSelectedParamKind == params::Kind::PATCH_CABLE)) {
+		float floatKnobPos = kMaxKnobPos * ((static_cast<float>(knobPos) + kMaxKnobPos) / (kMaxKnobPos * 2));
```
m-m-adams, line 3332: Ideally they'd show bipolar, maybe negative values fill in from the top. 

Currently this doesn't match the display of learning gold knobs to patch cables either (which is positive only). I think they need to match either way - so only positives in automation right now, or fix both

seangoodvibes: I will be fixing both so that the gold knobs outside automation view are bipolar as well. Thought it would be best to do that in a separate PR. It should be bipolar so that the middle of the LED indicators is 0, just like with pan.

m-m-adams (approved):

PR #1382, opened 2024-02-29
Allow Gold Knobs to Access Full Patch Cable / Mod Depth Range

seangoodvibes:
- [x] Allow gold knobs to access full patch cable / mod depth range (-50 to +50 / -128 to +128)
- [x] Update LED indicators to display full patch cable / mod depth range (-50 to +50 / -128 to +128)
- [x] Update mod encoder pop-ups so that they don't display if you're editing the same parameter in the menu
- [x] Fix bug where mod encoder pop-ups still display if you're editing the same patch cable / mod depth in the menu
- [x] Display patch cable source(s) and destination on the mod-encoder pop-up for OLED (e.g. Source -> Destination : ###)
- [x] Fixed bug where if root UI is automation view, switching parameter menu's by backing out of one menu into another one wouldn't refresh the background automation editor grid
- [x] Fix bug where LED indicators turn off when blinking the middle value (0)

There will be one more follow-up PR where the above changes will made for MIDI as well - specifically: I want to allow external midi that is learned to a patch cable to control the full range (-50 to +50).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,10 +75,11 @@ here: [Community Features](https://github.com/SynthstromAudible/DelugeFirmware/b
 - Updated the count-in setting to allow specifying the number of bars (1-4).
 - Added `VU Meter` rendering in the sidebar in Song / Arranger / Performance Views.
 - Added ability to save a synth/sample drum back to an instrument preset by holding audition and pressing save.
+- Mod (Gold) Encoders learned to the Mod Matrix can now access the full range of the Mod Matrix / Patch Cable parameters (e.g. from -50 to +50).
 
 In addition, a number of improvements have been made to how the OLED display is used:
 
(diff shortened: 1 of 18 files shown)


soymonitus:
When you say "for midi as well"... would you add the possibility to learn arpeggiator UnpatchedParams to the golden knobs of a midi clip? that would be neat

seangoodvibes:
> When you say "for midi as well"... would you add the possibility to learn arpeggiator UnpatchedParams to the golden knobs of a midi clip? that would be neat

When I said for midi as well. I meant being able to learn an external controller to the full mod depth range. I don't think that's currently possible.

I think learning your arp params to a mod encoder first requires that your midi arp menu's have the right midi learn functions in them. When I look at your midi arp params in the menu, they don't look learnable to me (eg there's no value bar on OLED)

stellar-aria (approved):

PR #1385, opened 2024-02-29
RecorderStatus enum class

stellar-aria:
Creates an enum class for the RecorderStatus

### DIFF
--- a/src/deluge/model/sample/sample_recorder.h
+++ b/src/deluge/model/sample/sample_recorder.h
@@ -30,13 +30,16 @@ enum class MonitoringAction {
 	SUBTRACT_RIGHT_CHANNEL = 2,
 };
 
-#define RECORDER_STATUS_CAPTURING_DATA 0
-#define RECORDER_STATUS_CAPTURING_DATA_WAITING_TO_STOP 1
-#define RECORDER_STATUS_FINISHED_CAPTURING_BUT_STILL_WRITING 2
-#define RECORDER_STATUS_COMPLETE 3
-#define RECORDER_STATUS_ABORTED                                                                                        \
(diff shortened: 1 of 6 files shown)


src/deluge/model/sample/sample_recorder.h
```diff
@@ -83,19 +88,19 @@ class SampleRecorder {
...
+	bool autoDeleteWhenDone = false;
 	bool keepingReasonsForFirstClusters;
 	uint8_t recordingNumChannels;
-	bool hadCardError;
-	bool reachedMaxFileSize;
+	bool hadCardError = false;
+	bool reachedMaxFileSize = false;
 	bool recordingExtraMargins;
 	bool pointerHeldElsewhere;
 	bool capturedTooMuch;
```
m-m-adams, line 106: should these be initialized too?

stellar-aria: probably but I don't know what they were/should be

m-m-adams: Yeah I think false is right, they'll all be false after a successful abort/finish recording

sapphire-arches, line 106:
```suggestion
	bool recordingExtraMargins = false;
	bool pointerHeldElsewhere = false;
	bool capturedTooMuch = false;
```

sapphire-arches: I pushed a fix for this, were there any other concerns?

stellar-aria: not that I know of, :shipit:

m-m-adams (approved):

PR #1392, opened 2024-03-01
Possibly fix bug with switching rapidly between configs

stellar-aria:


stellar-aria:
Needs to have fixes against unchecked std::optional access (`*` and `->`) before merging

src/deluge/dsp/delay/delay_buffer.h
```diff
@@ -120,7 +120,7 @@ class DelayBuffer {
 			return;
 		}
 		// If delay buffer spinning above sample rate...
-		if (resample_config_.value().actualSpinRate >= kMaxSampleValue) {
+		if (resample_config_->actualSpinRate >= kMaxSampleValue) {
```
m-m-adams, line 123: Will this change not add a problem in that it's now entirely unchecked?

stellar-aria: No, it's checked up at the top of the function and early returns if the optional doesn't have a value.

m-m-adams: Ok perfect

m-m-adams (approved):

m-m-adams:
Fix #1406

PR #1394, opened 2024-03-01
RAII FatFS minus the filesystem corruption

stellar-aria:
Let's try this again


stellar-aria:
Same PR as before, this time with checked access using `.value()`

src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -696,8 +696,8 @@ Error SampleRecorder::finalizeRecordedFile() {
 				if (firstSampleCluster->sdAddress == 0) {
 					FREEZE_WITH_ERROR("E268");
 				}
-				if ((firstSampleCluster->sdAddress - fileSystemStuff.fileSystem->database)
-				    & (fileSystemStuff.fileSystem->csize - 1)) {
+				if ((firstSampleCluster->sdAddress - fileSystemStuff.fileSystem.value().database)
```
m-m-adams, line 699: I don't think calling .value() is safe either - https://en.cppreference.com/w/cpp/utility/optional/value

Docs say it throws if the optional isn't present

m-m-adams: This isn't UB (unlike ->) but will crash deluge

stellar-aria: if the filesystem isn't valid I'd rather it crash than try to corrupt the SDCard or something.

stellar-aria:
I think it'll be better to merge this series as single commits so any bugs can be tracked down

m-m-adams (approved):

PR #1400, opened 2024-03-02
Make startup song feature more robust

joxihan:


src/deluge/deluge.cpp
```diff
@@ -452,23 +452,30 @@ void setupBlankSong() {
 /// Can only happen after settings, which includes default settings, have been read
 void setupStartupSong() {
 	auto startupSongMode = FlashStorage::defaultStartupSongMode;
```
m-m-adams, line 454: For safety against defaulting to a song that crashes the deluge, I think you could write a canary file here and delete it when the load is done

m-m-adams (approved):
Merging as this is a straight improvement otherwise

PR #1413, opened 2024-03-03
failsafe mechanism for startup song feature

joxihan:
failsafe description :
- we first check if SONGS/__STARTUP_OFF_CHECK_{SONG} exists on sd card, if so, we fallback.
- a popup will alert the user that STARTUP mode has been switched off for this given song.
- Two options for the user  :
- > if load/save another song, it will load that one, previous canary file will be ignored.
- > can remove the canary file from song browser.

if no canary file at startup :
- we create one
- process the load
- remove it


m-m-adams (approved):
Awesome thanks! successfully does not reopen a song that crashes on boot

PR #1427, opened 2024-03-04
default pad brightness : 4-100 value from UI with 4-step offset

joxihan:
not sure if it counts as a bug fix, but based on feedback i got from people using it on nightly, might be a good addition to 1.1 considering the low risk

### DIFF
--- a/src/deluge/gui/menu_item/defaults/pad_brightness.h
+++ b/src/deluge/gui/menu_item/defaults/pad_brightness.h
@@ -23,12 +23,21 @@ namespace deluge::gui::menu_item::defaults {
 class PadBrightness final : public Integer {
 public:
 	using Integer::Integer;
-	[[nodiscard]] int32_t getMinValue() const override { return kMinLedBrightness; }
-	[[nodiscard]] int32_t getMaxValue() const override { return kMaxLedBrightness; }
-	void readCurrentValue() override { this->setValue(FlashStorage::defaultPadBrightness); }
-	void writeCurrentValue() override {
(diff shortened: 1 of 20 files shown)


src/deluge/gui/menu_item/defaults/pad_brightness.h
```diff
@@ -23,12 +23,21 @@ namespace deluge::gui::menu_item::defaults {
 class PadBrightness final : public Integer {
 public:
 	using Integer::Integer;
-	[[nodiscard]] int32_t getMinValue() const override { return kMinLedBrightness; }
-	[[nodiscard]] int32_t getMaxValue() const override { return kMaxLedBrightness; }
-	void readCurrentValue() override { this->setValue(FlashStorage::defaultPadBrightness); }
+	void selectEncoderAction(int32_t offset) override {
+		offset = offset * 4;
```
m-m-adams, line 27: Offset is already multiplied by 5 if you're holding shift - this shouldn't be needed

joxihan: OK maybe I can add this in the doc as testers were complaining about it and removing the *4 offset

sapphire-arches: this *4 is serving a different purpose, preventing the selection of brightness values that don't map to a valid interval setting. I think it's ok as is.

m-m-adams: Right, ok that makes sense

m-m-adams (approved):

PR #1428, opened 2024-03-04
Add ENABLE_SYSEX_LOAD docs to CONTRIBUTING.md

sapphire-arches:


joxihan (approved):
can load the fw via midi following these steps, thanks! (minus the size error, which is another story :p)

PR #1432, opened 2024-03-04
[Draft] Implement more arpeggiator parameter modulations by MPE expression

soymonitus:
Now also Gate, Octaves, Ratchet Amount and Ratchet Probability are eligible to be modulated by MPE expressions (Z or Y). The modulation range is fixed for simplicity and would be:
- Base minimum value: the one you set up in the main Arpeggiator menu
- Maximum value: The maximum value for the parameter (usually 50)

Special case for “Octaves” which has a fixed range of 0 to 3 octaves note shifting.

NOTE: also to cherry-pick

### DIFF
--- a/src/deluge/modulation/arpeggiator.cpp
+++ b/src/deluge/modulation/arpeggiator.cpp
@@ -21,6 +21,7 @@
 #include "model/song/song.h"
 #include "playback/playback_handler.h"
 #include "storage/flash_storage.h"
+#include "util/fixedpoint.h"
 #include "util/functions.h"
 #include "util/lookuptables/lookuptables.h"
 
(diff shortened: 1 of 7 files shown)


src/deluge/modulation/arpeggiator.cpp
```diff
@@ -772,32 +826,27 @@ bool ArpeggiatorForDrum::hasAnyInputNotesActive() {
...
-		ratchetAmount = 1;
-	}
-	else { // > 0 -> 4
-		ratchetAmount = 0;
+		// Update live ratchetProbability value with the most up to date value from automation
+		ratchetProbability = ratchProb >> 16; // just 16 bits is enough resolution for probability
+
+		// Update live ratchetAmount value with the most up to date value from automation
+		// Convert ratchAmount to either 0, 1, 2 or 3 (equivalent to a number of ratchets: OFF, 2, 4, 8)
+		ratchetAmount = ratchAm >> 16;
```
m-m-adams, line 849: Can you fix this bug seperately so it can be cherry picked? I don't think the rest of these additions should be moved across, we should aim to generalize the ability to modulate things pre note on with random and LFO1 instead.

soymonitus: Hmmm I can't find the bug you mention. This seems fine. I am using the provided ratchAmount function parameter to convert it to discrete values of ratchetAmount
```

void ArpeggiatorBase::render(ArpeggiatorSettings* settings, int32_t numSamples, uint32_t gateThreshold,
                             uint32_t phaseIncrement, uint32_t sequenceLength, **uint32_t ratchAmount**, uint32_t ratchProb,
                             ArpReturnInstruction* instruction) {
	if (settings->mode == ArpMode::OFF || !hasAnyInputNotesActive()) {
		return;
	}

	// Update live Sequence Length value with the most up to date value from automation
	maxSequenceLength = (((int64_t)sequenceLength) * kMaxMenuValue + 2147483648) >> 32; // in the range 0-50

	// Update live ratchetProbability value with the most up to date value from automation
	ratchetProbability = ratchProb >> 16; // just 16 bits is enough resolution for probability

	// Update live ratchetAmount value with the most up to date value from automation
	// Convert ratchAmount to either 0, 1, 2 or 3 (equivalent to a number of ratchets: OFF, 2, 4, 8)
	**uint16_t amount = ratchAmount >> 16;
	if (amount > 45874) { // > 35 -> 50
		ratchetAmount = 3;
	}
	else if (amount > 26214) { // > 20 -> 34
		ratchetAmount = 2;
	}
	else if (amount > 6553) { // > 5 -> 19
		ratchetAmount = 1;
	}
	else { // > 0 -> 4
		ratchetAmount = 0;
	}**
```

Issue #1429, opened 2024-03-04
Sysex firmware loading broken by bootloader alignment workaround


sapphire-arches:
The alignment hack introduced in #1411 breaks sysex firmware loading because that process relies on the size of the .bin file to determine the size to send, but the sent size might not match the size declared in the binary itself (as parsed by the loadfw infrastructure) because the padding is not included.

The bootloader might also be loading garbage off the disk in this situation, which is probably fine since we zero and treat that as heap memory, but that probably needs to be verified.

See also #1084.

m-m-adams:
Yeah the bootloader is reading garbage from flash. The firmware updater uses the actual file size and not the size reported in the binary so it's fine, 

Two possible fixes:
1. Have the sysex loader use the file size instead of "end"
2. actually put something at the location of "end" so the linker pads the file out

PR #1446, opened 2024-03-06
Fix SysEx Firmware Load

M0R172:
See #1429

Got rid of the explicit size check, as there is still the CRC check, which seems unlikely to succeed if the size is wrong.

### DIFF
--- a/scripts/tasks/task-loadfw.py
+++ b/scripts/tasks/task-loadfw.py
@@ -1,4 +1,3 @@
-import sys
 import time
 import binascii
 import argparse
@@ -42,7 +41,7 @@ def argparser():
     return parser
 
 
-def pack87(src, dstsize):
+def pack_8_to_7_bits(src, dstsize):
     packets = (len(src) + 6) // 7
     dst = bytearray(dstsize)
     for i in range(packets):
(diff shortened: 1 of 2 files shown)


M0R172:
Certainly possible, but I do not think it is necessary:


The first packet currently is not "special" as a packet, but it contains the "code start address" and "code end address" fields from the beginning of the binary file - these are used by the sysex loader on the Deluge to determine the expected data size and allocate the buffer. 

Before padding was introduced this was identical to the size of the binary file.

With the now introduced padding, the buffer might be a bit larger (by max 65kB) than strictly necessary, but it can never be too short, as far as I understand it, so there should not be any problems arising from that.

After the last data packet, we send the "check and load" SysEx message, and *this* contains the size of the actual binary file, as well as a CRC of the binary, calculated on the sending computer. I use this size to tell the data length to the CRC function on the recieving Deluge (so *correct length is implicitly checked by correctness of the CRC*). In the past, this was *additionally* compared to the difference of "code start address" and "code end address", but with padding, this is no longer the same...


So, I think the mechanism is safe enough as it is...

sapphire-arches:
sounds good, thanks for the fix!

PR #1453, opened 2024-03-08
spread note ons over a few ticks

m-m-adams:
Experimenting with limiting the number of note ons per clip per step (at tick resolution)

This has performance improvements on medium songs (e.g. barely culling) and can avoid culling entirely by spreading out the allocations over a slightly longer period of time

~This has severe negative side effects on extremely heavy songs - if the deluge is being severely overloaded at note on (e.g. 15+ new notes), rather than culling almost all of them immediately (so they don't play at all) they will all attempt to play, and lead to cracks as they hard cull.~

No longer has side effects on heavy songs


m-m-adams (commented):
Going to switch to have actioning note ones (and anything else with allocations) be its own task outside of the audio routine entirely

PR #1460, opened 2024-03-09
session: do not show info names on 7seg

### LINKED ISSUE #1062: 7SEG lights flicker oddly when pressing certain buttons

7SEG only
Pressing certain pads (generally ones that don't emit text on the display) seems to illuminate random segments on the display.

Random segments shouldn't flicker when pressing certain pads.

_No response_

7SEG

Nightly, other

1.0.1, 1.1.0 - 1AD7933 (1/24)

In this case, SONG (Rows View), press audition pad repeatedly
or
SONG (Grid View), press audition pads repeatedly (any pad on the column where green and blue mode are)
https://github.com/SynthstromAudible/DelugeFirmware/assets/136033640/d921480a-d7db-4b34-957b-f2fd03b8af76

sapphire-arches:
Maybe fixes #1062

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -1826,12 +1826,6 @@ yesDoIt:
 }
 
 void SessionView::redrawNumericDisplay() {
-	UI* currentUI = getCurrentUI();
-	if (currentUI != &performanceSessionView) {
-		renderViewDisplay(currentUI == &arrangerView ? l10n::get(l10n::String::STRING_FOR_ARRANGER_VIEW)
-		                                             : l10n::get(l10n::String::STRING_FOR_SONG_VIEW));
-	}
-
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -1827,10 +1827,6 @@ void SessionView::renderOLED(uint8_t image[][OLED_MAIN_WIDTH_PIXELS]) {
 
 void SessionView::redrawNumericDisplay() {
 	UI* currentUI = getCurrentUI();
```
seangoodvibes, line 1829: I think this becomes stale now and no longer required unless you want to use currentUI down below in the function in place of the other getCurrentUI() calls (I would do that).

sapphire-arches: Fixed, thanks!

seangoodvibes (approved):

PR #1462, opened 2024-03-10
Bugfix - Pad LED Greyout Cols and Rows mixup

seangoodvibes:
Fixed erroneous swap of pad LED cols and rows for assessing greyout

Renamed function names to match order of pair (cols and rows)

### DIFF
--- a/src/deluge/hid/led/pad_leds.cpp
+++ b/src/deluge/hid/led/pad_leds.cpp
@@ -631,7 +631,7 @@ void renderExplodeAnimation(int32_t explodedness, bool shouldSendOut) {
 }
 
 void reassessGreyout(bool doInstantly) {
-	auto [newCols, newRows] = getUIGreyoutRowsAndCols();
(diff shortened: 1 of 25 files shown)


src/deluge/hid/led/pad_leds.cpp
```diff
@@ -631,7 +631,7 @@ void renderExplodeAnimation(int32_t explodedness, bool shouldSendOut) {
 }
 
 void reassessGreyout(bool doInstantly) {
-	auto [newCols, newRows] = getUIGreyoutRowsAndCols();
+	auto [newCols, newRows] = getUIGreyoutColsAndRows();
```
m-m-adams, line 634: aren't these still swapped? I don't see a fix for this

seangoodvibes: > aren't these still swapped? I don't see a fix for this

See ui.cpp line 70 for the real fix. all the rest are just name changes.

	return std::make_pair(cols, rows);

m-m-adams (approved):

PR #1467, opened 2024-03-10
More arpeggiator Rhythms (up to 50 different rhythms). And made param unpatched and assignable to gold knob

soymonitus:
I implemented all the different variations from 2 to 6 steps. As they were 50 variations, I converted the parameter to unpatchedParam so it is now assignable to golden knobs, to increment the fun of the arpeggiator!

NOTE: to cherry pick for 1.1


src/deluge/modulation/arpeggiator.h
```diff
@@ -28,29 +28,116 @@ class ParamManagerForTimeline;
...
 const std::array<char const*, NUM_PRESET_ARP_RHYTHMS> arpRhythmPatternNames = {
-    "0",    // <-
-    "00-",  // <-
-    "0-0",  // <-
-    "000-", // <-
-    "00-0", // <-
-    "0-00", // <-
-    "00--", // <-
-    "0--0", // <-
+    "0",      // <- No rhythm: play all notes
```
m-m-adams, line 90: A lot of these are rotated versions of each other - should we split to a rythm+rotation instead?

soymonitus: It would make managing it more difficult for the user. With just one knob (mappable to golden knobs) they can play with it directly

m-m-adams: I disagree - as a user I dislike that most of the rythms sound the same as I scroll through

soymonitus: Some user requested it to be assignable to gold knobs, and although they sound similar it allows you to give exactly the vibe you want and also to play with it live with the knobs

bnhrsch:
Would it be possible to make the Arp-Rythms a destination for the LFOs or ENVs? This way you could have a super slow LFO1 assigned to it to have a generative arpeggiator.

m-m-adams (approved):

PR #1479, opened 2024-03-13
provide default clion config

m-m-adams:
Provides a default clion config for jlink debuggers using the same mechanism as our vscode config defaults


jamiefaye:
Thanks!

PR #1480, opened 2024-03-14
Bipolar Params ~ Make Gold Knob LED indicators + Automation View Editor Bipolar

seangoodvibes:
For bipolar params:
1) Made Gold Knob LED indicators bipolar
2) Made automation view editor bipolar

Also added gold knob LED indicator functionality to automation view editor that makes blinks the middle value and makes it harder to turn the knob passed the middle value

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -155,6 +155,39 @@ const RGB rowBipolarDownTailColour[kDisplayHeight / 2] = {{53, 2, 2}, {38, 17, 2
 
 const RGB rowBipolarDownBlurColour[kDisplayHeight / 2] = {{79, 39, 39}, {77, 60, 48}, {73, 90, 62}, {71, 111, 71}};
 
+// lookup tables for the values that are set when you press the pads in each row of the grid
+const int32_t nonPatchCablePadPressValues[kDisplayHeight] = {0, 18, 37, 55, 73, 91, 110, 128};
+const int32_t patchCablePadPressValues[kDisplayHeight] = {-128, -90, -60, -30, 30, 60, 90, 128};
+
+// lookup tables for the min value of each pad's value range used to display automation on each row of the grid
+const int32_t nonPatchCableMinPadDisplayValues[kDisplayHeight] = {0, 17, 33, 49, 65, 81, 97, 113};
+const int32_t patchCableMinPadDisplayValues[kDisplayHeight] = {-128, -96, -64, -32, 1, 33, 65, 97};
+
+// lookup tables for the max value of each pad's value range used to display automation on each row of the grid
+const int32_t nonPatchCableMaxPadDisplayValues[kDisplayHeight] = {16, 32, 48, 64, 80, 96, 112, 128};
+const int32_t patchCableMaxPadDisplayValues[kDisplayHeight] = {-97, -65, -33, -1, 32, 64, 96, 128};
+
+// summary of pad ranges and press values (format: MIN < PRESS < MAX)
+// patch cable:
+// y = 7 ::   97 <  128 < 128
+// y = 6 ::   65 <   90 <  96
+// y = 5 ::   33 <   60 <  64
+// y = 4 ::    1 <   30 <  32
+// y = 3 ::  -32 <  -30 <  -1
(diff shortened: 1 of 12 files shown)


src/deluge/gui/views/view.cpp
```diff
@@ -1229,14 +1248,22 @@ void View::setKnobIndicatorLevel(uint8_t whichModEncoder) {
...
 	}
 
-	indicator_leds::setKnobIndicatorLevel(whichModEncoder, knobPos);
+	indicator_leds::setKnobIndicatorLevel(whichModEncoder, knobPos, isBipolar);
 }
 
 /// if you're dealing with a patch cable which has a -128 to +128 range
 /// we'll need to convert it to a 0 - 128 range for purpose of rendering on knob indicators
 int32_t View::convertPatchCableKnobPosToIndicatorLevel(int32_t knobPos) {
 	float floatKnobPos = kMaxKnobPos * ((static_cast<float>(knobPos) + kMaxKnobPos) / (kMaxKnobPos * 2));
```
m-m-adams, line 1257: this should just be (knob + 128) >> 1

src/deluge/hid/led/indicator_leds.cpp
```diff
@@ -178,15 +179,15 @@ void setMeterLevel(uint8_t whichKnob, uint8_t level) {
...
 	if (whichKnob == whichKnobMetering) {
 		uiTimerManager.setTimer(TimerName::METER_INDICATOR_BLINK, 500);
 	}
-	actuallySetKnobIndicatorLevel(whichKnob, level);
+	actuallySetKnobIndicatorLevel(whichKnob, level, isBipolar);
 }
 
 // Just set level
-void actuallySetKnobIndicatorLevel(uint8_t whichKnob, uint8_t level) {
+void actuallySetKnobIndicatorLevel(uint8_t whichKnob, uint8_t level, bool isBipolar) {
```
m-m-adams, line 190: This function can be massively simplified by calculating brightness before the loop, as in the original code

level >> 5 (e.g. divide by 32 and drop remainder) is the number of leds to light up

(level & 31) << 3 is equivalent to (level - (numLEDs * 32)) << 3, and is the remaining brightness (e.g. the remainder of the division)

Then you just need some simple logic to convert negative bipolar values to positive ones before feeding them into the function


As is this change makes the code wildly slower and it runs pretty often, so if we want to deliberately make it less efficient I'd like to see some benchmarks first

m-m-adams (changes requested):

src/deluge/gui/views/automation_view.cpp
```diff
@@ -3755,31 +3845,26 @@ void AutomationView::renderDisplayForMultiPadPress(ModelStackWithAutoParam* mode
...
-		// update LED indicators
-		// if you're dealing with a patch cable which has a -128 to +128 range
-		// we'll need to convert it to a 0 - 128 range for purpose of rendering on knob indicators
-		if (!onArrangerView && (clip->lastSelectedParamKind == params::Kind::PATCH_CABLE)) {
-			knobPosLeft = view.convertPatchCableKnobPosToIndicatorLevel(knobPosLeft);
-			knobPosRight = view.convertPatchCableKnobPosToIndicatorLevel(knobPosRight);
-		}
-		indicator_leds::setKnobIndicatorLevel(0, knobPosLeft);
-		indicator_leds::setKnobIndicatorLevel(1, knobPosRight);
+		setKnobIndicatorLevels(modelStackWithParam, knobPosLeft,
```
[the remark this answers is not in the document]
seangoodvibes: mostly to avoid calling view.convertPatchCableKnobPosToIndicatorLevel() twice if knobPosRight is the same as knobPosLeft.

src/deluge/hid/led/indicator_leds.cpp
seangoodvibes: a bit confused because I am still calculating brightness before the loop. 

As before I am calculating brightnessOutputValue in the loop.

the brightnessOutputValue depends on the indicator being looked at.

level >> 5 is useless for bipolar as it calculates max # of LED's lit up for non-bipolar

m-m-adams (approved):

PR #1484, opened 2024-03-15
Add storage manager reference to many param lists and call sites.

jamiefaye:
…elp disentangle.

This is an early step in a larger project to disentangle the serialization to/from XML code from the file management code. Since all root calls into the storage manager use "(storageManager, ", the code should work exactly like it always has.


jamiefaye:
Lets get this one in before many more merge conflicts show up.

seangoodvibes:
I started with reviewing change impacts to performance view and midi follow classes. Those look good.

src/deluge/storage/storage_manager.h
```diff
@@ -112,6 +110,8 @@ class StorageManager {
 	SyncLevel readAbsoluteSyncLevelFromFile(Song* song);
 	void writeAbsoluteSyncLevelToFile(Song* song, char const* name, SyncLevel internalValue, bool onNewLine = true);
 
+private:
+public:
```
m-m-adams, line 114: are these supposed to be private?

m-m-adams (commented):
As a sanity check can you rename the storage manager class? That should catch any instances of using the global instead of the passed in reference 

It would also probably be helpful to put the global storage manager in its own header that can be be included only in places that require it - as is it still relies on developers remembering not to use it, since the storage manager header still has to be included fairly broadly to support member calls

src/deluge/storage/storage_manager.h
jamiefaye: Renaming the storage manager class to shake-out unexpected global refs is definitely part of the plan for the next phase. I did get rid of the private/public issue.

m-m-adams (approved):

PR #1493, opened 2024-03-17
[Draft] Arpeggiator: add new Note Probability param

soymonitus:


soymonitus:
Probably, I need to fix first the current limitation of having to play one note so the doTick and render methods of the arp are called later. Right now this is hardcoded, so for the first arpeggiated note, it is always played, regardless of the probability setting, to keep the render/dotick flowing. There must be a way to have those method called without playing the first note..

bnhrsch:
Would love to have this feature! This would really elevate the Arpeggiator. Has the work on it been abandoned?

Issue #760, opened 2023-11-26
Kit with random number of sounds gets created when switching clip type from Audio to Kit


laikrodiz:
### Please describe the problem:

Kit with random number of sounds gets created when switching clip type from Audio to Kit.
Every time you repeat this, say convert Kit back to Audio, then back to Kit, the number of sounds in a Kit changes.

### What is the expected behavior?

Full kit should be loaded when switching clip type from Audio clip to Kit clip.

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

community

### What is the firmware name:

deluge-v1_0_0-beta+2023_11_23-4c959c6.bin

### If possible provide the steps to reproduce the issue and upload additional media:

Steps to reproduce:

1. Create new track, switch to Song mode
2. Create an empty audio clip
3. Hold the clip pad and press Kit. Audio clip is converted to Kit clip.
4. Enter the Kit clip to see, that it has random number of sounds loaded.

seangoodvibes:
Should fixed with PR #1532

Issue #1051, opened 2024-01-24
Converting empty Audio Clip to Kit crashes, in some scenarios it will reverse play direction of the newly created Kit.


ok-reza:
### Please describe the problem:

Converting a blank Audio Clip to Kit usually will generate a E379 error. There is a chance to convert this successfully, and in some cases it will reverse the play direction of the Kit.

### What is the expected behavior?

Converting Audio Clip to Kit should properly load the first available kit. Play direction should not change when converting.

### Is there a relevant Pull request?

Possibly related to #760

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

1.1.0 - 8FEA0F5

### If possible provide the steps to reproduce the issue and upload additional media:

For E379 error:
1. Turn on Deluge, enter SONG view
2. Press and hold row pad and convert to Audio Clip by pressing SELECT
3. Press and hold row pad of newly converted Audio Clip and press KIT
4. Crash should occur.

For Play Direction switch:
1. Turn on Deluge, enter SONG view
2. Create a new row and convert it to an Audio Clip
3. Press and hold row pad of newly converted Audio Clip and press KIT
4. If a crash doesn't occur, the play direction can reverse. Sometimes takes a couple tries.
5.  Additionally, the kit made doesn't properly load all the samples: in my case it loads 4 samples of the 808 preset, the first two and last two. (see https://github.com/SynthstromAudible/DelugeFirmware/issues/760)

m-m-adams:
I'm inclined to just block this - is there a reason to do it instead of deleting the audio clip and making a new clip?

ok-reza:
i think the use case is exactly that, to convert an empty audio clip to another clip type quicker than doing it manually. the use case is probably low specifically into KITS, but we are able to convert any empty clip to any other clip type currently so i think it's a nice to have. i don't think it would be a problem to have it blocked simply for the 1.1 release for stability purposes but addressed in the future.

the only other use case i can think of is if someone has gone ahead and made a template project with a variety of blank clips, and decides they don't need an audio clip and quickly wants to convert it.

thinking into the future, freezing/unfreezing tracks could work with selecting a sequenced clip's pad row and then pressing SELECT (like you would to create an audio clip, but in this case it freezes the eg. synth clip into an audio clip) and vice versa, but i imagine there may be some UX gremlins that pop up with this.

[...]

PR #1532, opened 2024-03-20
Bugfix - Don't allow converting audio clip to instrument clip

seangoodvibes:
Converting audio clip to instrument is very problematic, thus it is better to eliminate this behaviour and instead direct users to delete the audio clip and create a new clip instead of going through the "nice to have" conversion process.

This closes: 

https://github.com/SynthstromAudible/DelugeFirmware/issues/760
https://github.com/SynthstromAudible/DelugeFirmware/issues/1051

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -499,14 +499,8 @@ changeOutputType:
 			Clip* clip = getClipForLayout();
 
 			if (clip != nullptr) {
-				// If AudioClip, we have to convert back to an InstrumentClip
-				if (clip->type == ClipType::AUDIO) {
-					actionLogger.deleteAllLogs();
-					replaceAudioClipWithInstrumentClip(clip, newOutputType);
-				}
-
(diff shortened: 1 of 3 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -499,14 +499,8 @@ ActionResult SessionView::buttonAction(deluge::hid::Button b, bool on, bool inCa
...
-				// If AudioClip, we have to convert back to an InstrumentClip
-				if (clip->type == ClipType::AUDIO) {
-					actionLogger.deleteAllLogs();
-					replaceAudioClipWithInstrumentClip(clip, newOutputType);
-				}
-
-				// Or if already an InstrumentClip, changing Instrument type is easier
-				else {
+				// Don't allow converting audio clip to instrument clip
+				if (clip->type != ClipType::AUDIO) {
```
m-m-adams, line 503: if it's an audio clip we should flash CANT

m-m-adams (approved):

Issue #964, opened 2024-01-13
MIDI Learn Crash - When trying to play a hibernated instrument


seangoodvibes:
### Please describe the problem:

// Deluge crashes when you try the following steps

1) Create a brand new kit
2) Midi learn controller note to the single kit row that exists in this new kit
3) Hit play and record some notes in. 
4) Stop recording
5) Switch instrument types from kit to synth
6) Hit record and press learned note on controller again
7) Stop record
8) Switch back to kit instrument type
9) You will now notice a second kit row created with your recently recorded notes
10) Hit record again / play notes on midi controller
11) Deluge crashes

// Deluge also crashes when you try these steps:

1. Create a brand new kit
2. Midi learn controller note to the single kit row that exists in this new kit
3. Switch instrument types from kit to synth
4. Hit record and press learned note on controller
5. Stop record
6. Switch back to kit instrument type
7. You will now notice a second kit row created with your recently recorded notes
8. Hit record again / play notes on midi controller
9. Deluge crashes

<img width="652" alt="Screenshot 2024-01-12 at 8 07 14 PM" src="https://github.com/SynthstromAudible/DelugeFirmware/assets/138174805/80b2c37f-0772-4b77-af94-804db44d2992">

### What is the expected behavior?

Shouldn't be able to send/record notes to a hibernated instrument

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

1.1.0-DEV-1D2C31B0-DIRTY

### If possible provide the steps to reproduce the issue and upload additional media:

https://github.com/SynthstromAudible/DelugeFirmware/assets/138174805/cbc83235-8e72-4261-a03b-6c17ef41265c

seangoodvibes:
Closed as should be fixed with PR #1533

PR #1533, opened 2024-03-20
Restricted change of instrument clip types if the instrument clip is not empty

seangoodvibes:
Restricted change of instrument clip types to/from kit type if the instrument clip is not empty

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -1262,6 +1262,13 @@ void InstrumentClipView::changeOutputType(OutputType newOutputType) {
 		return;
 	}
 
+	// don't allow clip type change if clip is not empty
+	// only impose this restriction if switching to/from kit clip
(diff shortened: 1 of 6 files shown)


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -1262,6 +1262,11 @@ void InstrumentClipView::changeOutputType(OutputType newOutputType) {
 		return;
 	}
 
+	// don't allow clip type change if clip is not empty
+	if (!getCurrentInstrumentClip()->isEmpty()) {
+		return;
```
m-m-adams, line 1267: just add a check here and only return if the current type or the new type is kit

m-m-adams: otherwise popup cant

m-m-adams:
Closes #964

Possibly others, I think there's another kit conversion bug open

PR #1542, opened 2024-03-21
Feature / Add shortcut to remove audio clip time stretching

seangoodvibes:
Removed time stretching from samples that are loaded. Clip length is automatically set to the length of the sample.

Added new shortcut to remove timestretching from an audio clip and extend an audio clip without timestretching. 
  - Press `▼︎▲︎` + `◀︎▶︎` to set the Audio Clip length equal to the length of the audio sample. This will effectively remove timestretching from the audio sample.
  - Press `SHIFT` + `◀︎▶︎` + `turn ◀︎▶︎` to lengthen the audio clip without timestretching.

- [x] Get it working for the full audio sample length (when you first load audio sample and haven't touched green and red waveform markers yet)
- [x] Get it working for the adjusted audio sample length (using green and red waveform markers) 
- [x] Adjust code that sets new clip loop length to use the same functions as the current Shift + Turn <> combo)

### DIFF
--- a/src/deluge/gui/views/audio_clip_view.cpp
+++ b/src/deluge/gui/views/audio_clip_view.cpp
@@ -477,8 +477,9 @@ ActionResult AudioClipView::padAction(int32_t x, int32_t y, int32_t on) {
 							// Ok, move the marker!
 							int32_t newLength =
 							    (x + 1) * currentSong->xZoom[NAVIGATION_CLIP] + currentSong->xScroll[NAVIGATION_CLIP];
-
-							changeUnderlyingSampleLength(clip, sample, newLength);
+							int32_t oldLength = clip->loopLength;
+							uint64_t oldLengthSamples = clip->sampleHolder.getDurationInSamples(true);
+							changeUnderlyingSampleLength(clip, sample, newLength, oldLength, oldLengthSamples);
 
 							goto needRendering;
 						}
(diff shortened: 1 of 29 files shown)


src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -324,7 +325,46 @@ ActionResult AudioClipView::buttonAction(deluge::hid::Button b, bool on, bool in
...
+
+						// get sample end position (red waveform marker)
+						int32_t loopEndSample =
+						    (uint32_t)(guide.getBytePosToEndOrLoopPlayback() - sample->audioDataStartPosBytes)
+						    / (uint8_t)(sample->numChannels * sample->byteDepth);
+
+						uint32_t sampleLengthInSamples = std::abs(loopEndSample - loopStartSample);
+
+						// convert sample length in samples to ticks
+						float loopLength = (float)sampleLengthInSamples / playbackHandler.getTimePerInternalTickFloat();
```
m-m-adams, line 354: needs to be double - a 4gb audio file is longer than the largest int which can be stored in a float

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -324,7 +325,46 @@ ActionResult AudioClipView::buttonAction(deluge::hid::Button b, bool on, bool in
...
+				}
+
+				AudioClip* audioClip = getCurrentAudioClip();
+				SamplePlaybackGuide guide = audioClip->guide;
+				SampleHolder* sampleHolder = (SampleHolder*)guide.audioFileHolder;
+				if (sampleHolder) {
+					Sample* sample = (Sample*)sampleHolder->audioFile;
+					if (sample) {
+						// get sample start position (green waveform marker)
+						int32_t loopStartSample =
```
m-m-adams, line 342: can this just use SampleHolder::getLengthInSamplesAtSystemSampleRate()? This is recovering it from the actual sample file, and I believe it will be wrong if the sample is not 16bit/44khz

m-m-adams (approved):

sichtbeton:
Could we cherry pick this into RC3 from 1.1 maybe?

PR #1551, opened 2024-03-22
Bugfix - Disable sidebar actions while VU meter is shown

seangoodvibes:
Disabled interacting with sidebar while VU meter is displayed

Volume mod button now blinks when VU meter is on and being displayed


m-m-adams (commented):
I think we now need a way to disable to vu meter - maybe toggle it off if you hit the volume mod button again? 

nevermind that already works

m-m-adams (approved):

PR #1577, opened 2024-03-24
Bugfix/crossfade filters when activated

### LINKED ISSUE #932: Click when engaging filter from 0

When turning up a filter (HPF most notably) a click is heard. This  is not good for smooth filter modulations.

No click

No

OLED

Nightly

1.1.0-NIGHTLY-SESSEC7

Turn up the filter while a tone is playing

m-m-adams:
Adds a crossfade when filters are first turned on to avoid clicks and pops

Fixes #932 

In order to do that makes a number of other supporting changes:
1. Refactors filter functions to remove extraneous doFilter booleans, that is now passed by setting the filter off
2. Modifies getFilterDisplayName to not turn the filters back on
3. Adds new functions for toggling filter modes including off - not used in this PR, will be part of a future PR to enable filters in FM

Also marks rendering functions as hot, which adds a couple voices overall

### DIFF
--- a/src/deluge/dsp/filter/filter.h
+++ b/src/deluge/dsp/filter/filter.h
@@ -25,6 +25,8 @@
 namespace deluge::dsp::filter {
 constexpr uint32_t ONE_Q31U = 2147483648u;
 constexpr int32_t ONE_Q16 = 134217728;
+// fade in filter over 1/100th of a second to avoid click
+constexpr q31_t fadeIncrement = 1 * (ONE_Q31 / kSampleRate);
 extern q31_t blendBuffer[SSI_TX_BUFFER_NUM_SAMPLES * 2];
 /**
  *  Interface for filters in the sound engine
@@ -44,8 +46,8 @@ class Filter {
 public:
 	Filter() = default;
 	// returns a gain compensation value
-	q31_t configure(q31_t frequency, q31_t resonance, FilterMode lpfMode, q31_t lpfMorph, q31_t filterGain,
-	                q31_t dryLevel = 0) {
+	q31_t configure(q31_t frequency, q31_t resonance, FilterMode lpfMode, q31_t lpfMorph, q31_t filterGain) {
+
 		// lpfmorph comes in q28 but we want q31
 		return static_cast<T*>(this)->setConfig(frequency, resonance, lpfMode, lshiftAndSaturate<2>(lpfMorph),
 		                                        filterGain);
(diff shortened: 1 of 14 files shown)


src/deluge/dsp/filter/filter_set.h
```diff
@@ -59,12 +60,17 @@ class FilterSet {
...
+	// all filters share a state. This is fine since they just hold plain data and initialization is handled by
+	// reset/configure calls.  This is faster than using a variant at the cost of not throwing on unitialized access.
+	// However since there are no invariants to uphold, the worst case scenario is an audio glitch so whatever
+	// This cuts 250 bytes from the size of this class, which is fairly significant since every voice has a filterset
+	char lpfilterstate[lpfilterSize];
+	char hpfilterstate[hpfilterSize];
+	SVFilter& lpsvf = *(SVFilter*)lpfilterstate;
+	LpLadderFilter& lpladder = *(LpLadderFilter*)lpfilterstate;
+	HpLadderFilter& hpladder = *(HpLadderFilter*)hpfilterstate;
+	SVFilter& hpsvf = *(SVFilter*)hpfilterstate;
```
sapphire-arches, line 73: Can we use a `union` instead? Same no-throw behavior but without the questionable casting, I think.

src/deluge/dsp/filter/filter.h
```diff
@@ -54,23 +57,68 @@ class Filter {
...
+		if (fade) {
+			dryFade = 1;
+			wetLevel = 0;
+		}
+	}
+
+	inline void updateBlend() {
+		// fades over around 500 samples
+		dryFade = dryFade * 0.99;
+		wetLevel = (q31_t)(ONE_Q31 * (1 - dryFade));
```
sapphire-arches, line 120: does this eventually get all the way to `ONE_Q31`? I imagine so, but do we need an explicit clamp?

src/deluge/dsp/filter/filter.h
```diff
@@ -54,23 +57,68 @@ class Filter {
...
 	 */
-	void filterStereo(q31_t* startSample, q31_t* endSample) {
-		static_cast<T*>(this)->doFilterStereo(startSample, endSample);
-		;
+	[[gnu::hot]] void filterStereo(q31_t* startSample, q31_t* endSample) {
+		if (wetLevel == ONE_Q31) {
+			static_cast<T*>(this)->doFilterStereo(startSample, endSample);
+		}
+		else {
+			memcpy(blendBuffer, startSample, (endSample - startSample) * sizeof(q31_t));
```
sapphire-arches, line 91:
```suggestion
			std::copy(startSample, endSample, &blendBuffer[0]);
```
Double check the binary size but I think it should be fine.

src/deluge/dsp/filter/filter_set.h
m-m-adams: Smart - I thought you could only use POD structs in unions but looks like that may not be the case anymore

src/deluge/dsp/filter/filter.h
m-m-adams: yeah it does, it gets there by underflowing after about 500 samples

src/deluge/dsp/filter/filter.h
m-m-adams: we have a faster arm assembly memcpy implementation than libc/c++ and it makes a meaningful difference

It's in c_lib_alternatives.s

src/deluge/dsp/filter/filter_set.cpp
```diff
@@ -198,10 +192,9 @@ int32_t FilterSet::setConfig(int32_t lpfFrequency, int32_t lpfResonance, bool do
...
 void FilterSet::reset() {
-	hpladder.reset();
-	lpsvf.reset();
-	hpsvf.reset();
-	lpladder.reset();
-	noiseLastValue = 0;
+	hpfilter.ladder.reset();
+	lpfilter.svf.reset();
+	hpfilter.svf.reset();
+	lpfilter.ladder.reset();
```
sapphire-arches, line 198: this looks a little suspect now, since the filters share storage.

m-m-adams: oh good point - I don't think it's necessary at all

m-m-adams: It is necessary but zeroing the storage works - the reset for each filter just sets the state to 0 anyway

sapphire-arches: cool, I guess just leave a comment to the effect of "all zeros must represent a reset state" on the union? the rest of this PR looks fine.

m-m-adams: done, thanks

PR #1580, opened 2024-03-24
Keyboard retriggering

sapphire-arches:
This is especially important for velocity drums but it works for all the keyboard views.

Also required some special logic in the velocity drums view to make sure the most recent pad press is the one that determines the velocity.


sichtbeton:
i love it.

m-m-adams (approved):

ok-reza:
i missed this one, excellent work!

Issue #1544, opened 2024-03-21
Dotted arp rates can produce stuck notes if changing Rhythm from one length to another


soymonitus:
### Please describe the problem:

If you play with the Rhythm parameter and change for example from a 3-step rhythm to a 4-step rhythm or viceversa, and th sync rate is dotted, you can get stuck notes.

### What is the expected behavior?

To not get stuck notes

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### If possible provide the steps to reproduce the issue and upload additional media:

Play attached SONG49.XML song file

soymonitus:
[SONG49.XML.zip](https://github.com/SynthstromAudible/DelugeFirmware/files/14697375/SONG49.XML.zip)
This song will always get stuck notes.
It is set to dotted 8th, and rhythm is "0----". This way the first note is played, 4 notes omitted, and the next note, which is meant to play, falls right a 64th note before the end of the clip (that note will get stuck). The clip restarts from the beginning, and the next notes are silences, but there is one note stuck.

Dotted sync rates cause this weird behavior where the syncing restarts after the clip loops (you can't get it to drone). That, together with the rhythms with lots of silences, is causing the bug somehow

soymonitus:
@m-m-adams we need to reopen this one. I tested it with a different synth patch that had the gate shorter and it worked so I thought it was fixed, but once you increase the gate so the end of the gate goes beyond the end of the clip, you get the stuck note. The other bug fixed by my PR is really fixed so no problem with #1558

PR #1581, opened 2024-03-24
Miscellaneous arpeggiator bugfixing

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1544

After a deep testing I found not only one but several different bugs and inconsistencies. I fixed them as part of this PR to leave the arpeggiator module totally bug-free.

- Rhythm setting was not read nor written to song files
- Sidechain cloning method was forgetting about syncType param
- SyncType was not saved for arpeggiator sync rate, so if you set for example 8th Dotted, when loading, you got 8th Even.
- Fixed the bug with stuck notes (1544)
- Fixed first rhythm note repeated when loading a project and pressing play. The Rhythm, Ratchet and SeqLength params were only updated in the render method so they were not yet initialized when calling the first sequenced note in the doTick method. Fixed this by calling a new updateParams method before the doTick.


soymonitus:
Ready for review, please, cherry pick it

m-m-adams (approved):

PR #1588, opened 2024-03-25
UI Consistency - Update audio clip clearing string

seangoodvibes:
Updated string displayed when clearing audio clip with <> + back

If default FlashStrorage::automationClear is on it will only clear the audio sample and so "Sample Cleared" will be displayed instead of "Clip Cleared"

This makes the audio clip strings consistent with how they are displayed in an instrument clip (it displays "Notes cleared").


src/deluge/gui/l10n/english.cpp
```diff
@@ -812,6 +810,9 @@ PLACE_SDRAM_DATA Language english{
         {STRING_FOR_CHECKSUM_FAIL, "Checksum fail"},
         {STRING_FOR_WRONG_SIZE, "Wrong size?"},
         {STRING_FOR_BAD_KEY, "Incorrect Handshake Key"},
+
+        {STRING_FOR_CLIP_CLEARED, "Clip cleared"},
```
m-m-adams, line 814: Why Clip Cleared instead of Audio Clip Cleared?

m-m-adams: Ah I see it's just been moved, I thought it was a new addition

m-m-adams (approved):

PR #1596, opened 2024-03-25
Adding an slice mode option for short samples

sichtbeton:
For short slices, the current behaviour automatically switches the new kit item for audio samples to the playback mode "Once" when the sample is shorter then 2 seconds. 

This will provide an option to turn this behaviour off.


sichtbeton:
can be deleted. was a mistake.

PR #1601, opened 2024-03-26
Add the ability to turn off the automatic short slice detection when slicing audio in a kit

sichtbeton:
Currently when slicing audio into a new kit, a detection will change the playback mode of the new kit items to "ONCE" when the sample is under 2 seconds. 

-  PR will provide the ability to toggle this behaviour ON / OFF.
- if OFF then it will adhere to the new default slice mode that can be also set in the default menu
- you find the setting in "SELECT - DEFAULTS - Short Slice Mode"


sichtbeton:
can be deleted, was a mistake

PR #1603, opened 2024-03-26
[DRAFT] Short slice toggle

sichtbeton:
Add a toggle to switch off automatic short slice detection with playback mode "ONCE" in the audio slicer.


sichtbeton:
ready for review.

src/deluge/gui/ui/slicer.cpp
```diff
@@ -650,8 +650,14 @@ void Slicer::doSlice() {
 		uint32_t nextDrumStart = lengthInSamples / numClips;
 		firstRange->sampleHolder.endPos = nextDrumStart;
 
-		firstDrum->sources[0].repeatMode =
-		    (lengthMSPerSlice < 2002) ? SampleRepeatMode::ONCE : FlashStorage::defaultSliceMode;
+		if (FlashStorage::defaultShortSliceMode == true) {
```
sapphire-arches, line 653:
```suggestion
		if (FlashStorage::defaultShortSliceMode) {
```

You never need to compare with a boolean, it's already the right type.

src/deluge/storage/flash_storage.cpp
```diff
@@ -200,6 +201,8 @@ GridDefaultActiveMode defaultGridActiveMode;
 
 SampleRepeatMode defaultSliceMode;
 
+bool defaultShortSliceMode = true;
```
sapphire-arches, line 204: Maybe we should use a `SampleRepeatMode` here instead -- that gives maximum flexibility for essentially no cost in terms of code size or flash storage.

If you're feeling extra fancy, the default short and regular slice modes could be packed in to the same byte in the flash storage:

```cpp
static_assert(kNumRepeatModes < (1 << 4), "2 repeat modes are packed in to 1 byte");
buffer[167] = (util::to_underlying(defaultShortSliceMode) << 4) | util::to_underlying(defaultSliceMode);
```

And on the read side:
```cpp
uint8_t defaultSliceModeRaw = buffer[167] & 0xf;
uint8_t defaultShortSliceModeRaw = (buffer[167] >> 4) & 0xf;

if (defaultSliceModeRaw >= kNumRepeatModes) {
   defaultSliceMode = SampleRepeatMode::CUT;
} else {
  defaultSliceMode = static_cast<SampleRepeatMode>(defaultSliceModeRaw);
}

if (defaultShortSliceModeRaw >= kNumRepeatModes) {
   defaultShortSliceMode = SampleRepeatMode::ONCE;
} else {
  defaultShortSliceMode = static_cast<SampleRepeatMode>(defaultShortSliceModeRaw);
}
```

(make sure I got the settings right for matching the default behavior....)

src/deluge/gui/ui/slicer.cpp
sichtbeton: right thank you

src/deluge/storage/flash_storage.cpp
sichtbeton: i really dont know what any of that means. but i will check tomorrow again.

sichtbeton: it sounds like a fancy and good solution, but i am not capable of putting this all in the right place. the other comment i could just accept. is this also possible to do here?

sapphire-arches: it's a more complex change than can really fit in the framework of touchups to individual hunks (the diff sections) unfortunately. I can give implementation guidance or eventually get around to implementing it myself.

sichtbeton: Or is it ok to just leave it like this? Does it impact performance / mess with guidelines or anything?

sichtbeton:
i am not confident enough to make the suggested changes, but my method works. so could this be merged or anything that prevents that?

PR #1607, opened 2024-03-27
Automation View ~ Bug fixes and added select scrolling of patch cables

seangoodvibes:
- Added select scrolling of patch cables that exist (they are dynamically added to the end of the non-patch cable param scrolling list for synths and kit rows)
- Handled scenario where you selected a patch cable previously and it no longer exists when re-entering automation view. In this case it now sends you back to the automation overview.
- Fixed select scrolling bug that would cause interpolation shortcut to blink profusely 
- Updated patch cable name displayed to show both sources if it's a patch cable with two sources

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -3499,8 +3499,11 @@ bool AutomationView::selectPatchCableAtIndex(Clip* clip, PatchCableSet* set, int
 	// need to add patch cable source to the descriptor so that we can get the paramId from it
 	desc.addSource(cable->from);
 
-	// if we've previously selected a patch cable, we want to start scrolling from
-	// that patch cable
+	// if we've previously selected a patch cable, we want to start scrolling from that patch cable
+	// note: the reason why we can't save the patchCableIndex to make finding the previous patch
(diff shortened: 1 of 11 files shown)


src/deluge/gui/views/automation_view.cpp
```diff
@@ -3337,26 +3388,131 @@ void AutomationView::selectGlobalParam(int32_t offset, Clip* clip) {
...
+		foundCurrentPatchCable = true;
+	}
+	// if we found the patch cable we previously selected and we found another one
+	// or we hadn't selected a patch cable previously and found a patch cable
+	// select the one we found
+	else if ((foundCurrentPatchCable || (clip->lastSelectedParamKind != params::Kind::PATCH_CABLE))
+	         && (desc.data != clip->lastSelectedParamID)) {
+		clip->lastSelectedPatchSource = cable->from;
+		clip->lastSelectedParamID = desc.data;
+		clip->lastSelectedParamKind = params::Kind::PATCH_CABLE;
```
m-m-adams, line 3512: save the index somewhere so that you don't need to search for it again

m-m-adams (approved):

PR #1608, opened 2024-03-27
Os refactor/audio interrupt

m-m-adams:
Current status:

This works but culling is messed up, so need to reevaluate how to make those decisions when rendering is much more fixed


m-m-adams:
Too much non re entrant code for this to work, need to do a scheduler instead

PR #1614, opened 2024-03-28
Independent sample locking and improved lock rendering

### LINKED ISSUE #845: Waveform Loop Lock should have a visual indication that the loop is currently locked
Rather than just relying on the popups, we should draw the pad grid differently somehow. Bikeshed proposal: Render the upper and lower halves of the loop indicators with different colors (cyan/magenta on the top/bottom of the start, and magenta/cyan on the top/bottom of the end maybe?)

### LINKED ISSUE #1043: Cannot proceed with creating loop points on new sample row if loop lock is locked in some cases.

If loop lock is enabled, loop lock is on and remembered across all clips/rows. I think this actually can come in handy, but has some repercussions that need to be addressed.

In this case, if loop lock is enabled and the length of loop lock is say 2000 samples, if you enter another kit row's waveform view and the sample is smaller than 2000 samples, then when you try to create a loop start point (via green column press + a column press anywhere else). a loop start point is made but not visible and not adjustable (and has no effect).

I think the simplest and a sensible solution is to switch to FREE (loop lock disabled) when you try to create a loop start/end position on a sample that is shorter than the current loop lock length.

universal loop-lock is interesting (not even sure it was intentional) but has some odd effects. it seems that there needs to be both a universal loop lock and singular loop lock option.
i wonder if press and hold CYAN + press MAGENTA = unique loop lock (unique per that clip and kit row) and then press and hold MAGENTA + press CYAN = universal loop lock (do unique loop locks ignore universal loop locks?)

_No response_

OLED

Nightly

1.1.0 8FEA0F5

_No response_

sapphire-arches:
Fix #845 
Fix #1043


sapphire-arches:
These are in the commit messages, but just so github links this PR too:

Fixes #845 
Fixes #1043

m-m-adams (approved):

PR #1618, opened 2024-03-30
Maybe fix compressor gain staging

### LINKED ISSUE #1605: Master volume too loud

In the latest nightly, the master volume is way overdriven. On all the songs, this results in having to turn down volume to almost nothing.

Normal volume in all the tracks.

No

OLED

Nightly

_No response_

open older song
volume is overdriven in the reds
turn down volume to around 3-4 to get decent levels

sapphire-arches:
Fixes #1605

### DIFF
--- a/src/deluge/dsp/compressor/rms_feedback.cpp
+++ b/src/deluge/dsp/compressor/rms_feedback.cpp
@@ -90,7 +90,7 @@ void RMSFeedbackCompressor::render(StereoSample* buffer, uint16_t numSamples, q3
 	rms = calcRMS(buffer, numSamples);
 }
 
-float RMSFeedbackCompressor::runEnvelope(float current, float desired, float numSamples) {
+float RMSFeedbackCompressor::runEnvelope(float current, float desired, float numSamples) const {
 	float s{0};
 	if (desired > current) {
 		s = desired + std::exp(a_ * numSamples) * (current - desired);
@@ -110,8 +110,8 @@ float RMSFeedbackCompressor::calcRMS(StereoSample* buffer, uint16_t numSamples) 
 	q31_t offset = 0; // to remove dc offset
 	float lastMean = mean;
 	do {
-		q31_t l = thisSample->l - hpfL.doFilter(thisSample->l, a);
-		q31_t r = thisSample->r - hpfL.doFilter(thisSample->r, a);
+		q31_t l = thisSample->l - hpfL.doFilter(thisSample->l, hpfMovability_);
+		q31_t r = thisSample->r - hpfL.doFilter(thisSample->r, hpfMovability_);
 		q31_t s = std::max(std::abs(l), std::abs(r));
 		sum += multiply_32x32_rshift32(s, s) << 1;
 
--- a/src/deluge/dsp/compressor/rms_feedback.cpp
+++ b/src/deluge/dsp/compressor/rms_feedback.cpp
@@ -121,7 +121,7 @@ float RMSFeedbackCompressor::calcRMS(StereoSample* buffer, uint16_t numSamples) 
 
 	} while (++thisSample != bufferEnd);
 
-	float ns = float(numSamples) / 0.5;
+	float ns = float(numSamples * 2);
 	mean = (float(sum) / ONE_Q31f) / ns;
 	// warning this is not good math but it's pretty close and way cheaper than doing it properly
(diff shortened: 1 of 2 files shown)


src/deluge/dsp/compressor/rms_feedback.h
```diff
@@ -34,72 +34,144 @@ class RMSFeedbackCompressor {
...
-	constexpr int32_t getRatioForDisplay() { return ratio; }
+
+	/// Get the current ratio as a full-scale (0 to 2^31) number
+	[[nodiscard]] inline constexpr q31_t getRatio() const { return ratioKnobPos; }
+
+	/// Get the current ratio as an integer
+	///
+	/// @todo Thsi currently rounds (down) to the nearest integer, it would be nice to expose at least a little more of
+	/// the underlying precision.
+	[[nodiscard]] inline constexpr int32_t getRatioForDisplay() const { return ratio; }
```
m-m-adams, line 119: This should probably return a float so that we can display 2.x

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -79,8 +79,8 @@ void RMSFeedbackCompressor::render(StereoSample* buffer, uint16_t numSamples, q3
 		currentVolumeL += amplitudeIncrementL;
 		currentVolumeR += amplitudeIncrementR;
 		// Apply post-fx and post-reverb-send volume
-		thisSample->l = multiply_32x32_rshift32(thisSample->l, currentVolumeL) << 2;
-		thisSample->r = multiply_32x32_rshift32(thisSample->r, currentVolumeR) << 2;
+		thisSample->l = multiply_32x32_rshift32(thisSample->l, currentVolumeL) << 4;
```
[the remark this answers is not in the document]
m-m-adams: We can potentially apply 3 compressors to a sound (kit row, kit, song) which can each multiply by 8 for a total possible gain of 2^9 which is a bit dicy

I'm not sure the exact level that samples/waves render at but I think it's 2^20 or something which would definitely clip 

On the other hand it can be fixed by lowering volumes so we could chalk that up to user error

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -45,7 +45,7 @@ void RMSFeedbackCompressor::renderVolNeutral(StereoSample* buffer, uint16_t numS
 	// this is a bit gross - the compressor can inherently apply volume changes, but in the case of the per clip
 	// compressor that's already been handled by the reverb send, and the logic there is tightly coupled such that
 	// I couldn't extract correct volume levels from it.
-	render(buffer, numSamples, 2 << 26, 2 << 26, finalVolume >> 3);
+	render(buffer, numSamples, ONE_Q31 / 16, ONE_Q31 / 16, finalVolume >> 3);
```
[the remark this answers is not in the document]
sapphire-arches: I think this and the change to a right shift by 4 below are cancelling eachother out so the overall gain change is 0db. This gain reduction is a bit of a hack, but without it something clips internally and we get a high pitched whine -- maybe the RMS calculation, because that sums potentially 64 q31s?

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -110,14 +110,14 @@ float RMSFeedbackCompressor::calcRMS(StereoSample* buffer, uint16_t numSamples)
...
+		q31_t l = thisSample->l - hpfL.doFilter(thisSample->l, hpfMovability_);
+		q31_t r = thisSample->r - hpfL.doFilter(thisSample->r, hpfMovability_);
 		q31_t s = std::max(std::abs(l), std::abs(r));
-		sum += multiply_32x32_rshift32(s, s) << 1;
+		sum += multiply_32x32_rshift32(s, s);
 
 	} while (++thisSample != bufferEnd);
 
-	float ns = float(numSamples);
+	float ns = float(numSamples) / 0.5;
```
m-m-adams, line 120: nit: this should probably be *2 for clarity and in case the compiler does something dumb

src/deluge/dsp/compressor/rms_feedback.h
sapphire-arches: i think this should be done in a separate pr/as future work. the lack of precision in the ratio display has already confused people though, so we definitely should fix it.

m-m-adams (approved):

PR #1625, opened 2024-03-31
Refactor/machine readable menus

sapphire-arches:
#1620 should be merged first, this has a subset of those changes just to make it work.


docs/menus/oscillator/wave_index.md
```diff
@@ -0,0 +1,3 @@
+Index into the wavetable.
+
+This is in the standard 0-50 range, but modulation can reach values between the increments presented by this menu item.
```
m-m-adams, line 3: There's also some hacks to make sure that modulation with an LFO can always access the full range of the wavetable with depth 50

src/deluge/gui/menu_item/generate/dmenus/compressor.py
```diff
@@ -0,0 +1,51 @@
+from dmui.dsl import *
+
+threshold = Menu(
+    "audio_compressor::CompParam",
+    "threshold",
+    ["{name}", "{title}", "params::UNPATCHED_COMPRESSOR_THRESHOLD"],
+    "STRING_FOR_THRESHOLD",
```
m-m-adams, line 7: I think title should be explicitly stated too

src/deluge/gui/menu_item/generate/dmui/dsl.py
```diff
@@ -0,0 +1,314 @@
...
+        system is in a particular state. A value of None (the default)
+        indicates this menu item is always available.
+    """
+
+    def __init__(
+        self,
+        clazz: str,
+        cpp_name: str,
+        arg_template: list[str],
+        title: str,
```
m-m-adams, line 100: Is title required if it's not used in the cpp initialization arguments?

src/deluge/gui/menu_item/generate/dmenus/oscillator/sample.py
```diff
@@ -0,0 +1,87 @@
...
+    ["{name}", "{title}", "params::LOCAL_OSC_A_PITCH_ADJUST"],
+    "STRING_FOR_OSC_TRANSPOSE_MENU_TITLE",
+    "oscillator/sample/transpose.md",
+    name="STRING_FOR_TRANSPOSE",
+)
+
+pitch_speed = Menu(
+    "sample::PitchSpeed",
+    "samplePitchSpeedMenu",
+    ["{name}"],
```
m-m-adams, line 63: What's happening in this one, why isn't there a title?

sapphire-arches: I don't know, that's just how the C++ constructor is

m-m-adams: Ah ok, thought it was shenanigans with the dsl

m-m-adams:
Ok my only nit is that I think title should be optional, as is I feel like the syntax where title is the only positional argument is a bit odd

m-m-adams (approved):

PR #1628, opened 2024-03-31
dont render compressor when threshold is zero

m-m-adams:


src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -28,9 +28,9 @@ RMSFeedbackCompressor::RMSFeedbackCompressor() {
 // 16 is ln(1<<24) - 1, i.e. where we start clipping
 // since this applies to output
 void RMSFeedbackCompressor::updateER(float numSamples, q31_t finalVolume) {
-
-	// int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
-	float songVolumedB = logf(finalVolume);
+	// 33551360
+	//  int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
```
sapphire-arches, line 32:
```suggestion
	//  int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
	// We offset the final volume by a minuscule amount to avoid a finalVolume of zero resulting in NaNs propagating.
```

Issue #1609, opened 2024-03-27
Song does not save arp sync setting


colinmeyer:
### Please describe the problem:

Change arp sync setting (either by clip -> arp -> sync or by sync shortcut key in arp column), save song, reload song, and arp sync is back to default 16th notes

### What is the expected behavior?

Song should retain the arp sync setting, e.g. Off

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

1b376355

### If possible provide the steps to reproduce the issue and upload additional media:

1. new song
2. select button (clip menu) -> arp -> sync -> off
3. save song
4. load same song
5. select button (clip menu) -> arp -> sync (now shows 16th instead of off)

sapphire-arches:
I think this matches official behavior -- if the arpegiator itself is off, we don't save any of the arpegiator state, and the code for that check goes back to the initial OS release. I'm not opposed to changing this, but for now I'm going to remove it from the 1.1 milestone.

colinmeyer:
Rate and Gate get saved, regardless of Arp on or off. Sync and Octaves do not get saved

sapphire-arches:
can you upload a song file with a preset that doesn't save correctly? Also, what do you have `DEFAULTS > RESOLUTION` set to?

m-m-adams:
Rate and gate are internal automatable parameters in a synth that get passed to the arp, so they get saved differently - this is unintuitive but consistent with official

soymonitus:
Would it be bad to save it regardless of arp off/on state? @m-m-adams 
The code change seems just to remove the if condition here (and some other places more)

<img width="972" alt="image" src="https://github.com/SynthstromAudible/DelugeFirmware/assets/6472762/9fafbf00-e949-407b-890f-fb7e52a08d56">

PR #1632, opened 2024-03-31
Save arp settings regardless of arp being Off

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1609

To cherry pick!
Note: special attention to merging this with Release branch as it contains **bdsm** references


sapphire-arches:
Do we need to update the logic in `src/deluge/processing/sound/sound.cpp` as well?

soymonitus:
> src/deluge/processing/sound/sound.cpp

Not there, because there we are already saving the arpeggiator params even if arp is off. The bug is only for what it is saved for clips when saving a song. If you save a synth patch (SAVE + SYNTH), then it is already working.

sapphire-arches:
thanks for the fix!

Issue #1633, opened 2024-03-31
Midi arpeggiator parameters are not read/rendered correctly


soymonitus:
### Please describe the problem:

When you change some of the new midi arp params for midi clip: Ratchet amount, Ratchet prob, Seq length and Rhythm, if they have a value higher than 25, they are read as a negative number when accessing again those parameters.
Apart from that, Rhythm parameter doesn't fit the screen

### What is the expected behavior?

Values are read as positive numbers 0-50
Rhythm is displayed correctly

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

PR #1636, opened 2024-04-01
Arpeggiator param saving for Midi clips fixed

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1633

NOTE: To cherry pick!

Existing songs created with previous versions of 1.1 beta may need to go into arp settings of their MIDI clips and update the params to avoid crashes, as they are now in the range 0-50 instead of full integer.


m-m-adams:
Stick to saving/loading in the full param range. This will make it unnecessarily more complicated to integrate with other deluge parameter systems in the future

soymonitus:
> Stick to saving/loading in the full param range. This will make it unnecessarily more complicated to integrate with other deluge parameter systems in the future 

I tried to stick to full range ints but, i had a problem
In which existing values like Gate whose default value is 25 are fine but for values whose default value should be zero i have a problem. When multiplying the range 0-50 by 85899345 as you see in some parts of the code, it doesn't reach the maximum integer value, but 22 units less. When the center value is 25, equivalent to zero in the signed int32 range (as Gate and Rate use) that 22 unit diff is spread equally between each half of the integer range, in a way that the conversion from full int range is, minimum value is 0 and max value is 50. I tried making it start from minimum int so the default value is 0 and not 25 like Gate, but i found the problem that values are converted (rounded) as 1 unit less. For example, i set Rhythm value to 5, and when entering again the menu option and reading it, i see value 4. I set value 23 and i enter again and see 22. Etc. Do you know what i mean? Should i do some logic to make values default to zero if value is not present in the XML?

m-m-adams:
Maybe take a look at how the pan parameter gets converted to display as +-25? I think it would have the same complexities

soymonitus:
> Maybe take a look at how the pan parameter gets converted to display as +-25? I think it would have the same complexities

I have checked how Pan (unpatched param) is programmed and was surprised to find it to have a min and max values of -25 and 25 and have not conversion logic in the pan.h class. What am I missing...?:

namespace deluge::gui::menu_item::unpatched_param {

class Pan final : public UnpatchedParam {
public:
	using UnpatchedParam::UnpatchedParam;
	virtual void drawValue();

protected:
	[[nodiscard]] int32_t getMaxValue() const override { return kMaxMenuRelativeValue; }
	[[nodiscard]] int32_t getMinValue() const override { return kMinMenuRelativeValue; }
	int32_t getFinalValue() override;
	void readCurrentValue() override;
};
} // namespace deluge::gui::menu_item::unpatched_param

soymonitus:
@m-m-adams Fixed it with something even more simple from what I thought! It was a casting problem thing (uint32 vs int32). So I reverted all my funky changes and applied just the small fix

m-m-adams (approved):

PR #1639, opened 2024-04-01
Implement Universal SysEx Identity response

pmatila:
Be a good MIDI citizen and implement standard identity request (f07e7f0601f7) and response with manufacturer ID, device ID + firmware version string.

The first commit is just a minor refactor for const correctness.


sapphire-arches (approved):
Assuming you've tested this query with something that sends the standard request and interprets the response, this LGTM. Thanks for the contribution!

pmatila:
Tested with a local Python script, yes. I can add a dbt task for testing this if you'd prefer.

sapphire-arches:
good enough for me!

pmatila:
Eep, actually don't merge! I just realized the standard response has a stricter format than implemented here.

sapphire-arches:
whoops, sorry. please create a follow-up, I don't think having a broken implementation will create any problems between now and whenever that's merged.

pmatila:
Yeah no kittens will be killed by this, I'll followup. Apologies for the mess of a premature PR :-/

PR #1640, opened 2024-04-01
Make the SysEx identity reply standards compliant, doh

pmatila:
Commit d30653ebe2c831679dff36bb0b8a76b98ce7471e implemented the spirit but not the letter of the identity reply standard, which is a strictly defined 12 or 15 byte length (depending on manufacturer ID length) message. For the Deluge it looks something like:

	7e 7f 06 02 00217b 0100 0000 01020000

The second byte is supposed to be a sysex "device id" for which this device responds to. It'd need to be configurable to be meaningful, since it's not just reply with 7f aka "anything". The middle part is manufacturer ID followed by device family and member codes. The former is the official devical device id, and there are no other family members so that part is zero. The last four bytes are major-minor-patch-unused firmware version, 1.2.0 in the above example..


pmatila:
This is a follow-up to #1639 which was, ehm, rather creative when it came to actual standards as I realized a moment too late. Apologies for jumping the gun.

m-m-adams:
No worries, lgtm 

If you implement the other direction (sending an id request and parsing the response) that would be helpful for some other work going on to support specific midi devices like launchpads and the lumi keys!

pmatila:
Oh, ok, I'll keep that in mind. I do have additional plans in the universal sysex land, just don't know when I have time.

PR #1655, opened 2024-04-02
rms compressor: increase base makeup gain

sapphire-arches:
We still lose a small amount of volume when engaging the compressor but less than before.

### DIFF
--- a/src/deluge/dsp/compressor/rms_feedback.cpp
+++ b/src/deluge/dsp/compressor/rms_feedback.cpp
@@ -28,6 +28,7 @@ RMSFeedbackCompressor::RMSFeedbackCompressor() {
 // 16 is ln(1<<24) - 1, i.e. where we start clipping
 // since this applies to output
 void RMSFeedbackCompressor::updateER(float numSamples, q31_t finalVolume) {
+
 	// 33551360
 	//  int32_t volumePostFX = getParamNeutralValue(Param::Global::VOLUME_POST_FX);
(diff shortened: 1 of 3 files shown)


m-m-adams:
I think this is now too much, it makes it really easy to clip before the compressor's attack finishes. We might need to also add a saturation/soft clipping stage to avoid that

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -122,8 +140,8 @@ float RMSFeedbackCompressor::calcRMS(StereoSample* buffer, uint16_t numSamples)
 
 	} while (++thisSample != bufferEnd);
 
-	float ns = float(numSamples);
-	mean = (2 * float(sum) / ONE_Q31f) / ns;
+	float ns = float(numSamples * 2);
+	mean = (float(sum) / ONE_Q31f) / ns;
```
sapphire-arches, line 144: @m-m-adams is this intentional? Now that I'm looking at the math again I'm pretty sure it's correct at least.

src/deluge/dsp/compressor/rms_feedback.h
```diff
@@ -36,10 +36,11 @@ class RMSFeedbackCompressor {
 	}
 
 	/// Reset the state of the compressor so no gain reduction is applied at the start of the next render window.
-	void reset() {
+	void reset(StereoSample last = {0, 0}) {
```
sapphire-arches, line 39:
```suggestion
	void reset() {
```

m-m-adams (approved):

PR #1675, opened 2024-04-06
Add readme to l10n folder

seangoodvibes:
added read me for contributors


src/deluge/gui/l10n/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+For developers looking to add additional strings for use in the codebase, you will need to update the following three files:
+
+1) english.json
+2) seven_segment.json
+3) strings.h
+
+If the string for the OLED display is the same as on the 7SEG display, then there is no need to add the string to seven_segment.json.
+
+Do not update g_english.cpp or g_seven_segment.cpp as these are auto generated by a python script from the above english.json and seven_segment.json files.
```
sapphire-arches, line 11:
```suggestion
This folder contains the list of strings that can be localized for the OLED display and their abbreviations for the 7SEG display.

For developers looking to add additional strings for use in the codebase, you will need to update the following three files:

1) `english.json`
2) `seven_segment.json`
3) `strings.h`

If the string for the OLED display is the same as on the 7SEG display, then there is no need to add the string to `seven_segment.json`.

Do not update `g_english.cpp` or `g_seven_segment.cpp` as these are auto generated by a python script from the above `english.json` and `seven_segment.json` files.
```

stellar-aria (approved):

PR #1684, opened 2024-04-07
Bugfix/midi load error arranger

### LINKED ISSUE #1285: MIDI Clips not loading from song correctly

I have an E451 on song load. This error only appears in the case of the beta version. I tried the same song on Amadeus, on 4.1.4, on nightly from 2nd February and there song loads ok with no error. In the case of the beta version, the error manifests itself in the arranger, where instead of several consecutive mapped midi channels (1-6), only channel 1 is shown on all six rows. Also, all clips disappear for these midi rows. Midi channels 2-6 appear in the arranger in the rows below the other instruments and samples. When I look at the midi channel clips, they have completely messed up CC mapping.

[Cs1X Beast 9.zip](https://github.com/SynthstromAudible/DelugeFirmware/files/14364303/Cs1X.Beast.9.zip)

Load a song without E451

_No response_

OLED

other

Beta

_No response_

m-m-adams:
Fix #1285 , an incorrect tag name was set when internal was added as a midi destination

Stops the song loader from creating fake midi/cv clips if the song loads incorrectly, making errors like this more obvious. 

Also stops the session/arranger display from killing error messages during loading to ensure the error message is shown

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -1821,7 +1821,7 @@ nothingToDisplay:
 	setCentralLEDStates();
 }
 
-// render session view display on opening
+/// render session view display on opening
 void SessionView::renderViewDisplay(char const* viewString) {
 	if (display->haveOLED()) {
 		deluge::hid::display::OLED::clearMainImage();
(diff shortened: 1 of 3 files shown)


Balistix55:
@m-m-adams and @sapphire-arches thank you for your work. The issue is gone 🫶🏻

Issue #1683, opened 2024-04-07
Midi Follow - set output active clip to current clip if no other clips on output are active


seangoodvibes:
### Please describe the problem:

In midi follow, when you send notes, it sends them to the active clip

However the active clip may be wrong if it's for an output for which no other clips are active

### What is the expected behavior?

Set active clip to current clip if no other clips on an output are active, the same way that's currently being done if you press the audition pads in a clip

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1685, opened 2024-04-07
Bugfix - set current clip as active clip if possible

seangoodvibes:
In the midi follow context, when we're looking to control the active clip (e.g. for sending notes and mpe), check if active clip should be set to the current clip and set it if needed.

This is consistent with audition pad behaviour in an instrument clip.

If you have an output for which no clip is active, when auditioning a clip for that output, the active clip should be set to the current clip.

fix: https://github.com/SynthstromAudible/DelugeFirmware/issues/1683

### DIFF
--- a/src/deluge/io/midi/midi_follow.cpp
+++ b/src/deluge/io/midi/midi_follow.cpp
@@ -149,7 +149,8 @@ Clip* getSelectedClip() {
 }
 
 // returns activeClip for the selected output
-// special case for note and performance data where you want to let notes and MPE through to the active clip
+// special case for note and performance data where you want to let notes,
+// midi modulation sources (e.g. mod wheel), and MPE through to the active clip
 Clip* getSelectedClip(ModelStack* modelStack) {
 	// If you have an output for which no clip is active,
 	// when auditioning a clip for that output,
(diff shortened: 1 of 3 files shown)


src/deluge/io/midi/midi_follow.h
```diff
@@ -32,7 +32,7 @@ class ModelStackWithThreeMainThings;
 class ModelStackWithAutoParam;
 enum class MIDIMatchType;
 
-Clip* getSelectedClip(bool useActiveClip = false);
+Clip* getSelectedClip(bool useActiveClip = false, ModelStack* modelStack = nullptr);
```
m-m-adams, line 35: This call signature is misleading now - if you set useActiveClip to true then a modelstack is required, and a modelstack isn't needed if useActiveClip is false.

Use function overloading instead, so that you have getSelectedClip() and getSelectedClip(ModelStack*)

src/deluge/io/midi/midi_follow.cpp
```diff
@@ -109,9 +110,13 @@ void MidiFollow::initMapping(int32_t mapping[kDisplayWidth][kDisplayHeight]) {
 /// 1) pressing and holding a clip pad in arranger view, song view, grid view
 /// 2) pressing and holding the audition pad of a row in arranger view
 /// 3) entering a clip
-Clip* getSelectedClip(bool useActiveClip) {
+Clip* getSelectedClip(bool useActiveClip, ModelStack* modelStack) {
 	// special case for note and performance data where you want to let notes and MPE through to the active clip
 	if (useActiveClip) {
```
m-m-adams, line 115: split this to it's own function getSelectedClip(ModelStack* modelStack)

m-m-adams (approved):

PR #1687, opened 2024-04-08
Fix note-on ignoring during quantized record

### LINKED ISSUE #1422: Midi Notes lengths get truncated when recording in arranger mode

When recording a midi track (to play an external synth) in arranger mode, long notes get truncated to almost 0 length intermittently.  Only occurs with quantization on.  When quantization is off, notes record properly.  (This does not happen when in song mode.)

Note offs should not be quantized when quantization is enabled in arranger mode.  Note lengths should remain as played.

_No response_

OLED

other

C1.2.0

https://github.com/SynthstromAudible/DelugeFirmware/assets/147354026/01157925-d96c-44e2-ae15-76efa1cf9b87

sapphire-arches:
Rather than using a boolean that gets out of sync easily, we record the actual relative time stamp at which we should start paying attention.

There's probably gremlins in the reverse clip recording code, but I don't think this makes it any worse at least.

Fixes #1422.

### DIFF
--- a/src/deluge/model/note/note_row.cpp
+++ b/src/deluge/model/note/note_row.cpp
@@ -54,7 +54,7 @@ NoteRow::NoteRow(int16_t newY) {
 	drum = NULL;
 	firstOldDrumName = NULL;
 	soundingStatus = STATUS_OFF;
-	skipNextNote = false;
+	ignoreNoteOnsBefore_ = 0;
 	probabilityValue = kNumProbabilityValues;
 	loopLengthIfIndependent = 0;
 	sequenceDirectionMode = SequenceDirection::OBEY_PARENT;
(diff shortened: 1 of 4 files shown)


src/deluge/model/note/note_row.cpp
```diff
@@ -822,10 +822,12 @@ Error NoteRow::clearArea(int32_t areaStart, int32_t areaWidth, ModelStackWithNot
 void NoteRow::recordNoteOff(uint32_t noteOffPos, ModelStackWithNoteRow* modelStack, Action* action, int32_t velocity) {
 
 	// If we're just about to pass the actual recorded note, though, don't do it.
-	if (skipNextNote) {
+	if (noteOffPos < ignoreNoteOnsBefore_) {
 		return;
 	}
 
+	D_PRINTLN("rno %u", noteOffPos);
```
m-m-adams, line 829: Was this for debugging?

m-m-adams (approved):
Lgtm pending the print getting removed or made more legible

src/deluge/model/note/note_row.cpp
sapphire-arches: yeah, deleted it

PR #1690, opened 2024-04-08
Bugfix/fallback for files saved with incorrect tag on nightly builds

m-m-adams:
Fallback in midi instrument parser to use midiChannel as the channel if present. Was accidentally saved as midiChannel instead of channel for a few weeks

Fix to show error message on OLED (7seg already handles popups vs loading differently)

### DIFF
--- a/src/deluge/hid/display/display.h
+++ b/src/deluge/hid/display/display.h
@@ -12,7 +12,8 @@ class NumericLayerScrollingText;
 
 enum DisplayPopupType {
 	NONE,
-	GENERAL,     // Default popup type, if not specified
+	GENERAL,
+	LOADING,     // Default popup type, if not specified
 	PROBABILITY, // Popup shown when editing note or row probability
 	             // Note: Add here more popup types
 };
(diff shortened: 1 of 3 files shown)


src/deluge/hid/display/display.h
```diff
@@ -12,7 +12,8 @@ class NumericLayerScrollingText;
 
 enum DisplayPopupType {
 	NONE,
-	GENERAL,     // Default popup type, if not specified
+	GENERAL,
+	LOADING,     // Default popup type, if not specified
```
sapphire-arches, line 16:
```suggestion
	/// Default popup type, if not specified.
	GENERAL,
	/// Used for popups generated during file loading.
	LOADING,
	/// Popup shown when editing note or row probability.
```

src/deluge/hid/display/display.h
```diff
@@ -12,7 +12,8 @@ class NumericLayerScrollingText;
...
 	PROBABILITY, // Popup shown when editing note or row probability
 	             // Note: Add here more popup types
```
sapphire-arches, line 18:
```suggestion
	PROBABILITY,
	// Note: Add here more popup types
```

Issue #1048, opened 2024-01-23
Nudging two or more notes of the same row at the same time on pad grid can create inconsistencies.


ok-reza:
### Please describe the problem:

1) Enter two notes on the same row of a synth clip
2) Press and hold each note's pad (start from right to left to avoid note length extending.
3) Press and turn <> encoder to nudge clips, turn 3 times counter clockwise then 2 times clockwise
4) Press one pad and press <> encoder to check current nudge amount
5) Now press the other pad and press <> encoder to check current nudge amount
6) Note that one pad has a nudge of -2 and the other -3.

### What is the expected behavior?

Nudge amounts should be the same when editing at the same time. 

~~For note probability, one note will have latching on, not sure if that's default behavior or the same one knob off issue.~~

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

1.1.0 - 8FEA0F5

### If possible provide the steps to reproduce the issue and upload additional media:

This can also be observed by doing steps 1-3 but instead of turning 3 times, just turn counter clockwise and then clockwise back and forth, eventually one of the notes will start nudging farther and farther away while the other remains in the same place

m-m-adams:
The latching is intended behaviour, if you're holding multiple notes in a row the probability system interprets it as desiring them all to play/not play together

PR #1693, opened 2024-04-09
Remove partial nudge undo logic which is buggy for multiple notes held

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1048

Although the diff shows a lot of changes they are just indent changes. This commit is basically deleting a special case logic which was when you did nudge in one direction and immediately afterwards nudge in the opposite direction, it was doing a "partial" undo to restore note tails that may have been cut. While this could be a nice-to-have feature, it is just a very edge case (as the note tails are only cut if the nudging goes further so that the note start changes from pad at current zoom level) and the logic was not working properly for multiple notes being held. By removing this "partial undo" for when nudging back and forth, the multi-note nudging is working again consistently.


sapphire-arches:
I think I like this solution better - since you can already undo explicitly if the nudge does something you don't like, that seems better than something that behaves inconsistently depending on if the nudge is 1 or more notes.

Issue #1084, opened 2024-01-27
OLED is not correctly reset on boot


sapphire-arches:
Occasionally after flashing (either via sysex or via the SD bootloader) the OLED column pointer ends up somewhere offset and the whole display is aligned. In principle the `0x22` and `0x20` commands should reset the column pointer (according to to SSD1306 manuals), but apparently the display controller on our display is not a complete implementation.

m-m-adams:
From some testing adding a 100ms delay as soon as the timers are initialized works. I think debugging has the same issue as with yesterdays nightly where the dma hasn't completed. Since the dma from the previous run isn't stopped by the load we can wait for it to finish before resetting the dma controllers later in the startup routine

sapphire-arches:
We might be able to detect that the DMA engine is still active by checking the `EN` bit of the DMA engine - according to figure 9.3 of R01UH0437EJ0600 Rev.6.00 (RZ/A1L TRM) the DMA engine should reset that bit when the transfer completes. Adding a bit of code to `start.S` to wait for the DMA channel used by the bootloader and the main firmware (DMAC 4) to go idle.

https://github.com/SynthstromAudible/DelugeBootloader/blob/6ce8601eef539bcc3ad91432a1b8779a19aaaae5/src/drivers/oled/oled_low_level_all_cpus.h#L34

m-m-adams:
Technically still a problem but we don't really have a fix and the current state is fine

PR #1706, opened 2024-04-11
Maybe fix: OLED is not correctly reset on boot #1084 by waiting for Oled DMA to be idle at startup

M0R172:
I have not seen any OLED startup problems in a while, but is this PR what @sapphire-arches meant in #1084 by

> We might be able to detect that the DMA engine is still active by checking the EN bit of the DMA engine

Rather checking the TACT (transaction active) bit here, as IMO we do not want to see if the DMA channel is generally *enabled* but if it is still doing a transfer:

Manual:
> Bit 2 TACT Transaction Active
> Indicates whether the DMAC is active. This bit is intended to check that the channel is
> completely inactive.
> 0: DMA on Channel_n inactive
> 1: DMA on Channel_n active
> Set condition(s):
> • When a DMA transaction starts on Channel_n
> Clear condition(s):
> • When a DMA transaction is completed


m-m-adams:
Can you try removing the linker padding and see if it still boots without errors? The padding in the linker script also stops the boot loader transferring control while the dma transfer is running but is obviously not desirable

sapphire-arches:
Removing the `clearMainImage`  in `midi/sysex.cpp:firstPacket` could be another good way to test.

M0R172:
Thanks for the hints, indeed it does not seem to help when disabling the other fixes - will try some other variants later...

M0R172:
Well, actually...

> Can you try removing the linker padding and see if it still boots without errors?

It boots just fine without the linker padding: Updating via sysex, updating from sd card, or regular booting from sd card, all works fine - thing is: *it works fine without any other fix being applied*. Just remove the padding from the current community, and it still works!


> Removing the clearMainImage in midi/sysex.cpp:firstPacket could be another good way to test.

Removing *this* still gets the screen messed up every 5th or so time when updating via sysex, and waiting for DMA at startup does not help. FYI, what is actually sufficient to prevent this issue is just `uiTimerManager.unsetTimer(TimerName::OLED_SCROLLING_AND_BLINKING)`
which is called by `stopScrollingAnimation` called by `clearMainImage`.
(I would keep the  `clearMainImage`, as IMO it looks better to clear the screen than just freeze it.)


So maybe two issues (boot problem as fixed by padding, screen problem as fixed by stopping timer) are actually not that closely related? Any ideas how to reproduce the boot issues that are supposed to happen without padding? I even checked out commit e0f4e248, which is the one before the padding was introduced in #1411, and did not see any issues there - WTF?

OTOH, curiosity aside, none if this is causing real problems at the moment, so maybe I/we should not waste too much time on it... *shrug*

m-m-adams:
> It boots just fine without the linker padding: Updating via sysex, updating from sd card, or regular booting from sd card, all works fine - thing is: *it works fine without any other fix being applied*. Just remove the padding from the current community, and it still works!

It depends on the binary size so without the padding the bug comes and goes. The bootloader sends an image every 16k (I think) bytes that it loads and it only gets unaligned if it sends one and immediately starts the deluge before it completes

M0R172:
Thanks for the info!

As 

1. I could not find a way to reproduce the alignment problem (I tried things like align(0x4010), but to no avail)
2. There currently is no real issue besides wasting some kilobytes of memory on the alignment

I give up on this topic for now and close this PR.

PR #1708, opened 2024-04-11
Os/task handler

m-m-adams:
More work for the move to a central scheduler. This PR has the following:

- Restructure folders, start to seperate deluge app logic from the OS like stuff
- Adds a basic scheduler with a C api to add tasks and run it, along with accompanying tests. Can be enabled to replace the main loop by building with USE_TASK_MANAGER defined
- Restructures the tests workflows. There are now seperate cpputest builds for 32 bit and non-32bit tests to allow easier testing on non-linux systems


m-m-adams:
I swear the tests pass locally....

src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -0,0 +1,178 @@
...
+constexpr double rollTime = ((double)(UINT32_MAX) / DELUGE_CLOCKS_PERf);
+
+void TaskManager::runTask(taskID id) {
+	list[id].lastCallTime = getTimerValueSeconds(0);
+	list[id].handle();
+	double runtime = (getTimerValueSeconds(0) - list[id].lastCallTime);
+	if (runtime < 0) {
+		runtime += rollTime;
+	}
+	if (list[id].averageDuration < 0.000001 | list[id].averageDuration > 0.01) {
```
sapphire-arches, line 124: should be `||`

sapphire-arches (approved):

PR #1728, opened 2024-04-15
Refactor/rearrange tags written in XMLs so the songs load ok in official firmware

soymonitus:
This fixes the issue of not being able to load a song saved with community 1.1.0/1.2.0 with official firmware 4.1.2


soymonitus:
> I think we should bump the firmware tag - pretty certain official is still not loading properly even if this stops the crashing
> 
> 
> 
> Also conflicted on the move since I'd prefer to keep the filter params grouped 

So you mean that although it loads without crashing that maybe some params are not read?

m-m-adams:
Yeah unless I'm misreading it'll exit all blocks like this at the first unrecognized param name, so it won't finish reading all the arp params. Possibly the reverb too depending on where the new names were put. 

It's not clear to me why morph is crashing instead of just not being recognized though

m-m-adams:
Yeah that's probably a good idea

soymonitus:
> Yeah that's probably a good idea

What if I merge the changes from this PR https://github.com/SynthstromAudible/DelugeFirmware/pull/1727 into this one so all the fixes related to opening a song correctly with previous firmware are contained in one PR?

m-m-adams:
> Btw, confirmed that in other places we are writing first the syncType (new param) and then the syncLevel, and in the official firmware the syncLevel is never read as the syncType is unrecognized. Happens with LFOs, sidechain, delay, etc..

Yeah if you could add this fix and the arp ones into one PR that would be great!

soymonitus:
Ready for review

Issue #823, opened 2023-12-13
Cannot rename instruments in Arranger view via NAME shortcut (7SEG only)


ok-reza:
### Please describe the problem:

It seems renaming is blocked on 7SEG in Arranger view. The keyboard shows up but no input is allowed.

### What is the expected behavior?

To be able to rename songs. This may be an oversight with OLED/7SEG behavior on the same firmware?

### What hardware did you reproduce it with?

7SEG

### What firmware did you reproduce it with?

Release 1.0/Amadeus

### What is the firmware name:

c1.0.0

### If possible provide the steps to reproduce the issue and upload additional media:

Open a previous song or start a new one.
Enter Arranger mode
Hold Audition Pad of a given instrument + hit NAME shortcut.
Keyboard appears, no input received.

m-m-adams:
On one hand this is a bug, on the other hand renaming instruments breaks the synth preset loading logic in both community and official. So while this is probably fixable we should fix loading first

lburygin:
Actually you can rename, try to just type your instrument name and press enter. It is just the letters you typing don't show up on the screen, that's a bug, I found it too recently

vcastellm:
Same here, renaming works but it's not displayed.

bobtwinkles:
We should fix the display issue for 1.1, though we probably can't fix the browser bugs before then....

PR #1729, opened 2024-04-15
Fix renaming for 7SEG display for both Arranger tracks and kit rows

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/823

Now when renaming an arranger track or renaming a kit row, in 7SEG displays, it will show the name while you are typing it. Before, you were able to rename it anyways, the only problem was that the screen was not refreshing and showing the characters you were typing

To cherry pick

### DIFF
--- a/src/deluge/gui/ui/rename/rename_ui.cpp
+++ b/src/deluge/gui/ui/rename/rename_ui.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "gui/ui/rename/rename_ui.h"
+#include "gui/ui/qwerty_ui.h"
 #include "hid/display/oled.h"
 
 RenameUI::RenameUI() {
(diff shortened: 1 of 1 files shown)


src/deluge/gui/ui/rename/rename_ui.cpp
```diff
@@ -24,7 +25,10 @@ RenameUI::RenameUI() {
 }
 
 void RenameUI::displayText(bool blinkImmediately) {
-	renderUIsForOled();
+	if (display->haveOLED()) {
+		renderUIsForOled();
+	}
+	QwertyUI::displayText(blinkImmediately);
```
m-m-adams, line 31: should this be in an else block?

soymonitus:
I have added the "else" clause. It works well with both OLED and 7SEG, for kit rows and arranger tracks

PR #1732, opened 2024-04-18
bug fix - offer learned cc's for song params

seangoodvibes:
Fixed bug where you can learn a midi cc to a song param but the cc wasn't being offered to the song mod controllable by creating a new mod controllable audio function for offering the received cc to the song

### DIFF
--- a/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
+++ b/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
@@ -1307,9 +1307,10 @@ ModelStackWithThreeMainThings* ModControllableAudio::addNoteRowIndexAndStuff(Mod
 	return modelStackWithThreeMainThings;
 }
 
-bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber,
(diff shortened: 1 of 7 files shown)


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1307,9 +1307,9 @@ ModelStackWithThreeMainThings* ModControllableAudio::addNoteRowIndexAndStuff(Mod
 	return modelStackWithThreeMainThings;
 }
 
-bool ModControllableAudio::offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber,
-                                                          uint8_t value, ModelStackWithTimelineCounter* modelStack,
-                                                          int32_t noteRowIndex) {
+bool ModControllableAudio::offerReceivedCCToLearnedParams(
```
m-m-adams, line 1310: Can we refactor instead of bolting more into this function? This situation with two model stack arguments where one must be null is hard to follow and I think it'll lead to bugs down the road

A custom function just for offering CCs to song level gold knobs might be appropriate

seangoodvibes: Will do! Thx

PR #1737, opened 2024-04-18
Midi Follow ~ Add midi follow setting to make controlling song params optional

seangoodvibes:
Added new MIDI Follow default setting to make control of Song params optional. Controlling song params is enabled by default.

If control of song params is disabled and no clip is selected, MIDI Follow CC's received will be directed to control the active clip of the last instrument selected

Would like to cherry pick this to 1.1 if possible. I expect there will be some merge conflicts (e.g. string related) but I'll resolve them.


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -1454,68 +1454,64 @@ void ModControllableAudio::receivedCCFromMidiFollow(ModelStack* modelStack, Clip
...
-							    oldValue, modelStackWithParam);
-
-							// add 64 to internal knobPos to compare to midi cc value received
-							// if internal pos + 64 is greater than 127 (e.g. 128), adjust it to 127
-							// because midi can only send a max midi value of 127
-							int32_t knobPosForMidiValueComparison = knobPos + kKnobPosOffset;
-							if (knobPosForMidiValueComparison > kMaxMIDIValue) {
-								knobPosForMidiValueComparison = kMaxMIDIValue;
-							}
+						// get current value
```
m-m-adams, line 1457: why is this being changed? With the indentation change I'm not sure what the actual difference is

src/deluge/model/mod_controllable/mod_controllable_audio.h
```diff
@@ -91,7 +91,7 @@ class ModControllableAudio : public ModControllable {
 	virtual bool setModFXType(ModFXType newType);
 	bool offerReceivedCCToLearnedParams(MIDIDevice* fromDevice, uint8_t channel, uint8_t ccNumber, uint8_t value,
 	                                    ModelStackWithTimelineCounter* modelStack, int32_t noteRowIndex = -1);
-	void receivedCCFromMidiFollow(ModelStack* modelStack, Clip* clip, int32_t ccNumber, int32_t value);
+	static void receivedCCFromMidiFollow(ModelStack* modelStack, Clip* clip, int32_t ccNumber, int32_t value);
```
m-m-adams, line 94: Since this function can be static it probably makes more sense to be part of midi follow than modcontrollable

seangoodvibes: The original reason I had it in here was because that midi takeover calculation function is in here. I think I'll factor that out as well.

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
seangoodvibes: it's cause I removed this - this code basically checks that you're trying to control the same thing as the gold knobs. But since we're allowing controlling the last selected clip, this won't work anymore and isn't needed as we are controlling something different than the gold knobs.

```
if (modelStackWithParam->getTimelineCounter()
						    == view.activeModControllableModelStack.getTimelineCounterAllowNull()) {
```

m-m-adams (approved):

PR #1739, opened 2024-04-19
Add Note Row Play Direction to Synth / Midi / CV

seangoodvibes:
Added note row config of play direction for synth / midi / cv clips.

Hold audition pad while entering play direction menu

This also allows for quickly setting play direction for multiple note rows while in the note row play direction menu. Auditioning different note rows will refresh the play direction menu to show the play direction setting for each note row

I think we should merge this to 1.1 since we removed the previous kit workaround for setting note row play direction

### DIFF
--- a/src/deluge/gui/menu_item/sequence/direction.h
+++ b/src/deluge/gui/menu_item/sequence/direction.h
@@ -18,6 +18,7 @@
 #include "definitions_cxx.hpp"
 #include "gui/menu_item/selection.h"
 #include "gui/ui/sound_editor.h"
+#include "gui/views/instrument_clip_view.h"
 #include "model/clip/instrument_clip.h"
 #include "model/instrument/kit.h"
 #include "model/model_stack.h"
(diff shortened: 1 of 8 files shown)


src/deluge/gui/menu_item/sequence/direction.h
```diff
@@ -37,6 +38,20 @@ class Direction final : public Selection {
 				return clip->getNoteRowForDrum(modelStack, kit->selectedDrum); // Still might be NULL;
 			}
 		}
+		else if (clip->output->type != OutputType::KIT) {
+			if (soundEditor.selectedNoteRow) {
+				// get model stack with note row but don't create note row if it doesn't exist
+				ModelStackWithNoteRow* modelStackWithNoteRow =
+				    clip->getNoteRowOnScreen(instrumentClipView.lastAuditionedYDisplay, modelStack);
+				if (modelStackWithNoteRow != nullptr) {
```
m-m-adams, line 46: this can never be null, it'll be the original modelstack with a null note row if it fails

You check that in the next line so this one can be removed entirely

PR #1745, opened 2024-04-20
bump toolchain version

sapphire-arches:
fixes certifi install failure


jamiefaye (commented):
no need for change to dbtenv.sh  ?

PR #1749, opened 2024-04-21
Fix explode / implode animations to/from arranger / grid views

seangoodvibes:
Fixed explode / implode animations to / from arranger / grid views


seangoodvibes:
Converting to draft, still some work to do it seems.

m-m-adams (approved):

PR #1756, opened 2024-04-22
Code cleanup

seangoodvibes:
Refactored some code to make code more legible and remove some unnecessary checks

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -243,6 +243,14 @@ void SoundEditor::setLedStates() {
 ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) {
 	using namespace deluge::hid::button;
 
+	Clip* clip = nullptr;
+	bool isUIInstrumentClipView = false;
+
+	if (rootUIIsClipMinderScreen()) {
+		clip = getCurrentClip();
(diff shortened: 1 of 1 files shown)


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -383,9 +383,6 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 				}
 			}
 		}
```
seangoodvibes, line 385: Removing this disables toggling of affect entire while in menu

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -912,8 +909,7 @@ ActionResult SoundEditor::potentialShortcutPadAction(int32_t x, int32_t y, bool
...
 
 		// For Kit Instrument Clip with Affect Entire Enabled
-		else if (setupKitGlobalFXMenu && (getCurrentOutputType() == OutputType::KIT)
-		         && (getCurrentInstrumentClip()->affectEntire)) {
+		else if (setupKitGlobalFXMenu) {
```
seangoodvibes, line 912: setupKitGlobalFXMenu is only ever true if it's a kit and affect entire is enabled, so no need to check for kit and affect entire again...

m-m-adams:
This seems undesirable, testing this and it feels weird to have the gold knobs change when I enter a menu

m-m-adams (approved):

PR #1759, opened 2024-04-23
columns: save state of control columns per clip

bfredl:
Previously,
State for sidebars would just implicitly be associated with KeyboardLayout class. This moves the state to be specific per clip, just like other keyboard state.

Also implement loading/saving of this state as part as .xml song files

### DIFF
--- a/src/deluge/gui/ui/keyboard/column_controls/chord_mem.cpp
+++ b/src/deluge/gui/ui/keyboard/column_controls/chord_mem.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "chord_mem.h"
+#include "gui/ui/keyboard/layout/column_controls.h"
 #include "hid/buttons.h"
 #include "model/song/song.h"
 
@@ -23,9 +24,9 @@ namespace deluge::gui::ui::keyboard::controls {
 
 void ChordMemColumn::renderColumn(RGB image[][kDisplayWidth + kSideBarWidth], int32_t column) {
 	uint8_t otherChannels = 0;
-	for (int32_t y = 0; y < kDisplayHeight; ++y) {
+	for (int32_t y = 0; y < kDisplayHeight; y++) {
 		bool chord_selected = y == activeChordMem;
-		uint8_t chord_slot_filled = currentSong->chordMemNoteCount[y] > 0 ? 0x7f : 0;
+		uint8_t chord_slot_filled = chordMemNoteCount[y] > 0 ? 0x7f : 0;
 		otherChannels = chord_selected ? 0xf0 : 0;
 		uint8_t base = chord_selected ? 0xff : chord_slot_filled;
 		image[y][column] = {otherChannels, base, base};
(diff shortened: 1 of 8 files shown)


src/deluge/gui/ui/keyboard/layout/column_control_state.h
```diff
@@ -0,0 +1,65 @@
...
+	ChordColumn chordColumn{};
+	ChordMemColumn chordMemColumn{};
+	ScaleModeColumn scaleModeColumn{};
+	DXColumn dxColumn{};
+
+	ColumnControlFunction leftColFunc = VELOCITY;
+	ColumnControlFunction rightColFunc = MOD;
+	bool rightColSetAtRuntime = false;
+	ControlColumn* leftColPrev = &velocityColumn;
+	ControlColumn* rightColPrev = &modColumn;
```
soymonitus, line 55: leftColPrev and rightColPrev were deleted in the last commit/fix as they were not needed. They are not used in the code so can be deleted.

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
```diff
@@ -135,18 +142,19 @@ void ColumnControlsKeyboard::checkNewInstrument(Instrument* newInstrument) {
...
 	}
+	ColumnControlState& state = getState().columnControl;
 	auto* sound = (SoundInstrument*)newInstrument;
 	bool is_dx = (sound->sources[0].oscType == OscType::DX7);
 	if (!is_dx) {
-		if (rightColFunc == DX) {
-			rightColSetAtRuntime = false;
+		if (state.rightColFunc == DX) {
+			state.rightColSetAtRuntime = false;
 			// use rightColPrev if admissable??
```
soymonitus, line 151: Outdated comment

src/deluge/gui/ui/keyboard/layout/column_control_state.h
bfredl: updated, thanks!

src/deluge/gui/ui/keyboard/column_controls/chord_mem.cpp
```diff
@@ -64,4 +65,67 @@ void ChordMemColumn::handlePad(ModelStackWithTimelineCounter* modelStackWithTime
...
+		}
+	}
+
+	if (num == 0) {
+		return; // no state to save
+	}
+
+	writer.writeOpeningTag("chordMem");
+	for (int32_t y = 0; y < num; y++) {
+		writer.writeOpeningTag("chordSlot");
```
soymonitus, line 83: For my PR that saves chordmem to global song, I copied the same logic that you use here. The only thing that I changed was renaming "chordSlot" to just "chord" to make it shorter. Maybe good if you also update the tag to just "chord" so both are equal.

soymonitus:
I think it would be a good opportunity to also save the edits made to the other columns that can be edited, that is, Scales, Velocity and Modwheel.
Scales: save each scale (index) that has been assigned to each of the 8 pads
Velocity: save range (maximum and minimum velocities assigned to the upper and lower pads)
Modwheel:  same as velocity (min and max)

src/deluge/gui/ui/keyboard/column_controls/chord_mem.cpp
m-m-adams: Since yours is song level and this is clip level I like the separate names - duplicate names in parsing song vs. clip parameters has caused a few bugs already

m-m-adams:
Saving more side bars seems like a future addition more than something in this PR

m-m-adams (approved):

m-m-adams:
@soymonitus for integration I'd suggest adding logic so that the clip copies the song chords unless it has it's own set

soymonitus:
> @soymonitus for integration I'd suggest adding logic so that the clip copies the song chords unless it has it's own set

should i address that in a new PR after this is merged?

m-m-adams:
> > @soymonitus for integration I'd suggest adding logic so that the clip copies the song chords unless it has it's own set
> 
> should i address that in a new PR after this is merged?

yeah I think so

m-m-adams:
This appears to have picked up an issue preventing it from building on current head

bfredl:
Updated. it ignores the song-level chords now, I take @soymonitus plan to address that in a follow-up.

soymonitus:
Oh this was merged. Nice! I will work on the adaptation code to fallback to song state if clip state not found

PR #1765, opened 2024-04-24
Added High CPU usage default setting

seangoodvibes:
Added default high CPU usage indicator setting that let's you choose between Enabled and Disabled


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -438,6 +436,17 @@ inline void cullVoices(size_t numSamples, int32_t numAudio, int32_t numVoice) {
...
 			D_PRINTLN("under min voices but culling anyway");
 			cullVoice(false, FORCE, numSamples, nullptr);
+			cullType = CPUCullingIndicator::FORCE;
+		}
+	}
+	// blink LED to alert the user
+	if ((FlashStorage::defaultCPUCullingIndicatorMode != CPUCullingIndicator::OFF)
+	    && (cullType != CPUCullingIndicator::OFF)) {
+		if ((FlashStorage::defaultCPUCullingIndicatorMode == CPUCullingIndicator::ALL)
+		    || (FlashStorage::defaultCPUCullingIndicatorMode == cullType)) {
```
m-m-adams, line 446: this means that hard culls won't blink if you have it set to soft, which is probably not desirable

src/deluge/gui/menu_item/defaults/cpu_culling_indicator.h
```diff
@@ -0,0 +1,40 @@
...
+namespace deluge::gui::menu_item::defaults {
+class CPUCullingIndicatorMode final : public Selection {
+public:
+	using Selection::Selection;
+	void readCurrentValue() override { this->setValue(FlashStorage::defaultCPUCullingIndicatorMode); }
+	void writeCurrentValue() override {
+		FlashStorage::defaultCPUCullingIndicatorMode = this->getValue<CPUCullingIndicator>();
+	}
+	deluge::vector<std::string_view> getOptions() override {
+		return {
```
m-m-adams, line 33: I think we should just do on and off, and name it a high CPU usage indicator

m-m-adams (approved):

PR #1767, opened 2024-04-25
song: use 1.0 base makeup gain for compressor

sapphire-arches:


m-m-adams:
https://github.com/SynthstromAudible/DelugeFirmware/blob/772f1fdd33cc94e24241f2fd5e7272dd54febb46/src/deluge/processing/engines/audio_engine.cpp#L830

I think you also need to remove the >> 1 from the left and right volumes, that was added when the makeup gain was increased

sapphire-arches:
In that case, should we merge this as-is?

m-m-adams (approved):

PR #1769, opened 2024-04-25
Bugfix: Disable midi learning of global effectable params

seangoodvibes:
There is a bug with parsing midi learns for global effectables. Until this can be fixed, midi learning for global effectables has been disabled.


m-m-adams (changes requested):
This breaks patch cables and gold knob mappings

Gold knobs don't know what kind of param they're controlling while being read (patched or unpatched) and need the fall through in this function. Patch cables are using the same approach but can possibly be set to only read patched params

m-m-adams:
I'm repointing this to beta - we can leave it in the nightly while we find a fix

PR #1775, opened 2024-04-26
Workflow fixups - correct names and tags

m-m-adams:
Checkout the release branch before the nightly publish, otherwise it moves the beta tag back to community which looks dumb. This was previously handled by the build and publish being in the same job, but broke silently when I split them up

Give the two PR build checks different names for the type of build

Remove the review request from the cherry pick bot so I don't get notifications now that it's working smoothly

### DIFF
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -28,5 +28,3 @@ jobs:
           branch: release/1.1
           labels: |
             cherry-picked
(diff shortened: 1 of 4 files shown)


seangoodvibes (approved):

m-m-adams:
The required checks need to be updated to change "Build Check" to "Build check - Release" and "Build check - Debug", this PR gives them unique names

PR #1794, opened 2024-04-28
Bugfix/recover from not finding synth on card

### LINKED ISSUE #1789: Getting "file not found" when pressing LOAD+SYNTH or turning SELECT knob

After flashing c1.1.0 Beethoven when I try to load a preset or scoll to next i get "file not foun error"

To bring u a browse list or scroll to next/previous preset

_No response_

OLED

other

deluge-v1_1_0-beta+2024_04_28-cc25fff.bin

_No response_

### LINKED ISSUE #1792: "File not found" when adding tracks to projects started on Official firmware

Some (but not all) projects that were started with the official firmware will not allow me to create new synth / kit / midi tracks after flashing to the new C1.1.0 firmware. When I press a new row in Song view (or when I press the row button in Arranger view), it gives me a "file not found" error. I can, however, still copy existing tracks. 

Creating a new track.

I don't believe so

OLED

Built from community

Community C1.1 Beethoven Beta

_No response_

m-m-adams:
Fix #1792 
Fix #1789 

For projects where the synth is no longer in it's original location it will be loaded as "exists on card" but might not be in it's original location. In this case we'll cache that it's not on card but not return an error, as that's signalled through changing the existsOnCard attribute and the null file pointer.

### DIFF
--- a/src/deluge/storage/file_item.cpp
+++ b/src/deluge/storage/file_item.cpp
@@ -38,8 +38,9 @@ Error FileItem::setupWithInstrument(Instrument* newInstrument, bool hibernating)
 		tempFilePath.concatenate(filename.get());
 		bool fileExists = storageManager.fileExists(tempFilePath.get(), &filePointer);
 		if (!fileExists) {
-			D_PRINTLN("couldn't get filepath for file %d", filename.get());
-			return Error::FILE_NOT_FOUND;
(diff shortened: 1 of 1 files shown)


chewabledrapery:
I've checked the XML of my project file and 100% the path to all of the presets exist and are correct, any ideas?

PR #1820, opened 2024-05-03
Try to guide people to the per-release versions of community docs

pmatila:
Explain what the main community_features.md describes and add links to the corresponding release branches and changelog. Hoping to reduce the number of folks asking why feature X is not working for them after landing on the community_features doc from search engines or such.


m-m-adams (approved):
Great idea, thanks!

PR #1824, opened 2024-05-03
Feature/configure max voices

### LINKED ISSUE #1803: Maximum polyphony per synth is not configurable

The maximum polyphony for a track is currently either 16 (polyphonic) or 1 (monophonic/legato)

The check is here https://github.com/m-m-adams/DelugeFirmware/blob/be97feb547c92a6e9620bf34b05ee3dce9cca5b2/src/deluge/processing/engines/audio_engine.cpp#L1319 

It would be nice if this was a user configurable option in the polyphony menu, at least something like 1/4/8/16

#1751

OLED, 7SEG

Nightly, other

_No response_

_No response_

m-m-adams:
Close #1803 

Adds a submenu under the polyphonic option in the polyphony menu to set a max number of voices. It defaults to 8 for new synths and 16 for old synths where a max number is not configured.

Also makes more enums single byte, poor ordering and large enums was making sounds and voices ~20 bytes larger than needed

Fixes the IntegerWithOff class to also show off on 7seg - affects the MPE channel count menu, MPE velocity menu, count in bars, and Saturation menus.


src/deluge/gui/l10n/english.json
```diff
@@ -1,809 +1,731 @@
...
+    "STRING_FOR_TIME": "Time",
+    "STRING_FOR_MASTER": "Master",
+    "STRING_FOR_FILTERS": "Filters",
+    "STRING_FOR_CANT_TRANSPOSE": "Can't Transpose: No Scale Mode Clips",
+    "STRING_FOR_TRANSPOSE_INKEY": "In Key",
+    "STRING_FOR_TRANSPOSE_CHROMATIC": "Chromatic",
+    "STRING_FOR_TRANSPOSE_CHORD": "Chord",
+    "STRING_FOR_CANT_ENTER_SCALE": "Can't enter scale mode, MIDI transpose is chromatic",
+    "STRING_FOR_DEFAULT_HIGH_CPU_USAGE_INDICATOR": "High CPU Indicator"
+  }
```
sapphire-arches, line 729: Something weird happened here, maybe line endings?

src/deluge/processing/sound/sound.cpp
```diff
@@ -1253,6 +1253,11 @@ Error Sound::readTagFromFile(StorageManager& bdsm, char const* tagName, ParamMan
 		bdsm.exitTag("polyphonic");
 	}
 
+	else if (!strcmp(tagName, "maxVoices")) {
+		maxVoiceCount = bdsm.readTagOrAttributeValueInt();
+		bdsm.exitTag("maxVoices");
+	}
+
```
sapphire-arches, line 1260: Does this cause official to crash/not load subsequent settings?

sapphire-arches:
It would probably be good to have the voice count directly accessible under the `VOICE` menu as well -- honestly I might prefer that over the select-swap menu thing? That interaction pattern is already used for unison stereo spread but it feels kind of weird.

m-m-adams:
> It would probably be good to have the voice count directly accessible under the `VOICE` menu as well -- honestly I might prefer that over the select-swap menu thing? That interaction pattern is already used for unison stereo spread but it feels kind of weird.

Yeah I'll add it to the voice menu too. I like having it in the polyphony menu as well since it's semantically choosing "how polyphonic" and that gives a way to access it via shortcuts

src/deluge/gui/l10n/english.json
m-m-adams: Oops yeah - I'll check

src/deluge/processing/sound/sound.cpp
m-m-adams: yep, moved to the end

src/deluge/processing/sound/sound.cpp
```diff
@@ -4022,6 +4021,9 @@ void Sound::writeToFile(StorageManager& bdsm, bool savingSong, ParamManager* par
 	bdsm.writeClosingTag("modKnobs");
 
 	ModControllableAudio::writeTagsToFile(bdsm);
+	
+	// community attributes, new additions will stop official firmware from continuing to parse the sound
+	bdsm.writeAttribute("maxVoices", maxVoiceCount);
```
soymonitus, line 4026: Yes! glad you remembered it :)

m-m-adams: Well Reed did at least!

seangoodvibes:
This is a wonderful addition! Seems to work great!

PR #1827, opened 2024-05-04
Bugfix Midi Follow Feedback

seangoodvibes:
- Fixed midi follow feedback so it doesn't send feedback for song params by adding isClipContext function to check if we're dealing with clip context before sending midi follow feedback
- Removed midi follow feedback sending code from performance view as performance view controls song params
- Refactored getSelectedClip functions into more meaningful names
- Updated getSelectedClip to include clip selection when in arranger automation overview


src/deluge/io/midi/midi_follow.cpp
```diff
@@ -109,10 +109,36 @@ void MidiFollow::initMapping(int32_t mapping[kDisplayWidth][kDisplayHeight]) {
...
+				clip = output->activeClip;
+			}
+		}
+	}
+
+	return clip;
+}
+
+/// check if the current context is the song
+bool isClipContext() {
```
m-m-adams, line 135: not implemented?

seangoodvibes: Whoops, removed

m-m-adams (approved):

PR #1828, opened 2024-05-04
use midi follow cc checking for loopback midi

m-m-adams:
Makes using loopback feel integrated. If a cc is recieved from the loopback midi device on the main input channel for the clip, just use the midi follow mappings to choose the param. This allows using automation view in midi clips to automate synth clips without messing around with midi learn

The changes to midi follow are superficial - minor refactoring to allow explicitly passing in a modelStackWithTimelineCounter to handleReceivedCC that changed the indentation level


m-m-adams:
Unsure on cherry picking - I think this makes midi loopback a lot more usable but also it might have some interactions I missed

seangoodvibes (approved):
Seems to work great! Thanks for doing this, much simpler than I imagined.

Issue #1813, opened 2024-05-01
Mod Button Popup's don't display all mod assignments


seangoodvibes:
### Please describe the problem:

The mod button pop up's that show you what is currently assigned to the gold knob encoders only show you limited info for default knob assignments - e.g. filter, eq, delay, reverb, side chain, compressor

### What is the expected behavior?

The mod button pop up's should show you everything that is assigned to the gold knobs.

so if it's not filter, eq, delay, reverb, side chain, compressor, show the parameter name / patch cable assignment

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1834, opened 2024-05-05
Bug Fix Mod Button Popups

seangoodvibes:
Fixed mod button popup's so that all mod button presses now display the parameters assigned to the mod encoder.

fix https://github.com/SynthstromAudible/DelugeFirmware/issues/1813

### DIFF
--- a/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
+++ b/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
@@ -2050,7 +2050,24 @@ char const* ModControllableAudio::getSidechainDisplayName() {
 
 /// displays names of parameters assigned to gold knobs
 void ModControllableAudio::displayOtherModKnobSettings(uint8_t whichModButton, bool on) {
+	// the code below handles displaying parameter names on OLED and 7SEG
+
+	/* logic for OLED:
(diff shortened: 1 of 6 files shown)


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -2047,3 +2047,37 @@ char const* ModControllableAudio::getSidechainDisplayName() {
 		return l10n::get(STRING_FOR_FAST);
 	}
 }
+
+/// displays names of parameters assigned to gold knobs
+void ModControllableAudio::displayOtherModKnobSettings(uint8_t whichModButton, bool on) {
```
m-m-adams, line 2052: Could you add a doc comment explaining what this is doing for oled and non-oled contexts?

m-m-adams (commented):
lgtm, add a comment explaining the popup logic and I'm good to merge

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
seangoodvibes: > Could you add a doc comment explaining what this is doing for oled and non-oled contexts?

Done!

m-m-adams (approved):

PR #1863, opened 2024-05-08
dx7: support loading single-patch dx7 files

bfredl:
TODO:

- [ ] set SYNTH preset name to patch name from dx7 file if present


src/deluge/gui/menu_item/dx/cartridge.cpp
```diff
@@ -54,14 +56,14 @@ static bool openFile(const char* path, DX7Cartridge* data) {
 		goto close;
 	}
 	result = f_read(&file, buffer, fileSize, &numBytesRead);
-	if (numBytesRead < xx) {
+	if (numBytesRead < minSize) {
 		display->displayPopup("read error 2", 3);
```
sapphire-arches, line 60:
```suggestion
		display->displayPopup("file too small", 3);
```

or ideally this (and the other errors) should be localized strings.

bfredl:
updated with localization.

PR #1866, opened 2024-05-08
[Draft] Add getSquareTypeWithoutAction (prep for velocity editing view)

seangoodvibes:
Added new function to get info about note square type without actioning any changes to that note square

This is prep for new velocity editing automation view which which will render velocity colours based on type of note square


m-m-adams (commented):
This looks fine but I question it's actual utility - it's probably much more efficient to search the whole row at once and return a list of squares with something in them, rather than searching the row individually for each square

PR #1868, opened 2024-05-08
Refactor drawDrumName (prep for velocity editing view)

seangoodvibes:
Refactored drawDrumName function by separating code for getting drum name out into a new function called getDrumName

Prep for new velocity editing view which will render Drum name for kit note row selection on the display

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -3844,25 +3844,25 @@ void InstrumentClipView::drawNoteCode(uint8_t yDisplay) {
 }
 
 void InstrumentClipView::drawDrumName(Drum* drum, bool justPopUp) {
-	char drumName[50];
+	DEF_STACK_STRING_BUF(drumName, 50);
 
 	getDrumName(drum, drumName);
 
 	if (display->haveOLED()) {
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3844,26 +3844,51 @@ void InstrumentClipView::drawNoteCode(uint8_t yDisplay) {
...
+		if (drum->type == DrumType::MIDI) {
+			indicator_leds::blinkLed(IndicatorLED::MIDI, 1, 1);
+		}
+		else if (drum->type == DrumType::GATE) {
+			indicator_leds::blinkLed(IndicatorLED::CV, 1, 1);
+		}
+	}
+}
 
+void InstrumentClipView::getDrumName(Drum* drum, char* drumName) {
```
sapphire-arches, line 3880: This might benefit from taking a `StringBuf&` rather than a `char*`, since we could use the checked functions from that class.

seangoodvibes: > This might benefit from taking a `StringBuf&` rather than a `char*`, since we could use the checked functions from that class.

Changed it to StringBuf, can you let me know if that's what you meant? thx

PR #1879, opened 2024-05-09
Scheduler/runonce tasks

m-m-adams:
Adds an addOnceTask method to make a task that runs once after a specified delay

Fixes task removal to avoid a segfault when the number of tasks hits 0

Adds some more tests to cover additional functionality

### DIFF
--- a/src/OSLikeStuff/task_scheduler.cpp
+++ b/src/OSLikeStuff/task_scheduler.cpp
@@ -22,6 +22,7 @@
 #include <iostream>
 // currently 14 are in use
 constexpr int kMaxTasks = 20;
+constexpr double rollTime = ((double)(UINT32_MAX) / DELUGE_CLOCKS_PERf);
 struct Task {
 	TaskHandle handle;
 	uint8_t priority;
@@ -30,6 +31,7 @@ struct Task {
 	double minTimeBetweenCalls;
 	double targetTimeBetweenCalls;
 	double maxTimeBetweenCalls;
+	bool removeAfterUse;
 };
 
 struct SortedTask {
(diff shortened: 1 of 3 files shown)


src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -100,34 +103,48 @@ TaskID TaskManager::chooseBestTask(double deadline) {
...
+			list[id].averageDuration = runtime;
+		}
+		else {
+			list[id].averageDuration = (list[id].averageDuration + runtime) / 2;
+		}
 	}
 	else {
-		list[id].averageDuration = (list[id].averageDuration + runtime) / 2;
+		removeTask(id);
 	}
```
sapphire-arches, line 148: Switch if/else blocks rather than using `!`

src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -100,34 +103,48 @@ TaskID TaskManager::chooseBestTask(double deadline) {
...
 
 TaskID TaskManager::addRepeatingTask(TaskHandle task, uint8_t priority, double minTimeBetweenCalls,
                                      double targetTimeBetweenCalls, double maxTimeBetweenCalls) {
-	list[index] = (Task{task, priority, 0, 0, minTimeBetweenCalls, targetTimeBetweenCalls, maxTimeBetweenCalls});
+	list[index] = (Task{task, priority, 0, 0, minTimeBetweenCalls, targetTimeBetweenCalls, maxTimeBetweenCalls, false});
+	createSortedList();
+	return index++;
+}
+
+TaskID TaskManager::addOnceTask(TaskHandle task, uint8_t priority, double timeToWait) {
```
sapphire-arches, line 117: do we want to add some sanity checks of re-adding an already added once task?

m-m-adams: I think it might be useful to schedule the same task to run once at a couple different times

sapphire-arches (approved):

PR #1892, opened 2024-05-11
Os stuff/interrupt disabling

m-m-adams:
Fix interrupt disabling by adding memory and instruction blocks - cpsid isn't enough on it's own as memory operations or fetched instructions can be out of order

Fix reset handler enterring user mode - we don't use it, and it's not possible to go from usr mode back to sys mode so it just messes up the call stack in debuggers. Sys mode can only be entered from an ISR, in future this could be implemented by a syscall interrupt


sapphire-arches:
should this be cherry-picked?

m-m-adams:
Seems a little sketchy, it only really matters where we use them for atomics (just the chain load right now). In the audio routine it's just for performance and the actual improvement is under 1%

Issue #1864, opened 2024-05-08
rotating through different scales is destructive


nikodemus:
### Please describe the problem:

With scales containing different numbers of notes in the rotation, going through the whole cycle is destructive.

Suggested solution:

1. Require activating scales with other than 7 notes from the community menu.
2. If there are non-root notes in the current _song_, skip remove scales with different number of notes than the current one from the rotation. Ie. if I have picked a pentatonic scale and started a song, the scale button should rotate only between pentatonic scales, without showing the diatonic ones. (Could also define a more permissive mapping, but that would be more work.)
3. Add a Scale-menu, accessible with eg. SCALE + SELECT, from which any scale can be selected, even those with different number of notes. If there are non-root notes in the current song, and the number of notes is different in the new scale, ask for confirmation.

(As a bonus, the scale menu would make it easy to eg. switch back and forth between major and minor without having to rotate through several scales every time.)

### What is the expected behavior?

Going to a different scale and coming back to the original should restore the notes.

### Is there a relevant Pull request?

no

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1.0-beta-95acd60

### If possible provide the steps to reproduce the issue and upload additional media:

1. Start with any diatonic scale, and a 7th chord. Say MAJOR 1, 3, 5, 7.
2. Press SHIFT + SCALE to cycle through different scales, until you're back in MAJOR.
3. Observe that your chord is now 1, 3, 4, 5 instead.

soymonitus:
I would vote for a variation of option 3, which is:
- Override the shortcut Shift + Scale to show a popup like the one in FILL/ONCE clips with the list of scales (instead of just cycling), confirm selection with Select click. If the target scale has, for example, 5 notes but your sequence contains more than 5 rows with notes, it will print CANT, being consistent with the Keyboard Sidebar scale view.

soymonitus:
I will try to solve this by making the deluge  "remember" the notes placement in case you don't edit the sequence and just cycle all scales back to same scale

nikodemus:
So, I took a look at making the low effort "root notes only" patch for my own use if nothing else, but it doesn't really work:

- New song
- Pentatonic minor
- Write some notes
- Cycle through scales... oops, you can't return to pentatonic anymore, using the "only root notes" logic.

[...]

PR #1895, opened 2024-05-11
Make cycling from 5-note scale to 7-note scale non destructive (remember note positions)

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/1864

To cherry pick for 1.1

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -31,6 +31,7 @@
 #include "hid/led/indicator_leds.h"
 #include "hid/led/pad_leds.h"
 #include "hid/matrix/matrix_driver.h"
+#include "io/debug/log.h"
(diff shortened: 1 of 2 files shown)


soymonitus:
This is ready for review

nikodemus:
This makes from Major C with notes 1,3,5,7 works, excellent! (Ie. when I get back to major I have the same notes as I originally had.)

soymonitus:
> This makes from Major C with notes 1,3,5,7 works, excellent! (Ie. when I get back to major I have the same notes as I originally had.)

You can also try the "yellow mode" of keyboard side bars, to change scales.you can hold a yellow pad and turn updown encoder to set a different scale for that pad. Then you can make direct jumps from any scale to any scale. This fix also works in that scenario

PR #1898, opened 2024-05-12
Automation View ~ Refactor and update shortcut for pad selection mode

seangoodvibes:
Refactored and updated shortcut for pad selection mode

Shortcut is now `SHIFT` + `WAVEFORM` (the very top left pad in first column of grid).

more prep for velocity view


src/deluge/gui/views/automation_view.h
```diff
@@ -296,6 +298,8 @@ class AutomationView final : public ClipView, public InstrumentClipMinder {
 	uint8_t interpolationShortcutY;
 
 	bool padSelectionOn;
+	uint8_t padSelectionShortcutX;
+	uint8_t padSelectionShortcutY;
```
sapphire-arches, line 302: do these ever need to change at runtime? If not, they should be constants.

PR #1901, opened 2024-05-12
Fix automation pasting

### LINKED ISSUE #1209: [Automation View] Source automation Not Preserved When Copy Pasting Automation

I’ve found an issue when copy/paste automation. Something is happening to the page that you copy from after you paste the automation to the next page. The first automation point disappears. Also, the whole first page seems to not respond to the automation when first played. Also happens when copy pasting automation between different parameters e.g. master level to filter cutoff.

Note: On c1.0.1, the first automation point disappears, but once you redraw this the pattern/automation plays back fine.

Original data is preserved when copy pasting automation

_No response_

OLED

Nightly

1.1.0-NIGHTLY-9D1CBCB

1 - Empty synth clip
2- Double All, Double All to create 4 bar pattern
3- Add a long sustained note on each bar
4 - Press Clip to enter Automation View
5 - Select Master Level
6 - Press and <> to zoom to 16th notes
7 - Create automation on page 1 like below, where X is 100% and _ is 0%

X_X_XXX_X_X_XXX

8 - Hold Learn/Copy and press Gold Encoder to copy the automation
9 - Use <> to move to second page
10 - Shift+Learn/Copy and push Gold encoder to paste the automation
11- Now use <> to go back to page one. The first part of the automation is missing. It’s set at 0%
12 - Now paste automation into pages 3 and 4
13 - Reenter automation on first step of page 1. Set to 100%.
14 - Press Play. The first page of the automation is not heard. It seems to be set at 0% throughout

sapphire-arches:
Fixes #1209.

This turned out to not actually be memory corruption, just logic wonky enough to look like it was. The new behavior should feel much more consistent, as it does its best to build a solid fence around the region being pasted in to and prevent the new start/end values from spilling in to adjacent regions.

The PR contains some bonus refactoring and doc additions.


src/deluge/gui/positionable.h
```diff
@@ -21,7 +21,14 @@
...
+	Positionable() = default;
+	~Positionable() = default;
+
+	Positionable(Positionable const& other) = default;
+	Positionable(Positionable&& other) = default;
+
+	Positionable& operator=(Positionable const& other) = default;
+	Positionable& operator=(Positionable&& other) = default;
 
 	int32_t pos;
```
m-m-adams, line 33: Should this be initialized?

src/deluge/modulation/automation/auto_param.cpp
```diff
@@ -2058,49 +2058,78 @@ void AutoParam::paste(int32_t startPos, int32_t endPos, float scaleFactor, Model
...
+			int32_t resetPos = (ticksBeforeStart > startPos) ? (effectiveLength + startPos) - ticksBeforeStart
+			                                                 : (startPos - ticksBeforeStart);
+			int32_t resetI = nodes.search(resetPos, GREATER_OR_EQUAL);
+			ParamNode* resetNode = nodes.getElement(resetI);
+
+			if (!resetNode || resetNode->pos != resetPos) {
+				nodes.insertAtIndex(resetI);
+				resetNode = nodes.getElement(resetI);
+				resetNode->pos = resetPos;
+				resetNode->interpolated = resetNode != nullptr ? resetNode->interpolated : true;
```
m-m-adams, line 2099: If reset node is a null pointer should we be writing anything to it?

src/deluge/modulation/automation/auto_param.cpp
```diff
@@ -2058,49 +2058,78 @@ void AutoParam::paste(int32_t startPos, int32_t endPos, float scaleFactor, Model
...
+			}
+			else {
+				resetPos = wrappedEndPos;
+			}
+
+			int32_t resetI = nodes.search(resetPos, GREATER_OR_EQUAL);
+			ParamNode* resetNode = nodes.getElement(resetI);
+
+			if (!resetNode || resetNode->pos != resetPos) {
+				nodes.insertAtIndex(resetI);
```
m-m-adams, line 2123: Should these errors be checked?

src/deluge/gui/positionable.h
sapphire-arches: probably, but given the way `Positionable` is used the c++ constructor basically never run. I'll add it anyway.

m-m-adams (approved):

PR #1902, opened 2024-05-12
refactor clock status checks

nikodemus:
* Use `PlaybackHandler::isEitherClockActive()` consistently.
* Add corresponding `isInternalClockActive()` and `isExternalClockActive()`.


m-m-adams:
Can we now make the whichever_clock_active global variables not global too?

nikodemus:
> Can we now make the whichever_clock_active global variables not global too?

You mean `PLAYBACK_CLOCK_EXTERNAL_ACTIVE` &co? They're `#define"'d bitmask constants, but sure, I could eg. make them constants in `PlaybackHandler`?

m-m-adams:
Oh whoops I meant playbackState, but yeah moving the defines would be good too

m-m-adams (approved):

PR #1916, opened 2024-05-13
Bugfix/adding and removing tasks while running

m-m-adams:
Fixes a few more bugs in the scheduler:
1. If more than kMaxTasks scheduled it would overwrite the number of current tasks
2. If a task was remove while running it's vacant space in the task list wasn't reused, so tasks couldn't be added properly

### DIFF
--- a/src/OSLikeStuff/task_scheduler.cpp
+++ b/src/OSLikeStuff/task_scheduler.cpp
@@ -19,6 +19,7 @@
 #include "RZA1/ostm/ostm.h"
 #include "util/container/static_vector.hpp"
 #include <algorithm>
+#include <format>
 #include <iostream>
 // currently 14 are in use
 constexpr int kMaxTasks = 20;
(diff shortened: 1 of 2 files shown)


src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -152,7 +174,7 @@ void TaskManager::runTask(TaskID id) {
 	}
 }
 void TaskManager::clockRolledOver() {
-	for (int i = 0; i < index; i++) {
+	for (int i = 0; i < numActiveTasks; i++) {
```
sapphire-arches, line 177: shouldn't this also loop over all tasks, since if I understand correctly `list` can have holes?

src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -46,7 +45,7 @@ struct TaskManager {
 
 	std::array<Task, kMaxTasks> list{nullptr};
 	std::array<SortedTask, kMaxTasks> sortedList;
```
sapphire-arches, line 47: it would probably be a good idea to document what state the various entries in these lists can be in, when, and why.

src/OSLikeStuff/task_scheduler.cpp
m-m-adams: Yeah it should

PR #1921, opened 2024-05-13
tests: initial set of LFO tests

nikodemus:
* Some refactoring necessitated:

  * Split basic waveform generation from functions.h to waves.h. This gets rid of redundant dependencies of lfo.h, allowing it to be used in unit tests. There's plenty more that could follow, this is the minimal set to get going.

  * remove function.h inclusion from lookuptables.h: not needed.

  * getLFOInitialPhaseForZero/NegativeExtreme functions move to lfo.cpp.

* Tests: grab a couple of samples from every type of LFO, followed both global sync and local noteOn logic in phase resetting. ...the differences are interesting, but are they intentional?

### DIFF
--- a/tests/unit/lfo_tests.cpp
+++ b/tests/unit/lfo_tests.cpp
@@ -14,9 +14,9 @@ TEST(LFOTest, renderGlobalTriangle) {
 	LFOType type = LFOType::TRIANGLE;
 	// as per resyncGlobalLFO()
 	lfo.phase = getLFOInitialPhaseForZero(type);
-	CHECK_EQUAL(0, lfo.render(0, type, 0));
-	lfo.tick(10, 100);
-	CHECK_EQUAL(2000, lfo.render(0, type, 0));
+	int32_t numSamples = 10, phaseIncrement = 100;
+	CHECK_EQUAL(0, lfo.render(10, type, 100));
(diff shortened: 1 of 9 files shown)


tests/unit/lfo_tests.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+TEST(LFOTest, renderGlobalSine) {
+	LFO lfo;
+	LFOType type = LFOType::SINE;
+	// as per resyncGlobalLFO()
+	lfo.phase = getLFOInitialPhaseForZero(type);
+	CHECK_EQUAL(0, lfo.render(0, type, 0));
+
+	lfo.tick(10, 100);
+	CHECK_EQUAL(2412, lfo.render(0, type, 0));
```
m-m-adams, line 41: The "correct" result is (2^31)*sin(2*pi*1000/(2^32)) = 3141

i.e. convert the int phasor of 1000 to radians by multiplying by 2*pi/2^32
Then multiply by 2^31 to put it into a q31 fixed point integer

Maybe we should test for a total deviation from a proper sine? Otherwise check against a phase of 1024 which will yield the correct value, this one is just off because it's linear interpolation from a 256 entry table

nikodemus: I'll leave exhaustive testing for later... :)

PR #1925, opened 2024-05-13
TaskScheduler/FillBusyWaitTime

m-m-adams:
When the task manager has to wait to run a high priority task, find something else to go and do quickly


src/deluge/deluge.cpp
```diff
@@ -936,7 +940,7 @@ extern "C" void routineForSD(void) {
 
 	AudioEngine::logAction("from routineForSD()");
 	AudioEngine::routine();
-	playbackHandler.routine();
+	// playbackHandler.routine();
```
sapphire-arches, line 943: is this intended?

m-m-adams: nope

PR #1932, opened 2024-05-14
Load songs asynchronously

m-m-adams:
Load songs asynchronously to allow removing the playback handler call from the load function

Splits song loading into 4 pieces, each of which have a callback to continue when ready rather than busy waiting

### DIFF
--- a/src/deluge/gui/ui/load/load_song_ui.cpp
+++ b/src/deluge/gui/ui/load/load_song_ui.cpp
@@ -227,7 +227,7 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
 	return ActionResult::DEALT_WITH;
 }
 
-enum class LoadStatus { start, loadingSamples, waitingToEnd, newSongPlaying, finished };
+enum class LoadStatus { START, BEGIN_LOADING_SAMPLES, WAIT_FOR_SAMPLES, NEW_SONG_PLAYING, COMPLETE };
 StorageManager* bdsm{};
 // Before calling this, you must set loadButtonReleased.
 void LoadSongUI::performLoad(StorageManager& _bdsm) {
@@ -237,11 +237,11 @@ void LoadSongUI::performLoad(StorageManager& _bdsm) {
 void LoadSongUI::performLoadFixedSM() {
 	static int32_t count = 0;
 	Song* toDelete = nullptr;
-	static LoadStatus status = LoadStatus::start;
-	if (status == LoadStatus::finished)
-		status = LoadStatus::start;
+	static LoadStatus status = LoadStatus::START;
+	if (status == LoadStatus::COMPLETE)
+		status = LoadStatus::START;
 	switch (status) {
-	case LoadStatus::start: {
+	case LoadStatus::START: {
 		FileItem* currentFileItem = getCurrentFileItem();
 
 		if (!currentFileItem) {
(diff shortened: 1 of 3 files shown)


src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -226,233 +227,258 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
 	return ActionResult::DEALT_WITH;
 }
 
+enum class LoadStatus { start, loadingSamples, waitingToEnd, newSongPlaying, finished };
```
sapphire-arches, line 230:
```suggestion
enum class LoadStatus { START, BEGIN_LOADING_SAMPLES, WAIT_FOR_SAMPLES, NEW_SONG_PLAYING, COMPLETE };
```

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -226,233 +227,258 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+StorageManager* bdsm{};
```
sapphire-arches, line 231: These globals make me feel like we should teach the task manager how to use a `std::function` instead, but it's probably fine for an initial pass

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -226,233 +227,258 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+void LoadSongUI::performLoad(StorageManager& _bdsm) {
+	bdsm = &_bdsm;
+	performLoadFixedSM();
+}
+void LoadSongUI::performLoadFixedSM() {
+	static int32_t count = 0;
+	Song* toDelete = nullptr;
+	static LoadStatus status = LoadStatus::start;
+	if (status == LoadStatus::finished)
+		status = LoadStatus::start;
```
sapphire-arches, line 242:
```suggestion
	if (status == LoadStatus::finished) {
		status = LoadStatus::start;
	}
```

Style nits aside, should this reset happen in `performLoad` instead with a freeze if we try to start a load while another is ongoing?

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -226,233 +227,258 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
 	}
-
-	else {
-		playbackHandler.doSongSwap();
+	case LoadStatus::waitingToEnd: {
+		if (currentUIMode != UI_MODE_LOADING_SONG_NEW_SONG_PLAYING) {
+			addOnceTask([]() { loadSongUI.performLoadFixedSM(); }, 100, 0.05);
+			return;
+		}
+		status = LoadStatus::newSongPlaying;
```
sapphire-arches, line 446: do we need to add a task here? If not there should be a `[[fallthrough]];`

src/deluge/gui/ui/load/load_song_ui.cpp
m-m-adams: std::function is huge to accommodate arbitrary sized captures and the type erasure indirection makes it significantly slower than function pointers. We could explore an alternate implementation with limited capture size like this one - https://www.codeproject.com/Articles/7150/Member-Function-Pointers-and-the-Fastest-Possible

PR #1933, opened 2024-05-14
AudioClip: fix single-row rendering

sapphire-arches:
A *3 got missed so we were scribbling over other rows.


sapphire-arches:
Fixes #1933 1905

seangoodvibes (approved):

Issue #1893, opened 2024-05-11
Transition from Keyboard View to Grid View is not smooth


seangoodvibes:
### Please describe the problem:

Transition to grid view from keyboard view causes main grid and sidebar to flash

### What is the expected behavior?

Transition should be smooth from one view to the next

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### If possible provide the steps to reproduce the issue and upload additional media:

https://github.com/SynthstromAudible/DelugeFirmware/assets/138174805/ce6a69ec-5497-4003-90fa-4aa313db8913

PR #1934, opened 2024-05-15
Bug Fix keyboard screen and other animations

seangoodvibes:
Updated animations for entering / exiting keyboard screen from song row view / back to row view

Updated animations for entering / exiting keyboard screen from song grid view / back to grid view

Split the UI_MODE_EXPLODE_ANIMATION into two: EXPLODE and IMPLODE (UI_MODE_IMPLODE_ANIMATION)

Similar to the check's that were in place in the Clip View rendering functions to not render when INSTRUMENT_CLIP_COLLAPSING was happening (when exiting clip back to Song Row View), I added the same check for the IMPLODE UI mode. This eliminated flicker that I was noticing where it would clear the exiting view's grid and then re-render it again before finally transitioning to song view and rendering that view's grid.

I updated a number of rendering calls as well where I saw that the PadLED::image used was the wrong one (e.g. it should have been image[1] per the Official Firmware source code.

Did a lot of comparing with Official Firmware source code to update things that got changed by mistake.

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/1893

### DIFF
--- a/src/deluge/gui/ui/keyboard/keyboard_screen.cpp
+++ b/src/deluge/gui/ui/keyboard/keyboard_screen.cpp
@@ -88,7 +88,7 @@ ActionResult KeyboardScreen::padAction(int32_t x, int32_t y, int32_t velocity) {
 	// Pad pressed down, add to list if not full
 	if (velocity) {
 		// TODO: Logic should be inverted as part of a bigger rewrite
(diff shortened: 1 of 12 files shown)


sapphire-arches (commented):
does it ever matter if we're exploding or imploding? I don't immediately see a case but I'm only looking at the diff here and I'm sleepy. If not, we should just rename "explode" to something like `IN_SESSION_VIEW_TRANSITION`.

m-m-adams (approved):

Issue #855, opened 2023-12-18
Audition and Mute Pads do not get rendered correctly when re-entering synths after saving a song


sapphire-arches:
### Please describe the problem:

Loading the song file from acidaliex on discord (https://discord.com/channels/608916579421257728/1138924035657449482/1186028879807258715) and entering the column 2 or 3 synths (in grid mode), the note row mute pads do not get rendered correctly.

### What is the expected behavior?

Note row mute/audition pads get rendered

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Release 1.0/Amadeus, Nightly

### What is the firmware name:

fe31a4f9b112890b163157969097788eb81314ff

sichtbeton:
Had this but also a couple of times. When entering from grid mode, the sidebar goes blank. When you scroll up and down, it comes back.

m-m-adams:
I don't have a good solution but the problem happens [when the zoom is setup here](https://github.com/SynthstromAudible/DelugeFirmware/blob/3017c21195189b42787a8b8a80962a7d0b44be52/src/deluge/gui/views/timeline_view.cpp#L325) after the [fade](https://github.com/SynthstromAudible/DelugeFirmware/blob/3017c21195189b42787a8b8a80962a7d0b44be52/src/deluge/hid/led/pad_leds.cpp#L812)

At this point the sidebar has been rendered to the image store but not the image and gets wiped out. I believe that's intentional here so the sidebar probably just has to render again after the zoom?

sichtbeton:
I get this a lot. In every session. it's not a huge deal but kind of annoying.

PR #1939, opened 2024-05-15
Bugfix sidebar not rendering with explode animation

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/855

Looking at the official code, zooming was not something that was done when animating an explosion from arranger (and now grid view).

When the zooming code was added, it created a conflict where the clip got re-rendered at new zoom level while the explosion animation was finishing.

This created a situation where the sidebar got erased.

To fix this, you need to remove the duplicate actions from taking place (e.g. if you just zoomed in and re-rendered the clip, no need to also fade).

### DIFF
--- a/src/deluge/hid/led/pad_leds.cpp
+++ b/src/deluge/hid/led/pad_leds.cpp
@@ -798,18 +798,19 @@ void timerRoutine() {
 				memcpy(PadLEDs::imageStore, PadLEDs::image,
 				       (kDisplayWidth + kSideBarWidth) * kDisplayHeight * sizeof(RGB));
 
+				bool anyZoomingDone = false;
 				currentUIMode = UI_MODE_ANIMATION_FADE;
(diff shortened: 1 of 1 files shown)


src/deluge/hid/led/pad_leds.cpp
```diff
@@ -837,8 +838,17 @@ void timerRoutine() {
...
+				// if so, pause fade
+				if (!anyZoomingDone) {
+					recordTransitionBegin(130);
+					renderFade(0);
+				}
+				// if you zoomed in and re-rendered the sidebar,
+				// we'll continue the transition and fade after the next refresh
+				else {
+					uiTimerManager.setTimer(TimerName::MATRIX_DRIVER, UI_MS_PER_REFRESH);
+				}
```
sapphire-arches, line 851: should switch cases so the condition doesn't need to be inverted

Issue #1744, opened 2024-04-19
Midi follow feedback doesn't get sent properly


seangoodvibes:
### Please describe the problem:

Midi follow feedback isn't getting sent properly.

When you select a clip in song / arranger view -> it sends feedback (this is good)
When you enter a clip -> it sends feedback for the active clip, which may not be the same clip (this is not good)

### What is the expected behavior?

If you're in a clip, if it's not the active clip, we shouldn't send midi feedback for another clip -> send feedback for the current clip

Also, if playback is not running, should we always use current clip?

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #1945, opened 2024-05-15
Bugfix - Update midi follow feedback

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/1744

Updated midi follow feedback:
- removed a check for IsClipContext() which was preventing updated clip values from being sent while in song / arranger view 
  - e.g. upon release of a selected clip, it should have been sending feedback for the active clip, but it wasn't
- to send feedback for the current clip when you enter a clip view
  - it was sending feedback for the active clip which may not match the current clip you're in
- to send feedback when the active clip changes in song / arranger view
- to send feedback when playback starts 
  - because the feedback values previously received would have been for the last active clip processed which can differ in arranger (and maybe other areas also)
- to send feedback when launching a new clip in song view (which changes the active clip)
- to send feedback when arranger is changing from one clip to the next
  - this also works for sending feedback when you scroll in arranger to another clip and start playback on that clip

Cleaned up some bad comments wrapping I saw as well

### DIFF
--- a/src/deluge/playback/mode/arrangement.cpp
+++ b/src/deluge/playback/mode/arrangement.cpp
@@ -48,7 +48,7 @@ void Arrangement::setupPlayback() {
 
 	currentSong->setParamsInAutomationMode(true);
 
-	// It seems strange that I ever put this here. It's now also in PlaybackHandler::setupPlayback,
+	// Rohan: It seems strange that I ever put this here. It's now also in PlaybackHandler::setupPlayback,
 	// so maybe extra unneeded here - but that's not the only place that calls us, so have left it in for now
 	playbackHandler.swungTicksTilNextEvent = 0;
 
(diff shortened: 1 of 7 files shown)


src/deluge/playback/mode/arrangement.cpp
```diff
@@ -48,9 +48,9 @@ void Arrangement::setupPlayback() {
 
 	currentSong->setParamsInAutomationMode(true);
 
-	playbackHandler.swungTicksTilNextEvent =
-	    0; // It seems strange that I ever put this here. It's now also in PlaybackHandler::setupPlayback,
-	       // so maybe extra unneeded here - but that's not the only place that calls us, so have left it in for now
+	// It seems strange that I ever put this here. It's now also in PlaybackHandler::setupPlayback,
+	// so maybe extra unneeded here - but that's not the only place that calls us, so have left it in for now
```
m-m-adams, line 52: Can you add a note that this is a Rohan comment? It'll be you in git blame now

src/deluge/playback/playback_handler.cpp
```diff
@@ -475,15 +477,15 @@ void PlaybackHandler::endPlayback() {
...
 	bool wasRecordingArrangement = (recording == RecordingMode::ARRANGEMENT);
 
-	bool shouldDoInstantSongSwap =
-	    currentPlaybackMode
-	        ->endPlayback(); // Must happen after currentSong->endInstancesOfActiveClips() is called (ok I can't
-	                         // remember why I wrote that, and now it needs to happen before so that
-	                         // Clip::beingRecordedFrom is still set when playback ends, so notes stop)
+	// Must happen after currentSong->endInstancesOfActiveClips() is called (ok I can't
+	// remember why I wrote that, and now it needs to happen before so that
+	// Clip::beingRecordedFrom is still set when playback ends, so notes stop)
```
m-m-adams, line 482: Here too - basically just in case someone tries to figure out the answer to these questions

m-m-adams (approved):
Lgtm but if you could stick -Rohan on the sketchy comments that would be great

PR #1962, opened 2024-05-19
Ease timeline zoom level restriction

trappar:
Implements the feature I suggested here: https://github.com/SynthstromAudible/DelugeFirmware/discussions/314

## What

The maximum zoom level has been increased. Now, the maximum zoom is the point the point where the entire sequence is represented by a single grid cell.

## Why

Personally I never found this limitation to make sense. Why stop me from zooming out beyond my clip? If zoom only affected visibility, it would make sense... but zoom also affects entered note length. By allowing more freedom to zoom, you can access instant entry of note lengths that were previously not possible.

For example, this makes it super quick to enter basic chord progressions.

Given a brand new project, to create a 4 chord progression where each chord lasts for 1-bar:
1. Zoom all the way out to 1 bar per column (one quick twist)
2. Expand the sequence to 4 bars (3 clicks on the horizontal control)
3. Enter your 4 chords on the 4 active grid columns.

Prevously the fastest way I know to do this would be:
1. Hit shift + horizontal encoder twice (bit fiddly remembering that this equates to creating a 4-bar loop)
2. Create chords via pressing, holding, and pressing the note end (this takes 2x the number of presses as above)

Total inputs to make a 4-chord tri-tone chord progression with both methods:
* Restricted zoom: 27
* Unrestricted zoom: 16

### As a community feature?

I don't think this needs to be a community feature for the following two reasonsL
1. It only eases an existing restriction.
2. It does so in a way that shouldn't annoy users, thanks to the delay I introduced when crossing the previous max zoom level threshold. If you're zoomed in and want to return to the previous max zoom level, you can still hold the horizontal encoder down and quickly turn it to the left - it will catch at the prior max zoom for half a second.

### What even is "max zoom" anymore?

This PR degrades the ability to understand what clip's `getMaxZoom` means. The clip's max zoom isn't actually the timeline's max zoom anymore.

Ideally I would have liked to have updated the actual behavior of `Clip::getMaxZoom()` in `src/deluge/model/clip/clip.cpp`, but that function is used in many more places than the timeline.


src/deluge/gui/views/timeline_view.cpp
```diff
@@ -16,12 +16,14 @@
  */
 
 #include "gui/views/timeline_view.h"
+#include "algorithm"
```
m-m-adams, line 19: Style nit: '#include \<algorithm>' so that it's not confused with a local file

m-m-adams (commented):
Small nit but lgtm. As mentioned on discord I'm good to merge as is (in community features) or to add a "catch" at the old max zoom level and just have the feature always on

src/deluge/gui/views/timeline_view.cpp
```diff
@@ -22,9 +22,14 @@
...
 #include "hid/display/display.h"
 #include "hid/led/pad_leds.h"
+#include "model/settings/runtime_feature_settings.h"
 #include "model/song/song.h"
 #include "processing/engines/audio_engine.h"
 #include "string.h"
+#include <algorithm>
+
+uint32_t timeHorizontalZoomLastHitClipMax;
+int8_t delayedHorizontalScrollDirection;
```
trappar, line 32: Not sure if introducing these variables at this scope is the best way to do this (I'm not really a C/C++ dev).

Any thoughts on refactoring this for idiomaticy would be appreciated.

src/definitions_cxx.hpp
```diff
@@ -178,6 +178,7 @@ constexpr int32_t kScrollTime = 400;
 constexpr int32_t USE_DEFAULT_VELOCITY = 255;
 
 constexpr int32_t kMaxSequenceLength = 1610612736; // The biggest multiple of 3 which can fit in a signed 32-bit int32_t
+constexpr int32_t kMaxZoom = 100663296; // The zoom level in which the maximum sequence length is displayed on all pads
```
sapphire-arches, line 181: can this be calculated rather than additional magic?

src/deluge/gui/views/timeline_view.cpp
sapphire-arches: they should be members of the TimelineView class probably, though that replicates it across all timeline views... as long as it doesn't end up behaving weird I think it should go in the class.

src/deluge/gui/views/timeline_view.cpp
```diff
@@ -146,24 +151,36 @@ ActionResult TimelineView::horizontalEncoderAction(int32_t offset) {
...
+			uint32_t newZoom = zoomMagnitude == -1 ? oldXZoom >> 1 : oldXZoom << 1;
+			uint32_t clipLengthMaxZoom = getMaxZoom();
 
-			// Constrain to zoom limits
-			if (zoomMagnitude == -1) {
-				if (currentSong->xZoom[navSysId] <= 3) {
+			// When this timer is set, we may need to wait a bit before allowing zooming
+			if (timeHorizontalZoomLastHitClipMax != 0) {
+				bool isWithinZoomDelay =
+				    AudioEngine::audioSampleTimer - timeHorizontalZoomLastHitClipMax < kShortPressTime;
```
sapphire-arches, line 161: this should use `infinite_a_lt_b` from utils since it handles wrapping, otherwise we'll get stuck if there's a particularly unfortunate wrap.

src/definitions_cxx.hpp
trappar: That's how I had it before (`kMaxSequenceLength / 16`) but I thought maybe it was a bad idea to calculate this on every zoom. I don't really now what is and isn't a big deal when it comes to performance...

src/deluge/gui/views/timeline_view.cpp
trappar: Like this?

```suggestion
				    util::infinite_a_lt_b(AudioEngine::audioSampleTimer - timeHorizontalZoomLastHitClipMax, kShortPressTime)
```

I'm not sure I really understand the issue. Also, the method of doing this without `infinite_a_lt_b` is definitely used in a bunch of the existing code. I pretty much ripped this way of measuring a delay from a number of other examples that do the same thing.

src/definitions_cxx.hpp
sapphire-arches: `constexpr` expressions are calculated at compile time, so that should be no issue (well, technically the compiler is allowed to not but this is trivial constant propagation so it will in practice)

src/deluge/gui/views/timeline_view.cpp
sapphire-arches: yeah that function didn't exist until a couple months ago so a bunch of existing code doesn't use it. that code is all buggy it just doesn't matter often in practice since the wrap interval on the sample timer is so long.

src/definitions_cxx.hpp
trappar: Ohh... I get what you're saying. I interpreted the feedback wrong initially.

src/deluge/gui/views/timeline_view.cpp
```diff
@@ -146,24 +148,36 @@ ActionResult TimelineView::horizontalEncoderAction(int32_t offset) {
...
 			}
-			else {
-				if (currentSong->xZoom[navSysId] >= getMaxZoom()) {
-					goto getOut;
-				}
-				currentSong->xZoom[navSysId] <<= 1;
+
+			// Prevent quickly scrolling past the clip's max zoom level
+			if (newZoom == clipLengthMaxZoom && timeHorizontalZoomLastHitClipMax == 0) {
+				timeHorizontalZoomLastHitClipMax = AudioEngine::audioSampleTimer;
```
sapphire-arches, line 170: I think the `kShortPressTime` should be added here and then ClipMax used as a deadline, so we can compare it directly to the audio sample timer above rather than redoing math.

src/deluge/gui/views/timeline_view.h
```diff
@@ -70,4 +70,8 @@ class TimelineView : public RootUI {
 
 	// ui
 	UIType getUIType() { return UIType::TIMELINE_VIEW; }
+
+private:
+	uint32_t timeHorizontalZoomLastHitClipMax = 0;
+	int8_t delayedHorizontalScrollDirection;
```
sapphire-arches, line 76: this can only be -1 or +1 right? would be good to document that (and the meaning of the variable above as well)

sapphire-arches (commented):
2 minor nitpicks and this should get an entry in CHANGELOG.md, then lgtm

src/deluge/gui/views/timeline_view.h
trappar: How's what I adeed?

src/deluge/gui/views/timeline_view.cpp
trappar: Updated this along with the variable naming to be more clear (I think). Definitely open to further refinements though.

src/deluge/gui/views/timeline_view.h
```diff
@@ -70,4 +70,11 @@ class TimelineView : public RootUI {
 
 	// ui
 	UIType getUIType() { return UIType::TIMELINE_VIEW; }
+
+private:
+	// Used when scrolling horizontally to briefly catch on clip's max zoom
+	uint32_t delayHorizontalZoomUntil = 0;
+	// Horizontal scroll is only delayed in the direction that clip's max zoom was crossed in.
+	// This is the direction (-1 or +1).
+	int8_t delayHorizontalZoomMagnitude;
```
sapphire-arches, line 79:
```suggestion
	/// Used when scrolling horizontally to briefly catch on clip's max zoom
	uint32_t delayHorizontalZoomUntil = 0;
	/// Horizontal scroll is only delayed in the direction that clip's max zoom was crossed in.
	/// This is the direction (-1 or +1).
	int8_t delayHorizontalZoomMagnitude;
```

src/deluge/gui/views/timeline_view.cpp
sapphire-arches: looks good, thanks!

src/deluge/gui/views/timeline_view.h
sapphire-arches: left a suggestion so doxygen will pick it up, but otherwise looks fine. thanks!

YPares:
This feature doesn't seem present in the 1.1 firmware. Is it gated behind a setting?

trappar:
> This feature doesn't seem present in the 1.1 firmware. Is it gated behind a setting?

No, this will be in 1.2

PR #1977, opened 2024-05-20
toolchain: bump to v16

sapphire-arches:
Should fix the rtmidi problems


trappar (approved):
Tested on a Windows 11 machine and I was able to run the `loadfw` command which was previously failing!

Issue #813, opened 2023-12-12
LFO2 has a sync pad that's not (yet?) implemented


m-m-adams:
### Please describe the problem:

LFO1 has both rate and sync

LFO2 only has rate

LFO2 should also have a rate, to allow syncing to 1/4 or 1/8 notes while still being voice specific and reset by note-ons

### What is the expected behavior?

LFO2 should also have sync

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Release 1.0/Amadeus

nikodemus:
I'm still missing a lot of details, but I have a rough idea how to get this done nicely.

1. Move all LFO parameters to LFOConfig object or similar. Sound has then globalLFO, globalLFOConfig, and localLFOConfig. Voice has lfo (aka local LFO) just as before.

2. Put a pointer to the LFOConfig to the LFO itself: for sound just link globalLFO to globalLFOConfig when Sound() is initialized, for voice link/unlink to the correct sound when the voice is assigned/unassigned.

3. Move timeStartedSkippingRenderingLFO to the LFO object.

4. At this point resyncGlobalLFO() can become LFO::resync(), and setLFOGlobalSyncType/Level/WaveType can become LFO::setSyncType/Level/Wavetype. These methods will then be used to modify local LFO as well.

5. Add LFO2 sync stuff to the UI, hook it up just like for LFO1 pretty much.

---

So. Does this sound about right, or did someone envision a nicer way to do this?

I'm not super happy about linking LFOConfig as described above, but can't really see a way to avoid it while making code cleaner: otherwise we just end up with more semi-duplicated parameters in more places.

sapphire-arches:
for 2, you should be able to walk the existing `assignedToSound` pointer to find the LFOConfig for the local LFO. Otherwise, seems good to me!

I'm not sure I understand what `timeStartedSkippingRenderingLFO` is used for, but it does sound like it should live on the LFO.

nikodemus:
Re. 2, pointer walk means global and local LFO would work differently: I'm trying to organize things so that they could share the logic. 

However, since LFO2 retriggers, it's sync is going to be different anyhow... I'll need to think a bit more.

PR #1978, opened 2024-05-20
LFO2 sync draft

nikodemus:
As per #813 

TODO:

- [X] Figure out if LFO2 needs a resync like LFO1: right now the assumption is that since it retriggers with note-on, it's "good enough" without a resync. No idea if that's true. -->  *Decided that it's good enough for now at least: a drone with sync'ed LFO2 might experience some drift over time, though.*
- [x] Figure out if LFO1 and LFO2 should set initial phase the same way. Currently LFO1 has more logic about this. Applying that would change LFO2 behaviour, but maybe doing it when sync'ed would be ok? --> Retain old behaviour for unsynced LFO2, match LFO1 behaviour when synced.
- [x] Cleanup: do we need resyncing setters for LFO2? (Do we actually need them for LFO1, could we just wait for the next natural resync?) --> *Don't need them for LFO2, but should figure out a nicer way to handle them for LFO1.**
- [x] Implement initial phase behaviour matching LFO1 when synced.
- [x] Documentation.
- [x] Cleanup: we should be able to merge lfo::local::* and lfo::global::* classes.
- [x] Cleanup: nicer way to handle resyncing setters for LFO1. Do we need them, or could we wait for the next clock-driven resync instead? --> *Clock based resync is still an option, but now there are less redundant calls.*
- [x] Naming: localLFOConfig.waveType is a bit unwieldy. lfo1Config.waveType, etc, maybe? --> *`lfoConfig[]` array, as part of the Rate/Sync/Type menu class unification.*
- [x] Cleanup: should be able to merge logic for phase increment computation.
- [x] ~~Cleanup / optimization: cache phase increment, instead of recomputing it every time.~~ --> *Nope, decided this is out of scope: there's more than a couple of cases when the cache would need to be invalidated, so if this is done it's better done separately.* 
- [x] Missing sync menu for LFO2.
- [ ] More testing.


nikodemus:
#2004 supersedes this

PR #1997, opened 2024-05-22
Fix collect and save

### LINKED ISSUE #1995: Save and collect writes corrupt sample data to the SD card

Since 653b073a (bisected), running save and collect saves corrupt sample data in the created folder. The data is either garbage or random bits of XML of the song data.

Save and collect writes proper samples to the folder.

_No response_

7SEG

Nightly

1.2.0-dev-653b073a (tested up to 1.2.0-dev-2e09148d)

1. Turned on, loaded project
2. Saved with collect
3. Mounted the SD card on a computer, checked that the WAV files in the SONGS/x folder were corrupt

Azenet:
Naive fix that works the same way it used to work before the storage manager refactoring.

Fixes #1995

### DIFF
--- a/src/deluge/gui/ui/save/save_song_ui.cpp
+++ b/src/deluge/gui/ui/save/save_song_ui.cpp
@@ -364,7 +364,7 @@ fail3:
 							break; // Stop, on rare case where file ended right at end of last cluster
 						}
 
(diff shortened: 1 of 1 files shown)


m-m-adams:
I think this makes sense - the serializer isn't actually involved, it's just using its buffer as a copy buffer for the existing audio files on the card

m-m-adams (approved):

PR #2004, opened 2024-05-23
LFO: cleanups in preparation for LFO2 sync support

nikodemus:
- LFO parameters in Song have been folded into LFOConfigs[LFO_COUNT], allowing more uniform access.

- This in turn allows merging local and global LFO Type menu classes. LFO2 also switched to use the lfo::Rate class for consistency, in preparation for adding sync support.

- Initial phase setting logic has been moved to LFO from both Voice and Sound.

- Fix a latent bug where LFO sync type didn't get a default set when loading a song which didn't have it saved.

- LFO submenu variables renamed from lfo0Menu & lfo1Menu to lfo1Menu & lfo2Menu.

- Resyncing setters for LFO1 parameters have been removed, instead locations that make changes to them resyncs explicitly after all changes have been made, and isEitherClockRunning() check has been moved inside the resync function as a guard clause.

- resyncLFOsAndArpeggiators has been renamed to resyncLFOs, since it doesn't touch arpeggiators.

- A few instances of using SYNC_LEVEL_NONE instead of magic number 0.


src/deluge/model/song/song.cpp
```diff
@@ -3401,7 +3401,7 @@ void Song::addOutput(Output* output, bool atStart) {
 	}
 
 	// Must resync LFOs - these (if synced) will roll even when no activeClip
-	if (playbackHandler.isEitherClockActive() && this == currentSong) {
+	if (this == currentSong) {
```
m-m-adams, line 3404: is this ok? I thought they need a clock to know the tempo to sync to

src/deluge/modulation/lfo.cpp
```diff
@@ -22,3 +24,24 @@ uint32_t getLFOInitialPhaseForZero(LFOType waveType) {
...
+
+void LFO::setInitialPhase(const LFOConfig& config) {
+	// Historically LFO1 didn't do much initial phase setting unless it was sync'ed.
+	// LFO2 on the other hand didn't support sync, and always went for the negative
+	// extreme regardless of wave type.
+	//
+	// This logic matches the historical behaviour for synced LFO1 and unsynced LFO2,
+	// and will make synced LFO2 work the same as synced LFO1.
+	//
+	// TODO: Despite losing historical behaviour, consistent initial phase behaviour
```
m-m-adams, line 36: starting from negative extreme in the saw makes it more usable for lots of note modulation since it guarantees that the modulation will be active at note on. It'll be especially true if we add one shot LFOs

If we remove it by default we should make it configurable in an LFO phase menu or something

src/deluge/model/song/song.cpp
nikodemus: The guard is now in resyncGlobalLFO(), so no need to check here.

src/deluge/modulation/lfo.cpp
nikodemus: I was a little unclear.

Currently:
- LFO1 unsynced: ???
- LFO1 synced: from zero for sine and triangle, from negative extreme for others.
- LFO2 always: from negative extreme.

With this patch:
- LFO2 unsynced: from negative extreme.
- LFO1 synced: from zero for sine and triangle, from negative extreme for others.

My comment refers to using the sync logic for unsynced LFOs as well: starting from zero for sine and triangle, and keeping saw and others starting from negative.

If starting everything from negative for LFO2 makes it more usable, that's even simpler. It's the "do one thing for synced and another for unsynced" that seems a bit messy: I just wanted to match LFO1 behaviour for sync, assuming there was a good reason for it, without changing historical unsynced behaviour for LFO2.

m-m-adams: I'd lean towards always starting from negative so that they're all consistent, then changing saw to sine doesn't change the start of the note

nikodemus: Keep global initial phases as they are, or go to negative there as well?

m-m-adams: Probably leave them? I'm sure somebody is relying on the current behaviour

src/definitions_cxx.hpp
```diff
@@ -365,6 +365,15 @@ constexpr OscType kLastRingmoddableOscType = OscType::WAVETABLE;
 constexpr int32_t kNumOscTypesRingModdable = util::to_underlying(kLastRingmoddableOscType) + 1;
 constexpr int32_t kNumOscTypes = util::to_underlying(OscType::INPUT_STEREO) + 1;
 
+// LFO_ID is used exlusively is as an array index, so an enum class would
+// only add extra noise to get the underlying value in all places where this
+// is used.
+enum LFO_ID {
+	LFO1_ID = 0,
+	LFO2_ID = 1,
```
sapphire-arches, line 373: This might be a good place to document the global vs local nature of lfo 1 vs 2

src/deluge/gui/menu_item/lfo/rate.h
```diff
@@ -18,14 +18,18 @@
...
 class Rate final : public patched_param::Integer {
 public:
-	using Integer::Integer;
+	Rate(uint8_t lfoId, deluge::l10n::String name, deluge::l10n::String type, int32_t newP = 0)
+	    : Integer(name, type, newP), lfoID(lfoId) {}
 
 	bool isRelevant(ModControllableAudio* modControllable, int32_t whichThing) override {
 		Sound* sound = static_cast<Sound*>(modControllable);
-		return (sound->lfoGlobalSyncLevel == 0);
+		return (sound->lfoConfig[lfoID].syncLevel == SYNC_LEVEL_NONE);
```
sapphire-arches, line 29:
```suggestion
		return sound->lfoConfig[lfoID].syncLevel == SYNC_LEVEL_NONE;
```

src/deluge/gui/menu_item/lfo/rate.h
```diff
@@ -18,14 +18,18 @@
...
 	}
+
+private:
+	uint8_t lfoID;
```
sapphire-arches, line 33: For new code we use a `_` suffix for private variables, which will sort out the weirdness in the constructor variable names a bit.

src/deluge/gui/menu_item/lfo/sync.h
```diff
@@ -20,21 +20,28 @@
...
+		soundEditor.currentSound->lfoConfig[lfoID].syncLevel = menuOptionToSyncLevel(this->getValue());
+		// This fires unnecessarily for LFO2 assignments as well, but that's ok. It's not
+		// entirely clear if we really need this for the LFO1, even: maybe the clock-driven resyncs
+		// would be enough?
+		soundEditor.currentSound->resyncGlobalLFO();
 		soundEditor.currentSound->setupPatchingForAllParamManagers(currentSong);
 	}
+
+private:
+	uint8_t lfoID;
```
sapphire-arches, line 44: (same note as above for `_` suffix)

src/deluge/gui/menu_item/lfo/sync.h
```diff
@@ -20,21 +20,28 @@
 #include "model/song/song.h"
 #include "processing/sound/sound.h"
 
-namespace deluge::gui::menu_item::lfo::global {
+namespace deluge::gui::menu_item::lfo {
 
 class Sync final : public SyncLevel {
 public:
-	using SyncLevel::SyncLevel;
+	Sync(uint8_t lfoId, deluge::l10n::String name, deluge::l10n::String type) : SyncLevel(name, type), lfoID(lfoId) {}
```
sapphire-arches, line 27: Other note on the constructors: we usually keep the `l10n::String` parameters first.

nikodemus: Gotcha. I figured putting the shared ID in the beginning was easier to read:
```
lfo::Type lfo1TypeMenu{LFO1_ID, STRING_FOR_SHAPE, STRING_FOR_LFO1_TYPE};
lfo::Rate lfo1RateMenu{LFO1_ID, STRING_FOR_RATE, STRING_FOR_LFO1_RATE, params::GLOBAL_LFO_FREQ};
lfo::Sync lfo1SyncMenu{LFO1_ID, STRING_FOR_SYNC, STRING_FOR_LFO1_SYNC};
```
vs
```
lfo::Type lfo1TypeMenu{STRING_FOR_SHAPE, STRING_FOR_LFO1_TYPE,, LFO1_ID};
lfo::Rate lfo1RateMenu{STRING_FOR_RATE, STRING_FOR_LFO1_RATE, params::GLOBAL_LFO_FREQ, LFO1_ID};
lfo::Sync lfo1SyncMenu{STRING_FOR_SYNC, STRING_FOR_LFO1_SYNC, LFO1_ID};
```
Should I change to the latter?

sapphire-arches: I think so, yeah. That's also easier to handle when I eventually convert all this stuff to the new menu compiler logic...

PR #2006, opened 2024-05-23
Prep release 1.1 branch for release candidate 

m-m-adams:
- Disable the freeze on dangling cluster pointers in non-beta builds, it's a small memory leak and should get cleared at song change now anyway
- Turn off alpha/beta version
- Set build type to release
- pull out midi loopback due to the weird playhead issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   via affect entire in song/arranger mode, or by entering the `SONG MENU` by pressing `SELECT` in Song View and navigating to  `MASTER > VOLUME`
 - Fixed Stereo Unison Spread + Ringmod Synth Mode + Noise causing excessively loud output.
 - Fixed some bugs around the Waveform Loop Lock feature which allowed setting invalid loop points.
+- Fixed a bug in triangle LFO that could cause audible artefacts with some modulation targets, particularly when synced.
 
 ### User Interface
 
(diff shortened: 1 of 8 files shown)


src/deluge/model/sample/sample.cpp
```diff
@@ -1794,7 +1794,10 @@ void Sample::numReasonsDecreasedToZero(char const* errorCode) {
 
 		// LegsMechanical got, V4.0.0-beta2.
 		// https://forums.synthstrom.com/discussion/4106/v4-0-beta2-e078-crash-when-recording-audio-clip
+#if ALPHA_OR_BETA_VERSION
+		// this occurs very rarely and should only leak one cluster - better to not freeze in the official build
 		FREEZE_WITH_ERROR("E078");
+#endif
```
sapphire-arches, line 1800: this logic is already guarded

sapphire-arches (commented):
If we merge this to 1.1 we'll lose a bunch of sanity checks but maybe that's fine and we can reenable them after the RC build?

src/deluge/model/sample/sample.cpp
m-m-adams: Oh you're right, I missed that the whole function was wrapped in it

m-m-adams:
> If we merge this to 1.1 we'll lose a bunch of sanity checks but maybe that's fine and we can reenable them after the RC build?

Yeah, my thinking is that it's ok because we're only seeing E078 come up right now and we know in that case freezing is worse than leaking a small amount of memory. But yeah flipping it on again after making the RC is a good idea

sapphire-arches (approved):

PR #2029, opened 2024-05-26
Scheduler - reconfigure minimum time to be a gauranteed backoff between calls

m-m-adams:
Change the minimum time to measure from last return instead of last call

This makes it into a backoff that guarantees some time for lower priority tasks to run

### DIFF
--- a/src/OSLikeStuff/task_scheduler.h
+++ b/src/OSLikeStuff/task_scheduler.h
@@ -25,10 +25,12 @@ extern "C" {
 /// void function with no arguments
 typedef void (*TaskHandle)();
 typedef int8_t TaskID;
-/// highest priority is 0. Attempt to schedule the task every maxTime after minTime has passed, in priority order
-/// if runToCompletion is false the task will be interrupted by higher priority requirements, otherwise it will run
-/// until it returns. returns the index of the task in the global table
-uint8_t addRepeatingTask(TaskHandle task, uint8_t priority, double minTimeBetweenCalls, double targetTimeBetweenCalls,
+/// highest priority is 0.
+/// min time is the minimum time from when the task finishes to when it's next called
+/// target time is the desired time between calls
+/// max time is the maximum time between calls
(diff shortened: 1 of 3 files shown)


src/OSLikeStuff/task_scheduler.h
```diff
@@ -25,10 +25,12 @@ extern "C" {
...
 typedef int8_t TaskID;
-/// highest priority is 0. Attempt to schedule the task every maxTime after minTime has passed, in priority order
-/// if runToCompletion is false the task will be interrupted by higher priority requirements, otherwise it will run
-/// until it returns. returns the index of the task in the global table
-uint8_t addRepeatingTask(TaskHandle task, uint8_t priority, double minTimeBetweenCalls, double targetTimeBetweenCalls,
+/// highest priority is 0.
+/// min time is the minimum time from when the task finishes to when it's next called
+/// target time is the desired time between calls
+/// max time is the maximum time between calls
+/// The scheduler runs the lowest priority task that can finish before anything higher priority needs to start
```
sapphire-arches, line 32:
```suggestion
/// Schedule a task that will be called at a regular interval.
///
/// The scheduler will try to run the task at a regular cadence such that the time between completion of calls to the
/// task is approximately targetTimeBetweenCalls. It will never call the task sooner than backOffTime seconds after it
/// last completed.
///
/// Tasks are selected to run based on priority and expected duration (computed via a running average of previous
/// invocations of the task). The task with the lowest numerical priority that can complete before a task with higher
/// semantic priority needs to start will run, without violation of the backOffTime.
///
/// @param task The task to call
/// @param priority Priority of the task. Tasks with lower numbers are given preference over tasks with higher numbers.
/// @param backOffTime Minimum time between calls, in seconds.
/// @param targetTimeBetweenCalls Desired time between completion of the execution of calls to the task.
```

sapphire-arches (commented):
.

PR #2034, opened 2024-05-26
[DRAFT] persistent MIDI learn

nikodemus:
- Preserve learned MIDI controls across synth loads.
- Output::getMidiKnobs() -- Those subclasses that are also ModControllableAudio override the default.
- MidiKnobArray::cloneFrom() -- for cloning initial values from the previous output.
- StorageManager::loadInstrumentFromFile() -- add argument from which to copy MIDI settings,

TODO:

- [ ] When a new synth is loaded, it resets parameters from MIDI knobs to their default values. Current knob settings should instead be respected.
- [ ] Retain controls even when an entirely new synth is created.
- [ ] What happens when you have midi learned a control (eg. LPF) that the loaded synth does not suppot (eg. FM synth), and what should happen?
  - Partial answer: "nothing" happens, the mapped control does nothing, but is retained for
    the next synth. This is right now by accident, need to check that it works by design as well.
- [ ] CRASHES:
  - [ ] Map master volume, convert to kit, adjust master volume, BOOM.

### DIFF
--- a/src/deluge/gui/context_menu/load_instrument_preset.cpp
+++ b/src/deluge/gui/context_menu/load_instrument_preset.cpp
@@ -36,21 +36,13 @@ Sized<char const**> LoadInstrumentPreset::getOptions() {
 }
 
 bool LoadInstrumentPreset::acceptCurrentOption() {
-	Error error;
-
-	switch (currentOption) {
-	/*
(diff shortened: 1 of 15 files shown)


nikodemus:

PR #2045, opened 2024-05-28
[DRAFT] extended bitcrusher

nikodemus:
- Stock goes from 13 bits to 6 bits, this goes from 15 bits to 1 bit.

TODO:

- [x] Verify that I got the range right: I can't see or hear any reduction at Bitcrush=1 ? Is our sine made out of even numbers, or are my ears and scope 15 bits? (Uh, I need a scope right after Deluge, don't I?) 
- [ ] Change the UX to show number of bits crushed: "1 bts" displays well enough on 7 seg as well, I think.
- [ ] Preserve sound of saved instruments. New tag for new scale, handling for old tag.
  
MAYBE:

- [ ] The MOD-FX volume compensation is not right, maybe fix it. There's a too-obvious jump at one point.
- [x] The crushing is currently reduced by 1 whenever decimation (aka sample rate reduction) is used, why? What to do about that? If the UI shows "4 bit crush", it seems like it should be a 4 bit crush even with SRR? Maybe previously this made made sense when the minimum bitcrush was 3, but now when the minimum is 1, less so? (Could just document it, I guess?)
- [ ] Routing option to move bitcrush before the filters, something like:
  - DEFAULTS > FX > BITCRUSH > BEFORE/AFTER FILTERS
  - CLIP > FX > BITCRUSH > BEFORE/AFTER FILTERS

### DIFF
--- a/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
+++ b/src/deluge/model/mod_controllable/mod_controllable_audio.cpp
@@ -20,6 +20,7 @@
 #include "deluge/model/settings/runtime_feature_settings.h"
 #include "dsp/stereo_sample.h"
 #include "gui/l10n/l10n.h"
+#include "gui/menu_item/value_scaling.h"
 #include "gui/views/automation_view.h"
 #include "gui/views/performance_session_view.h"
 #include "gui/views/session_view.h"
@@ -602,9 +603,10 @@ void ModControllableAudio::processReverbSendAndVolume(StereoSample* buffer, int3
 	postReverbVolumeLastTime = postReverbVolume;
 }
 
-bool ModControllableAudio::isBitcrushingEnabled(ParamManager* paramManager) {
+int32_t ModControllableAudio::bitcrushingAmount(ParamManager* paramManager) {
 	UnpatchedParamSet* unpatchedParams = paramManager->getUnpatchedParamSet();
-	return (unpatchedParams->getValue(params::UNPATCHED_BITCRUSHING) >= -2113929216);
+	int32_t value = unpatchedParams->getValue(params::UNPATCHED_BITCRUSHING);
+	return computeCurrentValueForStandardMenuItem(value);
 }
 
 bool ModControllableAudio::isSRREnabled(ParamManager* paramManager) {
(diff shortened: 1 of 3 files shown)


src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -677,8 +683,8 @@ void ModControllableAudio::processSRRAndBitcrushing(StereoSample* buffer, int32_
 				                  + multiply_32x32_rshift32_rounded(currentSample->l, strength2 << 9);
 				grabbedSample.r = multiply_32x32_rshift32_rounded(lastSample.r, strength1 << 9)
 				                  + multiply_32x32_rshift32_rounded(currentSample->r, strength2 << 9);
-				grabbedSample.l &= bitCrushMaskForSRR;
-				grabbedSample.r &= bitCrushMaskForSRR;
+				grabbedSample.l &= bitCrushMask;
```
m-m-adams, line 686: It drops a bit during the multiplication above this line so the difference is to compensate for that. Dunno if it makes sense to do that though?

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -616,31 +616,37 @@ void ModControllableAudio::processSRRAndBitcrushing(StereoSample* buffer, int32_
...
-			*postFXVolume >>= (positivePreset - 4);
+		int32_t preset =  paramManager->getUnpatchedParamSet()->getValue(params::UNPATCHED_BITCRUSHING);
+		// Scale to 1-50, clamp to 15 -- use that as the reduction. Ie. reduce 1 bit per number
+		// on the user scale. Top end above 15 is all 1 bit sound.
+		//
+		// TODO: This changes sound of existing instruments with bitcrush. Need to address
+		// in load / save code.
+		int32_t crushAmount = std::min<int32_t>(15, preset / 85899345 + 25);
+		// TODO: What's the logic in the volume compensation here? Is this ad-hoc, or something
+		// that makes specific sense?
```
m-m-adams, line 633: seems ad hoc and I don't know that it matches the bitcrush volume increase all that well

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
```diff
@@ -616,31 +616,37 @@ void ModControllableAudio::processSRRAndBitcrushing(StereoSample* buffer, int32_
...
+			// Matches historical behaviour. Don't understand why, but ok.
+			bitCrushMask = 0xFFFFFFFF << (15 + crushAmount);
+		} else {
+			bitCrushMask = 0xFFFFFFFF << (16 + crushAmount);
 			StereoSample* __restrict__ currentSample = buffer;
 			do {
-				currentSample->l &= mask;
-				currentSample->r &= mask;
+				currentSample->l &= bitCrushMask;
+				currentSample->r &= bitCrushMask;
```
m-m-adams, line 647: if we're changing the effect anyway we can make the bit crush more intuitive by also shifting the sample left. Maybe alternate between shifting the sample and the mask so that we get 30 possible crush amounts?

nikodemus: I'll see what kind of sounds we get that way :)

src/deluge/model/mod_controllable/mod_controllable_audio.cpp
nikodemus: So, I think this makes sense. As you figured out, the SRR adds one bit of reduction. 0 to 1 is hard to hear, but each subsequent step is obvious. This keeps the audible crush the same with and without SRR - and old instruments sounding the same.

nikodemus:

PR #2047, opened 2024-05-29
Add names and usage stats to task manager

m-m-adams:
This is functional but I don't love the syntax for constructing tasks, would appreciate suggestions for refactoring


m-m-adams:
> We could have a type like `TaskSchedule` that packages the time parameters and priority, since we want to centralize all that data in a header anyway. That would at least make the argument list shorter.

Thanks, I think that works well

PR #2049, opened 2024-05-29
Automation view cleanup (prep for velocity)

seangoodvibes:
Some more automation view cleanup work and prep for velocity view


sapphire-arches:
This feels like we need to introduce an abstraction, something like `GridGraphEditable`, which handles the behaviors that are in the renamed methods. That way if we want to add another thing later (e.g. MPE params) it'll be a matter of implementing that abstract class rather than adding even more verbose method names to the automation view code.

seangoodvibes:
> This feels like we need to introduce an abstraction, something like `GridGraphEditable`, which handles the behaviors that are in the renamed methods. That way if we want to add another thing later (e.g. MPE params) it'll be a matter of implementing that abstract class rather than adding even more verbose method names to the automation view code.

Agreed.

This was a step towards that so that I could identify what is automation specific vs note specific and then be able to subsequently refactor / abstract stuff out to their own sub-classes or something.

sapphire-arches:
#2051 tracks the followup refactor, rename LGTM

PR #2052, opened 2024-05-29
oled: stop scrolling animation

sapphire-arches:
This used to be done implicitly by `clearMainImage`


src/deluge/gui/views/view.cpp
```diff
@@ -1966,6 +1966,7 @@ void View::drawOutputNameFromDetails(OutputType outputType, int32_t channel, int
 
 	if (display->haveOLED()) {
 		deluge::hid::display::oled_canvas::Canvas& canvas = hid::display::OLED::main;
+		deluge::hid::display::OLED::stopScrollingAnimation();
```
m-m-adams, line 1969: why are some of these using stop scrolling animation plus clear() and some using clearMainImage() instead?

sapphire-arches: because I realized later that I should just switch them all back and missed some

m-m-adams (approved):

PR #2056, opened 2024-05-30
refactoring: readCurrentValue() & friends

nikodemus:
- Pull out readCurrentValue() and getFinalValue() guts for testing, plus a few writeCurrentValue() ones, if they did not use getFinalValue().

- The inputs have an explicit range, which is short enough that we can enumerate it and check that current value undoes with final value did.

TODO:

- [ ] Test failures in arp-related versions. (1) Verify that the lifted functions produce the same values as original code. (2) If they do, debug trace what is going on at runtime. (3) Fix what's broken.

- [ ] Pull out all the rest of these functions.

- [ ] After everything works generate test vectors using current inputs and outputs, instead of just testing round-tripping.

- [ ] Eliminate duplicates: many of the implementations are functionally identical.

- [ ] With all non-identical use-cases in one place, replace them all with a single parametrized version.

- [ ] Celebrate.

### DIFF
--- a/src/deluge/gui/menu_item/arpeggiator/midi_cv/gate.h
+++ b/src/deluge/gui/menu_item/arpeggiator/midi_cv/gate.h
@@ -16,6 +16,7 @@
  */
 #pragma once
 #include "gui/menu_item/integer.h"
+#include "gui/menu_item/value_scaling.h"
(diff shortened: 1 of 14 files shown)


nikodemus:
Breaking into smaller parts, superseded by series startin from #2057

PR #2058, opened 2024-05-30
refactor: share value scaling between Integer and UnpatchedParam

nikodemus:
- Math is exactly the same in both, let's call it compute(Current|Final)ValueForStandardMenuItem.

- Add a few tests with explicit values in addition to the roundtripping.


nikodemus:
No rush, but FYI: I already have 6 sequential PRs in queue, ready to go after this - and each other.

Since this is a largish refactoring, I'm keeping the individual PRs bite sized for easier review, but that does create a bit of a queue... :)

PR #2060, opened 2024-05-31
containers: move open hash table tests to actual test binary

sapphire-arches:
also includes a bunch of mocks to bring up more of the software in the native test binary.


m-m-adams (approved):

m-m-adams:
Reverting - this only works on linux, it should be in the 32bit unit tests directory so they're excluded on other platforms

PR #2061, opened 2024-05-31
voice: use 2x regular release for fast release

sapphire-arches:
This is much more conservative than the original logic, avoiding note-off clicks for soft culls.

### DIFF
--- a/src/deluge/modulation/envelope.cpp
+++ b/src/deluge/modulation/envelope.cpp
@@ -86,6 +86,10 @@ considerEnvelopeStage:
 		break;
 
 	case EnvelopeStage::FAST_RELEASE:
+		if (fastReleaseIncrement < 2 * release) {
+			release = 2 * release;
+			fastReleaseIncrement = release;
+		}
 		pos += fastReleaseIncrement * numSamples;
 		if (pos >= 8388608) {
 			setState(EnvelopeStage::OFF);


nikodemus:
> i gotta be honest i have no idea why this works, but it helps with the House 15 song from chewabledrapery.

I'm not sure if this goes against project policy, but I at least would appreciate linking to the song in question. (Maybe also mentioning the reason/uncertainty of this in a comment, so one doesn't need to trawl commit messages to figure out why this is the way it is?)

src/deluge/modulation/envelope.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "modulation/envelope.h"
 #include "definitions_cxx.hpp"
+#include "io/debug/log.h"
```
nikodemus, line 20: Lefover include?

sapphire-arches:
> > i gotta be honest i have no idea why this works, but it helps with the House 15 song from chewabledrapery.
> 
> I'm not sure if this goes against project policy, but I at least would appreciate linking to the song in question. (Maybe also mentioning the reason/uncertainty of this in a comment, so one doesn't need to trawl commit messages to figure out why this is the way it is?)

yeah, i asked chewablesrapery to file a bug and it looks like they have, which has the song -- this PR is currently underbaked since i just wanted to get something for them to test today. I'll figure out why this works and retarget it to community before marking it ready for review

sapphire-arches:
Cleaned up the commit message and thought the logic through correctly, should be better now.

m-m-adams:
I believe this is actually making fast release at least twice as fast as normal release

A release of 0 has an increment of 8192 iirc. A fast release from polyphony limits (mono/auto/legato) is at an increment of 4096, which is doubled by this PR. A soft cull is at an increment of 65536, and so won't be affected 

Good to merge though since it does seem like a good change

m-m-adams (approved):

PR #2068, opened 2024-06-01
tests: CompParam::readCurrentValue() & getFinalValue()

nikodemus:
- Pull out the guts for testing.

- Rename standard menu item roundtrip test to "scaling test", since it tests a few specific values in addition to roundtripping.

### DIFF
--- a/src/deluge/gui/menu_item/audio_compressor/compressor_params.h
+++ b/src/deluge/gui/menu_item/audio_compressor/compressor_params.h
@@ -17,6 +17,7 @@
 #pragma once
 #include "definitions_cxx.hpp"
 #include "gui/menu_item/unpatched_param.h"
+#include "gui/menu_item/value_scaling.h"
 #include "gui/ui/sound_editor.h"
 #include "modulation/params/param_set.h"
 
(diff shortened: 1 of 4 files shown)


src/deluge/gui/menu_item/audio_compressor/compressor_params.h
```diff
@@ -26,21 +27,10 @@ class CompParam final : public UnpatchedParam {
 public:
 	using UnpatchedParam::UnpatchedParam;
 	void readCurrentValue() override {
-		auto value = (int64_t)soundEditor.currentParamManager->getUnpatchedParamSet()->getValue(getP());
-		this->setValue((value * (kMaxMenuValue * 2) + 2147483648) >> 32);
-	}
-	// comp params aren't set up for negative inputs - this is the same as osc pulse width
```
m-m-adams, line 32: Is this accurate? If so the function should probably be named for unipolar param instead of for comp param

nikodemus: There's another PR coming that deals with pulse width and does just that :)

sapphire-arches: A bunch of the unipolar params do use the full [-1,1) q31 range but are notionally (e.g. volume). HalfPrecisionUnipolar maybe?

m-m-adams (approved):

src/deluge/gui/menu_item/audio_compressor/compressor_params.h
nikodemus: I had InternallyUnipolar previously, but HalfPrecisionUnipolar is better!

PR #2073, opened 2024-06-02
Fix kit shortcut blinks

sapphire-arches:
Some bonus refactoring that saves some code size as well.

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -566,6 +566,7 @@ void SoundEditor::setupShortcutsBlinkFromTable(MenuItem const* const currentItem
 		for (auto y = 0; y < kDisplayHeight; ++y) {
 			if (items[x][y] == currentItem) {
 				setupShortcutBlink(x, y, 0);
+				return;
 			}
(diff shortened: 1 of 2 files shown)


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -560,6 +560,17 @@ void SoundEditor::updateSourceBlinks(MenuItem* currentItem) {
 	}
 }
 
+void SoundEditor::setupShortcutsBlinkFromTable(MenuItem const* const currentItem,
+                                               MenuItem const* const items[kDisplayWidth][kDisplayHeight]) {
+	for (auto x = 0; x < 15; ++x) {
+		for (auto y = 0; y < kDisplayHeight; ++y) {
+			if (items[x][y] == currentItem) {
+				setupShortcutBlink(x, y, 0);
```
m-m-adams, line 568: Should it return after this? There's cases where multiple pads have the same menu type, like audio clip reverse

sapphire-arches: ... yes

m-m-adams (approved):

PR #2085, opened 2024-06-03
tests: move OpenAddressingHashTable tests to a new test binary

sapphire-arches:
Still built using the 32-bit config, which should avoid problems on non-linux platforms.


m-m-adams:
Hm, looks like the test workflow breaks. Not sure where it's getting the path for the 32bit tests from but it seems to be wrong

sapphire-arches:
yeah, I thought it was using ctest but it looks like it's hardcoding all the paths. I'll have to update the workflow file.

sapphire-arches:
Should be fixed now.

m-m-adams (approved):

PR #2087, opened 2024-06-03
test/refactor: arpgeggiator::midi_cv::Gate values roundtrip

nikodemus:
- Pull out the guts for testing: the current value math is identical to standard, so share the machinery, but the final value computation produces numbers that are slightly smaller near the max values, and needs to remain distinct. They use-provided values roundtrip despite the asymmetry.

- Speculate on the reason things are the way they are.

### DIFF
--- a/src/deluge/gui/menu_item/value_scaling.h
+++ b/src/deluge/gui/menu_item/value_scaling.h
@@ -26,6 +26,7 @@
 ///
 /// Done:
 /// - audio_compressor::CompParam
+/// - arpeggiator::midi_cv::Gate
 /// - osc::PulseWidth
 /// - patched_param::Integer
 /// - patched_param::Pan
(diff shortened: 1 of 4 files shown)


src/deluge/gui/menu_item/value_scaling.h
```diff
@@ -68,3 +69,18 @@ int32_t computeCurrentValueForPan(int32_t value);
 /** Scales -25 to 25 range to INT32_MIN-INT32_MAX for storage and use.
  */
 int32_t computeFinalValueForPan(int32_t value);
+
+/** Scales INT32_MIN-INT32_MAX range to 0-50 for display.
+ *
+ * This roundtrips with the final value math despite being
+ * not being it's proper inverse.
```
sapphire-arches, line 76: duplicate `being`

PR #2089, opened 2024-06-03
smarter dbt loadfw, making things easier for new devs

nikodemus:
- Given build type, figure out the file.

- If no hex key is given, use a stored one, or input one interactively. Figure out the MIDI port automagically, if possible. Explicit instructions for finding the key.

- Just one positional arguments.

- This means that: a simple `./dbt loadfw relwithdebinfo` does the right thing.

- Minor one-time cost for old devs to update their local setups which probably have something like ./dbt loadfw 2 128793 build/relwithdebinfo/deluge.bin scripted somewhere: keeping old command-lines working would have meant more complicated command line logic to figure out which positional arguments are elided.

- Use the third MIDI port for in & output for loadfw and sysex logging both, not the first one: the third one is intended as a dedicated sysex port.

- For --help cases report available ports to stdout: argparser uses stdout for help messages.

- Prettier output for cases where there are no midi ports of a given type: say "none" instead of printing nothing.

### DIFF
--- a/scripts/tasks/task-loadfw.py
+++ b/scripts/tasks/task-loadfw.py
@@ -4,44 +4,53 @@ import argparse
 import util
 import itertools
 import rtmidi
+import os
 
 
 def argparser():
-    midiout = rtmidi.MidiOut()
-    available_ports = midiout.get_ports()
-    s = "\n\nusage example: dbt loadfw 123 abcd1243 /path/deluge.bin"
-    s += "\n\nloads /path/deluge.bin to the deluge with key abcd1243, connected on MIDI port # 123"
-    s += "\n\nAvailable MIDI ports:\n\n"
-    for i, p in enumerate(available_ports):
-        s += "Port # " + str(i) + " : " + str(p) + "\n"
-
     parser = argparse.ArgumentParser(
         prog="loadfw",
         formatter_class=argparse.RawDescriptionHelpFormatter,
         description="Send firmware to Deluge via MIDI",
-        epilog=s,
-        exit_on_error=True,
+        epilog="""\nusage example: dbt loadfw relwithdebinfo
+                  Send build/relwithdebinfo/deluge.bin to MIDI port identified with deluge,
+                  using hex key stored in .deluge_hex_key.""",
+        exit_on_error=False,
     )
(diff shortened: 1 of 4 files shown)


scripts/tasks/task-loadfw.py
```diff
@@ -122,31 +131,85 @@ def make_sysex_messages(binary, handshake):
...
+                output.send_message(msg)
                 time.sleep(0.001 * delay_ms)
 
 
+def try_read_hex_key():
+    name = ".deluge_hex_key"
+    if os.path.isfile(".deluge_hex_key"):
+        with open(name, "r") as f:
+            return f.read().strip()
+    else:
```
sapphire-arches, line 158: a `try/except` block night work better here, as generally filesystems are more of a "ask forgiveness than permission" sort of entity (consider e.g. a file that exists but it's not readable).

scripts/tasks/util.py
```diff
@@ -198,3 +198,32 @@ def get_environment_from_batch_command(env_cmd, initial=None):
...
 def get_dbt_version():
     return open(get_git_root() / "toolchain" / "REQUIRED_VERSION").readline().rstrip()
+
+
+def ensure_midi_port(type, midi, port):
+    if port is None:
+        for i, p in enumerate(midi.get_ports()):
+            if str(p).startswith("Deluge"):
+                port = i
+                break
```
sapphire-arches, line 208: the intention is for port 3 to be the dedicated sysex virtual cable, and I think this logic selects port 1 instead? maybe not critical to fix now but it would be good to do while we're here.

nikodemus: I've no idea about the dedicated virtual port, but yes, this selects the first or "main" port labelled Deluge. This one works, though.

m-m-adams: There's cases that aren't handled on the deluge side if there's a second midi stream going at the same time. Using port 3 just mitigates that

scripts/tasks/task-loadfw.py
m-m-adams: Also consider the file might exist while checking but not while running if this is ran in parallel with a git clean or something similar

Issue #2066, opened 2024-06-01
Sidebar Scales mode allows you to change pads to duplicate scales.


ok-reza:
### Please describe the problem:

By pressing and holding a sidebar's scales pad and turning the `Vertical Encoder`, you can change the scale that pad switches to (this allows users to put the newer scales in scales mode, like 'Blues'). However, there is no way to prevent the user from assigning multiple pads to the same scale, when selected, multiple pads will be lit white. 

### What is the expected behavior?

Don't allow multiple pads able to be assigned to the same scale. I think it can create confusion and unnecessary busyness.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1.0-beta-F4A2506

Issue #2065, opened 2024-06-01
Sidebar 'Scales' mode doesn't return back to the current scale if the scale was set without the sidebar itself


ok-reza:
### Please describe the problem:

If you use sidebar's 'Scales' mode with a scale that's not `Major`, when you first press and hold another scale via the sidebar, on release it defaults to `Major` (presumably since it's the bottom-most pad).

If you've already manually latched a scale via sidebar, this issue doesn't occur. 

### What is the expected behavior?

Sidebar Scales mode should remember that the first current scale is the current latched scale. If you start in Dorian, you'll see the pad lit up for Dorian on the sidebar, but if you press and hold any other pad in that column, on release it will revert back to Major until you manually latch a new scale for the first time.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1.0-beta-F4A2506

### If possible provide the steps to reproduce the issue and upload additional media:

quick reproduce:
1) enter keyboard view with scale on and sidebar set to 'scales'
2) change keyboard scale to something other than major via shift+scale
(note how sidebar 'scales' is updating the active white pad as you do so)
3) now, press and hold any other pad for sidebar scales
4) release pad, and see that it reverts to 'major' the bottommost pad.

PR #2090, opened 2024-06-03
Fix scale sidebar column UI inconsistencies

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/2066
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/2065

To cherry pick for 1.1

### DIFF
--- a/src/deluge/gui/ui/keyboard/column_controls/scale_mode.cpp
+++ b/src/deluge/gui/ui/keyboard/column_controls/scale_mode.cpp
@@ -18,7 +18,6 @@
 
 #include "scale_mode.h"
 #include "gui/ui/keyboard/layout/column_controls.h"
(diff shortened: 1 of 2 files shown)


src/deluge/gui/ui/keyboard/column_controls/scale_mode.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include "scale_mode.h"
 #include "gui/ui/keyboard/layout/column_controls.h"
+#include <cstdbool>
```
m-m-adams, line 21: this shouldn't be needed, in cpp it just provides the macro  __bool_true_false_are_defined

m-m-adams (approved):

PR #2093, opened 2024-06-04
[DRAFT] Feature/ DX7: receive/send params as sysex

bfredl:
This works to edit a dx7 patch on the deluge using e.g. the Dexed desktop app, although it is a bit of a mess to setup.


src/deluge/playback/playback_handler.cpp
```diff
@@ -2937,6 +2937,23 @@ void PlaybackHandler::aftertouchReceived(MIDIDevice* fromDevice, int32_t channel
 	}
 }
 
+void PlaybackHandler::yamahaSysexReceived(MIDIDevice* fromDevice, uint8_t* data, int32_t len) {
+	int32_t substatus = data[2] >> 4;
+	int32_t channel = data[2] & 0x0f;
+
+	display->popupText("huga");
+
```
sapphire-arches, line 2945: debug code?

sapphire-arches (commented):
would be good to add docs on how to set this up to the dx7 markdown.

docs/features/dx_synth.md
```diff
@@ -93,6 +93,24 @@ However when feedback is activated on algorithm 4 and 6 (those will multi-operat
...
+## MIDI integration
+
+The synth supports sending and receiving param edits as sysex messages, in the standard format supported by DX7 and many
+software implementations such as dexed. It is possible to use the Dexed UI for editing patch parameters.
+Here is an example for how setting up [Dexed](https://asb2m10.github.io/dexed/)
+
+- Connect the Deluge to the computer with USB
+- Start dexed in "Standalone" mode.
+- Click the "options" button in the top left.
+    - Make sure "Deluge MIDI 1" is selected as an "Active MIDI input". Also set "MIDI output" to "Deluge MIDI 1"
```
m-m-adams, line 105: these instructions should probably go to midi port 3 - we don't implement midi sysex stream interruption so I think it could cause bugs if you're playing notes and editing at the same time

m-m-adams:
This has a bunch of conflicts now - are you still working on it?

bfredl:
Rebased and changed preferred sysex port to "Deluge MIDI 3"

mweigt:
First of all: Thank you for the fantastic Work! Enabling DX7 emulation on Deluge is an incredible added value!

As testing was requested: Just jammed for one hour using dexed as sound control surface while playing a track on the deluge. Everything seems to work. 

Some nits :
- I have to actually *enter a single parameter sub-menu* (i.e. "algorithm" or "sync") for the "learn" assign midi channel to work, *not just the "DX7 parameters" menu*. That might be the intended behaviour, but the Doc is a bit ambiguous here.
- I *have to* activate midi follow mode for the deluge to react to dexed param changes. Again, this is probably the (currently) intended behaviour, but the docs could be interpreted as this being optional. 
- Is there/will there be a way to assign a fixed channel without using midi follow? 
- When using dexed to control the deluge, only one paramter seems to be sent at each edit: switching to a different patch on dexed is not reflected on the deluge. I do not know if dexed is not sending the data or deluge not recieving/processing this.
- Is there/will there be a way to additionally learn single parameters to normal midi CCs?

m-m-adams (approved):

m-m-adams (approved):

PR #2101, opened 2024-06-06
Refactor InstrumentClipView::adjustVelocity

seangoodvibes:
Refactored InstrumentClipView::adjustVelocity

Added comments to the code, factored out some of the code to separate functions

Prep for velocity view

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -2047,12 +2047,14 @@ void InstrumentClipView::checkIfAllEditPadPressesEnded(bool mayRenderSidebar) {
 	}
 }
 
+// adjust a note's velocity when pressing and holding a pad with a note in it and turning the horizontal encoder <>
 void InstrumentClipView::adjustVelocity(int32_t velocityChange) {
 
 	int32_t velocityValue = 0;
 
 	Action* action;
-	if (display->haveOLED() || display->hasPopup()) {
+	if (display->hasPopup()) {
+		// we're only going to adjust velocity when there's a pop-up so no need to get an action otherwise
 		action = actionLogger.getNewAction(ActionType::NOTE_EDIT, ActionAddition::ALLOWED);
 		if (!action) {
 			return; // Necessary why?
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2083,87 +2085,114 @@ void InstrumentClipView::adjustVelocity(int32_t velocityChange) {
...
 					noteI++;
 					note = noteRow->notes.getElement(noteI);
 				}
 
-				editPadPresses[i].intendedVelocity =
-				    velocitySumThisSquare / numNotesThisSquare; // Get the average. Ideally we'd have done this when
-				                                                // first selecting the note too, but I didn't
+				// Rohan: Get the average. Ideally we'd have done this when first selecting the note too, but I didn't
+
+				// Sean: not sure how getting the average when first selecting the note would help because the average
```
m-m-adams, line 2105: I think this is "when first selecting the note" like when you first press it, before starting to adjust velocity. That first press auditions at the velocity of the first note in the square

m-m-adams: Just remove the comments though since I don't think it matters

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -2083,87 +2085,114 @@ void InstrumentClipView::adjustVelocity(int32_t velocityChange) {
...
+	displayVelocity(velocityValue, velocityChange);
+
+	reassessAllAuditionStatus();
+}
+
+// determines whether or not you're trying to adjust the velocities of multiple notes
+// with different starting velocities (prior to adjustment)
+// used to determine whether to display the updated velocity value or a generalized
+// "velocity increased / decreased" message
+void InstrumentClipView::getVelocityValue(int32_t& velocityValue, int32_t velocity) {
```
m-m-adams, line 2138: Nit:

For better clarity the name should be updateVelocityValue and the second argument should be newVelocity

Should have a doc comment too like
///sets velocityValue to newvolicity if velocityValue is 0. Keep it at velocityValue if they're the same. Set it to 255 if they're different

m-m-adams (approved):

PR #2105, opened 2024-06-07
[DRAFT] Refactor Codebase for 64 Bit System Compatibility

seangoodvibes:


src/deluge/memory/general_memory_allocator.cpp
```diff
@@ -159,12 +158,12 @@ void* GeneralMemoryAllocator::alloc(uint32_t requiredSize, bool mayUseOnChipRam,
 }
 
 uint32_t GeneralMemoryAllocator::getAllocatedSize(void* address) {
-	uint32_t* header = (uint32_t*)((uint32_t)address - 4);
-	return (*header & SPACE_SIZE_MASK);
+	uintptr_t header = (uintptr_t)address - 4;
```
m-m-adams, line 161: This still needs to be a pointer - it's a size_t * (pointer to a size_t) since it's the address of where the allocation size is stored

tests/CMakeLists.txt
```diff
@@ -10,8 +10,7 @@ endif()
 enable_testing()
 #needs to support -m32 since the memory tests assume a 32 bit architecture
 if (UNIX AND NOT APPLE)
-    add_compile_options(-m32 -Og -ggdb3)
-    add_link_options(-m32)
+    add_compile_options(-Og -ggdb3)
     add_subdirectory(32bit_unit_tests)
```
m-m-adams, line 14: You can just keep this out of the if Block and delete the rest - it'll let you run these tests locally on non Linux systems while developing (./dbt test)

src/resetprg.c
```diff
@@ -126,7 +126,7 @@ static void relocateSDRAMSection(uint32_t* start, uint32_t* end) {
  * Return Value : none
  *******************************************************************************/
 void resetprg(void) {
-	emptySection(&__frunk_bss_start, &__frunk_bss_end);
+	emptySection(__frunk_bss_start, __frunk_bss_end);
```
m-m-adams, line 129: These still need to be references

src/resetprg.c
```diff
@@ -184,12 +184,12 @@ void resetprg(void) {
...
 
 #if !defined(NDEBUG)
-	const uint32_t SDRAM_SIZE = EXTERNAL_MEMORY_END - EXTERNAL_MEMORY_BEGIN;
+	const size_t SDRAM_SIZE = EXTERNAL_MEMORY_END - EXTERNAL_MEMORY_BEGIN;
 	memset((void*)EXTERNAL_MEMORY_BEGIN, 0, SDRAM_SIZE);
 #endif
 
-	relocateSDRAMSection(&__sdram_text_start, &__sdram_text_end);
-	relocateSDRAMSection(&__sdram_data_start, &__sdram_data_end);
+	relocateSDRAMSection(__sdram_text_start, __sdram_text_end);
```
m-m-adams, line 191: These still need to be references

src/resetprg.c
```diff
@@ -100,20 +100,20 @@ void _fini(void) {
...
-		*dst = 0;
+		dst = 0;
 		++dst;
 	}
 }
 
-static void relocateSDRAMSection(uint32_t* start, uint32_t* end) {
-	uint32_t* src = (uint32_t*)((uint32_t)&__heap_start + ((uint32_t)start - PLACEMENT_SDRAM_START));
-	uint32_t* dst = start;
+static void relocateSDRAMSection(uintptr_t start, uintptr_t end) {
```
m-m-adams, line 111: These should still be pointers - they're the addresses of the start and end of the section. I think they'd switch to uint* if anything

src/resetprg.c
```diff
@@ -100,20 +100,20 @@ void _fini(void) {
...
+	uintptr_t src = (__heap_start + (start - PLACEMENT_SDRAM_START));
+	uintptr_t dst = start;
 	while (dst < end) {
-		*dst = *src; // Copy to SDRAM
```
m-m-adams, line 115: They still need to be dereferenced, this is supposed to copy and clear values through the pointers

PR #2108, opened 2024-06-08
Refactoring: Update sequence/mpe/automation clear/shift/nudge code and documentation

seangoodvibes:
Updated implementation of the automation defaults which dictate the behaviour of clearing / shifting / nudging sequences on automation parameters and notes/sample and MPE

The refactoring in this PR extends the previous refactoring done for clearing a clip by passing "clear/shift/nudgeAutomation" and "clearSequenceAndMPE" as variables to the various clear/shift/nudge functions as opposed to checking in those functions for the automation defaults or whether you're in automation view, etc.

Also refactored the code for shifting audio clip automation into MPE and non-MPE automation.

Update documentation as well where appropriate.

### DIFF
--- a/src/deluge/model/clip/audio_clip.cpp
+++ b/src/deluge/model/clip/audio_clip.cpp
@@ -1229,11 +1229,11 @@ bool AudioClip::currentlyScrollableAndZoomable() {
 }
 
 void AudioClip::clear(Action* action, ModelStackWithTimelineCounter* modelStack, bool clearAutomation,
-                      bool clearNotesAndMPE) {
-	Clip::clear(action, modelStack, clearAutomation, clearNotesAndMPE);
+                      bool clearSequenceAndMPE) {
+	Clip::clear(action, modelStack, clearAutomation, clearSequenceAndMPE);
 
-	// do not clear sample when you are in automation view
-	if (getRootUI() != &automationView) {
+	// if clearSequenceAndMPE is true, clear sample
+	if (clearSequenceAndMPE) {
 		// If recording, stop that - but only if we're not doing tempoless recording
 		if (recorder) {
 			if (!playbackHandler.isEitherClockActive()) {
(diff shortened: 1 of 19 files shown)


src/deluge/model/clip/audio_clip.cpp
```diff
@@ -1229,11 +1229,11 @@ bool AudioClip::currentlyScrollableAndZoomable() {
 }
 
 void AudioClip::clear(Action* action, ModelStackWithTimelineCounter* modelStack, bool clearAutomation,
-                      bool clearNotesAndMPE) {
-	Clip::clear(action, modelStack, clearAutomation, clearNotesAndMPE);
+                      bool clearSequenceAndMPE) {
+	Clip::clear(action, modelStack, clearAutomation, clearSequenceAndMPE);
```
m-m-adams, line 1233: readability: maybe accept an enum whatToClear that could be {SEQUENCE_MPE, AUTOMATION, EVERYTHING} instead of the two bools so that it's immediately clear when looking at a function call

src/deluge/model/clip/audio_clip.cpp
```diff
@@ -1270,50 +1270,83 @@ void AudioClip::setPos(ModelStackWithTimelineCounter* modelStack, int32_t newPos
...
 }
 
-bool AudioClip::shiftHorizontally(ModelStackWithTimelineCounter* modelStack, int32_t amount) {
-	// No horizontal shift when recording
-	if (recorder)
-		return false;
-	// No horizontal shift when no sample is loaded
-	if (!sampleHolder.audioFile)
-		return false;
+bool AudioClip::shiftHorizontally(ModelStackWithTimelineCounter* modelStack, int32_t amount, bool shiftAutomation,
```
m-m-adams, line 1273: same enum thing here

src/deluge/modulation/params/param_manager.cpp
```diff
@@ -556,7 +556,11 @@ void ParamManagerForTimeline::tickSamples(int32_t numSamples, ModelStackWithThre
 void ParamManagerForTimeline::nudgeAutomationHorizontallyAtPos(int32_t pos, int32_t offset, int32_t lengthBeforeLoop,
                                                                Action* action,
                                                                ModelStackWithThreeMainThings* modelStack,
+                                                               bool nudgeAutomation, bool nudgeMPE,
```
m-m-adams, line 559: same enum thing - it's just hard to immediately see what the flags are setting when they're only true/false

m-m-adams (commented):
Actual changes look good, one readability concern and lgtm

m-m-adams (approved):

Issue #2112, opened 2024-06-10
Shortcut to edit Song Transpose number of semitones interferes with scrolling + instant mutes workflow


soymonitus:
### Please describe the problem:

While performing in row mode, i am used to do scrolling through the rows and doing instante mute/unmute actions (shift + mute pads). While holding Shift to be quicker to mute/unmute, I can't scroll like before because instead of scrolling the Deluge changes the TRANSPOSE number of semitones for the new global transpose feature

### What is the expected behavior?

Instead of using shortcut "SHift + Up/Down encoder", I suggest moving it to "Shift + hold and turn up/down encoder"(currently unused), keeping the existing "Hold and turn up/down encoder" to do the actual transpose action, so it doesn't interfere with scrolling and having shift engaged for instant mute/unmutes

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1 beta

PR #2113, opened 2024-06-10
Change Transpose Interval shortcut to avoid interfering with Row Mode instant mute workflow

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/2112

To cherry pick for 1.1


m-m-adams (approved):
thanks, this feels good to use

Issue #2064, opened 2024-06-01
Holding top sidebar value then switching sidebar mode should not cause latching.


ok-reza:
### Please describe the problem:

If you hold the top sidebar pad then turn the `<>` encoder to switch to a different sidebar mode, velocity stays latched to 127, the topmost value you pressed before switching. You can confirm this by either setting the adjacent sidebar's mode back to velocity and see it's set to 127, or simply recording after switching modes and checking note velocity.

### What is the expected behavior?

The value shouldn't latch since a user is forced to press that topmost pad to switch modes. It should switch back to the previous (or default) latched setting.

This is only an issue for Velocity and Modulation.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1.0-beta-F4A2506

m-m-adams:
Clarity: it's latched until you next enter velocity/mod wheel

PR #2116, opened 2024-06-11
Fix latched velocity and modwheel when changing sidebar column mode

soymonitus:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/2064
For nightly branch.

### DIFF
--- a/src/deluge/gui/ui/keyboard/column_controls/velocity.cpp
+++ b/src/deluge/gui/ui/keyboard/column_controls/velocity.cpp
@@ -57,6 +57,15 @@ bool VelocityColumn::handleVerticalEncoder(int8_t pad, int32_t offset) {
 	return false;
 };
 
+void VelocityColumn::handleLeavingColumn(ModelStackWithTimelineCounter* modelStackWithTimelineCounter,
+                                         KeyboardLayout* layout) {
+	if (!FlashStorage::keyboardFunctionsVelocityGlide) {
+		// Restore previously set Velocity
+		vDisplay = velocity32;
+		layout->velocity = (velocity32 + kHalfStep) >> kVelModShift;
+	}
+};
+
 void VelocityColumn::handlePad(ModelStackWithTimelineCounter* modelStackWithTimelineCounter, PressedPad pad,
                                KeyboardLayout* layout) {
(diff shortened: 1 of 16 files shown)


src/deluge/gui/ui/keyboard/column_controls/velocity.cpp
```diff
@@ -57,6 +57,15 @@ bool VelocityColumn::handleVerticalEncoder(int8_t pad, int32_t offset) {
 	return false;
 };
 
+void VelocityColumn::handleLeavingColumn(ModelStackWithTimelineCounter* modelStackWithTimelineCounter,
+                                         KeyboardLayout* layout) {
+	if (!FlashStorage::keyboardFunctionsVelocityGlide) {
```
sapphire-arches, line 62: I think we always want to reset to the held value, though I'm open to arguments in favor of this behavior.

soymonitus: If the Glide is enabled there is no coming back to old value. Regardless of if you short press or long press the value is replaced. But if glide is disabled it must jump to the last short pressed pad

m-m-adams: Is there any way to grab the previous value when glide is enabled? If not we should add one even if just for the purpose of holding the top pad to change columns

soymonitus: Ohh I know what you mean now. I think that if i remove the if condition here it will work as you want. I am going to test it...

soymonitus: @m-m-adams @sapphire-arches  I made a change to remove the logic that was there to send the pad release event to the previous column that was there before changing the column type. And now the pad release is "forced" at the exact moment you change columns. Also removed the IF condition in velocity and modwheel so it also respect the last set velocity and modwheel when you leave those columns. That way, even in glide mode, you can set a lower velocity for example, hold the top pad to change the column type, and still have that low velocity set and not replaced by 127 because you held the top pad

m-m-adams (approved):

PR #2145, opened 2024-06-18
Don't send multiple clocks at once

### LINKED ISSUE #2144: Severe Trigger Clock Dropouts

Alternate title: Trigger Clock Output Fails to Meet Realtime Deadline

When using the analog clock at the gate output, it will randomly stop transmitting for long periods.

On 8aa506f this happens after a few seconds. According to @m-m-adams it happens after a few minutes on **Release Candidate 3** 564cba31cd14d9fbfc16a28114ab6d4dba9441e6.

The trigger clock putters along more-or-less constantly, within a margin of error that most humans (and downstream sequencers) consider indistinguishable from lock-step. 

_No response_

OLED

Nightly

8aa506f

1. Turn on
2. Configure output 4 for clock
3. Set desired PPQN (happens at any but more obviously with larger PPQN)
4. Hook up something to watch the clock output
5. Press play
6. Wait a few seconds (typically less than 30) and the clock output will stop and then restart, this will continue to happen over and over again at apparently random intervals

## Additional Info

Tested with multiple downstream devices, but it is so severe that it is visible with the naked eye.

Watching a blinking LED connected to the output is sufficient to detect the dropouts at higher PPQNs, it is that major! The LED (or any downstream sequencer/clock divider) will receive no signal at all for up to nearly a second(?!) before the Deluge resumes transmitting, leaving the downstream devices hopelessly out of sync.

The improvements between the latest nightlies and the current community branch is likely due to improvements in the yield/scheduler code that have gone in, but it remains. I'm not sure what commit introduced the issue.

**NOTE: This is *different* to the other analog clock output bug that I reported.** 

- This happens *immediately* after being powered on and presents as randomly the clock not outputting and then resuming. 
- This bug does not occur with the official firmware.

# UPDATE 2024-06-19

Some attempts at code fixes have improved the situation, so that the dropouts occur less often, sometimes only after dozens of minutes (depending on the BPM). But the issue persists and the underlying cause has not yet been identified. What is suspected is that the flag which tells the Deluge when another clock pulse is needed gets double set, resulting in triggers being delayed. @m-m-adams noticed that something similar happens in MIDI, where the number of ticks per minute is correct but get delayed and bunched up. This may be the same behaviour happening with gates, but the gate circuitry isn't fast enough to send all of those pulses so close together so they all get blended together, resulting in the clock not being transmitted in discrete pulses but instead as a single gate.

m-m-adams:
Fix #2144 

Bug occurs because clocks are only scheduled when within numSamples worth of time at the end of the audio routine, but this causes a bug because the next render call is not guaranteed to be before numSamples have passed (there's about 64 samples of extra buffer on average). This would not occur during the main loop in official but was possible in routineForSD, making the bug take much longer to reproduce on official and 1.1. As the sequencer is only updated for the next numSamples it cannot be scheduled further in advance without sequencer changes.

When the clock should have been done before the render started but wasn't scheduled yet it gets added to the gate and midi output buffers. They aren't flushed until the next time the clock ISR runs, at which point two clocks get sent at once.

As a solution I have changed the clock logic to flush anything that's previously in the buffer when a new clock is added. This ensures that clocks go out on time (usually sample accurate, always within 64ish samples) 

Future work should probably re examine the whole output timing issue - clocks and notes/ccs shouldn't be linked in this way but decoupling them would be very difficult.

### DIFF
--- a/src/deluge/playback/playback_handler.cpp
+++ b/src/deluge/playback/playback_handler.cpp
@@ -629,6 +629,7 @@ maybeDoTriggerClockOutputTick:
 
 			if (fractionNextAnalogOutTick <= fractionLastTimerTick) {
 				doTriggerClockOutTick();
+				D_PRINTLN("double trigger");
 				goto maybeDoTriggerClockOutputTick;
 			}
 
--- a/src/deluge/playback/playback_handler.cpp
+++ b/src/deluge/playback/playback_handler.cpp
@@ -623,16 +623,23 @@ void PlaybackHandler::actionTimerTickPart2() {
 			uint32_t analogOutTicksPer;
 			getAnalogOutTicksToInternalTicksRatio(&internalTicksPer, &analogOutTicksPer);
 			uint64_t fractionLastTimerTick = lastTimerTickActioned * analogOutTicksPer;
-
+			static bool triggered = false;
 maybeDoTriggerClockOutputTick:
 			uint64_t fractionNextAnalogOutTick = (lastTriggerClockOutTickDone + 1) * internalTicksPer;
 
 			if (fractionNextAnalogOutTick <= fractionLastTimerTick) {
+				// theory - if we do this instead of scheduling it (handled at end of current audio routine), and the
+				// next tick isn't in this window, then this tick is picked up at the start of the audio routine and not
+				// by an ISR
 				doTriggerClockOutTick();
-				D_PRINTLN("double trigger");
-				goto maybeDoTriggerClockOutputTick;
+				if (triggered == true) {
+					D_PRINTLN("double trigger");
+				}
+				triggered = true;
+				// goto maybeDoTriggerClockOutputTick;
+				fractionNextAnalogOutTick += internalTicksPer;
 			}
-
+			triggered = false;
 			// Schedule another trigger clock output tick
 			scheduleTriggerClockOutTickParamsKnown(analogOutTicksPer, fractionLastTimerTick, fractionNextAnalogOutTick);
 		}
--- a/src/deluge/playback/playback_handler.cpp
+++ b/src/deluge/playback/playback_handler.cpp
@@ -623,13 +623,10 @@ void PlaybackHandler::actionTimerTickPart2() {
 			uint32_t analogOutTicksPer;
 			getAnalogOutTicksToInternalTicksRatio(&internalTicksPer, &analogOutTicksPer);
 			uint64_t fractionLastTimerTick = lastTimerTickActioned * analogOutTicksPer;
-			static bool triggered = false;
 			uint64_t fractionNextAnalogOutTick = (lastTriggerClockOutTickDone + 1) * internalTicksPer;
 
 			if (fractionNextAnalogOutTick <= fractionLastTimerTick) {
-
 				doTriggerClockOutTick();
-
 				fractionNextAnalogOutTick += internalTicksPer;
 			}
 			// Schedule another trigger clock output tick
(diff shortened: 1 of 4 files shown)


m-m-adams:
Bug is caused by updating the clock status before the previous clock has been sent out - on gates this causes a missed clock, on midi it's a double clock (two sent at same time). Last commit identifies the bug location

No solution yet

acook:
I'm seeing pretty significant jitter with a Deluge's 24 PPQN clock being detected by my Pam's as rapidly shifting between 118-121 BPM from a target 120 BPM.

This 3 BPM variation would be equivalent to a wow & flutter of 2.5%, well above the 0.2-0.25% threshold often said to be the audible minimum. This of course obliterates tempo-synced delays and makes recordings effectively impossible to align with other takes or layers.

What's interesting though is that if I have a set of 128th notes playing in a synth clip *with arpeggiator enabled*, the BPM seems to stabilize. This stabilizing effect is noticeable up until around 8th notes with synth clips. I'm sure the jitter is still there, but the synth is forcing syncs often enough that the Pam's can smooth it out.

I sometimes noticed the clock locking in with CV clips, but it was not as reliable. Drum kits seem to have no effect even at very fast trigger rates. Either way, if the clip is not actively playing notes then the stabilizing effect is absent. 

So it is like you said, notes and clocks are clearly connected.

m-m-adams:
I'd expect our jitter to be +-1.5ms anyway. The BPM impact of that will depend entirely on your tempo and PPQN. What do you have it set to? 

There's probably space for some more improvements but this does solve the dropout problem which is a large improvement

acook:
I don't understand what you mean. To clarify:

PPQN is 24

This jitter is absent in the official firmware.

This jitter is suppressed by the arpeggiator.

m-m-adams:
Oh sorry - this PR is only targeting the dropouts. The jitter is unrelated so with this fix I'd still expect an average jitter of about 1.5ms (32 samples at 44.1kHz) with occasional spikes up to 3ms if there's a long render time.

At 24 PPQN and 120BPM each pulse is 20.8 ms, and our range will be 20-22ms (116-125 bpm) during spikes. That's likely what PPW is seeing and then adding some smoothing. You might have better results with a lower PPQN. 

Feel free to open a new issue for the jitter since it should be possible to improve, but it's off topic for this PR

m-m-adams:
image

Measured and plotted jitter, confirming the expectation for how it should be behaving right now

src/deluge/playback/playback_handler.cpp
```diff
@@ -623,15 +623,15 @@ void PlaybackHandler::actionTimerTickPart2() {
 			uint32_t analogOutTicksPer;
 			getAnalogOutTicksToInternalTicksRatio(&internalTicksPer, &analogOutTicksPer);
 			uint64_t fractionLastTimerTick = lastTimerTickActioned * analogOutTicksPer;
-
-maybeDoTriggerClockOutputTick:
+			static bool triggered = false;
 			uint64_t fractionNextAnalogOutTick = (lastTriggerClockOutTickDone + 1) * internalTicksPer;
 
 			if (fractionNextAnalogOutTick <= fractionLastTimerTick) {
+
```
sapphire-arches, line 630: extra line?

src/deluge/playback/playback_handler.cpp
```diff
@@ -623,15 +623,15 @@ void PlaybackHandler::actionTimerTickPart2() {
 			uint32_t analogOutTicksPer;
 			getAnalogOutTicksToInternalTicksRatio(&internalTicksPer, &analogOutTicksPer);
 			uint64_t fractionLastTimerTick = lastTimerTickActioned * analogOutTicksPer;
-
-maybeDoTriggerClockOutputTick:
+			static bool triggered = false;
```
sapphire-arches, line 626: debug leftover or is it used somewhere outside the diff?

sapphire-arches (approved):

PR #2148, opened 2024-06-20
The next installment of the reforms of the Storage Manager

jamiefaye:
In which we split the StorageManager.cpp into 3 files. We also are slowly eliminating "globalism" by allowing multiple Serializers and Deserializers to co-exist. The writeJsonFlag controls which file format we save in. RIght now it is bool const = false which may allow some of the code to be stripped away.


seangoodvibes:
First comment: 

I see use of both f_close and closeFile. I think we should be consistent and use one or the other and not both?

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -284,7 +284,8 @@ void LoadSongUI::performLoad(StorageManager& bdsm) {
 
 someError:
 		display->displayError(error);
-		storageManager.closeFile();
+		f_close(&smDeserializer.readFIL);
```
m-m-adams, line 287: This feels like it should still be done through the storage manager

src/deluge/gui/ui/save/save_song_ui.cpp
```diff
@@ -237,7 +237,7 @@ bool SaveSongUI::performSave(StorageManager& bdsm, bool mayOverwrite) {
 				// converted
 
 				// Open file to read
-				FRESULT result = f_open(&fileSystemStuff.currentFile, sourceFilePath, FA_READ);
+				FRESULT result = f_open(&smSerializer.writeFIL, sourceFilePath, FA_READ);
```
m-m-adams, line 240: Same here - I think we should keep the actual file system calls centralized

src/deluge/gui/ui/load/load_song_ui.cpp
jamiefaye: I moved file closing into the FileReader and FileWriter classes.

src/deluge/gui/ui/save/save_song_ui.cpp
jamiefaye: Also handled by above change.

PR #2163, opened 2024-06-24
exit editor on left column press

### LINKED ISSUE #2027: Can't exit Sound or Settings Menus by pressing anywhere on the mute/unmute/launch column when in either Keyboard View.

Normally, the mute/unmute column is always available to press to exit any menu open on the Deluge. This is incredibly useful since the topmost mute/launch pad is right below the SHIFT button and is easily accesible.

With the new Keyboard View sidebar controls, this now doesn't allow users to access this shortcut when in a menu in Synth Keyboard View. Additionally, despite no sidebar controls being in Drum Keyboard View, the unused launch column also has no effect on exiting menus.

Pressing a launch column should be consistent with normal Deluge behavior and exit the current menu before functioning as it's sidebar duties.

I have a feeling that this may slightly disrupt the use of the sidebar controls in Synth Keyboard View, as you may be editing a sound parameter via the Sound Menu and want to utilize auditioning notes along with the sidebar features. If the expected behavior was implemented, then while doing so you would then exit the open sound param menu for your first sidebar press, then resume as normal. However, the best way to find out if this feels natural and isn't bothersome is to implement it and see. 

Another option would be to include an 'EMPTY' sidebar control option where the given column has no use aside from exiting a menu (if it's the launch column?). However, this means that a user won't be able to utilize a consistent Deluge feature unless they set their sidebar column each time they boot up, which is undesirable.

_No response_

OLED, 7SEG

other

1.1 beta

Synth Keyboard View:
1) Enter Synth Keyboard View
2) Enter a param shortcut (SHIFT+LPF)
3) Hit the first sidebar column to exit sound parameter menu.
4) No exit is made, velocity may adjust to 127.

Drum Keyboard View:
1) Enter Drum Keyboard View
2) Enter a param shortcut (SHIFT+LPF)
3) Hit the first sidebar column on the left to exit sound parameter menu.
4) No exit is made.

m-m-adams:
I don't like this personally - it looks and feels like a horrible waste of UI space

Fix #2027

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -151,7 +151,9 @@ bool SoundEditor::getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) {
 	switch (uiType) {
 	case UIType::KEYBOARD_SCREEN:
 		// don't greyout keyboard screen
-		doGreyout = false;
+		doGreyout = true;
+		// left column is instant exit so grey it out
+		*cols = 0x00000002;
 		break;
 	case UIType::AUTOMATION_VIEW:
 		// only greyout if you're not in automation editor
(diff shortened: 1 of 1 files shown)


sichtbeton:
cool. but couldnt this made like this: when deep in a menu = exit right away. then resume as normal.

so you wouldnt waste the sidebar in normal use but still keep this very valuable shortcut.

PR #2166, opened 2024-06-24
exit UIs on long press back

m-m-adams:
Implements long press back to exit menus as a fast and consistent everywhere alternative to pressing the sidebar as mentioned in #2027 

1. Make exitUI a virtual function in UI rather than separate per class functions

2. When back is pressed start a timer to exit the current UI

3. UIs can exit on long press back by implementing exitUI. Currently the following UIs exit (basically everything that has an instant exit ability already) 

- Browsers (song, sample, synths, etc.)
- Sound editing
- Various renaming menus

Also reduces the long hold from 500 to 400 - this feels good for the existing long press functions and 500 felt too long for holding back to exit. Open to putting that back though


seangoodvibes:
This works really well. Love this. Can we add an entry to change log and community features (general improvements) for this?

seangoodvibes (approved):

PR #2179, opened 2024-06-25
[draft] refactor: replace Song::modeNotes and numModeNotes with a NoteSet

nikodemus:
- NoteSet::operator[] gets note value of a scale degree.
- NoteSet::fromScaleNotes() constructs a NoteSet from scale notes.
- NoteSet::applyChanges() applies an array of scale degree semitone offsets.
- NoteSet::addUntrusted() implments safeguards that loading from song file uses.
- NoteSet::clear() clears a set.
- NoteSet::highest() returns the highest note in the set.

There are no known issues with this patch, but given its widespread nature I want write more comprehensive tests and run it a while before saying it's done. 😅

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -3358,9 +3358,9 @@ int32_t InstrumentClipView::getYVisualFromYDisplay(int32_t yDisplay) {
 int32_t InstrumentClipView::getYVisualWithinOctaveFromYDisplay(int32_t yDisplay) {
 	int32_t yVisual = getYVisualFromYDisplay(yDisplay);
 	int32_t yVisualRelativeToRoot = yVisual - currentSong->rootNote;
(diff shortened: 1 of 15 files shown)


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -4600,9 +4600,10 @@ ActionResult InstrumentClipView::verticalEncoderAction(int32_t offset, bool inCa
 
 			// If shift button not pressed, transpose whole octave
 			if (!Buttons::isShiftButtonPressed()) {
-				// If in scale mode, an octave takes numModeNotes rows while in chromatic mode it takes 12 rows
```
seangoodvibes, line 4603: As this transpose code is common to both automationView and instrumentClipView, I would suggest factoring out the code to an InstrumentClipView:: function that can be called from automation view so we don't have code duplication.

But maybe that can be a different PR after this one is merged :)

nikodemus: In #2189 while thinking about this one still.

nikodemus:
#2201 is it

Issue #2151, opened 2024-06-21
Deluge “Freezing” when doubling clip length 


lmcmulle:
### Please describe the problem:

updated: doubling clip causes freeze - requiring power cycle - even when sequencer is not playing - various tested scenarios in comment below - 

tested with jun 21 nightly - 

### old version. 
scenario: playing an arrangement under moderate cpu load (cpu light not flashing)  doubling clip length will cause the machine to freeze requiring a power cycle 

consistent  problem - happened twice with the jun 19 nightly build- only while playing - but not every time.  i will run some tests to see if i can narrow the fault conditions 

additional tests performed see comment 

L

### What is the expected behavior?

clip will duplicate while playing and not freeze the machine 

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

jun 19th nightly 

### If possible provide the steps to reproduce the issue and upload additional media:

playing an arrangement with several tracks placing the machine under moderate to heavy load (no apparent voice stealing ) 
duplicating a clip of a synth track which is playing caused the machine to freeze

lmcmulle:
issue persists with Jun-21 nightly -
- issue is now replicated across multiple tracks
- high cpu light is NOT flashing 
- only occurs when sequencer is playing (it seems) 
- if the sequencer is playing - and the track being doubled is not playing  - the issue still occurs 
- there does not seem to be a pattern to when it happens sometimes it’s in the first, but as many as 4 doubles before it happens 
- -track length of the track being doubled  does not affect behavior - 
- tested to see if resulting size (eg 4,8,16 bars) was a factor it was not -

lmcmulle:
additional testing:
- doubling one clip - backing out to arranger - doubling a variation of the same clip - freeze occurs 
- double one clip - switching instruments - double that clip - freeze occurs 


###net it out: doubling the clip causes the freeze - somewhere between the 2nd and 4th double - regardless of instrument /track changes or  other state of the machine eg playing or not-

m-m-adams:
I have had this happen so it's definitely a bug. I'm not having luck deliberately reproducing while attached to a debugger so far though.

m-m-adams:
Most reliable repro - make a couple tracks using large samples (like BoD), add notes, press play, then alternate doubling tracks until a freeze happens

Freeze happens in memcpy in resizeableArray::generateRepeats which appears to clobber everything 

Non optimized builds do not freeze

[...]

PR #2188, opened 2024-06-26
Catch error before infinite copy

m-m-adams:
Reliably freezes with error on #2151 before the infinite memcpy

It is extremely unclear why but popupMaxX and popupMaxY appear to have garbage in them

I'm not familiar with the image code at all so I'm going to stop here and hope someone else picks it up

### DIFF
--- a/src/deluge/hid/display/oled.cpp
+++ b/src/deluge/hid/display/oled.cpp
@@ -309,7 +309,14 @@ void copyRowWithMask(uint8_t destMask, uint8_t sourceRow[], uint8_t destRow[], i
 
 void copyBackgroundAroundForeground(ImageStore backgroundImage, ImageStore foregroundImage, int32_t minX, int32_t minY,
                                     int32_t maxX, int32_t maxY) {
-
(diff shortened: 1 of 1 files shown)


nikodemus:
Since they're globals they're pretty ripe for memory overwrites - and a different layout between relaease and relwithdebinfo explains why this doesn't happen with both.

Trying to find out what else is near them in memory might give a clue, but textually `ConsoleItems` is nearby, and if `drawConsoleTopLine` is called with `numConsoleItems = 0` it would underflow, possibly causing this?

m-m-adams:
image

They're in the declared order so it's probably a logic bug somewhere, otherwise all 4 would be garbage

m-m-adams:
> If nothing else, adding a check for `numConsoleItems > 0` to `drawConsoleTopLine` might be worth it?

That seems to stop the bug entirely which I think implies that there's UB somewhere

m-m-adams:
@nikodemus @saphire_arches the closest array looks to be the console canvas itself, so probably an out of bounds write in one of the console rendering routines

m-m-adams:
Gonna merge this, want to rule out that this is all the freezes and there aren't multiple causes

PR #2189, opened 2024-06-26
refactor: simplify call sites of nudgeNotesVertically()

nikodemus:
- Move the "how many rows do we shift"-logic to the function instead of duplicating it on call sites: instead pass in an enum indicating the kind of nudge requested.

- Interpret the integer argument as direction instead of a count, making it explicit we're either moving by one row or octave.

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -3697,9 +3697,8 @@ ActionResult AutomationView::verticalEncoderAction(int32_t offset, bool inCardRo
 		else if (!currentUIMode && outputType != OutputType::KIT) {
 			actionLogger.deleteAllLogs();
 
-			clip->nudgeNotesVertically(
-			    offset, Buttons::isShiftButtonPressed() ? VerticalNudgeType::ROW : VerticalNudgeType::OCTAVE,
-			    modelStack);
+			auto nudgeType = Buttons::isShiftButtonPressed() ? VerticalNudgeType::ROW : VerticalNudgeType::OCTAVE;
+			clip->nudgeNotesVertically(offset, nudgeType, modelStack);
 
 			instrumentClipView.recalculateColours();
 			uiNeedsRendering(this, 0, 0xFFFFFFFF);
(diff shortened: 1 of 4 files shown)


src/deluge/model/clip/instrument_clip.h
```diff
@@ -78,7 +80,7 @@ class InstrumentClip final : public Clip {
 	void replaceMusicalMode(uint8_t numModeNotes, int8_t changes[], ModelStackWithTimelineCounter* modelStack);
 	void seeWhatNotesWithinOctaveArePresent(NoteSet&, int32_t, Song* song, bool deleteEmptyNoteRows = true);
 	void transpose(int32_t, ModelStackWithTimelineCounter* modelStack);
-	void nudgeNotesVertically(int32_t, ModelStackWithTimelineCounter* modelStack);
+	void nudgeNotesVertically(int32_t, VerticalNudgeType, ModelStackWithTimelineCounter* modelStack);
```
m-m-adams, line 83: Can we name the first parameter for clarity?

src/deluge/gui/views/automation_view.cpp
```diff
@@ -3697,21 +3697,10 @@ ActionResult AutomationView::verticalEncoderAction(int32_t offset, bool inCardRo
 		else if (!currentUIMode && outputType != OutputType::KIT) {
 			actionLogger.deleteAllLogs();
 
-			offset = std::min((int32_t)1, std::max((int32_t)-1, offset));
+			clip->nudgeNotesVertically(
+			    offset, Buttons::isShiftButtonPressed() ? VerticalNudgeType::ROW : VerticalNudgeType::OCTAVE,
```
m-m-adams, line 3701: nit but can we put the nudge type in a variable so this call isn't split on 3 lines?

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -4594,22 +4594,11 @@ ActionResult InstrumentClipView::verticalEncoderAction(int32_t offset, bool inCa
 			char modelStackMemory[MODEL_STACK_MAX_SIZE];
 			ModelStackWithTimelineCounter* modelStack = currentSong->setupModelStackWithCurrentClip(modelStackMemory);
 
-			InstrumentClip* instrumentClip = getCurrentInstrumentClip();
-
-			offset = std::min((int32_t)1, std::max((int32_t)-1, offset));
+			InstrumentClip* clip = getCurrentInstrumentClip();
+			clip->nudgeNotesVertically(
+			    offset, Buttons::isShiftButtonPressed() ? VerticalNudgeType::ROW : VerticalNudgeType::OCTAVE,
```
m-m-adams, line 4599: here too

m-m-adams (approved):

PR #2190, opened 2024-06-27
refactor: share value scaling between arpeggiator::midi_cv::Ratchet and Rhythm

nikodemus:
- The math is exactly the same, because NUM_PRESET_ARP_RHYTHMS _must_ in the current state of the codebase be kMaxMenuItem+1.

- Make that dependecy explicit, rename to kMaxPresetArpRhythm to get rid of extraneous - 1 calcs.

- Name the shared value scaling compute(Current|Final)ValueForArpMidiCVRatchetsOrRhythm. There doesn't seem to be a real reason they are shared, and the scaling seems awkward for Rhythm: awkward name reflects that un-nicely.

- Pull out the rhythm patterns into a separate header, so they don't pull in resizable array to unit tests.

- Use the compute-function in arpeggiator.cpp where appropriate.


tests/unit/value_scaling_tests.cpp
```diff
@@ -59,18 +61,22 @@ TEST(ValueScalingTest, arpMidiCvGateValueScaling) {
 	CHECK_EQUAL(50, computeCurrentValueForArpMidiCvGate(INT32_MAX));
 }
 
-TEST(ValueScalingTest, arpMidiCvRatchetValueScaling) {
+TEST(ValueScalingTest, consistentArpAndMenuMaxValues) {
+	CHECK_EQUAL(kMaxMenuValue, kMaxPresetArpRhythm);
```
sapphire-arches, line 65: this should probably be a `static_assert` near the definition of `kMaxPresetArpRhythm`

nikodemus: On reflection, I think the test needs to be separate:

Adding a static assert next to

```const int32_t kMaxPresetArpRhythm = kMaxMenuValue;```

is just redundant. Anyone tempted to change that assignment will also change the static assert.

The test being lexically distant is an advantage, because it will not be modified at the same time without any consideration.

I'll add some commentary about _why_ they need to be the same, though.

m-m-adams: I think this has the same issue, it'll just get modified to pass the test. Do any of the other tests fail if this isn't true?

nikodemus: I'll test and see. Still, a comment where the values are defined and an explicit test referring to the comment is IMO a reasonable precaution against accidental modification of depenedent parameters.

I mean, they're dependent already. This is just a tiny bit safety for when someone is neat that footgun.

nikodemus: I think the test is the way to go: no other test fails, and the build succeeds if I just increase the number of arp rhythms by one, despite that messing up menus.

I can change it to the a static assert instead, but as said I believe being textually distant is an advantage in this case: more chance of actually stopping and thinking instead of working on autopilot.

Just let me know which way you want it.

PR #2193, opened 2024-06-27
Release 1.2 - Update Midi Follow Feedback for kit drum selection

seangoodvibes:
Added one additional scenario where Midi Follow Feedback should be sent:

- when you've selected a kit clip for midi follow control and you're not currently in that kit clip (e.g. in song view), and affect entire is disabled in that kit clip, changing drum selection will necessitate sending updated midi feedback to advise controller of the changed kit drum selection context

Will create a separate PR for Release 1.1


m-m-adams:
I don't think we should merge. Midi feedback updating sends 70 messages, if you play a simple kick/snare/hat pattern with feedback enabled it will wreck the midi link for all other uses

This should wait until we have prioritized midi

PR #2194, opened 2024-06-27
Release 1.1 - Update midi follow feedback for kit drum selection

seangoodvibes:
Added one additional scenario where Midi Follow Feedback should be sent:

- when you've selected a kit clip for midi follow control and you're not currently in that kit clip (e.g. in song view), and affect entire is disabled in that kit clip, changing drum selection will necessitate sending updated midi feedback to advise controller of the changed kit drum selection context


m-m-adams:
https://github.com/SynthstromAudible/DelugeFirmware/pull/2193#issuecomment-2195894840

Same thoughts but to me putting this in 1.1 is an absolute no without extensive testing

PR #2199, opened 2024-06-28
Performance/note on speed

m-m-adams:
Improve the speed of note ons by marking the happy path (no arp, no probability, no failures as happy to improve caching

Lower the impact of note ons by not rendering the voice in the same window that the note was created in

Combine for a 10%ish reduction in window size during note ons for synth voices, both changes are about half of that independently. Multisamples are improved enough to play an extra voice in the same render length. 

Synth before (66 default voice note ons):
image

After:
image

Multisample before (29 multisampled piano note ons):
image

Multisample after:

image

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -826,7 +826,7 @@ skipDoingSumTo100:
 			}
 
 			// Otherwise...
-			else {
+			else [[unlikely]] {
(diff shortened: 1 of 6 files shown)


nikodemus:
I assume the mispredictions from [[unlikely]] are noise compared to the real costs of the unhappy paths?

m-m-adams:
Yeah I've attempted to only put them in places where the unlikely path is fast enough to not need caching. Like calculating probability and iterations is common but only changes the next branch if the note ends up not sounding, in which case it's fast anyway

PR #2200, opened 2024-06-28
Performance/improve render caching, fix unitialized variable use

m-m-adams:
Significant improvement in voices by tagging likely/unlikely branches. Can likely improve a lot more by caching some repeated checks. 

Also removes rendering directly to the output buffer - the extra time to check and set up for that is worse than an extra memcpy

Fixes a bug where stereo wavefolding used the non-anti aliased fold function

same before as here - #2199

Synth after:
image

Multisample after:

image

### DIFF
--- a/src/deluge/model/voice/voice.cpp
+++ b/src/deluge/model/voice/voice.cpp
@@ -129,7 +129,7 @@ bool Voice::noteOn(ModelStackWithVoice* modelStack, int32_t newNoteCodeBeforeArp
 		sourceValues[util::to_underlying(PatchSource::NOTE)] = 2147483647;
 	}
 	else if (newNoteCodeAfterArpeggiation <= 0) {
-		sourceValues[util::to_underlying(PatchSource::NOTE)] = -2147483648u;
+		sourceValues[util::to_underlying(PatchSource::NOTE)] = -2147483648;
 	}
 	else {
 		sourceValues[util::to_underlying(PatchSource::NOTE)] = ((int32_t)newNoteCodeAfterArpeggiation - 64) * 33554432;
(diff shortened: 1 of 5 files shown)


src/deluge/model/voice/voice.cpp
```diff
@@ -1090,49 +1092,7 @@ skipAutoRelease: {}
...
-			// If still here, no mismatch, so go for it
-			renderingDirectlyIntoSoundBuffer = true;
-		}
-		else {
-			// If got here, we're rendering in mono
-			renderingDirectlyIntoSoundBuffer = (soundRenderingInStereo == false);
-		}
-	}
-decidedWhichBufferRenderingInto:
+	bool renderingDirectlyIntoSoundBuffer = false;
```
sapphire-arches, line 1095: I think this can be deleted? The corresponding downstream code is gone now.

src/deluge/model/voice/voice.cpp
```diff
@@ -1550,7 +1486,7 @@ skipUnisonPart: {}
 			int32_t* const oscBufferEnd = oscBuffer + (numSamples << 1);
 			// fold
 			if (paramFinalValues[params::LOCAL_FOLD] > 0) {
-				dsp::foldBuffer(oscBuffer, oscBufferEnd, paramFinalValues[params::LOCAL_FOLD]);
+				dsp::foldBufferPolyApproximation(oscBuffer, oscBufferEnd, paramFinalValues[params::LOCAL_FOLD]);
```
sapphire-arches, line 1489: I don't see this change mentioned in the PR description, what impact does it have?

m-m-adams: Oops - it uses the non-aliasing fold for stereo rendering. This render path got missed when I added it originally

src/deluge/model/voice/voice.cpp
m-m-adams: I think it's still in use, it gets passed into some of the other functions

m-m-adams: I could make it constexpr?

m-m-adams: I got it all removed, it's used less than I thought

PR #2214, opened 2024-07-02
refactor: combine modeNotes and rootNote into a MusicalKey-class

nikodemus:
- MusicalKey is mostly a POD class for now, but intent is to pull most root & scale
  logic from song.cpp here, as long as it can be disentangled from code dealing with
  song and clip structures.


soymonitus:
"Key" seems a pretty generic term for referring to the current scale. Could it me renamed to something like MusicalKey or something like that? wdyt?

PR #2217, opened 2024-07-03
[DRAFT] Feature/trace actions

m-m-adams:


m-m-adams:
this is too hacky, I'll think of ways to do this less hackily

PR #2219, opened 2024-07-03
define getName() for UI subclasses and add opt debug to matrixdriver

mnantel:
definitions_cxx.hpp -> #define MATRIX_DRIVER_DEBUG enables additional debug output
names are the file name of the class, usually the same as the class name and the UI mode name, but lowercase
debug prints out getName from UI subclass, pad x/y, velocity

the purpose of this change is to facilitate learning the code workflow


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3470,6 +3470,11 @@ void InstrumentClipView::auditionPadAction(int32_t velocity, int32_t yDisplay, b
 		actionLogger.closeAction(ActionType::NOTE_NUDGE);
 	}
 
+	if (shiftButtonDown && velocity) {
+		D_PRINT("AUTOSAMPLE", velocity);
+		display->displayPopup("AUTOSAMPLE");
+		return;
+	}
```
sapphire-arches, line 3477: Leftover from some earlier change?

src/definitions_cxx.hpp
```diff
@@ -27,6 +27,8 @@
 
 #define AUTOMATED_TESTER_ENABLED (0 && ALPHA_OR_BETA_VERSION)
 
+#define MATRIX_DRIVER_DEBUG 0
+
```
sapphire-arches, line 31: This could be plumbed in to the build system - look for how `ENABLE_SYSEX_LOAD` works for an example.

mnantel: Removed the gremlin. I can see about the build system element.

mnantel:
Ok, I think I am good with this PR. Lesson learnt: keep those as drafts until I am done :) Sorry about the spam and numerous commits. Live and learn!

m-m-adams:
This looks good to me. It takes up about 800 bytes for the extra strings but I think removing unused strings in release builds is probably a more general optimization we should look into

m-m-adams (approved):

PR #2221, opened 2024-07-03
Bugfix (c1.1.1) - save load midi preset in keyboard view

seangoodvibes:
Fixed bug where you couldnt save or load a midi preset while in keyboard view

### DIFF
--- a/src/deluge/gui/ui/keyboard/keyboard_screen.cpp
+++ b/src/deluge/gui/ui/keyboard/keyboard_screen.cpp
@@ -535,8 +535,8 @@ ActionResult KeyboardScreen::buttonAction(deluge::hid::Button b, bool on, bool i
 		}
 	}
 
-	else if (b == MIDI) {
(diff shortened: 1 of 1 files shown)


src/deluge/gui/ui/keyboard/keyboard_screen.cpp
```diff
@@ -535,8 +535,8 @@ ActionResult KeyboardScreen::buttonAction(deluge::hid::Button b, bool on, bool i
 		}
 	}
 
-	else if (b == MIDI) {
-		if (on && currentUIMode == UI_MODE_NONE) {
+	else if (b == MIDI && currentUIMode == UI_MODE_NONE) {
```
m-m-adams, line 538: Should probably preemptively do this for cv too

seangoodvibes: Done!

PR #2223, opened 2024-07-03
[DRAFT] Add bulk logging of function invocations. misc DX

tastycode:
leaving in draft while I get back up to speed the process for proposing changes. Most of this I wrote 6 months ago, but it was n't relevant enough to risk bundling it in with some other changes. 

This is 2 additional DBT tasks along with supporting documentation changes and an additional macro. Nothing user facing.

1. Allow devs to bulk annotate entire directory-trees of function calls with logging calls that record each invocation of a function into the debug log (without arguments)
2. a convenience script for parsing deluge crash reporter output.

while experiencing a bout of ADHD, I added a macro `D_POPUP` which is a convenient `sprintf` enabled dev shortcut.
- Ensures the popup only exists in dev builds
- Allows sprintf-style invocation e.g. `D_POPUP("%d bottles of beer on the wall", 99)` 
- Always logs to the debug log
- Does not try to display log messages longer than 100 char
- Falls back in case `display->displayPopup` is not available.

### DIFF
--- a/scripts/tasks/task-annotate.py
+++ b/scripts/tasks/task-annotate.py
@@ -1,0 +1,199 @@
+import os
+import clang.cindex
+import argparse
+import subprocess
+from iterfzf import iterfzf
+
+
+FN_BLACKLIST = ["graphicsRoutine"]
+tags_file = "tags"  # Replace with the actual path to your tags file
+
+
+def argparser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="annotate", description="Adds print statements everywhere"
+    )
+    parser.add_argument("action", choices=["add", "rm"], help="The action to perform")
+    parser.add_argument("-p", "--path", help="the directory to process")
+    parser.add_argument(
+        "-c",
+        "--comment",
+        help="A comment added to the annotation to help with search/removal",
+        default="DBT:ANNOTATE",
+    )
+    parser.add_argument("-m", "--max", help="Maximum number of files to annotate")
+    parser.add_argument(
+        "-t",
+        "--template",
+        help="A template for the log statement",
+        default='D_PRINTLN("{function}");',
+    )
+
+    return parser
+
+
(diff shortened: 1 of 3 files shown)


scripts/tasks/task-annotate.py
```diff
@@ -0,0 +1,165 @@
...
+                    "src/deluge/model/consequence",
+                    "src/deluge/model/drum",
+                    "src/deluge/model/instrument"]
+    print(f"invoking {' '.join(ctags_invocation)}")
+    result = subprocess.run(ctags_invocation, capture_output=True)
+    if result.returncode == 0:
+        with open(tags_file, "w+") as tags_file_pointer:
+            tags_file_pointer.write(result.stdout.decode('utf-8'))
+    else:
+        print(f"Failed to invoke {" ".join(ctags_invocation)}")
```
sapphire-arches, line 88:
```suggestion
        print(f"Failed to invoke {' '.join(ctags_invocation)}")
```
Looks like the python format check picked up an actual bug in this format string.

scripts/toolchain/requirements.txt
```diff
@@ -2,4 +2,6 @@ certifi==2024.6.2
 setuptools==70.1.1
 pyserial==3.5
 ansi==0.3.7
-python-rtmidi==1.5.8
\ No newline at end of file
+python-rtmidi==1.5.8
+iterfzf
+libclang
```
sapphire-arches, line 7: If we're going to add these dependencies they should be version locked.

sapphire-arches (commented):
You'll need to use a `black` compatible with the workflow version (currently 24.4.2) to get the Python format check to pass - it looks like it picked up an actual bug in the format strings though.

I like libclang/libtooling as a basis for some stuff, though I'm a little concerned it'll add like 300 megabytes to the toolchain build next time we run it. Do you have an idea of the size it adds to the site-packages when installed?

PR #2240, opened 2024-07-05
refactor: move preset scales from lookuptables.h|cpp

nikodemus:
- define them using definition-expansion styleas well, so presetScale* arrays and NUM_PRESET_SCALES stay in sync automatically.


src/deluge/model/scale/preset_scales.cpp
```diff
@@ -0,0 +1,14 @@
+#include "model/scale/preset_scales.h"
+
+std::array<char const*, NUM_PRESET_SCALES> presetScaleNames = {
+#define DEF(id, name, notes) name,
+    DEF_SCALES()
+#undef DEF
+};
+
+// Note: Value of note intervals taken from here: https://www.apassion4jazz.net/scales2.html
```
sapphire-arches, line 9: this note should move to the definition of `DEF_SCALES`

src/deluge/model/scale/preset_scales.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * The actual definitions expand this macro, picking the thing they need:
+ * - enum PresetScales
+ * - presetScaleNames
+ * - presetScaleNotes
+ *
+ * Use by defining DEF(id, name, notes) before invoking DEF_SCALES() and undefining it
+ * after.
+ *
+ * DEF_NOTES manages the notes array intact on the trip through the preprocessor, since
```
sapphire-arches, line 21: keeps the notes array intact?

PR #2244, opened 2024-07-05
refactoring: iterators replacing almost all traverseClips-gotos

nikodemus:
- AllClips, InstrumentClips, and Audioclips -classes are iterators across different sets of clips. Their instances are built using static ::everywhere(), ::inSession(), and ::inArrangementOnly() methods. Examples:

      for (Clip* clip : AllClips::inSession(currentSong)) { ... }

      for (InstrumentClip* clip : InstrumentClips::everywhere(currentSong)) { ... }

- This does away with the most gotos and casts - or rather, moves the casts to the iterators.

- Some gotos remain: in Output-related functions there's an entirely different logic for iterating over "all clips in session and arranger for specific output", that uses sound->clipInstances for arrangement-only clips. The reason isn't obvious, and neither is it clear if that code can be directly replaced with the new iterators.

- Tests are probably the most questionable part here, since there's a minor #ifdef leakage into the application code, in order to avoid pulling Song, Clip, etc as dependencies.

- Codesize 1706264 on release build is a bit disappointing vs tip of community at 1705776 -- this was intended to be neutral or positive change, but maybe not too bad considering the improved clarity.

### DIFF
--- a/src/deluge/model/song/clip_iterators.h
+++ b/src/deluge/model/song/clip_iterators.h
@@ -1,0 +1,94 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+#include "definitions_cxx.hpp"
+
+#ifdef IN_UNIT_TESTS
+#include "clip_iterator_mocks.h"
+#else
+#include "deluge.h"
+#include "model/clip/clip.h"
+#include "model/clip/clip_array.h"
+#include "model/song/song.h"
+#endif
+
+// This is an incomplete iterator, missing operator*() and operator->(): the templated iterators inherit
+// from this and implement just those for minimal template bloat.
+class ClipIteratorBase {
+public:
+	__attribute__((noinline))
+	ClipIteratorBase(ClipArray* array_, uint32_t index_, ClipArray* nextArray_, std::optional<ClipType> clipType_);
+	__attribute__((noinline)) ClipIteratorBase& operator++();
+	__attribute__((noinline)) ClipIteratorBase operator++(int);
+	/** Deletes the clip pointed to by the iterator and advances it. */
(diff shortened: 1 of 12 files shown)


sapphire-arches:
could the iterators yield references instead of pointers? that might be best left to another PR depending on the diff size though...

nikodemus:
> could the iterators yield references instead of pointers? that might be best left to another PR depending on the diff size though...

I would make all the sites where this is used change much more, so probably better later. This should be much easier to review for equivalent behaviour than if that's combined.

src/deluge/model/song/clip_iterators.h
```diff
@@ -0,0 +1,94 @@
...
+	__attribute__((noinline))
+	ClipIteratorBase(ClipArray* array_, uint32_t index_, ClipArray* nextArray_, std::optional<ClipType> clipType_);
+	__attribute__((noinline)) ClipIteratorBase& operator++();
+	__attribute__((noinline)) ClipIteratorBase operator++(int);
+	/** Deletes the clip pointed to by the iterator and advances it. */
+	__attribute__((noinline)) void deleteClip(InstrumentRemoval instrumentRemoval);
+	__attribute__((noinline)) friend bool operator==(const ClipIteratorBase& a, const ClipIteratorBase& b) {
+		return (a.array == b.array) && (a.index == b.index) && (a.clipType == b.clipType);
+	}
+	__attribute__((noinline)) friend bool operator!=(const ClipIteratorBase& a, const ClipIteratorBase& b) {
```
sapphire-arches, line 30: We've generally been using the official C++ attribute syntax, so these should be `[[gnu::noinline]]` rather than `__attribute__`

PR #2248, opened 2024-07-06
Bugfix/dont count stats if the audio routine doesn't run

### LINKED ISSUE #2247: Trigger Clock Jitter in Community Editions

There is clock jitter in the community edition that can cause noticeable BPM instability in downstream devices. This jitter is not present on the official firmware versions.

---

I'm seeing pretty significant jitter with a Deluge's 24 PPQN clock being detected by my Pam's as rapidly shifting between 118-121 BPM from a target 120 BPM.

This 3 BPM variation would be equivalent to a wow & flutter of 2.5%, well above the 0.2-0.25% threshold often said to be the audible minimum. This of course obliterates tempo-synced delays and makes recordings effectively impossible to align with other takes or layers.

---

What's interesting though is that if I have a set of 128th notes playing in a synth clip with arpeggiator enabled, the BPM seems to stabilize. This stabilizing effect is noticeable up until around 8th notes. I'm sure the jitter is still there, but the synth is forcing syncs often enough that the Pam's can smooth it out.

I sometimes noticed the clock locking in with CV clips, but it was not as reliable. Drum kits seem to have no effect even at very fast trigger rates.

Either way, if the clip is not actively playing notes then the stabilizing effect is absent.

---

Created per https://github.com/SynthstromAudible/DelugeFirmware/pull/2145#issuecomment-2211442859

---

Jitter may also be present on MIDI, but is less noticeable.

Clock jitter should ideally be less 0.2%.

_No response_

OLED

Release 1.1/Beethoven

cedf24a

1. Enable trigger clock
2. Connect gate output to desired downstream clock with BPM readout
3. Set both to 24 PPQN (other values also show this effect, but 24 may make it more obvious)
4. Press `play` on the Deluge
5. Observe inconsistent BPM

m-m-adams:
Should fix #2247

Fixes a bug where calling the audio routine while it was locked out still "counted" for call frequency, leading to occasional drops in audio call frequency at low cpu load

Before:
image

After:
image

### DIFF
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -953,6 +953,7 @@ void routine() {
 
 	if (audioRoutineLocked) {
 		logAction("AudioDriver::routine locked");
+        ignoreForStats();
 		return; // Prevents this from being called again from inside any e.g. memory allocation routines that get
 		        // called from within this!
 	}
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -953,7 +953,7 @@ void routine() {
 
 	if (audioRoutineLocked) {
 		logAction("AudioDriver::routine locked");
-        ignoreForStats();
+		ignoreForStats();
 		return; // Prevents this from being called again from inside any e.g. memory allocation routines that get
 		        // called from within this!
 	}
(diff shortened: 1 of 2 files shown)


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -953,6 +953,7 @@ void routine() {
 
 	if (audioRoutineLocked) {
```
sapphire-arches, line 954: could this check become a run condition using the logic added for song loading instead?

m-m-adams: Not yet, when the CPU is busy run conditions aren't checked often enough for the audio routine since checking them is the lowest possible priority. 

Something similar should be doable but not possible yet

PR #2249, opened 2024-07-06
refactor: audio clip attack value scaling

nikodemus:
- fix wrong value in computeFinalValueForArpMidiCvRate() comment

- math is same as in with arp::midi_cv::Gate, rename the shared logic to compute(Current|Final)ValueForSemiStandardMenuItem().

- move the thin wrappers to be inline functions

- delete dead code: computeFinalValueForCompParam()

### DIFF
--- a/src/deluge/gui/menu_item/value_scaling.cpp
+++ b/src/deluge/gui/menu_item/value_scaling.cpp
@@ -16,16 +16,13 @@ int32_t computeCurrentValueForPan(int32_t value) {
 	return ((int64_t)value * (kMaxMenuRelativeValue * 2) + 2147483648) >> 32;
 }
 
-int32_t computeCurrentValueForArpMidiCvGate(int32_t value) {
-	return computeCurrentValueForStandardMenuItem(value);
-}
-
 int32_t computeCurrentValueForArpMidiCvRatchetsOrRhythm(uint32_t value) {
 	return ((int64_t)value * kMaxMenuValue + 2147483648) >> 32;
 }
 
-int32_t computeCurrentValueForArpMidiCvRate(int32_t value) {
-	return computeCurrentValueForStandardMenuItem(value);
+int32_t computeFinalValueForSemiStandardMenuItem(int32_t value) {
+	// (2147483648 / kMidMenuValue) == 85899345
+	return (uint32_t)value * 85899345 - 2147483648;
 }
 
 int32_t computeFinalValueForStandardMenuItem(int32_t value) {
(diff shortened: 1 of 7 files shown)


src/deluge/gui/menu_item/value_scaling.h
```diff
@@ -72,6 +74,31 @@ int32_t computeCurrentValueForStandardMenuItem(int32_t value);
...
+inline int32_t computeCurrentValueForSemiStandardMenuItem(int32_t value) {
+	return computeCurrentValueForStandardMenuItem(value);
+}
+
+/** Scales 0-50 range to INT32_MIN-(INT32_MAX-45) for storage and use.
+ *
+ * It is unclear when this is used intentionally, and when by accident.
+ *
+ * - arpeggiator::midi_cv::Gate: possibly so that gate
+ *   will momentarily go down even at 50: the values produced create a 2.5ms
```
m-m-adams, line 93: this is enforced by the CV engine so it shouldn't matter. I'm inclined to think this is always an accident

nikodemus: I'll update the comment, and leave the "how can we make things more consistent" for later.

m-m-adams: Worth testing but I'd be surprised if it even changes sounds. It's a 2e-8% difference

src/deluge/gui/menu_item/value_scaling.cpp
```diff
@@ -16,16 +16,13 @@ int32_t computeCurrentValueForPan(int32_t value) {
...
-	return computeCurrentValueForStandardMenuItem(value);
-}
-
 int32_t computeCurrentValueForArpMidiCvRatchetsOrRhythm(uint32_t value) {
 	return ((int64_t)value * kMaxMenuValue + 2147483648) >> 32;
 }
 
-int32_t computeCurrentValueForArpMidiCvRate(int32_t value) {
-	return computeCurrentValueForStandardMenuItem(value);
+int32_t computeFinalValueForSemiStandardMenuItem(int32_t value) {
```
m-m-adams, line 23: For clarity - computeFinalValueWithoutRoundToOne?

That explains what it does better than just being semi standard

nikodemus: Done, though that naming scheme has the minor issue of not being symmetrical:

`computeCurrentValueWithoutRoundingToOne()`
`computeFinalValueWithoutRoundingToOne()`

Unless the reader has fully internalized Q31 usage it's really not obvious what rounding it's speaking about. 😅 ...but since I don't expect these functions will live long after the complete value scaling scheme has been extracted for examination, I don't think that's an issue. :)

src/deluge/gui/menu_item/value_scaling.h
nikodemus: Yeah, no measurable difference for gate.

m-m-adams (approved):

PR #2261, opened 2024-07-07
tasks: infrastructure for API-specific matchers

sapphire-arches:
All the APIs expose port names differently, so we'll need different regexes for all the rtmidi platforms.

This is kind an awful hack, maybe the utils file should be split in to ones that depend on rtmidi and those that don't.

### DIFF
--- a/scripts/tasks/util.py
+++ b/scripts/tasks/util.py
@@ -203,11 +203,14 @@ def get_dbt_version():
 
 def ensure_midi_port(type, midi, port):
     if port is None:
-        r = re.compile("MIDI(OUT|IN)3 \\(Deluge\\)")
+        deluge_ports = []
         for i, p in enumerate(midi.get_ports()):
-            if r.match(str(p)):
-                port = i
-                break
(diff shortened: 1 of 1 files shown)


sapphire-arches:
The `deluge` and the `3` are in different orders on your original platform (windows?) and at least linux/alsa

sapphire-arches:
that might still be better though, rather than having dedicated ones

nikodemus:
Two regexes? One checking for "deluge" and another for "3".

sapphire-arches:
that might give a false positive if the device index is present in the string representation on some platform but would work for the two we have so far at least.

nikodemus:
Actually, counting and picking the third one labeled "deluge" might be pretty close to exact, and robust to all kinds of strange schemes?

sapphire-arches:
yeah, seems reasonable. That's what the latest version implements.

m-m-adams (approved):
I'm using a network midi interface and deluge 3 is first. On the other hand I have an in depth enough midi setup that I'm aware of that and this probably targets the average use case better

PR #2263, opened 2024-07-07
refactor: split SessionView::SelectEncoderAction() into parts

nikodemus:
- each distinct action has its own function, with action_ prefix

- use switch to dispatch on the UI mode for more clarity

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -1161,7 +1161,7 @@ void SessionView::drawSectionRepeatNumber() {
 	}
 }
 
-void SessionView::action_changeSectionRepeats(int8_t offset) {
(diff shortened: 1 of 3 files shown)


sapphire-arches:
How do you feel about using `handle` as the prefix, rather than `action_`? The latter would be a new style of function name which kinda feels bad.

nikodemus:
> How do you feel about using `handle` as the prefix, rather than `action_`? The latter would be a new style of function name which kinda feels bad.

I'm not religious about it, but I do have a reason for the new style: these are explicitly actions (or commands) triggered from the HID, with no "what should I do now" logic left the specific encoder value.

They're not handling something, they're _doing_ something. How about `command` (without _) as an alternative?

sapphire-arches:
I like that, thanks!

PR #2264, opened 2024-07-07
swing interval shortcut

nikodemus:
- Hold TAP TEMPO and turn TEMPO to adjust the current swing interval.
- Hold TAP TEMPO and press TEMPO to show the current swing interval.
- On OLED, both swing interval and swing amount are always displayed together, with a two-row popup. On 7seg only the main value for each action is shown.
- Swing Interval menu moved from SETTINGS to SONG.
- Explicit NUM_SWING_INTERVALS value instead of SYNC_TYPE_TRIPLET, which is a bit too magical.
- new utility function mod(): like %, but always non-negative
- Extract sync level and type logic from SyncLevel menu item to more reusable location, add tests.
- For both swing amount and interval when using the tempo encoder, on first encoder click display current value instead of changing it
- Splits out some of the new test mocks into more reusable pieces.
- DEFAULTS > SWING AMOUNT, SWING INTERVAL instead of just amount. Default menus for Swing amount and interval both say "Default swing" at the top, because longer text would not fit, and the menu label      does not scroll. On 7-seg SWIN becomes SWIA, and swing interval is denoted by SWII.


nikodemus:
Why do I always forget to format before submitting?

src/deluge/model/song/song.cpp
```diff
@@ -148,7 +148,7 @@ Song::Song() : backedUpParamManagers(sizeof(BackedUpParamManager)) {
 
 	swingAmount = 0;
 
-	swingInterval = 8 - insideWorldTickMagnitude; // 16th notes
+	swingInterval = FlashStorage::defaultSwingInterval;
```
m-m-adams, line 151: does this work if you change song resolution?

nikodemus: Kinda no, kinda yes.

BUG: Defaults shows the swing interval in relationship to current song magnitude, not the default magnitude. If you shink default magnitude by one the default swing interval will look the same until you actually have a new song with the new magnitude, at which point it will have doubled -- and that will be the new song's initial swing interval. (I have a fix so that default swing interval is based on the default magnitude, so defaults shows the right value.) 

FEATURE OR BUG? With the above bug fixed, changing default magnitude still changes the default swing interval. We could mostly adjust the swing interval automatically when changing magnitude, but if you start with an extreme swing interval and change magnitude in the wrong direction, then we can't. (I don't have this done, but I can do it.)

m-m-adams (approved):

PR #2267, opened 2024-07-08
fix first build after clean

nikodemus:
- Clean nukes the g_english.cpp & friends, so the glob won't see them, The dependency on deluge_strings still generates them, but for them to be built they need to be in place before the source glob. Just moving the add_dependency() earlier is not enoough: they're not generated until the build is running, and by that time we need the sources list already.


sapphire-arches:
does `g_menus.inc` suffer from the same problem?

nikodemus:
> does `g_menus.inc` suffer from the same problem?

Nope, because it's included by a file in our source tree: no separate compilation step required.

Issue #2257, opened 2024-07-06
Performance View - add Bass Frequency and Treble Frequency as assignable params


seangoodvibes:
### Please describe the problem:

You can't assign Bass Frequency and Treble Frequency to the Performance View grid.

### What is the expected behavior?

Make Bass Frequency and Treble Frequency assignable

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

PR #2270, opened 2024-07-10
Add Bass and Treble Freq to Perf View

seangoodvibes:
Added two additional song params (Bass and Treble Frequency) that were missing from the list of assignable params in performance view

I think we should cherry pick to 1.1.1 but I'm ok with not doing that as well

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@
 ### MIDI
 - Added Universal SysEx Identity response, including firmware version.
 
+## c1.1.1 Beethoven
+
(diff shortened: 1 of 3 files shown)


sapphire-arches:
probably needs a changelog entry (new 1.1.1 section too?) otherwise lgtm

seangoodvibes:
> probably needs a changelog entry (new 1.1.1 section too?) otherwise lgtm

Added to new 1.1.1 section!

PR #2280, opened 2024-07-11
Bugfix/make activeClip private, fix associated bugs

m-m-adams:
Makes activeClip private to enforce getter/setter use

Fix bugs where activeClip could be changed without updating the external state

Fix bugs when activeClip is set to null without telling the output

Update all setActiveClip methods to handle null pointers

### DIFF
--- a/src/deluge/processing/audio_output.cpp
+++ b/src/deluge/processing/audio_output.cpp
@@ -394,7 +394,7 @@ bool AudioOutput::wantsToBeginArrangementRecording() {
 
 bool AudioOutput::setActiveClip(ModelStackWithTimelineCounter* modelStack, PgmChangeSend maySendMIDIPGMs) {
 	if (activeClip
-	    && (activeClip != modelStack->getTimelineCounter()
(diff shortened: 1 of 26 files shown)


src/deluge/processing/audio_output.cpp
```diff
@@ -394,7 +394,7 @@ bool AudioOutput::wantsToBeginArrangementRecording() {
 
 bool AudioOutput::setActiveClip(ModelStackWithTimelineCounter* modelStack, PgmChangeSend maySendMIDIPGMs) {
 	if (activeClip
-	    && (activeClip != modelStack->getTimelineCounter()
+	    && (!modelStack || activeClip != modelStack->getTimelineCounter()
```
seangoodvibes, line 397: should this be getActiveClip()?

m-m-adams: Nah this is an output, it owns the activeClip

seangoodvibes (approved):

PR #2281, opened 2024-07-11
delete dead code: NowRow::noteOn() and noteOff()

nikodemus:
- declarations only, definitions don't exist


nikodemus:
I'll add more small cleanups in one go.

PR #2293, opened 2024-07-14
Clip Names

KrisGelero:
Clip Names support for the Deluge

- Name KIT & Synth Clips to select their names on the arrangement or grid view.
- Drum KIT can be set for clip name (affect entire) or per note row changing the name of sample loaded. 
- Auto clip name for AUDIO Clips. (extract the filepath of the selected sample).

Tried to make the codebase as small as possible and tested with different scenarios. It's my first C++, please let me know if there are errors, so I can learn from it. 

Many thanks!


KrisGelero:
The code has a serious bug that I need to resolve :) Will dive into it.

src/deluge/gui/views/arranger_view.cpp
```diff
@@ -3290,3 +3309,40 @@ void ArrangerView::requestRendering(UI* ui, uint32_t whichMainRows, uint32_t whi
...
+
+	int32_t yPos = OLED_MAIN_TOPMOST_PIXEL + 3;
+	int32_t textSpacingX = kTextTitleSpacingX;
+	int32_t textSpacingY = kTextTitleSizeY;
+
+	canvas.drawStringCentred("ARRANGER", yPos, kTextSpacingX, kTextSpacingY);
+
+	if (output->type == OutputType::AUDIO) {
+		AudioClip* audioclip = (AudioClip*)clip;
+		clip->clipName.set(&audioclip->sampleHolder.filePath);
```
m-m-adams, line 3325: this shouldn't be handled by the view, it should be set when the audio clip's file path is set or recording is finalized

src/deluge/gui/views/arranger_view.cpp
```diff
@@ -1538,6 +1543,12 @@ void ArrangerView::editPadAction(int32_t x, int32_t y, bool on) {
 
 				if (x == xPressed && y == yPressedEffective) {
 
+					hid::display::OLED::clearMainImage();
+					ClipInstance* clipInstance = output->clipInstances.getElement(pressedClipInstanceIndex);
+					drawClipName(output, clipInstance->clip);
```
m-m-adams, line 1548: should we be drawing clips on release? It's no longer selected at this point

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -916,8 +917,32 @@ ActionResult SoundEditor::potentialShortcutPadAction(int32_t x, int32_t y, bool
...
+
+					if (clip) {
+						renameClipNameUI.clip = clip;
+						openUI(&renameClipNameUI);
+						return ActionResult::DEALT_WITH;
+					}
+				}
+			}
+			// if else then just keep it like the old way to rename the kit row item
+			goto doNextAfterName;
```
m-m-adams, line 941: this is in-extensible (nobody can add another condition to the if else chain without breaking the goto)

Could you put all the conditions into the if statement? Split this block out to a function that returns true if the editing took place and false otherwise so it continues down the chain. Then just return dealt with if the condition is met

m-m-adams: I also don't see why audio clips are excluded

src/deluge/gui/views/arranger_view.cpp
KrisGelero: I tested some scenarios, this was the best to still view the latest selected clip. Otherwise, on the arranger its still difficult to see what's the last clip been added or changed.

src/deluge/gui/views/arranger_view.cpp
KrisGelero: Going to change this part and look where the entry point is ( maybe on a few different places )

src/deluge/gui/ui/sound_editor.cpp
KrisGelero: Agreed, was a bit of a dirty patch here with the label. Will warp it in a function.

m-m-adams (approved):

PR #2294, opened 2024-07-15
Update pad selection mode instructions in automation_view.md

baymud:
The instructions for entering pad selection mode were out of date. The image of button shortcuts still needs edited though


baymud:
The doc link was wrong, so this is no longer relevant

PR #2299, opened 2024-07-15
Add New Clip Type Menu and Remove Yellow Grid Mode

seangoodvibes:
## Description

Added new clip type context menu that opens on creation of a new clip in song view which enables you to select the type of clip before the clip is created, removing the need to create a synth clip and then change its type afterwards.

This frees up the press clip and press select encoder shortcut to be used to selecting the clips launch style.

Removed grid view yellow mode (from https://github.com/SynthstromAudible/DelugeFirmware/pull/1494) because you can now select clip launch style in all grid view modes.

## Usage: 

Add new clip by pressing an empty row or adding a clip to an empty column in grid view.

A menu will open enabling you to select the type of clip that you wish to create. The options are:

1. Audio clip
2. Synth clip
3. Kit clip
4. MIDI clip
5. CV clip

Select the type by scrolling to it in the menu and pressing the select encoder to create a clip with that type.

Alternatively, while the menu is open, you can also select Synth/Kit/MIDI/CV by pressing the corresponding deluge buttons.

If you exit the menu (by pressing back or pressing the grid) then no clip will be created.

### DIFF
--- a/docs/community_features.md
+++ b/docs/community_features.md
@@ -291,6 +291,15 @@ as the micromonsta and the dreadbox nymphes.
  - Submenu's on OLED for automatable parameters (e.g. LPF Frequency) render the current parameter value at the end. You still need to click on `SELECT` to edit the parameters value / edit modulation depth and patch cables.
  - All other submenu's on OLED are rendered with a ">" at the end to indicate that it is a submenu.
 
+#### 3.27 Updated UI for Creating New Clips in New Tracks in Song View
+- ([#2299]) Added `NEW CLIP TYPE` menu that opens on creation of a `NEW CLIP` in `SONG VIEW` which enables you to select the type of clip before the clip is created.
+  - In `SONG (GRID) VIEW`, this menu will only appear when you add a `NEW CLIP` to a `NEW TRACK` (empty column).
+  - This menu will not appear if you are cloning clips (e.g. pressing one clip and then pressing an empty pad).
+- Removed the `SONG VIEW` shortcut of Pressing `CLIP` + `SELECT` to convert an Empty `INSTRUMENT CLIP` to an `AUDIO CLIP`. 
+
+#### 3.28 Updated UI for Setting Clip Launch Style in Song View
+- ([#2299]) Pressing `CLIP` + `SELECT` in `SONG VIEW` will now always open the `LAUNCH STYLE` menu so you can change the Clip launch style between `DEFAULT`, `FILL` and `ONCE`.
+
 ## 4. New Features Added
 
 Here is a list of features that have been added to the firmware as a list, grouped by category:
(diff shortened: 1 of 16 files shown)


docs/community_features.md
```diff
@@ -330,8 +330,7 @@ Here is a list of features that have been added to the firmware as a list, group
 #### 4.1.3 - Fill Clips and Once Clips
 
 - ([#196] and [#1018]) In Song View (Rows mode), holding the status pad (mute pad) for a clip and pressing `SELECT`
-  brings up a clip type selection menu. For Grid Mode there is a dedicated Config Mode in which this menu can be
-  reached, see `Yellow Mode` in [Grid View](#415grid-view).
+  brings up a clip type selection menu. For Grid Mode, hold a clip and press `SELECT` to enter a `CLIP SETTINGS` menu where you can convert a clip to an audio clip and set the clip's launch style.
```
trappar, line 333: Is this true at this point? I can't find the "convert a clip to an audio clip" option. I just see the launch style options.

seangoodvibes: > Is this true at this point? I can't find the "convert a clip to an audio clip" option. I just see the launch style options.

Thanks that was outdated. I've updated the documentation now!

m-m-adams (approved):

m-m-adams (approved):

schlesstones:
Hi, this is just one person's opinion, but I've used this UI change for a few days and I find it very cumbersome for my workflow, mostly because we have removed the ability to convert a non-audio clip to an audio clip:

- I often start a project with an empty audio clip for looping (using auto-tempo), which means the audio clip has to be the only clip in the song.  So, while I used to just press and hold the clip and then press select, I now have to add an audio clip and then delete the synth clip because there is no longer a way to convert it to an audio clip. 
- Because there is no dedicated "Audio" button or gesture, I must use the scroll menu to create an audio clip which pulls me out of the "flow" of audio creation.  It sounds silly but avoiding menus while making music is one of the things I love the most about Deluge.

I do understand wanting to make the UI consistent for "press clip and press select encoder", so I have a suggestion:
- Could SHIFT->CV be the shortcut to turn a non-audio clip to an audio clip?
- For those of us who don't use CV at all, could there be a community menu item to make the CV button a dedicated "audio button"?

Anyway thanks for considering!  Let me know if there is an alternate place I should be placing beta test feedback for community UI features.  Thanks!

m-m-adams:
Hey we're looking at adding back the audio conversion option (will be a popup menu)

That said I'll offer two suggestions 

1. You can create a default song DEFAULT.xml and enable the startup song feature. That will let you create a new default with an audio clip already created in new songs
2. The popup menu for creating a new clip can be totally ignored, the exact same hold empty row and press select action will still make an audio clip. Similarly holding the row and pressing synth/kit/midi/cv buttons immediately creates the appropriate clip

There's some more discussion on github but someone might move forward with repurposing the cv button for audio clips and relegating cv clips to a menu

schlesstones:
Hi thanks for responding!  This all makes sense.

Put me down as a strong vote for repurposing the CV button for audio tracks!  :-)

todd-gochenour:
I just tested this feature.   I was hoping that the clip name would replace the current preset name shown in large type on the UI screen.  Especially when I've created a new synth 'Synt1' and give it a name, I see:

         Synth
        SYNT1
1: My Named Clip

centered in the UI leaving no space for anything else.   I would like to see the key and scale information here on the third line, as that information is more important when interacting with the clip than the name 'Synt1' or 'Rich Saw Bass'.

I would like to see:

1/16      Synth       # 120
 1: My Named Clip
C-2 Harmonic Minor

where the upper left corner reveals the current zoom level, the upper right corner reports BPM, the lower left corner reveals the key and scale and the lower right corner is free for something else.

PR #2306, opened 2024-07-18
new dbt command: midi-clock

nikodemus:
- Listens to MIDI clock from Deluge, prints out the time intervals, or saves them to file.


nikodemus:
Now that the clock issue solution is known not sure if this is really needed, though.

m-m-adams (approved):

PR #2311, opened 2024-07-19
[DRAFT] global filters for synths, ie. not per voice

nikodemus:
- SoundInstrument has a globalEffectable, Sound::render() accepts one, and uses it to filter post-fx if it's provided

- Sound menu has a new entry: Global filters, right under Filters. (As side effect, it's now SONG > GLOBAL FILTERS as well, not SONG > FILTERS. Ehmn.)

- [ ] Didn't really think about how this interacts with kits almost at all yet.

- [ ] FIX PARAMS. Params are currently just plain wrong. UnpatchedSound and UnpatchedGlobal enums overlap, so using eg. grabbing UNPATCHED_LPF_FREQ from sound is actually going to get UNPATCHED_PORTAMENTO. An entirely separate param vector for the SoundInstrument::fx would just waste space, since it has entries for things we don't need, like the EQ settings and sidechain shape.
  - LFO1, and SIDECHAIN modulation should be available for global filter, so they're not really unpatched params in the long run.
  - Therefore they should Global instead.

- [ ] Testing, including perf testing. Right now all sounds use the filter, even if it's fully open. That should at least needs to be fixed.

- [ ] Saving as part of SYNTH.

- [ ] documentation

- [ ] Even if more FX from global effectable are added, this feels off, because GlobalEffectable has all these UX-related functions as well, like modEncoderButtonAction. Do I smell a need to split the audio processing from GlobalEffectable/ModControllable into a separate FXProcessor class, maybe?

- [ ] UX design. Where should the menu be? What should it say? GLOBAL LPF and HPF under filters menu for sound? Should they be in the gold knob rotation?

- [ ] STRETCH: because they're _right_there_, second delay and mod-fx as well.

PS! If you try this out, because of the above global LPF setting all sound is gone until you go and open up the filters! Give a try with taking a nice sound, adding bitcrush and flanger, and then play with LPF. :)


nikodemus:

PR #2315, opened 2024-07-20
Render Clip and Section Launch Playhead in Song Grid and Performance Views

seangoodvibes:
A white playhead is now rendered in Song Grid and Performance Views that lets you know when a clip or section launch event is scheduled to occur. The playhead only renders the last 16 notes before a launch event.
 - Note: this playhead can be turned off in the Community Features submenu titled: `Enable Launch Event Playhead (PLAY)`
 
 The display also now shows the number of Bars (or Notes for the last bar) remaining until launch in all Song views (Grid, Row, Performance).


soymonitus:
Was the playhead finally moved to the last row? If no, please, consider this idea even though the playhead shows on top for the keyboard view, as the top row in grid mode is usually full of clips.

Besides, why rendering red on quarters? It makes the sequence more crazy. The playhead already is easy to follow as it renders 16th notes,.. no need for quarters. It would be needed if you were only to render the quarter notes and not the 16th notes

m-m-adams (approved):

seangoodvibes:
> Was the playhead finally moved to the last row? If no, please, consider this idea even though the playhead shows on top for the keyboard view, as the top row in grid mode is usually full of clips.
> 
> Besides, why rendering red on quarters? It makes the sequence more crazy. The playhead already is easy to follow as it renders 16th notes,.. no need for quarters. It would be needed if you were only to render the quarter notes and not the 16th notes

Updated it so it's just a white playhead

As for the "playhead in top row" let's play with it first. maybe we can make it customizable as part of the community feature toggle.

PR #2318, opened 2024-07-21
Bugfix/respect record arming

m-m-adams:
Non audio clips let you toggle the record arming state but didn't do anything with it. This bugfix has them respect their arming state while recording

### DIFF
--- a/src/deluge/modulation/automation/auto_param.cpp
+++ b/src/deluge/modulation/automation/auto_param.cpp
@@ -111,7 +111,8 @@ void AutoParam::setCurrentValueInResponseToUserInput(int32_t value, ModelStackWi
 	if (isPlaying) {
 
 		// If recording...
-		if (playbackHandler.recording != RecordingMode::OFF && modelStack->timelineCounterIsSet()) {
+		if (playbackHandler.recording != RecordingMode::OFF && modelStack->timelineCounterIsSet()
+		    && modelStack->getTimelineCounter()->armedForRecording) {
 
 			// If in record mode and shift button held down, delete automation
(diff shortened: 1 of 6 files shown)


seangoodvibes:
@m-m-adams what's the behaviour this fixes exactly? (e.g. how can I check current behaviour vs fixed behaviour?) also should this be cherry picked? Thx

m-m-adams:
You can arm and disarm clips for recording by holding record and pressing the clip. This worked for audio clips but instrument clips never checked if they were armed and recorded even if they're disarmed

m-m-adams:
I don't think it's ever worked in official so dunno if we need to cherry pick it, I just noticed while working on looping

src/deluge/modulation/automation/auto_param.cpp
```diff
@@ -111,7 +111,8 @@ void AutoParam::setCurrentValueInResponseToUserInput(int32_t value, ModelStackWi
 	if (isPlaying) {
 
 		// If recording...
-		if (playbackHandler.recording != RecordingMode::OFF && modelStack->timelineCounterIsSet()) {
+		if (playbackHandler.recording != RecordingMode::OFF && modelStack->timelineCounterIsSet()
+		    && modelStack->getTimelineCounter()->armedForRecording) {
```
sapphire-arches, line 115: could we downcast to `Clip` here instead of moving the arm state to timeline counter? I'm not sure which is better really

m-m-adams: I think the plus side of this approach is that it applies to song automation too if someone adds a way to toggle it later - then you could tweak away without worrying about record being on/off

PR #2330, opened 2024-07-24
Add stem export config menu for normalization

seangoodvibes:
This is a follow-up to the Stem Export menu PR. It adds a new configuration sub-menu with a single configuration setting (to start) to Enable or Disable the normalization of stems prior to export.

Refactored the recorder functions to accept "allowNormalization" as a flag from the caller to exclude normalization from the exported wav file.


sichtbeton:
this is great!

m-m-adams (approved):

PR #2333, opened 2024-07-26
Allow grouping clips in a section on a single output

m-m-adams:
Current implementation focused on audio clips but the skeleton is there for all clips

The idea is that clip groups are represented as a doubly linked list. When an output is rendered, it renders all the clips in the active group before applying it's clip effects

This significantly reduces the clutter of looping in grid view

TODO:

- [x] In audio clip view, render the layers on the sidebar (like kit rows)
- [x] Clarify UI - only one set of fx is rendered, but all can be adjusted. Done - all linked clips now functionally share a param manager
- [x] Make launching from grid view launch the whole group

also probably some refactoring to come


nikodemus:
Just reading the description, the FX sounds a bit confusing and potentially hard to manage.

Any chance of Kit-like FX where each clip has its own FX + container has global via affect-entire?

m-m-adams:
I've been thinking about that, it would be doable by moving the global effectable from the audio output to the audio clip. They're fairly large but at the moment layers work by giving each clip its own (larger) output so it wouldn't be a net increase

src/deluge/model/clip/clip.h
```diff
@@ -37,8 +37,10 @@ class StorageManager;
 class Serializer;
 class Deserializer;
 
-enum class ClipGroupType { NONE, EXCLUSIVE, SHARED };
-
+// Only shared is implemented, plan is exclusive would act like the official behaviour and select one clip, sequential
+// would play through them one after another
```
nikodemus, line 41: Could this comment describe SHARED, since it's the one implemented? (Or refer elsewhere for the description?)

src/deluge/model/clip/clip.h
```diff
@@ -210,8 +215,43 @@ class Clip : public TimelineCounter {
...
 	// returns the first clip in the group, or the clip itself if there's no group
 	Clip* getHeadOfGroup() { return head; }
-	bool isInGroup() { return groupType == ClipGroupType::SHARED; }
+	bool isInGroup() { return groupType != ClipGroupType::NONE; }
+	// deliberately doesn't check the list - there can only be one per section per output to not break grid view
+	// this allows it to be used to relink clips on load
 	bool isInGroupWith(Clip* clip) { return isInGroup() && clip->output == output && clip->section == section; }
+	ClipGroupType getGroupType() { return groupType; };
+
+	void insertAfter(Clip* newNextNode, ClipGroupType type) {
```
nikodemus: nit: this seems pretty beefy for implicit inline from the header?

src/deluge/model/song/song.cpp
```diff
@@ -2228,6 +2228,17 @@ Error Song::readClipsFromFile(Deserializer& reader, ClipArray* clipArray) {
 				delugeDealloc(memory);
 				return error;
 			}
+			// find and link to its group if needed
+			if (newClip->isInGroup()) {
+				auto c = clipArray->getNumElements();
+				for (c; c >= 0; c--) {
+					auto prevClip = clipArray->getClipAtIndex(c);
```
nikodemus: I feel like I must be misreading this, but isn't this iterating backwards from N, instead of N-1? Ie. isn't the first access out of bounds?

I would have expected

```
for (auto c = clipArray->getNumElements() - 1; c >= 0; c--) {
```

m-m-adams: I think you might have been reviewing a specific commit and not the whole pr - you're correct that this was a bug but I fixed it already haha 


Same for moving the insert function into the cpp file instead of the header

m-m-adams:
Making this draft, likely to close. There's a bunch of weirdness that comes from the clips existing in both the clip group and the session.

I'll probably re approach this later with a new "container clip" type that holds other clips and acts as a mini session itself

PR #2348, opened 2024-07-29
Render more menu's as toggles + Render checkbox and arrow icons

seangoodvibes:
- Some selection class menu's are actually toggles (e.g. most in community features, arp mode).

  This PR enables those selection class menu's to behave like toggles and thus display a checkbox (on OLED) and dot (on 7SEG)

- Replaced ascii checkbox "[ ]" and arrow ">" with a check box and arrow icons.

- Refactored logic for calculating render start x pixel position for rendering icons and param values


m-m-adams:
Some non toggle menus are displaying a dot on 7seg - I've noticed PREV and CURS do it so far but there's probably more

seangoodvibes:
> Some non toggle menus are displaying a dot on 7seg - I've noticed PREV and CURS do it so far but there's probably more

Fixed thx!

m-m-adams (approved):

PR #2365, opened 2024-07-31
learning user scales

nikodemus:
## InstrumentClip::seeWhatNotesWithinOctaveArePresent() cleanup

- pass in the key instead of song and root note
- deleteEmptyNoteRows was unused, delete
- TODO: there's a comment in one place about that: make sure it being unused wasn't just a bug.

## learn + scale learns current notes as custom scale

- extract scale-related commands into methods of InstrumentClipView.
- make AUDITION + SCALE and SCALE + AUDITION both work the same way in and out of scale mode. (Previously AUDITION + SCALE in scale mode was the odd one out.)
- LEARN + SCALE learns notes from current & scale mode clips as a new scale.

## user scale as part of SHIFT + SCALE cycling

- Song::userScale stores user scale (there can be only one)
- scaleId() -> NUM_PRESET_SCALES for user scales
- scale name is always fetched with getScaleName()
- OTHER_SCALE and CUSTOM_SCALE_WITH_MORE_THAN_7_NOTES both replaced with USER_SCALE, or CANT_CHANGE_SCALE as appropriate.
- propagating the correct type for scale id / index a bit furhter, still remains as int32_t farther out.


src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -1495,9 +1491,12 @@ int32_t InstrumentClip::getYNoteFromYVisual(int32_t yVisual, Song* song) {
...
 	NoteSet notesPresent;
 
-	seeWhatNotesWithinOctaveArePresent(
-	    notesPresent, 0, song,
-	    false); // Don't delete anything yet, since we're still going to make use of the noteRowsOnScreen!
+	seeWhatNotesWithinOctaveArePresent(notesPresent, song->key);
+	// TODO: there used to be "deleteUnusedRows" argument, and this passed in false, but that
+	// didn't actually do anything. Check that it wasn't just a bug, and not deleting would make
+	// things right.
+	//
```
sapphire-arches, line 1498: The two users I can see are both in `Song` -- one in `Song::setRootNote` and the other in `Song::setPresetScale`. However, in the implementation, that argument is unused. I don't think this is a bug strictly speaking, but it *is* a potential place to save memory and gain some performance. If we delete the now-empty `NoteRows` those won't be processed for sequencing and we can reuse that memory for something else later. It's probably wise to actually implement that in another PR if we're going to do it.

src/deluge/model/scale/preset_scales.h
```diff
@@ -87,11 +86,13 @@ extern std::array<char const*, NUM_PRESET_SCALES> presetScaleNames;
...
+#define CANT_CHANGE_SCALE 255
 #define PRESET_SCALE_RANDOM 254
 #define PRESET_SCALE_NONE 255
 // These offsets allow us to introduce new 7, 6 and 5 note scales in between the existing
 // ones keeping the decreasing order and without breaking backwards compatibility for
 // defaults stored in flash memory.
 #define OFFSET_6_NOTE_SCALE 64
 #define OFFSET_5_NOTE_SCALE 128
+
+const char* getScaleName(uint8_t scale);
```
sapphire-arches, line 98: As long as we're touching all the places that pass around scale IDs, maybe it's worth `typedef`'ing it in case we want to make it more strongly typed later.

src/deluge/model/clip/instrument_clip.cpp
nikodemus: We already do delete the empty note rows, unconditionally. This wanting to keep them is what led me to wonder.

src/deluge/model/scale/preset_scales.h
nikodemus: No objections to making it less of a magic index, but I think that instead of adding any more preset scales to the firmware new scales should be on the SD card as scala files, which leads scale id / index being an insufficient identifier. Haven't thought this fully through yet.

nikodemus: So that ballooned a bit... Made a separate PR here, https://github.com/SynthstromAudible/DelugeFirmware/pull/2381 but I can update this one as well.

PR #2367, opened 2024-08-01
Tempo automation

m-m-adams:
Adds tempo automation for songs in arranger mode

Not integrated with arranger automation view - @seangoodvibes if you could do that it would be great, I was thinking the sync light could blink in automation view if it's automated?


sapphire-arches (approved):
changelog/community features entry needed, otherwise the code LGTM

PR #2371, opened 2024-08-02
Feature/record from outputs

m-m-adams:
Allows choosing a specific output to record into audio clips

Select TRACK as the source, then there's a menu added for track selection to choose an output


src/deluge/processing/audio_output.h
```diff
@@ -80,8 +80,12 @@ class AudioOutput final : public Output, public GlobalEffectableForClip {
...
 	int32_t overrideAmplitudeEnvelopeReleaseRate;
 
-	AudioInputChannel inputChannel;
-	bool echoing; // Doesn't get cloned - we wouldn't want that!
+	AudioInputChannel inputChannel{AudioInputChannel::UNSET};
+	Output* outputRecordingFrom{nullptr};
+	// only used during loading - index changes as outputs are added/removed and this won't get updated. Pointer stays
+	// accurate through those changes
+	int16_t outputRecordingFromIndex{-1}; // int16 so it fits with the bool and because that should be enough outputs
+	bool echoing;                         // Doesn't get cloned - we wouldn't want that!
```
sapphire-arches, line 88:
```suggestion
	/// Audio channel used for recording and monitoring.
	AudioInputChannel inputChannel{AudioInputChannel::UNSET};
	/// Which output to record from. Only valid when inputChannel is AudioInputChannel::SPECIFIC_OUTPUT.
	Output* outputRecordingFrom{nullptr};
	/// Only used during loading - index changes as outputs are added/removed and this won't get updated. Pointer stays
	/// accurate through those changes.
	///
	/// int16 so it packs nicely with `echoing` below
	int16_t outputRecordingFromIndex{-1}; // int16 so it fits with the bool and because that should be enough outputs
	/// When true, this output is monitoring its input.
	///
	/// Does not get copied when this Output is cloned, as that would result in undesirable doubling of the monitored audio.
	bool echoing;
```

src/deluge/storage/audio/audio_file_manager.cpp
```diff
@@ -322,17 +321,33 @@ Error AudioFileManager::getUnusedAudioRecordingFilePath(String* filePath, String
...
+			return error;
+		}
+		highestUsedAudioRecordingNumber[folderID]++;
 	}
-	error = filePath->concatenate(".WAV");
-	if (error != Error::NONE) {
-		return error;
+	else {
+		char namedPath[255]{0};
+		snprintf(namedPath, sizeof(namedPath), "%s/%s/%s.wav", filePath->get(), songName->get(), channelName);
```
sapphire-arches, line 342: we should probably always use the indexed format below, otherwise the first recording will be named with a different format which feels undesirable.

src/deluge/storage/audio/audio_file_manager.cpp
```diff
@@ -322,17 +321,33 @@ Error AudioFileManager::getUnusedAudioRecordingFilePath(String* filePath, String
...
+		int i = 1;
+		while (storageManager.fileExists(namedPath)) {
+			snprintf(namedPath, sizeof(namedPath), "%s/%s/%s_%d.wav", filePath->get(), songName->get(), channelName, i);
```
sapphire-arches, line 345:
```suggestion
			snprintf(namedPath, sizeof(namedPath), "%s/%s/%s_%03d.wav", filePath->get(), songName->get(), channelName, i);
```
Fixed-length format would make alphabetical sorting happier, at least until someone has over 1000 takes for a single instrument.

src/deluge/model/output.h
```diff
@@ -160,9 +161,27 @@ class Output {
...
+		}
+		return false;
+	}
 
 protected:
 	virtual Clip* createNewClipForArrangementRecording(ModelStack* modelStack) = 0;
 
-	Clip* activeClip;
+	Clip* activeClip{nullptr};
+	SampleRecorder* recorder{nullptr};
```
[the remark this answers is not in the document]
m-m-adams: I'm planning to support instrument outputs in a future PR, just not sure of the UI for how to show the layers yet

src/deluge/storage/audio/audio_file_manager.cpp
m-m-adams: Oops this is a different PR - I agree but I'll make the change there

src/deluge/model/output.h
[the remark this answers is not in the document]
m-m-adams: Wait mixed up PRs - this one already supports instruments, it records the rendered audio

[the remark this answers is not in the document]
m-m-adams: Oh misunderstood - you meant MIDI and CV. Yeah I don't see an obvious way to do that

PR #2374, opened 2024-08-02
Bugfix - respect record arming for auditioning

m-m-adams:
Makes auditioning and keyboard view respect record arming properly


sapphire-arches:
this should get a changelog entry, since it could confuse people. otherwise LGTM

m-m-adams:
> this should get a changelog entry, since it could confuse people. otherwise LGTM

It's got one already, this is just a bugfix so it applies to auditioning as well as midi

PR #2376, opened 2024-08-03
add ACTIVE SCALES menus, and some USER scale polish

nikodemus:
This grew a bit more sprawling than I like, so if it's better to split this into smaller chunks, just let me know.

- SONG > ACTIVE SCALES, and DEFAULTS > SCALE > ACTIVE SCALES. USER is always considered enabled. (This is to keep the flash storage requirement for disabled scales to 2 bytes: we have exactly 16 preset scales right now.)
- Toggling a scale off in ACTIVE SCALES removes it from the scale rotation, and from consideration as a random scale.
- Keyboard scale mode allows access to USER scale & respects disabled scales as well.
- Active scales & USER scale saved explicitly as part of song.
- USER scale removed from default scales: we don't save the user scale in flash, so having it as a default doesn't really work, and instead of spending 2 bytes of a scarce resource on it, let's work on support for scales saved on the SD card instead.
- Translation between flash storage codes and scale ids moved to one place.
- Support for fixed dots with scrolling text on 7-segment display.


nikodemus:
(On current community loadfw of this fails due to the UB issue, but rebasing on top of eg 8e43e8809d730cef7f907e0f92dffdcba5e29d3c works.)

seangoodvibes (approved):

PR #2377, opened 2024-08-03
Bugfix - Make community features sync scaling menu a regular setting menu and not a toggle

seangoodvibes:
Made the community features sync scaling action menu a regular setting menu and not a toggle menu as it's not a regular toggle


nikodemus (approved):
Thanks!

PR #2385, opened 2024-08-05
Feature: chord keyboard

madeline-scyphers:
# Description

It adds a chord keyboard as an additional keyboard option, accessible the same way as other keyboards instrument supporting keyboards (press keyboard button, then hold it while turning select knob). 

Each column is a note (chormatic, all 12 notes), with infinite scrolling with horizontal encoder. Each pad is a different chord. So the first row is just roots, second is major triads, 3rd is minor triads, etc. Vertical encoder scrolls up to more chords. 

Holding a note on a row and pressing in the horizontal encoder and turning changes the voicing for all the chords on that row (hold any major triad, pressing and turning horizontal encoder changes ALL major triad voicings, not just the one you are holding).

## Future Improvements

- Allow changing chord rows (vertical enc and pad press, kind of like in song or clip mode?)
- User defined chords, voicings, and chord orderings
- allow moving between chord families easily, maybe holding a pad and pressed in vert enc switches from G7 to G9 to G7b9, etc
- Have the pads be dimmer or brighter based on something that makes sense to help with navigation (key or tonal centers, just light up certain columns or rows, etc)
- Make the bottom row always be the root might be nice so that way you can always play a melody or bass line no matter how far up in the chords you are
- Maybe allow a bottom row step sequencer to punch in chords instead having to record them

https://github.com/user-attachments/assets/c7cf2054-8b1f-49a9-a60d-81a68652698d


src/deluge/gui/ui/keyboard/chords.h
```diff
@@ -0,0 +1,170 @@
...
+	 * @return The voicing
+	 */
+	Voicing getChordVoicing(int32_t chordNo);
+
+	void adjustChordRowOffset(int32_t offset);
+	void adjustVoicingOffset(int32_t chordNo, int32_t offset);
+
+	Chord chords[kUniqueChords];
+	int32_t voicingOffset[kUniqueChords] = {0};
+	uint32_t chordRowOffset = 0;
```
madeline-scyphers, line 162: Would it be better put these two state trackers in the `KeyboardStateChord` struct or here? If they are kept there, then we could more easily reuse the `ChordList` class for the chord columns if we want to add more functionality to those. If that isn't important, then just unifying those chord columns with these chord definitions, without the `ChordList` class could be enough. Thoughts?

m-m-adams: I'd go with your gut feel. Either way it doesn't look super difficult to change in the future if needed, and I think you might be the most likely person to extend it into chord columns anyway!

src/deluge/gui/ui/keyboard/layout.h
```diff
@@ -51,7 +51,9 @@ class KeyboardLayout {
 	virtual void handleVerticalEncoder(int32_t offset) = 0;
 
 	/// Will be called with offset 0 to recalculate bounds on clip changes
-	virtual void handleHorizontalEncoder(int32_t offset, bool shiftEnabled) = 0;
+	virtual void handleHorizontalEncoder(int32_t offset, bool shiftEnabled,
```
m-m-adams, line 54: This is the only change that I'm not sure about since it sort of breaks the metaphor of scrolling the keyboard under your fingers

That said I'm also not sure how useful that is beyond familiarity since you'll usually just have one pad held in this keyboard anyway, and your discord posts about it really speak to how useful this is so I think we should give it a try

src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+	if (noteCodeIsSharp[noteCodeWithinOctave]) {
+		*thisChar = display->haveOLED() ? '#' : '.';
+		thisChar++;
+		drawDot = 255;
+	}
+	else {
+		drawDot = 0;
+	}
+	*thisChar = '\0';
```
m-m-adams, line 124: nit but maybe just null initialize noteName={0} and then just use noteName[0] for the note code and noteName[1] for the sharp/flat? It's a bit confusing to read iterator code that's not really iterating and that avoids having noteName[2] be garbage for natural notes

src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+	// Create a buffer to hold the full chord name and format it
+	char fullChordName[300];
+	if (voicingName[0] == '\0') {
```
m-m-adams, line 128: just 0 is usually better, it avoids potential mistakes like forgetting an escape and using '0'

m-m-adams (commented):
Just some nits but code looks good!

src/deluge/gui/ui/keyboard/layout.h
madeline-scyphers: For documentation, this is what I put in the discord

"The reason I wasn't doing that was with the other keyboards SHIFT + :hturn:  had more to do with the keyboard layout.

If it is SHIFT, and you have to be pressing a chord to change voicings, it pretty much requires sticky SHIFT. It would be near impossible to hold SHIFT  + :hturn:  + :padpress: . I don't know how much that is a draw back.

Then considerations for future next steps,

I want to add some more functionality like changing Chord order, which should be fairly easy to do if handleVerticalEncoder is passed PressedPad presses[kMaxNumKeyboardPadPresses], then you could just drag them like you drag notes in clip view or song view.

The bigger thing is that I would like to maybe add an ability to change notes from one chord to the next inside a family. If you are on a G7, you could move to G9, move to G7b9, etc. This would potentially allow reducing the height of the chord keyboard, maybe. But then I don't know how to do that controls wise. I thought about maybe :vpressturn: + :padpress:  or SHIFT + :vturn:/:hturn:   + :padpress:  could be used for this. It feels like the chord keyboard might just have more functionality than the other keyboards.

If we had horizontal dimmer or white bars (every 4 rows or whatever) for easier navigation, that could be configurable with shift + :hturn:  or we could use SHIFT + :hturn:  to offset the root note row octave wise or allow more rows of the root note for playing a bass line while doing chords. 

Just things I have thought about for other useful features and controls with the chord keyboard for the next round. I didn't want to make this initial PR too big
Also since I have so many future plans for the chord keyboard feature and control wise. I was thinking if it does make it into 1.2, putting it behind a community features menu for now. I don't think it needs to be behind one forever since it doesn't impact basically anything when it isn't in use, but if it is behind that menu it allow it to be more fluid to change until it is finalized"

src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
madeline-scyphers: This helped me figure out why I was why I was getting garbage if I just used `noteCodeToString` straight like elsewhere in the code base, and why I partially reimplemented it. So thanks!

m-m-adams (approved):

PR #2410, opened 2024-08-08
fix for note rows without drums

m-m-adams:


m-m-adams:
same as #2402 but the cherry pick failed

PR #2412, opened 2024-08-08
Bugfix/audio track recording with null output

m-m-adams:
When an audio output is cloned it could have its source set as internal but the internal recorder would still be null

Fix menu display if that happens, and fix cloning so it doesn't happen


seangoodvibes (approved):

m-m-adams:
gonna make it error when creating the recorder just in case

PR #2421, opened 2024-08-11
Feature/overdub in place

m-m-adams:
Add a new overdub method for audio clips based on resampling their output. This allows for more traditional looper workflows. LOOP command auto extends and the existing sample loops as you play. LAYER does not auto extend but continues adding layers. This is functionally the same as the existing clip cloning method but avoids weirdness in the grid by staying on a single output. 

I've elected to not use this new method in rows mode for old times sake and so that people who like the individual clip per layer approach can continue to use it.

Exclusively works in grid mode for audio outputs with monitoring right now (midi controller required). There's some work to do to enable it for outputs without monitoring (namely adding the input to the rendered audio, and therefore recording pre effects) but this basic usecase of recording post fx and with monitoring on is fully functional

This is largely ready to extend to instrument clips as well - the missing feature is automatically copying the previous clip contents as the loop extends so it's disabled until that's done. Turning it on now just leads to the existing clip playing once and then auto extending with blank space.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 - Added blend control to compressors
 - Added ability to record from a specific track's output. Set an audio clips input to TRACK, then in the audio clip menu
 use the TRACK menu to select the specific track to record from
+  - To hear the instrument through the audio clip's FX set the input to TRACK (FX)
 - Added filters in FM synth mode. They're set to OFF by default, enable by changing them to any other mode using the menu or db/oct shortcut.
+- Audio clips with monitoring (or FX) active in grid mode now support in place overdub using the global midi loop/layer commands
 
 ### User Interface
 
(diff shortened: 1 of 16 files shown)


m-m-adams:
Absolutely but I'm not sure of the UI yet. The looping command for rows mode is to hold record and press a row below the one you want to loop but that doesn't work in the grid (record+pad in green does record arming, it's open in blue but that breaks the 'edit mode' metaphor)

seangoodvibes:
@m-m-adams so I tried this out and it doesn't seem to work for me.

- I created a new song
- I created an audio clip with monitoring on (to record external audio in)
- With the audio clip selected, I send the global midi command loop to record the first audio loop, and hit loop again to stop
- When the audio starts looping, I hit loop or layer (both have the same issue) to record the next audio layer and I get an error saying "INSUFFICIENT RAM" on the display, so it's unable to record that second audio layer. The first audio loop keeps looping, no issues there. 

So it just seems like the creation of the layer is failing.

m-m-adams:
RAM issues are unrelated

seangoodvibes (approved):

PR #2425, opened 2024-08-11
Feature - allow running instrument outputs into audio outputs for fx and recording

m-m-adams:
Extends the specific output recording to allow use as an fx processor

When an output is processed in this way it is just rendered into the audio output's buffer instead of being rendered by the song


seangoodvibes:
So I tested it, seems to work good, one main observation though: delay doesn't work as an effect on the monitored audio.

Could we also document this :D? Thx

seangoodvibes (approved):

PR #2429, opened 2024-08-13
remember last selection and exit on pad release

m-m-adams:
Caches last created clip type, immediately create it if you release the pad to make clip creation fast again

The menu opens with the clip you are about to create selected for visual clarity


seangoodvibes (approved):
As discussed, approving, will follow this-up with second PR to clean things up.

PR #2431, opened 2024-08-13
always skip multi-range select if there's only one option

nikodemus:
- `checkPermissionToBeginSessionForRangeSpecificParam()` used to have an `automaticallySelectIfOnlyOne`-argument, which did exactly what it says on the tin.

- That function was called by 4 MenuItems, 2 of which passed true, and 2 of which passed false:
  - `FileSelector` & `AudioRecorder`: false
  - `LoopPoint` & `sample::Transpose`: true

- Only time this code has been touched by community was code-reorg by Kate, and there's no clear reason for the different usages.

- This commit removes that argument, and behaves as if it was always true. Either we'll find out why, or the UX is slightly smoother.

### DIFF
--- a/src/deluge/gui/menu_item/file_selector.cpp
+++ b/src/deluge/gui/menu_item/file_selector.cpp
@@ -72,6 +72,6 @@ MenuPermission FileSelector::checkPermissionToBeginSession(ModControllableAudio*
 		return MenuPermission::NO;
 	}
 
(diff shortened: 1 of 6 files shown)


nikodemus:
Converted to draft, as I realized there's an apparent simplification in the area, that will either make it more obvious why this is ok, or ... not. :)

nikodemus:
Changed my mind, this is better without the refactoring.

seangoodvibes (approved):

PR #2439, opened 2024-08-14
Update UI for creating and converting audio clips in song grid view

seangoodvibes:
UI changes for Song Grid View: this is a follow up to #2429 and #2299

Updated UI for creating new audio clips
- when creating a new clip in a new track, pressing cross screen will select the audio clip type

Updated UI for converting an empty instrument clip to an audio clip
- Instead of holding a pad and pressing select, you now hold a pad and press cross screen

When you select an audio clip, the cross screen button will also now flash to indicate that you've selected an audio clip


m-m-adams:
I think there's a couple conflicts with the cross screen LED in corner cases, right now it's just creating audio clips in arranger view but audio clips themselves will too if get cross screen automation working at some point 

Otherwise lgtm, the worst "normal" downside I've noticed is that it highlights the lack of an actual audio led a bit

m-m-adams:
@seangoodvibes big bug, pressing back when creating kits in green mode doesn't seem to cancel properly. You end up with a synth getting created without any clips and get stuck in the copy clips mode

PR #2446, opened 2024-08-15
[DRAFT] horizontal menus paging menus with editable values

nikodemus:
- Visual idea for OLED, works transparetnly on 7-seg.

      LOW-PASS FILTER
      -------------------
      FREQ RESO DRIV MODE
        10    5    0 12DB

  If there's > 4 items some of them will flow to the next page:

      MOD-FX          [1]2
      --------------------
      TYPE RATE FBCK DEPTH
      GRAI    5    2    35

- Some menu structures have been fiddled with, in the spirit of experimentation: mostly order has been adjusted to work better with the horizontal menus, but eg. sample-related items have been split from oscillator menus to another, correspoding to the grouping of shortcuts. Perhaps even more controversially, Osc1 & Osc2 volumes have been grouped with NOISE into a MIXER menu under Sound, matching common front-panel org in synths.

- Active value displayed by inverting the column, editable by holding SHIFT and turning SELECT. Step-by-5-when-holding-shift is disabled when diting values in this view: clicking through to the single value editor makes step-by-5 work normally.

- This "editing from menu" is available on 7-seg as well: holding SHIFT will display the current value for the menu item, and allow editing via select while it is held.

- When a shortcut (like LPF FREQ) is opened via pads, on OLED it opens the parent horizontal menu instead, with the right item activated. If you select an inactive shortcut, the relevant menu is still opened, but with first item in focus instead. This makes the pads a bit faster to use and learn, I think. 7-seg behaviour for shortcuts is unchanged!

  NOTE: Might be interesting to experiment with 7-seg behaviour:
  - DON'T add the "open the menu even if item is inactive", but DO open the submenu for active items -> value is directly editable as long as you hold shift, when you let go it will pop you in the menu. Exact same number of button presses needed as before, but more flexibility?

- Implemented via HorizontalMenu class, which parallels Submenu.

- HorizontalMenu for the parent menus associated with all shortcuts. Buttons with missing disabled shortcuts still take to the relevant menu, as long as _some_ shortcuts remain there.

  TODO:
  - Sound FM shortcuts
  - Other views than sound?

- Some menus get special visuals:
  - Filter route is rendered as an arrow.
  - LFO type is rendered as the waveform.

- Some fresh duplication of menu item classes where previously same menu item was used eg. for Env1 and Env2, etc.

- [ ] Sound shortcuts for FM pads missing.

- [ ] Instread of being a parallel class to Submenu, HorizontalMenu should probably be its subclass, and old Submenu should be an explicit VerticalMenu.

- [ ] Envs1 & 2 magic should be restored, and as much of the parts pulled out of the menu compiler for other this too should be restored.

- [ ] Make the edit-modifier configurable in COMMUNITY menu:
  - [ ] NONE
  - [ ] SHIFT
  - [ ] Hold and turn SELECT

- [ ] Refactor renderInHorizontalMenu() methods could share more code.

- [ ] Maybe split this into smaller parts?

- [ ] Review names visible. Should we have shortName(), name(), and title() separately?


nikodemus:
Superseded by https://github.com/SynthstromAudible/DelugeFirmware/pull/3118/files

PR #2448, opened 2024-08-16
Allow ColumnControlsKeyboards to opt out of sidebar columns

madeline-scyphers:
A keyboard subclass can override allowSidebarType to skip certain sidebar cols.

Encountering bug where if Chord Keyboard has default constructor, then changing sidebar col after entering chord keyboard with cause a crash, but if using an empty constructor will cause a crash when changing to in-key keyboard after entering the chord keyboard

### DIFF
--- a/src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
+++ b/src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
@@ -145,7 +145,7 @@ void KeyboardLayoutChord::drawChordName(int16_t noteCode, const char* chordName,
 }
 
 bool KeyboardLayoutChord::allowSidebarType(ColumnControlFunction sidebarType) {
-	if ((sidebarType == ColumnControlFunction::CHORD) || (sidebarType == ColumnControlFunction::DX) ||
+	if ((sidebarType == ColumnControlFunction::CHORD) ||
 	    // TODO, when scales for chord keyboard are implemented, add this back in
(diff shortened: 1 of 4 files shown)


src/deluge/gui/ui/keyboard/layout/chord_keyboard.cpp
```diff
@@ -137,4 +143,14 @@ void KeyboardLayoutChord::drawChordName(int16_t noteCode, const char* chordName,
 		display->setScrollingText(fullChordName, 0);
 	}
 }
+
+bool KeyboardLayoutChord::allowSidebarType(ColumnControlFunction sidebarType) {
+	if ((sidebarType == ColumnControlFunction::CHORD) || (sidebarType == ColumnControlFunction::DX) ||
+	    // TODO, when scales for chord keyboard are implemented, add this back in
+	    (sidebarType == ColumnControlFunction::SCALE_MODE)) {
```
madeline-scyphers, line 150: I figured these 3 were the ones it should skip for now. Does that seem right? `CHORD` for sure, `SCALE_MODE` until  the chord keyboard supports scales, and would `DX` be good to skip too? Any others?

m-m-adams: I think the DX sidebar should stay, it's skipped unless you're on a DX7 patch anyway

m-m-adams (commented):
Lgtm except removing the DX sidebar, its allowed in the other ones so I think it should stay here too

m-m-adams (approved):

PR #2450, opened 2024-08-16
Submenu doesn't need SoundEditor::menuCurrentScroll

nikodemus:


nikodemus:
FYI: This is the one that matters most as pre-factoring for putting Horizontal menu functionality into Submenu.

seangoodvibes (approved):

PR #2451, opened 2024-08-16
Enumeration & Selection don't need SoundEditor::menuCurrentScroll

nikodemus:


nikodemus:
I'll fold this with Selection, easier to look at them together.

src/deluge/gui/menu_item/enumeration.cpp
```diff
@@ -1,15 +1,18 @@
...
 namespace deluge::gui::menu_item {
 void Enumeration::beginSession(MenuItem* navigatedBackwardFrom) {
 	Value::beginSession(navigatedBackwardFrom);
-	if (display->haveOLED()) {
-		soundEditor.menuCurrentScroll = 0;
-	}
-	else {
-		drawValue();
-	}
+	drawValue();
```
seangoodvibes, line 9: is this a mistake? drawValue was only called before if you had 7SEG, not OLED. or does drawValue also handle OLED rendering?

nikodemus: Yes, it does! This seems to be a common pattern in the code, presumably left over from original OLED work: there's an if-OLED check before drawValue, then drawValue does the same check again.
```
void Enumeration::drawValue() {
	if (display->haveOLED()) {
		renderUIsForOled();
	}
	if (display->have7SEG()) {
		display->setTextAsNumber(getValue());
	}
}
```

seangoodvibes (approved):

Issue #2437, opened 2024-08-14
Three issues with the stem recording


thesoundfarmer:
### Please describe the problem:

To summarise and better explain the issues I'm having with stem export

Issue 1: Exporting in both Arranger and Clip/Song view on particular songs
Deluge seems to freeze when exporting stems on older tracks but the stem recording can be cancelled so the Deluge hasn't crashed. Just hangs on 0/11 stems.

Issue 2: Exporting in Arranger view
Most but not all stems export
Those that do are all the same file size (49mb) however some are only 5 seconds long whilst some are full length but silent
Others export fine

Issue 3: Exporting in song and clip view
Most but not all stems export

### What is the expected behavior?

Had hoped it would export all instruments

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

deluge-v1_2_0-nightly+2024_08_12-31d29ed and deluge-v1_2_0-nightly+2024_08_16-a0f4498

### If possible provide the steps to reproduce the issue and upload additional media:

Issue 1:  

Turned on
Loaded track in clip view
Pressed Save and Rec
Message reads "Exporting 0/11"
Fails to proceed

Issue 2:

Turned on
Loaded arranger view
Pressed Save and Rec
Stems export happily
Open on computer
Only some of the instruments have saved.  Files are there but empty

Issue 3:

Turned on
Loaded track in clip view
Pressed Save and Rec
Most but not all stems exported

thesoundfarmer:
I should add Issue 1 is exporting in clip mode
Issue 2 is exporting in song mode

m-m-adams:
I suspect 1 might be related to empty clips, they might start linear recording with auto length expansion

thesoundfarmer:
Hi,

The only muted instruments are the duplicates doing different parts of the melodies.  I'd have thought they'd unmute when they're cued

seangoodvibes:
> Hi,
> 
> The only muted instruments are the duplicates doing different parts of the melodies. I'd have thought they'd unmute when they're cued

Are you in arranger view?

In arranger view stem export will exclude muted tracks, if you want to export all tracks, unmute them all before stating stem export

If you're doing it in song view, it will export all clips regardless of mute state.

thesoundfarmer:
[...]

PR #2456, opened 2024-08-16
Make exporting stems to silence and including/excluding song FX configurable + Bug fixes

seangoodvibes:
- Added new stem export configuration menu option to disable exporting stems to silence

This should allow stems reported in issue https://github.com/SynthstromAudible/DelugeFirmware/issues/2437 to be exported, however, it does not resolve the issue as to why silence is not reached with those songs.

- Made including / excluding Song FX configurable

- Updated VU meter + Silence check to check levels before Song FX if exporting stems with Song FX disabled

- Added blinking of back button to indicate that you can cancel stem export

- Fixed bug where cancelled stem export would finish with non-offline rendering

- Fixed bug where display would get stuck with a pop-up after cancelling stem export

- Fixed bug where cancelling stem export process would not cancel the recording right away which could lead to problems where recording never finishes.

- Fixed bug where pads and encoders could still be interacted with while stem export process was underway.

- Added some troubleshooting steps to the documentation


src/deluge/gui/views/arranger_view.cpp
```diff
@@ -935,6 +935,10 @@ void ArrangerView::auditionEnded() {
 }
 
 ActionResult ArrangerView::padAction(int32_t x, int32_t y, int32_t velocity) {
+	// do not interpret pad actions when stem export is underway
```
m-m-adams, line 938: can this be done at the matrix driver instead of spreading these checks everywhere?

src/deluge/model/song/song.cpp
```diff
@@ -2337,6 +2338,10 @@ void Song::renderAudio(StereoSample* outputBuffer, int32_t numSamples, int32_t*
 #endif
 	}
 
+	if (stemExport.processStarted && !stemExport.includeSongFX) {
```
m-m-adams, line 2341: could you just skip rendering the song fx instead?

m-m-adams (approved):

PR #2464, opened 2024-08-19
Stem Export: Added 60 second recording silence timeout

seangoodvibes:
Added 60 second recording timeout when exporting stems until silence.

If approxRMSLevel does not drop to mutable noise floor (< 9) within 60 seconds of playback turning off then track / clip is deemed silent and recording is stopped.

### DIFF
--- a/docs/features/stem_export.md
+++ b/docs/features/stem_export.md
@@ -105,7 +105,7 @@ If you are using the `EXPORT TO SILENCE` feature, it may take longer for your tr
 
 As a safety measure, if your track does not become silent within 60 seconds of playback stopping, then the recording will stop automatically.
 
-If you do not want to wait potentially 60 seconds, you will need to press `BACK` to cancel the stem export. If you press `BACK` after the track's loop length is finished, it will save the stem, however the stem's length may not match the loop length.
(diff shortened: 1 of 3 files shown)


docs/features/stem_export.md
```diff
@@ -88,21 +89,23 @@ This menu allows you to start a stem export and configure various settings relat
...
+#### Scenario: The track has an exceptionally long tail that doesn't not drop to silence
 
-In this case, you will need to press `BACK` to cancel the stem export. If you press `BACK` after the track's loop length is finished, it will save the stem, however the stem's length may not match the loop length.
+If you are using the `EXPORT TO SILENCE` feature, it may take longer for your track to export because the recorder is waiting for your track to become silent. Thus, you will see that playback has turned off but the record button continues to blink rapidly because it is still waiting for silence before stopping the recording.
 
-Possible solution:
+As a safety measure, if your track does not become silent within 60 seconds of playback stopping, then the recording will stop automatically.
+
+If you do not want to wait potentially 60 seconds, you will need to press `BACK` to cancel the stem export. If you press `BACK` after the track's loop length is finished, it will save the stem, however the stem's length may not match the loop length.
```
m-m-adams, line 108: why is this? Intuitively it feels like we should be able to set the correct loop length anyway and only the ending would be shortened

PR #2470, opened 2024-08-21
Fix #2463 Drumrandomizer ignore mac dot files

### LINKED ISSUE #2463: Drum Randomizer should ignore  Mac .dot files

Just reorganized all my drum samples to take advantage of the Drum Randomization community feature, which was such a good idea! I renamed all samples in each folder numerically (1.wav, 2.wav, etc).

But, when I hold hit AUDITION + RANDOM, it sometimes loads "._#.wav" (._1.wav, ._57.wav, etc.) which are hidden mac files in that directory, which don't contain any audio, so the drum randomizer has ~50% of actually working.

[BONUS IDEA!] Respect the setting to audition samples that the deluge already has, where you can set to always audition, or audition only if playback stopped, so if playback is stopped... audition the sound when RANDOM is pressed. Thx!

AUDITION + RANDOM selects a news sample from the directory the pad's sample is in, ignoring the mac .dot files ("._blah.wav")

_No response_

7SEG

Release 1.1/Beethoven

c1.1.0

Mac only
1. Initialize a new kit
2. Load a sample from a directory that has other samples in it
3. Hold [AUDITION] + [RANDOM] to randomize the sample
4. Check display to see if an actual sample was loaded, or if a .dot file was loaded erroneously

M0R172:
Fix #2463

isAudioFilename now correctly rejects filenames starting with dot

(isAifffile was not changed because it is used in a different context)

### DIFF
--- a/src/deluge/util/functions.cpp
+++ b/src/deluge/util/functions.cpp
@@ -1301,7 +1301,7 @@ void getThingFilename(char const* thingName, int16_t currentSlot, int8_t current
 }
 
 bool isAudioFilename(char const* filename) {
-	if (*filename == '.') {
+	if (filename[0] == '.') {
 		return false;
 	}
 	char* dotPos = strrchr(filename, '.');
--- a/src/deluge/util/functions.cpp
+++ b/src/deluge/util/functions.cpp
@@ -1301,9 +1301,11 @@ void getThingFilename(char const* thingName, int16_t currentSlot, int8_t current
 }
 
 bool isAudioFilename(char const* filename) {
+	if (*filename == '.') {
+		return false;
+	}
 	char* dotPos = strrchr(filename, '.');
-	return (dotPos != 0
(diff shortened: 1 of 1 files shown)


src/deluge/util/functions.cpp
```diff
@@ -1301,9 +1301,11 @@ void getThingFilename(char const* thingName, int16_t currentSlot, int8_t current
 }
 
 bool isAudioFilename(char const* filename) {
+	if (*filename == '.') {
```
m-m-adams, line 1304: style nit: filename[0] is clearer 
```suggestion
	if (filename[0] == '.') {
```

stirman:
Thank you!

src/deluge/util/functions.cpp
M0R172: Fully agreed!

m-m-adams (approved):

PR #2508, opened 2024-08-27
Cleanup: Adjust clip titles

seangoodvibes:
Cleaned up titles for MIDI, CV and Audio clips

### DIFF
--- a/src/deluge/gui/views/view.cpp
+++ b/src/deluge/gui/views/view.cpp
@@ -1952,7 +1952,7 @@ void View::drawOutputNameFromDetails(OutputType outputType, int32_t channel, int
 			break;
 		case OutputType::MIDI_OUT:
 			if (channel < 16) {
-				outputTypeText = "MIDI channel";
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/view.cpp
```diff
@@ -1952,7 +1952,7 @@ void View::drawOutputNameFromDetails(OutputType outputType, int32_t channel, int
 			break;
 		case OutputType::MIDI_OUT:
 			if (channel < 16) {
-				outputTypeText = "MIDI channel";
+				outputTypeText = "MIDI";
 			}
 			else if (channel == MIDI_CHANNEL_MPE_LOWER_ZONE || channel == MIDI_CHANNEL_MPE_UPPER_ZONE) {
 				outputTypeText = "MPE zone";
```
m-m-adams, line 1958: should just be MPE - "zone" is analogous to "channel"

seangoodvibes: Done good catch!

PR #2550, opened 2024-09-01
Create looping_in_grid_view.md

seangoodvibes:
Documented the two looping techniques in grid view


m-m-adams:
This is almost all covered in the looping section of the official manual. I was going to do a community specific guide once I have it closer to complete

My plan is to get rid of the clip cloning style looping from grid view entirely (will be moved to AUDIO SAMPLER clips or something) but there's a lot more to do first

PR #2552, opened 2024-09-02
[DRAFT] Feature/disable midi monitoring

m-m-adams:
Disables midi monitoring if the device and channel match the midi output

Also fixes some inheritance style stuff (mostly so I could refactor signatures automatically)

Needs a menu to turn this on/off, currently will never send midi in this situation


bukru (approved):

bukru:
ops sorry wrong manipulation i didn't approve the change i didn't verify nothing, i'm just agree with this functionality  and i though it will increase this counter ...
vote

PR #2562, opened 2024-09-02
Bugfix/open copy buffer before use

m-m-adams:
This fixes a (nightly only) storage manager bug caused by using the write buffer for both writing the actual song file and for copying samples to/from the collect all location


seangoodvibes (approved):

seangoodvibes:
oops I approved it but probably best if Jamie looks it over :D

Issue #2564, opened 2024-09-02
Update stem export file path


seangoodvibes:
### Please describe the problem:

Recapping stem export changes for myself:

Currently we export stems to the following file path:

SAMPLES/STEMS/SONGNAME#####/FILENAME#####.WAV

The agreed upon change is to export stems to a single Song folder with no number incrementing after the Song name

Within each folder there will two folders for Arranger and Song exports

Arranger and Song exports will have a 2 digit number appended to the end for incremental stem exports

File names within each Arranger and Song folder will have a 3 digit number appended to them representing the unique ID of the track or clip being exported

Thus the stem file path will change to the following:

SAMPLES/STEMS/SONGNAME/ARRANGER##/FILENAME###.WAV

SAMPLES/STEMS/SONGNAME/SONG##/FILENAME###.WAV

In the future, when kit drum exporting is added, a KIT## folder will be added underneath ARRANGER## or SONG## depending on the type of exporting being done (eg is it being done from arranger, from song view or from within a kit clip)

So we would have the following file paths for Kit drums:

SAMPLES/STEMS/SONGNAME/ARRANGER##/KIT##/FILENAME###.WAV

SAMPLES/STEMS/SONGNAME/SONG##/KIT##/FILENAME###.WAV

### What is the expected behavior?

See above

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

PR #2572, opened 2024-09-03
Update Stem Export Folder Path and File Names

seangoodvibes:
### Description

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2564

> Currently we export stems to the following file path:
> 
> SAMPLES/STEMS/SONGNAME#####/FILENAME#####.WAV
> 
> The agreed upon change is to export stems to a single Song folder with no number incrementing after the Song name
> 
> Within each folder there will two folders for Arranger and Song exports
> 
> Arranger and Song exports will have a 2 digit number appended to the end for incremental stem exports
> 
> File names within each Arranger and Song folder will have a 3 digit number appended to them representing the unique ID of the track or clip being exported
> 
> Thus the stem file path will change to the following:
> 
> SAMPLES/STEMS/SONGNAME/ARRANGER##/FILENAME###.WAV
> 
> SAMPLES/STEMS/SONGNAME/SONG##/FILENAME###.WAV
> 
> In the future, when kit drum exporting is added, a KIT## folder will be added underneath ARRANGER## or SONG## depending on the type of exporting being done (eg is it being done from arranger, from song view or from within a kit clip)
> 
> So we would have the following file paths for Kit drums:
> 
> SAMPLES/STEMS/SONGNAME/ARRANGER##/KIT##/FILENAME###.WAV
>
> SAMPLES/STEMS/SONGNAME/SONG##/KIT##/FILENAME###.WAV

### DIFF
--- a/src/deluge/processing/stem_export/stem_export.h
+++ b/src/deluge/processing/stem_export/stem_export.h
@@ -81,7 +81,7 @@ public:
 	Error getUnusedStemRecordingFilePath(String* filePath, AudioRecordingFolder folder);
 	Error getUnusedStemRecordingFolderPath(String* filePath, AudioRecordingFolder folder);
 	int32_t highestUsedStemFolderNumber;
-	String lastSongNameForStemExport;
+	String lastFolderNameForStemExport;
 	void setWavFileNameForStemExport(StemExportType type, Output* output, int32_t fileNumber);
 	String wavFileNameForStemExport;
 	bool wavFileNameForStemExportSet;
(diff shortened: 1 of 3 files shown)


src/deluge/processing/stem_export/stem_export.h
```diff
@@ -81,7 +81,7 @@ class StemExport {
 	Error getUnusedStemRecordingFilePath(String* filePath, AudioRecordingFolder folder);
 	Error getUnusedStemRecordingFolderPath(String* filePath, AudioRecordingFolder folder);
 	int32_t highestUsedStemFolderNumber;
-	String lastSongNameForStemExport;
+	String lastFolderNameForStemExport;
```
m-m-adams, line 84: does this need to be initialized to an empty string?

seangoodvibes: > does this need to be initialized to an empty string?

Yes that's a good idea! Done thx!

m-m-adams (approved):

PR #2575, opened 2024-09-04
[DRAFT] Enhancement / Performance View ~ Allow use of gold knobs to edit pad values in value editing mode

seangoodvibes:
Add a new DEFAULTS > UI > PERFORMANCE menu to enable the use of gold knobs to edit pad values in the performance view value editing mode

Also took the opportunity to move the AUTOMATION defaults menu under the UI menu.


m-m-adams:
Is this really necessary? It seems pretty niche and we need to be mindful of binary size and code complexity, especially with menus

seangoodvibes:
> Is this really necessary? It seems pretty niche and we need to be mindful of binary size and code complexity, especially with menus

eh not absolutely necessary no

someone requested being able to perform sweeps more easily than the pads allow in a live context, so I suggested the value editing mode, to which they said it works but using gold knobs would be better because doing sweeps with select encoder is difficult

m-m-adams:
Can they not just use the gold knobs outside of value editing mode? Trying to use an editing mode for performance just doesn't sound like a good idea

Issue #2069, opened 2024-06-02
When auditioning a sample in Drum Keyboard View the gold knobs/param LEDs don't reflect the last auditioned drum sample.


ok-reza:
### Please describe the problem:

If you enter Drum Keyboard View, audition any pad, the gold knob values and assignments do not switch to reflect the newly auditioned sample. Once you switch back to Kit Clip View, the last auditioned pad will then have it's gold knob parameters update.

Currently, if you adjust, say, LPF in Drum Keyboard View (with `Affect Entire` off) you will simply adjust the sample row that was last active while in Kit Clip View despite toggling different samples. Pressing a new sample in Drum Keyboard View doesn't update the gold params to that sample.

### What is the expected behavior?

Similarly to auditioning different samples in Kit Clip View, the gold knobs and assignments should update based on the last auditioned sample in Drum Keyboard View

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

other

### What is the firmware name:

1.1.0-beta-F4A2506

PR #2601, opened 2024-09-06
Bugfix Set Selected Drum when pressing drum pads in kit keyboard view

seangoodvibes:
Updated drum velocity keyboard view to set selected drum when you press drum pads

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2069

Pressing drum pads in kit keyboard view will now update the drum selection so that you can edit parameters of each drum directly from the kit keyboard view when affect entire is disabled

- Note: if you use the `MIDI FOLLOW FEEDBACK` feature, no MIDI Feedback data will be sent when you change drum selections in this view because it will send too much MIDI data and affect the Deluge's performance.


m-m-adams (commented):
I don't want to cherry pick this - drumming on the keyboard and sending midi follow output on every press seems likely to cause issues with how rough we know the midi stack currently is

seangoodvibes:
> I don't want to cherry pick this - drumming on the keyboard and sending midi follow output on every press seems likely to cause issues with how rough we know the midi stack currently is

Added code to not send midi feedback when changing drum selection using this view. I think that it addresses that concern.

m-m-adams (approved):

PR #2611, opened 2024-09-08
Refactor deluge_main for readability

M0R172:
Extracted methods and functions, moved some stuff around, cleaned up and added comments, grouped stuff logically (I hope)

I think this is now *much* more readable, functionality should be the same...


src/deluge/deluge.cpp
```diff
@@ -878,59 +907,68 @@ extern "C" int32_t deluge_main(void) {
...
-#endif
+	currentPlaybackMode = &session; // ?
+
+	setOutputState(CODEC.port, CODEC.pin, 1); // Enable Audio Output
+
+	AudioEngine::init();
+
+	audioFileManager.init();
+
+	tests_2();
```
m-m-adams, line 946: Bit of a nit but maybe keep the checks around these? 

Or just remove them, I'm not sure that they actually work anymore

m-m-adams:
Thanks for the PR! Definitely more readable like this

src/deluge/deluge.cpp
M0R172: Removed the checks for readability. I'm pretty sure the compiler will remove the empty function calls when the defines are negative.

Did not want to remove them in case they were still being used by someone - For now, I added comments to check them for possible future removal...

m-m-adams: Yeah it will, I think it's more readable to immediately see that they won't run though. Maybe just more descriptive names? 

tests_1 -> automatedActionTests and tests_2 -> hardwareTests

M0R172: Seems that most of theses tests do not even compile when activated, and some seem to hang/crash.

For now I just removed the invocations in deluge_main, 
but left the test code in place, 
as some of it is also called in other ways (e.g. ramTestLED can be started by pressing the three first mod buttons). Also some of the test code might be fixed or at least serve as inspiration for a future test suite?

m-m-adams: > Seems that most of theses tests do not even compile when activated, and some seem to hang/crash.
> 
> For now I just removed the invocations in deluge_main, but left the test code in place, as some of it is also called in other ways (e.g. ramTestLED can be started by pressing the three first mod buttons). Also some of the test code might be fixed or at least serve as inspiration for a future test suite?

Yeah some of them have already been factored out into the 32bit unit tests suite, it would be nice to move more that way down the road

m-m-adams (approved):

PR #2615, opened 2024-09-08
Enhancement/ Play and Restart Arranger From Held Pad

seangoodvibes:
Added ability to start / restart arrangement playback from the clip pad you're holding in arranger.

https://github.com/user-attachments/assets/adc8b653-5c73-4fe2-942f-58c894bac0a1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### User Interface
 
 - Added Vibrato and Sidechain patch cables to Automation View Overview and Grid Shortcuts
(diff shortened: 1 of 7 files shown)


soymonitus:
This is so cool

m-m-adams:
Is it possible to also fix cross screen start position to be from the previous playback position instead of the left edge of screen? As is it doesn't work properly if you scroll while in cross screen mode so that the cursor is in the middle or somewhere else in the grid

PR #2619, opened 2024-09-09
Bugfix / arranger performance view playback restart behaviour

seangoodvibes:
The playback restart changes made for arranger view were not carrying over to arranger performance view

This was due to:
- arranger performance view not being recognized as an "arranger view"
- performance view was also not returning the correct NavSysId previously which would mean it would not restart arranger from the right scroll position


m-m-adams (approved):

seangoodvibes:
Converted to draft because just remembered arranger automation view also needs to be included like performance view

PR #2627, opened 2024-09-09
fix initialization orders and defer reading community settings

m-m-adams:
Some of the PIC commands seem to need time in between them so put them back where they used to be

Also defer reading the files on the SD card until the SD card is ready rather than just hoping


seangoodvibes (approved):

M0R172:
Damn! Sorry, forgot to test it with cold-booting.

It worked nicely when using loadfw - but then the hardware is kind of initialized already. Something to keep in mind!

PR #2642, opened 2024-09-13
Revert pr's causing nightly crash on boot

seangoodvibes:
Revert PR's #2611 and #2627

#2611 introduced a crash bug in the nightly firmware. #2627 tried to fix it but it didn't work entirely.

FYI @M0R172 @m-m-adams


m-m-adams (approved):
Lgtm

m-m-adams:
@M0R172 fysa it crashes on oled, the fix I added afterwards made it work on 7seg

If you want to resubmit a PR I think the PIC commands need to be sent with some delay between them, I suspect they only worked in the original code due to their ordering with the other startup functions

Issue #2632, opened 2024-09-09
Preset and sample browser should have the inverted highlighted section extended by 1px.


ok-reza:
### Please describe the problem:

The inversion occurs at the same point as the first pixel of the first character. Depending on the character, it can make no left inverted border present.

![IMG_7472](https://github.com/user-attachments/assets/4cb2521b-3a43-4367-8525-5c27395f7503)

### What is the expected behavior?

The inverted area should be extended 1px to the left of the title's start point. This will be consistent with how the bottom and top of the highlighted inversion has a 1px space of white around the text.

Super minor no doubt, but this would make things a hair more presentable.

This has always been present in any OLED firmware.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

seangoodvibes:
Agreed will take care of this too

Issue #2631, opened 2024-09-09
Menus with chevron and check marks have titles cut off earlier than needed


ok-reza:
### Please describe the problem:

Menus with chevron and check mark icons on the right hand side also have a gap between the end of the title and the icon.

### What is the expected behavior?

![IMG_7470](https://github.com/user-attachments/assets/34999a5f-54fe-43f4-a7f3-09abd7ef64c8)

The red space marked above could be filled out more. I think one extra character can fit while still leaving a little space before the icons. This would partly reduce the need of scrolling and look more presentable/intentional imo.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.2-beta

seangoodvibes:
Agreed, I'll fix this

PR #2648, opened 2024-09-14
Fix/ submenu scrolling + Qwerty menu inverted area

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2632
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2631

@ok-reza

### DIFF
--- a/src/definitions_cxx.hpp
+++ b/src/definitions_cxx.hpp
@@ -926,6 +926,9 @@ constexpr int32_t kTextBigSizeY = 13;
 constexpr int32_t kTextHugeSpacingX = 18;
 constexpr int32_t kTextHugeSizeY = 20;
 
+// submenu icons
+constexpr int32_t kSubmenuIconSpacingX = 7;
+
 constexpr int32_t kNoteForDrum = 60;
 
 enum BendRange {
--- a/src/deluge/gui/context_menu/context_menu.cpp
+++ b/src/deluge/gui/context_menu/context_menu.cpp
@@ -89,12 +89,12 @@ void ContextMenu::renderOLED(deluge::hid::display::oled_canvas::Canvas& canvas) 
 		}
 
 		if (isCurrentOptionAvailable()) {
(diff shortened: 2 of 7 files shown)


ok-reza:
this is looking really great ! looks very clean now

i found 2 minor nits:
1) in the song/preset browsers, the right side of the inverted highlighted square is 1px short. if it were to be extended 1px towards the right, then it will properly add a white border on the right side when a song/preset's name has the most amount of characters without triggering text scrolling. interestingly enough, if a name is long enough to trigger scrolling, once scrolling, the inverted highlight has that 1px extension on the right side.

2) in the regular menus, the text boundary of the entries should be 1px shorter on the right side (not the highlighted region but the text bounding box). due to the current 1px extra, it means it effectively includes a 1px space after the last visible character. once the highlighted entry starts to scroll, the scrolling text's bounds looks like it increases by 1px on the right side. shortening the bounding box by 1px will remove this, albeit insanely subtle, effect

seangoodvibes:
> this is looking really great ! looks very clean now
> 
> i found 2 minor nits:
> 
> 1. in the song/preset browsers, the right side of the inverted highlighted square is 1px short. if it were to be extended 1px towards the right, then it will properly add a white border on the right side when a song/preset's name has the most amount of characters without triggering text scrolling. interestingly enough, if a name is long enough to trigger scrolling, once scrolling, the inverted highlight has that 1px extension on the right side.
> 2. in the regular menus, the text boundary of the entries should be 1px shorter on the right side (not the highlighted region but the text bounding box). due to the current 1px extra, it means it effectively includes a 1px space after the last visible character. once the highlighted entry starts to scroll, the scrolling text's bounds looks like it increases by 1px on the right side. shortening the bounding box by 1px will remove this, albeit insanely subtle, effect

I fixed the first one. Can't make heads or tails of the second one. Probably because I don't see what you're seeing. I posted videos on discord.

Issue #2630, opened 2024-09-09
Adjust TAP TEMPO to set new project tempo after 4 presses rather than 2.


ok-reza:
### Please describe the problem:

Currently, pressing TAP TEMPO twice will generate a new BPM and set it for the project. This implementation personally has never been an issue for me in the past.

However, now that we can use SYNC-SCALE as a FILL button, I've found that if I'm triggering FILL often (via SYNC-SCALE) during a live performance, I can mistakenly also press TAP TEMPO (which happens to be right above sync-scale). If it's pressed twice within a short window of time, then the whole BPM changes for the song.

### What is the expected behavior?

I think this can be avoided if it takes 4 tap tempo presses to initiate a BPM change for the project.

As far as I know, this change from 2 taps to 4 taps shouldn't have any negative implications for peoples workflows. As normal, any subsequent tap after 4 will continue to update the BPM normally per tap.

Not the end of the world if this isn't addressed. I think a workaround is to hit UNDO and the tempo can be reverted back, but it would be a nice QoL fix, especially for those that don't see it coming.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.2-beta

sinewave440hz:
I've also experienced this is an inconvenience on occasion, so I'm adding a 👍 to this :) Personally I've done it by accident when activating and de-activating the metronome.

PR #2649, opened 2024-09-14
[DRAFT] Fix/ Adjust number of tap tempo presses to engage tap tempo

seangoodvibes:
Adjusted number of `TAP TEMPO` button presses to engage `TAP TEMPO` to `FOUR (4)` to avoid mistakingly changing tempo.

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2630

@ok-reza 

To do:
- [ ] add community features toggle


m-m-adams:
I'm not a huge fan of this - I like that tap tempo engages immediately, 4 presses feels too slow

m-m-adams:
Moving to 1.3 - I don't want to introduce a UI change this late in the beta. Even if it can be argued as a fix it's not from anything in this beta period or even the previous release, so I don't feel it justifies skipping going through nightly (and probably a community feature to disable it) 

The downside is it makes using tap tempo deliberately twice as slow so I don't want to just drop it

seangoodvibes:
> Moving to 1.3 - I don't want to introduce a UI change this late in the beta. Even if it can be argued as a fix it's not from anything in this beta period or even the previous release, so I don't feel it justifies skipping going through nightly (and probably a community feature to disable it) 
> 
> 
> 
> The downside is it makes using tap tempo deliberately twice as slow so I don't want to just drop it

Yup I'm good with this going to 1.3

Issue #2628, opened 2024-09-09
Enabling metronome is blocked while record count-in is occurring.


ok-reza:
### Please describe the problem:

This has always been present, since official.

If RECORD is armed, and you press PLAY, the count-in bars begin. During this count down, I often realize I want the metronome to be enabled before the recording starts. However, SHIFT+TAP TEMPO has no effect while record count-in is occurring.

This would just be a nice-to-have QoL improvement from when you forget last minute to arm the metronome before recording.

### What is the expected behavior?

SHIFT+TAP TEMPO enables or disabled metronome while RECORD COUNT-IN is occuring.

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

other

### What is the firmware name:

Official to Community Firmware

PR #2653, opened 2024-09-14
Fix/ Enable toggling of metronome during record count in

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2628

Enabled toggling of `METRONOME` during `RECORD COUNT-IN`.

@ok-reza


src/deluge/gui/views/view.cpp
```diff
@@ -158,7 +158,7 @@ ActionResult View::buttonAction(deluge::hid::Button b, bool on, bool inCardRouti
 				endMidiLearnPressSession();
 			}
 		}
-		else if (currentUIMode == UI_MODE_NONE) {
+		else if (currentUIMode == UI_MODE_NONE || currentUIMode == UI_MODE_RECORD_COUNT_IN) {
```
m-m-adams, line 161: does tap tempo work during the count in? I feel like changing tempo mid count in is unlikely to be intentional

seangoodvibes: > does tap tempo work during the count in? I feel like changing tempo mid count in is unlikely to be intentional

Good catch, just blocked tap tempo during record count in

PR #2667, opened 2024-09-16
Fix offline rendering in memory limited songs

### LINKED ISSUE #2573: Stem Exporting again

Hi, me again

I am still having trouble with stem exporting.

I'm in arranger view wanting to export the 12 instruments that make my up track.

Normalise is off (I've tried with on too)
Export to silence is off
FX are off

I press save and record and the Deluge exports.  12 files appear on the sd card as the same file size (in this case 64mb). 
 However when I import them into Reaper or ProTools, the majority of them are 5 seconds long.  Occasionally one is as long as the track should be, the rest are still 5 seconds long and silent.

Nothing is muted in Arranger view until each track is processed.

To be able to fully export all the instruments in arranger  view

_No response_

OLED

Nightly

deluge-v1_2_0-nightly+2024_09_03-36b2b72

Screenshot 2024-09-03 at 21 57 55

Screenshot 2024-09-03 at 22 04 01

m-m-adams:
Finishes all reads/writes before continuing to render during offline rendering

Fix #2573

### DIFF
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -981,18 +981,24 @@ void routine() {
 		}
 	}
 	else {
-		auto timeNow = getSystemTime();
-		while (getSystemTime() < timeNow + 32 / 44100.) {
-			size_t numSamples = 32;
-			int32_t timeWithinWindowAtWhichMIDIOrGateOccurs;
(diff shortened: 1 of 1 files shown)


m-m-adams:
I checked this against the songs in the issue and it seems to be a complete fix

seangoodvibes (approved):

PR #2682, opened 2024-09-18
[DRAFT] Enhancement/ Enable moving tracks and clips within tracks in grid view

seangoodvibes:
Added ability to move tracks in grid view by holding a selected clip and turning the Horizontal Encoder (<>)

Added ability to move clips within a track in grid view by holding a selected clip and turning the Vertical Encoder

Moved shortcut to change track colour in grid view to Shift + Hold Clip + Turn <>

Moved shortcut to change track colour gradient in grid view to Shift + Hold Clip + Turn Vertical Encoder

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -678,6 +678,11 @@ ActionResult SessionView::padAction(int32_t xDisplay, int32_t yDisplay, int32_t 
 	}
 
 	if (currentSong->sessionLayout == SessionLayoutType::SessionLayoutTypeGrid) {
+		// if we just moved a clip or a track, we want to ignore other presses, other than the release
+		// of the clip or track we selected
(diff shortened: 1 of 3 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -4278,41 +4296,263 @@ char const* SessionView::getMacroKindString(SessionMacroKind kind) {
...
+		else if (offsetX != 0) {
+			track->colour = static_cast<int16_t>(track->colour + (colourStep * offsetX) + 192) % 192;
+		}
+		requestRendering(this);
+	}
+
+	return ActionResult::DEALT_WITH;
+}
+
+ActionResult SessionView::gridHandleClipMove(int32_t offsetY) {
```
m-m-adams, line 4349: This is extremely difficult to read, and breaking the semantic meaning of the grid presses will lead to bugs (e.g. x/y no longer represent where a users finger is)

m-m-adams (commented):
This is a good idea but I think the approach needs to be changed

PR #2689, opened 2024-09-19
BugFix/ Remove Midi Loopback

seangoodvibes:
Removed the midi loop back documentation

To do:
- [x] Disable the feature
- [x] Remove loopback code altogether

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@
 
 ### MIDI
 
-#### <ins>General</ins>
+#### <ins>Sysex</ins>
 - Added Universal SysEx Identity response, including firmware version.
 
 #### <ins>MPE</ins>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,9 +44,16 @@
 - Added filters in FM synth mode. They're set to OFF by default, enable by changing them to any other mode using the menu or db/oct shortcut.
 
 ### MIDI
+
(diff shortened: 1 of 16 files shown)


m-m-adams:
I think we should delete the code - it's unlikely we'll be able to bring it back in its current form

seangoodvibes:
> I think we should delete the code - it's unlikely we'll be able to bring it back in its current form

Done

CHANGELOG.md
```diff
@@ -347,9 +347,6 @@
     - Clips not in scale mode are unaffected (similar to the existing transpose behaviour from the encoders).
     - Configureable in `SETTINGS > MIDI > TRANSPOSE` between chromatic and in-scale transposition.
 
-#### <ins>Loopback</ins>
-- Added a `MIDI LOOPBACK` mode, accessible in the SONG menu, which directs MIDI data from internal MIDI clips back to the Deluge input.
-
```
sapphire-arches, line 352: instead of just deleting the entry, we should add an entry that says it was removed.

seangoodvibes: > instead of just deleting the entry, we should add an entry that says it was removed.

Since we never officially released this feature I don't think we need to, unless it's a note for folks that use in the nightly and beta builds.

For 1.1 we removed the feature. 

It got left in the 1.1 notes in the nightly by mistake and should've been moved up to the 1.2 section.

PR #2691, opened 2024-09-20
[DRAFT] Enhancement/ Add selected clip pulsing in Song Grid View

seangoodvibes:
Draft until https://github.com/SynthstromAudible/DelugeFirmware/pull/2682 is merged :)

https://github.com/user-attachments/assets/f8209543-dad9-4ffd-bab1-9f85fadc0248

- The selected clip in `Song Grid View` (which is the last clip you interacted with), now shows a `pulsing / breathing` animation whereby the clip clip will faded in and out between it's clip colour and a blurred version of the clip's colour.

- I checked the voice count for a kit song that was culling and I did not notice any change between toggling the pulsing on or off.


src/deluge/gui/colour/rgb.h
```diff
@@ -178,6 +178,17 @@ class RGB {
...
+	 * @brief Compare two colours to determine if they're the same
+	 *
+	 * @param sourceA The first colour
+	 * @param sourceB The second colour
+	 * @return bool The true / false result if they're the equal
+	 */
+	[[nodiscard]] static constexpr bool isEqual(RGB sourceA, RGB sourceB) {
+		return ((sourceA.r == sourceB.r) && (sourceA.g == sourceB.g) && (sourceA.b == sourceB.b));
+	}
+
```
sapphire-arches, line 191: should implement `operator==` instead (and in this cause, you can probably do that using:
```suggestion
	bool operator==(RGB const &) = default;
```

sapphire-arches (commented):
We should check voice count with this PR, excessive pad rerending and sending to the PIC has caused us pain in the past.

src/deluge/gui/colour/rgb.h
seangoodvibes: gives me an error:

/Users/sean/Documents/GitHub/DelugeFirmware/src/deluge/gui/colour/rgb.h:185:14: error: defaulted 'bool RGB::operator==(const RGB&)' must be 'const'
  185 |         bool operator==(RGB const &) = default;

seangoodvibes:
> We should check voice count with this PR, excessive pad rerending and sending to the PIC has caused us pain in the past.

Will do

src/deluge/gui/colour/rgb.h
sapphire-arches: oops, need a `const` before the `=`

seangoodvibes: Awesome worked thanks!

PR #2696, opened 2024-09-21
[DRAFT] Polish/ Clip Settings Menu - Blink Selected Clip Pad

seangoodvibes:
Just some nice polish for the clip settings menu to identify menu context

While in the Clip Settings menu in grid view (which is accessed by holding an existing clip pad and pressing select), the pad for the clip selected for that menu will blink

### DIFF
--- a/src/deluge/gui/context_menu/clip_settings/clip_settings.cpp
+++ b/src/deluge/gui/context_menu/clip_settings/clip_settings.cpp
@@ -4,6 +4,7 @@
 #include "gui/l10n/l10n.h"
 #include "gui/ui/rename/rename_clipname_ui.h"
 #include "gui/ui/root_ui.h"
+#include "gui/ui_timer_manager.h"
 #include "gui/views/session_view.h"
 #include "hid/display/display.h"
 #include "model/clip/clip.h"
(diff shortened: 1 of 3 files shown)


src/deluge/gui/context_menu/clip_settings/clip_settings.cpp
```diff
@@ -62,6 +70,9 @@ bool ClipSettingsMenu::acceptCurrentOption() {
 			openUI(&launchStyle);
 		}
 		else {
+			// disable shortcut blinking in the rename UI because you don't see
+			// the clip pad on the grid
+			uiTimerManager.unsetTimer(TimerName::SHORTCUT_BLINK);
```
sapphire-arches, line 75: isn't there a specific function for clearing the blink state? otherwise I think the pad can get stuck on white.

seangoodvibes: > isn't there a specific function for clearing the blink state? otherwise I think the pad can get stuck on white.

nope for this one it's different. because the blinking uses the function PadLEDs::flashMainPad(x,y) I don't think it can get stuck on white.

src/deluge/gui/context_menu/clip_settings/clip_settings.h
```diff
@@ -14,8 +14,11 @@ class ClipSettingsMenu final : public ContextMenu {
 	bool setupAndCheckAvailability();
 	bool canSeeViewUnderneath() override { return true; }
 	bool acceptCurrentOption() override; // If returns false, will cause UI to exit
+	void focusRegained() override;
 
 	Clip* clip = nullptr;
+	int32_t clipX = -1;
```
m-m-adams, line 20: There's already tracker variables for the last grid presses, no need to create more

seangoodvibes: > There's already tracker variables for the last grid presses, no need to create more

I did it this way because I was thinking maybe we'd want to add the clip settings menu to the row view and blink the mute pad.

if we did that, then passing the X and Y instead of referencing session view's variables would make more sense

PR #2702, opened 2024-09-22
Feature - looper and sampler types for audio outputs

m-m-adams:
Creates modes for audio clips, player, sample and looper. Modes are changed by turning select in an audio clip view or while holding an audio clip in song view.

Player: Monitoring is off, overdubs work by cloning. This is intended to be used for playing a static audio file or recording without monitoring.

Sampler: Monitoring is on, monitoring turns off after recording, overdubs work by cloning. This is intended to be used for sampling from an ongoing audio stream where you only want to hear the part you're looping afterwards

Looper: Monitoring is on and remains on. Overdubs do a real overdub in place. This is intended for live looping or use as an fx processor

I've slightly modified copying as well - if the clip is empty then it works the same way as current copying (target clip becomes a player), if the clip has contents then the original clip becomes a player and the new clip takes its sampler/looper settings. This makes copying an output to make new loops much smoother

Removed: Moving a clip to another output from rows view. This didn't do anything useful in rows view (can't change section of the clip or move it to an output that already has a clip in the section) and it's still possible from arranger.

### DIFF
--- a/src/deluge/gui/context_menu/audio_input_selector.cpp
+++ b/src/deluge/gui/context_menu/audio_input_selector.cpp
@@ -28,27 +28,15 @@ namespace deluge::gui::context_menu {
 
 enum class AudioInputSelector::Value {
 	OFF,
-
 	LEFT,
-	LEFT_ECHO,
-
 	RIGHT,
-	RIGHT_ECHO,
-
 	STEREO,
-	STEREO_ECHO,
-
 	BALANCED,
-	BALANCED_ECHO,
-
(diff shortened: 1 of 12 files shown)


seangoodvibes (approved):

Evilpilli:
Might be my favorite new feature! A monitor type would be fantastic too.
Monitor always on
When copied, the pasted clip becomes player. 
Selecting the mode will remove any effects/filters/automation etc. To provide "clean" audio signal.

PR #2708, opened 2024-09-23
enable grain fx all the time

m-m-adams:
Grain isn't particularly heavy with the scheduler so just enable it all the time


m-m-adams:
It's actually even better post the refactoring in my modfx pr so let's wait. I'm not sure we really want it in modcontrollable either, we might want to move it to global effectable since the grain object itself is a few hundred bytes

PR #2712, opened 2024-09-25
Feature/warble

m-m-adams:
Adds an LFO I've named "Warble", it's a second order filtered sample and hold that produces a random wave something like this:

image

Adds a Mod fx type I've called "Warbler", which uses this LFO shape to make a flanger/chorus/whatever based modulation abomination that makes super cool lofi warbley noises. It essentially consists of two things, a randomly drifting vibrato and a comb filter. Controls are the normal rate/depth/feedback/offset. 

To make a tape warble type thing set rate to 15ish, depth to 5-10, feedback 0-2 and offset at 25 

With depth at 0 and feedback at 0 it's not doing anything. If you turn up feedback you'll start to comb filter the signal, offset controls the length of the comb filter. Turning depth up adds modulation to that delay time (like a chorus or flanger) but since the modulation is unpredictable it sounds more "warm" "analog" "chewy" "lofi" 

Turning rate up makes it go faster

Turning feedback up makes it get super weird

Turning offset adds phasing (unless feedback is 0, in which case it does nothing)

Also labels the grainfx controls properly since I was touching those menus anyway


seangoodvibes (approved):

seangoodvibes:
lgtm once you add documentation thx!

PR #2719, opened 2024-09-28
Feature/ Make default and last clip type configurable (#2716)

seangoodvibes:
* Make default and last clip type configurable

Add defaults menu to configure default clip type and whether to use last clip type

* Update CHANGELOG.md

### DIFF
--- a/src/deluge/gui/menu_item/defaults/ui/clip_type/default_new_clip_type.h
+++ b/src/deluge/gui/menu_item/defaults/ui/clip_type/default_new_clip_type.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2014-2023 Synthstrom Audible Limited
+ *
+ * This file is part of The Synthstrom Audible Deluge Firmware.
+ *
+ * The Synthstrom Audible Deluge Firmware is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+#pragma once
+#include "definitions_cxx.hpp"
+#include "gui/l10n/l10n.h"
+#include "gui/menu_item/selection.h"
+#include "storage/flash_storage.h"
+#include "util/misc.h"
+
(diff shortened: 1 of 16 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -94,6 +94,7 @@ SessionView sessionView{};
 
 SessionView::SessionView() {
 	xScrollBeforeFollowingAutoExtendingLinearRecording = -1;
+	createClip = false;
```
m-m-adams, line 97: what is this used for?

src/deluge/gui/menu_item/defaults/ui/clip_type/default_new_clip_type.h
```diff
@@ -0,0 +1,37 @@
...
+namespace deluge::gui::menu_item::defaults::ui::clip_type {
+class DefaultNewClipType final : public Selection {
+public:
+	using Selection::Selection;
+	void readCurrentValue() override { this->setValue(FlashStorage::defaultNewClipType); }
+	void writeCurrentValue() override { FlashStorage::defaultNewClipType = this->getValue<OutputType>(); }
+	deluge::vector<std::string_view> getOptions() override {
+		using enum l10n::String;
+		return {
+		    "SYNTH", "KIT", "MIDI", "CV", "AUDIO",
```
m-m-adams, line 33: this should be using the output type enum, this will break if they're reordered

src/deluge/gui/views/session_view.cpp
seangoodvibes: It's used after the clip creation yield to know if you should proceed with clip creation or not - eg if you cancelled out then it doesn't create a clip

src/deluge/gui/menu_item/defaults/ui/clip_type/default_new_clip_type.h
seangoodvibes: > this should be using the output type enum, this will break if they're reordered

will fix this is in the main branch and cherry pick after

src/deluge/gui/views/session_view.cpp
m-m-adams: I meant the line - it's already initialized to false (that's what createClip{false} does). Is there a case where that's not working?

src/deluge/gui/menu_item/defaults/ui/clip_type/default_new_clip_type.h
m-m-adams: don't bother cherry picking, this branch isn't getting future development so it doesn't matter

src/deluge/gui/views/session_view.cpp
seangoodvibes: > I meant the line - it's already initialized to false (that's what createClip{false} does). Is there a case where that's not working?

ah yes not necessary, just an oversight i'll remove it

m-m-adams:
I'd like another review before we merge this just because we're skipping nightly and any user testing

PR #2721, opened 2024-09-29
[DRAFT] Feature/ Add midi cc labels + midi cc #'s / name's to mod encoder popups

seangoodvibes:
- Cleaned up rename UI classes
- Added ability to rename MIDI CC's in MIDI clips. Changed are saved by Instrument (e.g. per MIDI channel). Changes can be saved to a MIDI preset and with the song.
  - To rename a `MIDI CC`, enter `Automation View` and select the CC you wish to rename using the `Select Encoder`.
  - With the `MIDI CC` selected, press down either `Gold (Mod) Encoder` and then press the `Name` grid shortcut to open the `MIDI CC Renaming UI`. Enter a new name and press `Select` or `Enter`.
  - The new `MIDI CC` name will be immediately visible in `Automation View` and you can assign that CC to a `Gold Knob` in the MIDI clip and it will show that name when you press down on the `Gold (Mod) Encoder`.
- Added MIDI CC numbers and labels to `Gold (Mod) Encoder` popups.

### DIFF
--- a/src/deluge/gui/ui/rename/rename_midi_cc_ui.cpp
+++ b/src/deluge/gui/ui/rename/rename_midi_cc_ui.cpp
@@ -15,30 +15,43 @@
  * If not, see <https://www.gnu.org/licenses/>.
  */
 
-#include "gui/ui/rename/rename_clipname_ui.h"
+#include "gui/ui/rename/rename_midi_cc_ui.h"
 #include "definitions_cxx.hpp"
 #include "extern.h"
 #include "gui/l10n/l10n.h"
-#include "gui/views/instrument_clip_view.h"
+#include "gui/views/automation_view.h"
 #include "hid/buttons.h"
 #include "hid/display/display.h"
 #include "hid/led/pad_leds.h"
+#include "model/instrument/midi_instrument.h"
 #include "model/output.h"
 #include "model/song/song.h"
 
-RenameClipNameUI renameClipNameUI{};
+RenameMidiCCUI renameMidiCCUI{};
 
-RenameClipNameUI::RenameClipNameUI() {
-	title = "Clip Name";
+RenameMidiCCUI::RenameMidiCCUI() {
+	title = "CC Name";
 }
 
-bool RenameClipNameUI::opened() {
+bool RenameMidiCCUI::opened() {
 	bool success = QwertyUI::opened();
 	if (!success) {
 		return false;
 	}
 
-	enteredText.set(&clip->clipName);
+	Clip* clip = getCurrentClip();
+
+	int32_t cc = clip->lastSelectedParamID;
+
+	// if we're not dealing with a real cc number
+	// then don't allow user to edit the name
+	if (cc < 0 || cc >= kNumRealCCNumbers) {
+		return false;
+	}
+
+	MIDIInstrument* midiInstrument = (MIDIInstrument*)clip->output;
+
+	midiInstrument->getNameFromCC(cc, &enteredText);
 
 	displayText();
 
@@ -47,12 +60,12 @@ bool RenameClipNameUI::opened() {
 	return true;
 }
 
-bool RenameClipNameUI::getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) {
+bool RenameMidiCCUI::getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) {
 	*cols = 0b11;
 	return true;
 }
 
-ActionResult RenameClipNameUI::buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) {
+ActionResult RenameMidiCCUI::buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) {
 	using namespace deluge::hid::button;
 
 	// Back button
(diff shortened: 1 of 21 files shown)


m-m-adams:
Could we make the pad assignable as well? So use the midi follow CC layout as a default but let other pad to cc mappings be created

seangoodvibes:
> Could we make the pad assignable as well? So use the midi follow CC layout as a default but let other pad to cc mappings be created

Yes very good idea! Do you want that as part of this PR? I want to do that next.

src/deluge/gui/ui/rename/rename_midi_cc_ui.cpp
```diff
@@ -0,0 +1,146 @@
...
+		return false;
+	}
+
+	Clip* clip = getCurrentClip();
+
+	int32_t cc = clip->lastSelectedParamID;
+
+	// if we're not dealing with a real cc number
+	// then don't allow user to edit the name
+	if (cc < 0 || cc >= kNumRealCCNumbers) {
```
m-m-adams, line 48: CC1 needs to be blocked too (it's mod wheel and stored internally as cc 130 or something)

src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -1121,3 +1164,29 @@ void MIDIInstrument::sendNoteToInternal(bool on, int32_t note, uint8_t velocity,
...
+int32_t MIDIInstrument::getCCFromName(String* name) {
+	for (int32_t i = 0; i < kNumRealCCNumbers; i++) {
+		if (ccNames[i].equalsCaseIrrespective(name)) {
+			return i;
+		}
+	}
+	return 255;
+}
+
+void MIDIInstrument::getNameFromCC(int32_t cc, String* name) {
```
m-m-adams, line 1179: Since this can't error it would be clearer to return a string instead of taking a return pointer as an input

src/deluge/model/instrument/midi_instrument.h
```diff
@@ -45,6 +45,14 @@ class MIDIInstrument final : public NonAudioInstrument {
 	bool readTagFromFile(Deserializer& reader, char const* tagName);
 	Error readModKnobAssignmentsFromFile(int32_t readAutomationUpToPos,
 	                                     ParamManagerForTimeline* paramManager = nullptr);
+
+	// cc names
+	String ccNames[kNumRealCCNumbers];
```
m-m-adams, line 50: Couple things - 

1. This is a 600 byte array, it would be better to store it externally instead of as part of the output
2. If you make it a std array then it can use iterators instead of a for loop to check all the names. Then this can change to a pointer to an array, and you can allocate the actual array in the midi output constructor
3. This should be private to enforce access via the api functions you wrote

seangoodvibes: > Couple things -
> 
> 1. This is a 600 byte array, it would be better to store it externally instead of as part of the output
> 2. If you make it a std array then it can use iterators instead of a for loop to check all the names. Then this can change to a pointer to an array, and you can allocate the actual array in the midi output constructor
> 3. This should be private to enforce access via the api functions you wrote

Will do thanks!

alifeinbinary:
Hey @seangoodvibes, I pulled this PR and configured some synths. It's working really well, great job! The only snag I found was when clearing out a value I would get a warning saying "No duplicates" or something to that effect. Are the values placed in a hash set? I suppose it makes sense to check for duplicate values, however, null value or empty string should be allowed in case a user enters a value at the incorrect key and needs to clear it out. Otherwise, I think this implementation is solid. I've been anticipating this feature for some time, so I really appreciate you putting the time in to make it possible. Thank you!

seangoodvibes:
> Hey @seangoodvibes, I pulled this PR and configured some synths. It's working really well, great job! The only snag I found was when clearing out a value I would get a warning saying "No duplicates" or something to that effect. Are the values placed in a hash set? I suppose it makes sense to check for duplicate values, however, null value or empty string should be allowed in case a user enters a value at the incorrect key and needs to clear it out. Otherwise, I think this implementation is solid. I've been anticipating this feature for some time, so I really appreciate you putting the time in to make it possible. Thank you!

Thanks

There's still a bunch of work to do on this unfortunately. The recommended implementation is to save the labels to a file separate from the midi preset file and song so that they can be shared.

Not 100% sure how I'm going to do all that yet.

So I guess this PR is more of a proof of concept of how labels could work and be edited on the deluge.

Hopefully I can get back to this soon-ish.

alifeinbinary:
@seangoodvibes What problems are there with this current approach? Copying/creating MIDI presets in the ./MIDI directory and then assigning them within the Deluge UI works well currently. I created a few preset files for my devices and had ChatGPT fill in the values matching the keys to CC# in the respective PDF manuals for said devices and then loaded them, worked great. It makes sense to abstract the <modKnobs /> tag into the preset file as you have done to keep all MIDI preset variables in one location. I would add that a label property on <modKnob /> would make development of a preset editor tool more robust. 

I'll continue rocking this PR and see if I can stress test it.

seangoodvibes:
I was ok with this approach but a couple others voiced concerns about on
the discord:

You can see what was discussed here:

https://discord.com/channels/608916579421257728/1211653958662234184

Basically I think the chief concern is that the labels are saved in a
preset file that contains other midi instrument settings that you may not
want. So you have no way of loading the labels into the deluge without
loading those other settings if you save it in the preset as it is
currently done.

There was also concern about wanting to be able to update labels and have
the changes reflected in every song that uses that device. The current
approach would not allow that.

The recommendation is to:

- save the labels to the song and use them if the device label file cannot
be found
- if a device label file can be found, use that instead.

So I need to add saving the labels to another file

And then I need to somehow connect the song / midi preset files to that
other file.

So that’s the work that I need to do.

Nothing else would change from the perspective of how the labels work once
they’re loaded or how you would edit them on the deluge - eg I would still
use automation view as the UI for editing, I would still show labels in
automation view and on the gold knobs, etc etc

seangoodvibes:
> @seangoodvibes What problems are there with this current approach? Copying/creating MIDI presets in the ./MIDI directory and then assigning them within the Deluge UI works well currently. I created a few preset files for my devices and had ChatGPT fill in the values matching the keys to CC# in the respective PDF manuals for said devices and then loaded them, worked great. It makes sense to abstract the `<modKnobs />` tag into the preset file as you have done to keep all MIDI preset variables in one location. I would add that a label property on `<modKnob />` would make development of a preset editor tool more robust. 
> 
> 
> 
> I'll continue rocking this PR and see if I can stress test it. 

I was ok with this approach but a couple others voiced concerns about on the discord:

You can see what was discussed here:

https://discord.com/channels/608916579421257728/1211653958662234184

Basically I think the chief concern is that the labels are saved in a preset file that contains other midi instrument settings that you may not
want. So you have no way of loading the labels into the deluge without loading those other settings if you save it in the preset as it is currently done.

There was also concern about wanting to be able to update labels and have the changes reflected in every song that uses that device. The current
approach would not allow that.

The recommendation is to:

- save the labels to the song and use them if the device label file cannot be found
- if a device label file can be found, use that instead.

So I need to add saving the labels to another file

And then I need to somehow connect the song / midi preset files to that other file.

So that’s the work that I need to do.

Nothing else would change from the perspective of how the labels work once they’re loaded or how you would edit them on the deluge - eg I would still use automation view as the UI for editing, I would still show labels in automation view and on the gold knobs, etc etc

alifeinbinary:
Fair point. I appreciate your response. Having consistency across projects would be a high priority, I agree. Hadn't realised that was a shortfall of this approach as I'd only used my Default.XML file when testing this PR. I'll resume this in the Discord thread and read what the others have written.

PR #2723, opened 2024-09-29
UI Consistency/ Swapped grid view section and mode columns

seangoodvibes:
Swapped section and mode sidebar columns for further consistency with `SONG ROW VIEW`.

### DIFF
--- a/src/deluge/gui/context_menu/configure_song_macros.cpp
+++ b/src/deluge/gui/context_menu/configure_song_macros.cpp
@@ -32,7 +32,7 @@ namespace deluge::gui::context_menu {
 ConfigureSongMacros configureSongMacros{};
 
 bool ConfigureSongMacros::getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) {
-	*cols = 0x01; // Only mode (audition) column
+	*cols = 0x01; // Only section (audition) column
 	*rows = 0x0;
 	return true;
 }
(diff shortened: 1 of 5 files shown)


m-m-adams:
I don't think this makes anything more consistent until we've removed the modes and loop buttons. As is it loses the visual connection to the rest of the grid, the column of modes breaks up the grid metaphor and it doesn't look like the section column on the right is associated with the main pads

src/deluge/gui/context_menu/configure_song_macros.cpp
```diff
@@ -83,11 +83,11 @@ ActionResult ConfigureSongMacros::padAction(int32_t x, int32_t y, int32_t on) {
 	if (sdRoutineLock) {
 		return ActionResult::REMIND_ME_OUTSIDE_CARD_ROUTINE;
 	}
-	// don't allow user to switch modes
```
m-m-adams, line 86: this looks wrong? Why is the behaviour changing so we can switch modes but not sections now

src/deluge/gui/context_menu/midi_learn_mode.cpp
```diff
@@ -86,10 +86,10 @@ ActionResult MidiLearnMode::padAction(int32_t x, int32_t y, int32_t on) {
 		return ActionResult::REMIND_ME_OUTSIDE_CARD_ROUTINE;
 	}
 	// don't allow user to switch modes
-	if (x <= kDisplayWidth) {
+	if (x < kDisplayWidth || x > kDisplayWidth) {
```
m-m-adams, line 89: !=

src/deluge/gui/views/performance_session_view.cpp
```diff
@@ -1021,38 +1021,33 @@ ActionResult PerformanceSessionView::padAction(int32_t xDisplay, int32_t yDispla
...
+							// long press from grid mode - it just peeks performance view)
+							if (!on
+							    && ((AudioEngine::audioSampleTimer - timeGridModePress) >= FlashStorage::holdTime)) {
 								releaseViewOnExit(modelStack);
 								changeRootUI(&sessionView);
-								sessionView.gridHandlePads(xDisplay, yDisplay, on);
 							}
 						}
+						// if you pressed the green or blue mode pads, go back to grid view and change mode
+						else if ((yDisplay == 7) || (yDisplay == 6)) {
```
m-m-adams, line 1046: use the grid mode enum

PR #2728, opened 2024-09-29
[DRAFT] Refactor how the 'iteration' parameter is saved for each note from a list of fixed presets to fully customizable iteration

soymonitus:
The idea is to allow later for a view similar to velocity view where you can tap and select the divisor and the enabled steps, allowing you to enter more complex iteration configs for each note like: 1 and 8 out of 8 (=play first and last bars out of 8) ,... or 2, 3 and 4 out of 4 (=play last 3 bars but not the first), etc.

Plan is to use an int16 for the iteration value and store the divisor in the last 8 bits, and use the first 8 bits as booleans to tell if an iteration is enabled or not:

To encode the two previous examples, they will look like this:
000 1000 1000 0001 = this would be divisor = 8, and play bars 1st and 8th
000 0100 0000 1110 = this would be divisor = 4, and play bars 2nd, 3rd and 4th

- [ ] Move the custom iterance menu inside the iterance parameter when clicking select while CUSTOM is selected
- [x] Check if all menus update when changing selected note or row
- [ ] Cleanup
- [ ] Documentation


soymonitus:
@seangoodvibes  this is the idea i had for the iteration parameter saved for a note, let me know what you think

seangoodvibes:
> @seangoodvibes this is the idea i had for the iteration parameter saved for a note, let me know what you think

Seems like you have a good idea of what to do. It does seem like it will be a bit complicated to ensure backwards compatibility but I think it will be worth it to have more options for iterance.

We'll definitely need to test it out a lot to ensure it's working correctly once you're done as I imagine you'll need to do some refactoring of how the existing logic works.

Regarding changing the variable to int16_t. It would worthwhile running that by the other folks on the team to ensure they agree with the logic behind it.

Thanks

soymonitus:
Do we have to keep backwards compatibility with current nightly? I thought nightly was ever changing. Anyway if there is a need to be backwards compatible with nightly it is easy as an int8 will translates in a divisor of value zero which would mean that the value is an old preset and must be transformed into the new custom value

seangoodvibes:
> Do we have to keep backwards compatibility with current nightly? I thought nightly was ever changing. Anyway if there is a need to be backwards compatible with nightly it is easy as an int8 will translates in a divisor of value zero which would mean that the value is an old preset and must be transformed into the new custom value

No not the nightly, was referring to backwards compatibility with 1.2, 1.1, etc.

good to know you have an idea on how to translate the old value into the new value system :)

src/deluge/model/note/note_row.cpp
```diff
@@ -3351,9 +3362,18 @@ Error NoteRow::readFromFile(Deserializer& reader, int32_t* minY, InstrumentClip*
...
+					fill = hexToIntFixedLength(&hexChars[26], 2);
+					iterance = sanitizeIterance(hexToIntFixedLength(&hexChars[22], 4));
+					probability = hexToIntFixedLength(&hexChars[20], 2);
+					lift = hexToIntFixedLength(&hexChars[18], 2);
+					if (lift == 0 || lift > 127) {
+						goto useDefaultLift;
+					}
+				}
+				else if (noteHexLength == 26) { // if nightly firmware 1.3 with no custom iterances
+					fill = hexToIntFixedLength(&hexChars[26], 2);
```
seangoodvibes, line 3375:
```suggestion
					fill = hexToIntFixedLength(&hexChars[24], 2);
```

I think this is supposed to be 24

src/deluge/model/note/note_row.cpp
```diff
@@ -3351,9 +3362,18 @@ Error NoteRow::readFromFile(Deserializer& reader, int32_t* minY, InstrumentClip*
 				uint8_t velocity = hexToIntFixedLength(&hexChars[16], 2);
 				uint8_t lift, probability, iterance, fill;
 
-				if (noteHexLength == 26) { // if reading iterance and fill
-					fill = hexToIntFixedLength(&hexChars[24], 2);
-					iterance = hexToIntFixedLength(&hexChars[22], 2);
+				if (noteHexLength == 28) { // if reading custom iterance and fill
```
seangoodvibes, line 3365: I think you need to update what sets the noteHexLength further up

src/deluge/model/note/note_row.cpp
```diff
@@ -3471,6 +3491,12 @@ getOut: {}
 			goto doReadNoteData;
 		}
 
+		// Notes stored as hex data including iterance and fill (community firmware 1.3 onwards)
```
seangoodvibes, line 3494:
```suggestion
		// Notes stored as hex data including custom iterance and fill (community firmware 1.3 onwards)
```

src/deluge/model/note/note_row.cpp
```diff
@@ -3471,6 +3491,12 @@ getOut: {}
 			goto doReadNoteData;
```
seangoodvibes, line 3491: non-custom iterance and fill*

src/deluge/model/note/note_row.cpp
soymonitus: iterance now has 2 bytes, so that's why the code is different

soymonitus:
@seangoodvibes I have everything implemented, you can now edit custom iterance from the menu. The thing is that i added it under the same level of iterance, fill and probability. But I want to move it to when you click Select knob while CUSTOM is selected as iterance. I know I have to do it using the **MenuItem* selectButtonPress() override** (it is commented out in iterance.h). But I don't know exactly how to create the **noteCustomIteranceRootMenu** submenu so I can reference it inside iterance class. Could you help me by proposing changes or even making commits?

PR #2743, opened 2024-10-01
[DRAFT] Update OLED to stop scrolling during playback

seangoodvibes:
Updated OLED display to stop scrolling song names (in session/arranger) and preset names (in clip views) while playback is running

- [ ] Scroll once before stopping scroll


soymonitus:
Does this do at least one scrolling for presets if playback active? I would like to see the full name while playback is on, at least once

seangoodvibes:
> Does this do at least one scrolling for presets if playback active? I would like to see the full name while playback is on, at least once

Ah yes it was supposed to do that. Need to add that

PR #2771, opened 2024-10-05
Feature/cv expression velocity

m-m-adams:
Adds the ability to set a CV instrument to use both 1 and 2 channels, which makes the cv2 source selectable between mod wheel, velocity, and aftertouch


seangoodvibes:
I documented the CV expression PR, there's just one more menu bug left that I want to fix before it's merged

It relates to that += 1 line you added for skipping over pitch mode

it's causing the menu to not select the correct current mode

The way it should work is:

if you're incrementing the menu item, skip over pitch mode only if current mode is OFF
if you're decrementing the menu item, skip over pitch mode only if current mode is MOD

otherwise don't do the +/- 1 adjustment

And if you're loading the menu for the first time and the mode is pitch for some reason, i would have it set the mode to off

seangoodvibes (approved):

PR #2788, opened 2024-10-06
Added LPF filter to MI reverb

markyounq:
WIP adding LPF filter to the Mutable Instruments reverb engine.


src/deluge/dsp/reverb/reverb.hpp
```diff
@@ -103,5 +110,6 @@ class Reverb : reverb::Base {
 	float damping_;
 	float width_;
 	float hpf_;
+	float lpf_;
```
m-m-adams, line 113: This isn't initialized before it's used by setModel to set the lpf frequency - that's probably why changing the initializer values doesn't work

m-m-adams:
Could you add saving/loading code before we merge?

It would also be nice to have this as a feature on both reverbs as a follow up. It would require refactoring so that the specific reverb process function writes to an intermediate buffer where we run the filters

src/deluge/dsp/reverb/mutable/reverb.hpp
```diff
@@ -143,6 +147,12 @@ class Mutable : public Base {
 
 	[[nodiscard]] float getHPF() const { return hp_cutoff_val_; }
 
+	void setLPF(float f) {
+		lp_cutoff_val_ = f;
+		lp_cutoff_ = calcFilterCutoff(f);
```
m-m-adams, line 152: This will make the maximum cutoff 520 Hz which is probably not intended - it's specific to turning a 0-1 knob turn into an appropriate HPF setting of 20-520 Hz with an exponential curve

m-m-adams: you could move the scaler variables to be function parameters and use the same function for both?

markyounq:
> Could you add saving/loading code before we merge?

I'm not sure what you mean. Saving/loading the reverb LPF cutoff? 

> It would also be nice to have this as a feature on both reverbs as a follow up. It would require refactoring so that the specific reverb process function writes to an intermediate buffer where we run the filters

Agreed but I don't think I'll cover it in this PR

m-m-adams:
> I'm not sure what you mean. Saving/loading the reverb LPF cutoff?

Yeah just needs to be added to the reverb block in the song-> read/write from file functions here https://github.com/SynthstromAudible/DelugeFirmware/blob/ba1581ca4a00a9712b483084a3eb10c7a9f17775/src/deluge/model/song/song.cpp#L1144

src/deluge/dsp/reverb/base.hpp
```diff
@@ -14,12 +14,25 @@ struct Base {
 		amplitude_right_ = amplitude_right;
 		amplitude_left_ = amplitude_left;
 	}
-
-	static constexpr float calcFilterCutoff(float f) {
+	enum class FilterType {
+		LowPass,
+		HighPass
+	};
+	static constexpr float calcFilterCutoff(float f, FilterType filtertype) {
```
stellar-aria, line 21:
```suggestion
	template <FilterType filtertype>
	static constexpr float calcFilterCutoff(float f) {
```
Instead of passing FilterType as a parameter, this should be a template parameter so that the branches below can be constexpr and not eval'd at runtime

src/deluge/dsp/reverb/base.hpp
```diff
@@ -14,12 +14,25 @@ struct Base {
...
+		float minFreq;
+		float maxFreq;
 		// this exp will be between 1 and 4.48, half the knob range is about 2
-		// the result will then be from 0 to 500Hz with half the knob range at 300hz
+		// for the HPF the result will then be from 0 to 500Hz with half the knob range at 300hz
+		// for the LPF the result will be from 2300 to 20000 with half the knob range at 7978.537hz
 		// then shift to 20-520Hz as there is a low end buildup in the reverb that should always be filtered out
-		float fc_hz = 20 + (std::exp(1.5f * f) - 1) * 150;
+		if (filtertype == FilterType::LowPass) {
```
stellar-aria, line 28:
```suggestion
		if constexpr (filtertype == FilterType::LowPass) {
```

src/deluge/dsp/reverb/base.hpp
```diff
@@ -14,12 +14,25 @@ struct Base {
...
+			minFreq = 2300.0f;
+			maxFreq = 5083.74f;
+		} else if (filtertype == FilterType::HighPass) {
```
stellar-aria, line 31:
```suggestion
		} else if constexpr (filtertype == FilterType::HighPass) {
```

src/deluge/dsp/reverb/mutable/reverb.hpp
```diff
@@ -142,14 +142,16 @@ class Mutable : public Base {
 
 	void setHPF(float f) {
 		hp_cutoff_val_ = f;
-		hp_cutoff_ = calcFilterCutoff(f);
+		FilterType type = FilterType::HighPass;
+		hp_cutoff_ = calcFilterCutoff(f, type);
```
stellar-aria, line 146:
```suggestion
		hp_cutoff_ = calcFilterCutoff<FilterType::HighPass>(f);
```

src/deluge/dsp/reverb/mutable/reverb.hpp
```diff
@@ -142,14 +142,16 @@ class Mutable : public Base {
...
 	}
 
 	[[nodiscard]] float getHPF() const { return hp_cutoff_val_; }
 
 	void setLPF(float f) {
 		lp_cutoff_val_ = f;
-		lp_cutoff_ = calcFilterCutoff(f);
+		FilterType type = FilterType::LowPass;
+		lp_cutoff_ = calcFilterCutoff(f, type);
```
stellar-aria, line 154:
```suggestion
		lp_cutoff_ = calcFilterCutoff<FilterType::LowPass>(f, type);
```

stellar-aria (changes requested):

src/deluge/dsp/reverb/base.hpp
m-m-adams: Oh nice, I've been trying to find a clean way to do this

markyounq:
@stellar-aria thanks for the suggestions! That works great. For some reason the 4th suggestion commit didn't work, so I added it manually and re-pushed.

stellar-aria (commented):
You might need to run format to clean up a couple indentation things, but otherwise LGTM

stellar-aria (approved):

PR #2789, opened 2024-10-07
Enable shift in note editor

seangoodvibes:
Enable shift in note editor


m-m-adams:
This makes pressing the button work but still isn't affecting the scroll speed

PR #2796, opened 2024-10-09
Add exporting arrangement as single stem option

seangoodvibes:
Added stem export configuration option to export a single arrangement stem composing of all unmuted tracks in arranger view.


src/deluge/processing/stem_export/stem_export.cpp
```diff
@@ -316,6 +329,42 @@ int32_t StemExport::exportInstrumentStems(StemExportType stemExportType) {
 	return totalNumOutputs;
 }
 
+/// iterates through all instruments, checking if there's any that should be exported (unmuted)
+/// then exports them all as a single arrangement stem
+/// simulates the button combo action of pressing record + play twice to enable resample
+/// and stop recording at the end of the arrangement
+int32_t StemExport::exportArrangementStem(StemExportType stemExportType) {
+	// prepare all the instruments for stem export
+	int32_t totalNumOutputs = disarmAllInstrumentsForStemExport();
```
m-m-adams, line 338: it would be nice to add master track as a new stem export type and skip this entirely. That would avoid the messy logic in the disarming function

src/deluge/processing/stem_export/stem_export.cpp
```diff
@@ -246,14 +250,23 @@ int32_t StemExport::disarmAllInstrumentsForStemExport() {
 				else {
 					output->exportStem = false;
 				}
-				output->mutedInArrangementModeBeforeStemExport = output->mutedInArrangementMode;
```
m-m-adams, line 249: better to leave this logic alone and skip the disarming step entirely using a new export type

m-m-adams:
Seems to work, just a bit hard to follow. I made some suggestions that will probably make it easier to support more export types in future if needed

m-m-adams (approved):

PR #2808, opened 2024-10-14
Feature/ Threshold recording

seangoodvibes:
- Added `THRESHOLD RECORDING` default setting which can be turned on in the `SETTINGS > RECORDING > THRESHOLD RECORDING (THRE) > MODE` submenu
  - When this is enabled, recording of external audio input (e.g. Line In / Microphone) will only begin when it detects an audio signal according the Recording Threshold mode that has been set. There are four Threshold Recording modes that can be selected from:
    - `DISABLED (OFF)`: Standard Deluge behaviour. Recording starts right away.
    - `LOW`: Sets a lower threshold for lower input signals.
    - `MEDIUM`: Sets a medium threshold that is good for most signals (including the internal microphone with gain off).
    - `HIGH`: Sets a high threshold that is good for noisy signals / microphones with gain
  - You can also view and temporarily change the current threshold recording setting as follows:
    - Press and hold `RECORD` + turn  `SELECT`
    - Enter the Song menu while in Song or Arranger View by pressing `SELECT` and entering the `SONG > THRESHOLD RECORDING (THRE) > MODE` submenu


src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -1049,9 +1092,22 @@ void SampleRecorder::feedAudio(int32_t* __restrict__ inputAddress, int32_t numSa
...
+			}
+			// if we're threshold recording and didn't detect audio in previous cycles
+			// check if there's any audio in this cycle
+			else {
+				StereoFloatSample approxRMSLevel =
+				    sample->envelopeFollower.calcApproxRMS((StereoSample*)beginInputNow, numSamplesThisCycle);
+				if (std::max(approxRMSLevel.l, approxRMSLevel.r) > startValueThreshold) {
+					writePos = writePosNow;
+					numSamplesCaptured += numSamplesThisCycle;
+					sample->audioStartDetected = true;
```
m-m-adams, line 1108: You probably want to reset the envelope follower too

src/deluge/model/sample/sample.h
```diff
@@ -140,6 +143,8 @@ class Sample final : public AudioFile {
 
 	SampleClusterArray clusters;
 
+	AbsValueFollower envelopeFollower{};
```
m-m-adams, line 146: How big are these? We dont really need one per sample, might be better to just allocate one when record is activated. Maybe even just have a single one that's always re used for threshold sampling?

src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -1049,9 +1095,23 @@ void SampleRecorder::feedAudio(int32_t* __restrict__ inputAddress, int32_t numSa
...
+			// if we're threshold recording and didn't detect audio in previous cycles
+			// check if there's any audio in this cycle
+			else {
+				StereoFloatSample approxRMSLevel =
+				    envelopeFollower.calcApproxRMS((StereoSample*)beginInputNow, numSamplesThisCycle);
+				if (std::max(approxRMSLevel.l, approxRMSLevel.r) > startValueThreshold) {
+					writePos = writePosNow;
+					numSamplesCaptured += numSamplesThisCycle;
+					sample->audioStartDetected = true;
+					envelopeFollower = {};
```
m-m-adams, line 1112: I think there's a reset method, it's just to set its current level to zero. Otherwise it would still be high when the next sample is started

seangoodvibes: > I think there's a reset method, it's just to set its current level to zero. Otherwise it would still be high when the next sample is started

mm I don't see a reset method in the AbsValueFollower class

I see a setup function, would you just feed 0 to that?

// takes in knob positions in the range 0-ONE_Q31
void AbsValueFollower::setup(q31_t a, q31_t r) {
	setAttack(a);
	setRelease(r);
}

m-m-adams: Oops, guess it didn't make it over from the compressor code. it's just something like this 

``` 
	/// Reset the state of the follower
	void reset() {
		lastMeanl = 0;
		lastMeanr = 0;
	}
```

seangoodvibes: > Oops, guess it didn't make it over from the compressor code. it's just something like this
> 
> ```
> 	/// Reset the state of the follower
> 	void reset() {
> 		lastMeanl = 0;
> 		lastMeanr = 0;
> 	}
> ```

perfect, implemented that thanks!

m-m-adams (approved):

PR #2810, opened 2024-10-15
Feature // Load next song MIDI command

soymonitus:
This adds a new learnable midi command (in button LOAD) that will autoload next song in the current folder

- [x] Fix issue where the loaded song still displays the info of the previous song
- [x] Fix crash when receiving the midi command while song is in the process of being loaded
- [x] Make it ignore commands while processing another auto-next command
- [x] Make it skip folders
- [x] Add docs
- [x] Refactor LoadUI so it loading code is decoupled from UI code


soymonitus:
Ready for review!

src/definitions_cxx.hpp
```diff
@@ -152,6 +152,7 @@ enum class MidiLearn : uint8_t {
 	SECTION,
 	INSTRUMENT_INPUT,
 	DRUM_INPUT,
+	LOAD_BUTTON,
 };
```
seangoodvibes, line 156: Do we really want to add this learn + load button shortcut to learn the global midi command? We don't do this with any of the other global midi commands.

soymonitus: I thought it was handy as we have other learnable commands on buttons like Record on the record button. But if you feel it is not needed i can remove it

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -227,9 +234,60 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+		// in case we wrapped around the whole list of files and didn't find a song,
+		// we just exit without doing anything else
+
+		if (!songFound) {
+			// if song not found, close the UI
+			exitThisUI();
+		}
+
+		// force re-render display
+		uiNeedsRendering(this, 0xFFFFFFFF, 0xFFFFFFFF);
```
seangoodvibes, line 284: this doesn't re-render the display, it re-renders the grid, is that what you meant to do here? comment doesn't match

src/deluge/gui/views/view.cpp
```diff
@@ -203,6 +204,22 @@ ActionResult View::buttonAction(deluge::hid::Button b, bool on, bool inCardRouti
 		}
 	}
 
+	// Load button for MIDI learn
```
seangoodvibes, line 207: per my comment above, not sure we need this shortcut given that all other global midi commands don't have shortcuts like this. It's a learn once type thing. I really don't think we need a shortcut for it.

seangoodvibes:
@m-m-adams this implementation looks solid but would appreciate a second lens on it. nothing stands out for me.

src/deluge/gui/ui/load/load_song_ui.cpp
soymonitus: Yes, the grid is what I want to rerender

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -227,9 +234,60 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+		} while (currentFileIndexSelected != fileIndexSelected && !songFound);
+		// in case we wrapped around the whole list of files and didn't find a song,
+		// we just exit without doing anything else
+
+		if (!songFound) {
+			// if song not found, close the UI
+			exitThisUI();
+		}
+
+		// force re-render display
```
seangoodvibes, line 283:
```suggestion
		// force re-render grid
```

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -227,9 +234,60 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+	if (currentUIMode == UI_MODE_LOADING_SONG_UNESSENTIAL_SAMPLES_ARMED
+	    || currentUIMode == UI_MODE_LOADING_SONG_UNESSENTIAL_SAMPLES_UNARMED
+	    || currentUIMode == UI_MODE_LOADING_SONG_ESSENTIAL_SAMPLES
+	    || currentUIMode == UI_MODE_LOADING_SONG_NEW_SONG_PLAYING || skipAnimations) {
+		// While in the process of loading a song, don't do anything
+		return;
+	}
+
+	skipAnimations = true; // disable animations while auto-loading a song
+	if (openUI(&loadSongUI)) {
```
m-m-adams, line 247: is it possible to bypass opening the load song UI for this? Just seems weird and I wonder if there's a cleaner way

src/deluge/gui/views/view.cpp
```diff
@@ -203,6 +204,22 @@ ActionResult View::buttonAction(deluge::hid::Button b, bool on, bool inCardRouti
...
+	else if (b == LOAD && currentUIMode == UI_MODE_MIDI_LEARN) {
```
m-m-adams, line 208: this should probably be done in the global midi commands menu instead

soymonitus: It's done there already. But you have this as an alternative, that's it, holding Load +learn and sending some midi message. It can be removed from this place if you think it shouldn't be here and leave it just in the main menu for midi commands.

m-m-adams: Oh ok, must have overlooked it

src/deluge/gui/ui/load/load_song_ui.cpp
soymonitus: Not sure if possible without replicating code. There are lots of things going on when you open the browser so not sure if the refactoring needed is worth it.

soymonitus:
Ok i removed the possibility to learn the command from holding LOAD so it is now only accessible from settings menu. @m-m-adams

src/deluge/gui/ui/load/load_song_ui.cpp
m-m-adams: yeah I think it would need a split like for synth/kit loading, where the UI can be used without actually being opened. If you take a look at how they handle turning the select encoder to change presets something like that would probably make more sense. My concern is almost purely that I think this will lead to bugs in the future since it's not clear that the load song UI will need to support being used while not being the active UI, and since this is likely a niche use it could probably silently break for a while before being noticed

soymonitus: Can you point me in the direction where that split for synth/kit loading is done? Is there any specific PR that refactored the synth browser to be separate? thanks!

m-m-adams: https://github.com/SynthstromAudible/DelugeFirmware/blob/6f3409c0b87d6f25cf4c775eb7825b09334f7cc9/src/deluge/gui/ui/load/load_instrument_preset_ui.cpp#L1402

This function is used when scrolling presets without having the file browser open. I think it was there originally 

Here it is in action - https://github.com/SynthstromAudible/DelugeFirmware/blob/a5ab97bbb5687514c3e62962a74b263cba7476f0/src/deluge/model/song/song.cpp#L4655 

Something like this is cleaner since it's clear what it does and doesn't have the chance of weird interactions with the UI stack in the future

soymonitus: @m-m-adams could you take a look at latest changes? now it doesn't pop the UI

soymonitus: I am still getting freezes with loading song from stopped playback state, but I wanted to check if what I have already modified is the direction in which you wanted this PR to go

soymonitus: If I touch the grid when the next song is loading due to my new feature I got this once:

0x200e1755: Kit::renderOutput(ModelStack*, StereoSample*, StereoSample*, long, long*, long, long, bool, bool) at :?
0x200e170d: Kit::renderOutput(ModelStack*, StereoSample*, StereoSample*, long, long*, long, long, bool, bool) at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/model/instrument/kit.cpp:655
0x2012ff5b: Song::renderAudio(StereoSample*, long, long*, long) [clone .constprop.0] at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/OSLikeStuff/timers_interrupts/timers_interrupts.h:44
 (inlined by) Song::renderAudio(StereoSample*, long, long*, long) [clone .constprop.0] at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/model/song/song.cpp:2421
0x2015a958: RMSFeedbackCompressor::render(StereoSample*, unsigned short, long, long, long) at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/dsp/compressor/rms_feedback.cpp:129 (discriminator 1)

why would it crash the deluge? could it be something related to timer interrupts? Maybe the code that loads the next song should be declaring some sort of thread or timer that i am missing?

soymonitus: This is another crash, do you find any clue?

0x200e1755: Kit::renderOutput(ModelStack*, StereoSample*, StereoSample*, long, long*, long, long, bool, bool) at :?
0x20133d51: Voice::renderBasicSource(Sound*, ParamManagerForTimeline*, long, long*, long, bool, long, bool*, long, bool, unsigned long*, unsigned long*, long, unsigned long*, bool, long) [clone .constprop.0] at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/model/voice/voice.cpp:2463
0x2015b08c: DelayBuffer::isActive() const at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/dsp/../dsp/delay/delay_buffer.h:67
 (inlined by) Delay::process(std::span<StereoSample, 4294967295u>, Delay::State const&) at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/dsp/delay/delay.cpp:282
0x200e6657: ModControllableAudio::processFX(StereoSample*, long, ModFXType, long, long, Delay::State const&, long*, ParamManager*, bool, long) at /Users/raul/Development/Projects/Deluge/DelugeFirmware/src/deluge/model/mod_controllable/mod_controllable_audio.cpp:177 (discriminator 2)

soymonitus: it seems several things depend on detecting if the UI is open:
image
Is it really necessary to refactor this as part of this PR? seems like overkill :(

soymonitus:
@m-m-adams I finally was able to do the refactor and get rid of the crash. The code is much simpler now.

src/deluge/gui/ui/load/load_song_ui.cpp
```diff
@@ -228,9 +230,70 @@ ActionResult LoadSongUI::buttonAction(deluge::hid::Button b, bool on, bool inCar
...
+	if (loadingSongInProgress) {
+		return;
+	}
+	loadingSongInProgress = true;
+
+	outputTypeToLoad = OutputType::NONE;
+	currentDir.set(&currentSong->dirPath);
+
+	if (fileIndexSelected == -1) {
+		return;
```
m-m-adams, line 251: this needs to set loading song in progress to false before returning

It might be better to just move everything after the checks into another function so that this one only handles checking/setting the lock and then calling the actual queueNextSong function

m-m-adams:
Other than that lgtm though

soymonitus:
@m-m-adams Applied your feedback suggestion and also added some extra safety check

soymonitus:
@m-m-adams Applied your feedback suggestion, and also added the right safety checks after testing several options for that.

PR #2825, opened 2024-10-19
Pulse Selected Clip in Grid View

seangoodvibes:
https://github.com/user-attachments/assets/f8209543-dad9-4ffd-bab1-9f85fadc0248

- The selected clip in `Song Grid View` (which is the last clip you interacted with), now shows a `pulsing / breathing` animation whereby the clip clip will faded in and out between it's clip colour and a blurred version of the clip's colour.

- I checked the voice count for a kit song that was culling and I did not notice any change between toggling the pulsing on or off.


src/deluge/gui/views/session_view.cpp
```diff
@@ -3303,18 +3317,29 @@ RGB SessionView::gridRenderClipColor(Clip* clip) {
...
+	if (renderPulse && (x != gridFirstPressedX || y != gridFirstPressedY)) {
+		resultColour = resultColour.adjust(255, 2);
+	}
+
+	// Black phase of arm flashing
+	if (view.clipArmFlashOn && clip->armState != ArmState::OFF) {
+		return colours::black;
+	}
+	// If we're currently pulsing this clip and we have the current pulse colour, use that instead
+	else if (!viewingRecordArmingActive && renderPulse && gridSelectedClipPulsing && (clip == selectedClipForPulsing)
```
m-m-adams, line 3338: checking gridSelectedClipPulsing is redundant (if it's null then it won't equal clip)

I think checking that the colour isn't black is also redundant - the only time it's the case while pulsing is active is if reset is called, and using the base colour instead will just cause a 50ms blip of the wrong colour

src/deluge/gui/views/session_view.cpp
```diff
@@ -4609,3 +4638,157 @@ Output* SessionView::getOutputFromPad(int32_t x, int32_t y) {
...
+	return false;
+}
+
+static const uint32_t pulseUIModes[] = {UI_MODE_CLIP_PRESSED_IN_SONG_VIEW, UI_MODE_CREATING_CLIP, 0};
+
+void SessionView::gridPulseSelectedClip() {
+	if (gridSelectedClipPulsing) {
+		// check if we should stop pulsing
+		if (gridCheckForPulseStop()) {
+			return gridStopSelectedClipPulsing();
```
m-m-adams, line 4696: The return should be on its own line, otherwise this will silently cause bugs if gridStopSelectedClipPulsing ever has a return value added (e.g. maybe an error code)

src/deluge/gui/views/session_view.cpp
```diff
@@ -4609,3 +4638,157 @@ Output* SessionView::getOutputFromPad(int32_t x, int32_t y) {
...
+				uint32_t selectedClipX;
+				uint32_t selectedClipY;
+				gridCoordsFromClip(currentClip, selectedClipX, selectedClipY);
+				// don't render pulse if clip is scrolled out of view (coords = - 1)
+				if (selectedClipX != -1 && selectedClipY != -1) {
+					// pulse if:
+					// A) you're not holding a pad
+					// or
+					// B) you're holding a pad that's different from the selected clip
+					if (currentUIMode == UI_MODE_NONE
```
m-m-adams, line 4721: why not pulse if you're holding the current clip's pad?

src/deluge/gui/views/session_view.cpp
```diff
@@ -4609,3 +4638,157 @@ Output* SessionView::getOutputFromPad(int32_t x, int32_t y) {
...
+					    || (gridFirstPressedX != selectedClipX || gridFirstPressedY != selectedClipY)) {
+
+						// save the selected clip so that we don't override the pulsed colour when re-rendering grid
+						selectedClipForPulsing = currentClip;
```
m-m-adams, line 4725: Why do these need to be different? If we're only ever pulsing the selected clip then it would make more sense to just check the same variable

src/deluge/gui/views/session_view.cpp
```diff
@@ -4609,3 +4638,157 @@ Output* SessionView::getOutputFromPad(int32_t x, int32_t y) {
...
+
+						// get the clip's colour, which can be different from the track colour
```
m-m-adams, line 4727: I think all of this should be moved into a setupPulsing function or similar - this is a lot to calculate on every pulse update and it only changes when the selected clip is changed

src/deluge/gui/views/session_view.cpp
```diff
@@ -3350,9 +3350,9 @@ RGB SessionView::gridRenderClipColor(Clip* clip, int32_t x, int32_t y, bool rend
 	if (view.clipArmFlashOn && clip->armState != ArmState::OFF) {
 		return colours::black;
 	}
+
 	// If we're currently pulsing this clip and we have the current pulse colour, use that instead
-	else if (!viewingRecordArmingActive && renderPulse && gridSelectedClipPulsing && (clip == selectedClipForPulsing)
-	         && (gridSelectedClipRenderedColour != RGB::monochrome(0))) {
+	if (!viewingRecordArmingActive && renderPulse && (clip == selectedClipForPulsing)) {
```
seangoodvibes, line 3355: might need to confirm whether removing the "gridSelectedClipRenderedColour != RGB::monochrome(0)" check affects anything. I forget why I specifically checked for that.

src/deluge/gui/views/session_view.cpp
stellar-aria: Fixed

src/deluge/gui/views/session_view.cpp
m-m-adams: I don't see the point in the selectedClipForPulsing field at all. We always want it to be the selected/active clip so we should just get it from getCurrentClip instead of  caching it in a field that we have to regularly update and keep synced

seangoodvibes: @m-m-adams i think I did this for rendering reasons so that it only renders the pulse when we're ready to. I also believe without it it results in choppy rendering and weird glitches. Can try swapping it out if you want. 

I honestly can't remember why I did it but I did put in a comment saying "save selected clip to not override the pulse when re rendering the grid".

I believe there's also a lot of situations where the grid fully re renders and it messes up the rendering of the pulse.

seangoodvibes:
Switched this back to draft as it isn’t working properly atm

m-m-adams (approved):

PR #2833, opened 2024-10-21
Update Ne10 library

stellar-aria:
Updates the Ne10 library by removing the custom in-source inclusion and moving to an external FetchContent library with (minimal) customizations.

Note: needs size and performance testing (particularly around wavetable loading).


lib/CMakeLists.txt
```diff
@@ -10,3 +10,14 @@ FetchContent_Declare(argon
 )
 FetchContent_MakeAvailable(argon)
 
+add_compile_definitions(NDEBUG)
+set(NE10_ASM_OPTIMIZATION ON)
+set(NE10_TARGET_ARCH armv7)
+
+FetchContent_Declare(Ne10
+  GIT_REPOSITORY https://github.com/stellar-aria/Ne10
```
m-m-adams, line 18: Is it possible to pin a specific version or commit?

stellar-aria: done!

PR #2835, opened 2024-10-21
Fix loops remaining popup

seangoodvibes:
Fixed display of loops remaining popup when launching a clip so that it doesn't show when you're pressing on a clip in song view

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2836

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -2352,8 +2352,8 @@ void SessionView::displayPotentialTempoChange(UI* ui) {
 /// display number of bars or quarter notes remaining until a launch event
 int32_t SessionView::displayLoopsRemainingPopup(bool ephemeral) {
 	int32_t sixteenthNotesRemaining = session.getNumSixteenthNotesRemainingTilLaunch();
-	// don't show pop-up if you're pressing a clip in song view
-	if (currentUIMode != UI_MODE_CLIP_PRESSED_IN_SONG_VIEW) {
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -2352,29 +2352,32 @@ void SessionView::displayPotentialTempoChange(UI* ui) {
...
 int32_t SessionView::displayLoopsRemainingPopup(bool ephemeral) {
 	int32_t sixteenthNotesRemaining = session.getNumSixteenthNotesRemainingTilLaunch();
-	if (sixteenthNotesRemaining > 0) {
-		DEF_STACK_STRING_BUF(popupMsg, 40);
-		if (sixteenthNotesRemaining > 16) {
-			int32_t barsRemaining = ((sixteenthNotesRemaining - 1) / 16) + 1;
-			if (display->haveOLED()) {
-				popupMsg.append("Bars Remaining: ");
+	// don't show pop-up if you're pressing a clip in song view
+	if (currentUIMode != UI_MODE_CLIP_PRESSED_IN_SONG_VIEW) {
```
m-m-adams, line 2356: Can we check to be in UI mode none? This seems likely to cause bugs later

seangoodvibes: > Can we check to be in UI mode none? This seems likely to cause bugs later

Done

m-m-adams (approved):

PR #2843, opened 2024-10-26
Fix bug with persist midi learn

seangoodvibes:
Logic for persisting midi knob array between preset changes should only apply to synth and kit instruments.

This fixes a crash that happens when trying to load a midi preset.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -41,6 +41,7 @@
 #include "model/consequence/consequence_clip_existence.h"
 #include "model/instrument/cv_instrument.h"
 #include "model/instrument/midi_instrument.h"
+#include "model/mod_controllable/mod_controllable_audio.h"
 #include "model/sample/sample_recorder.h"
 #include "model/scale/preset_scales.h"
 #include "model/scale/scale_change.h"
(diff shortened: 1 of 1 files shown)


src/deluge/model/song/song.cpp
```diff
@@ -3311,21 +3311,25 @@ void Song::replaceInstrument(Instrument* oldOutput, Instrument* newOutput, bool
...
+		// only sound instruments have a midi knob array
+		if (newOutput->type == OutputType::SYNTH || newOutput->type == OutputType::KIT) {
+			// Migrate any midi learned params for synth and kit clip preset changes
+			// For kit clips it will migrate only kit affect entire midi learned params
+			// Scenario:
+			// - you create a clip
+			// - you midi learn a controller to that clip's params
+			// - you then go to change the preset for that clip
+			// - you expect that you can continue controlling the same params for the new preset
+			Sound* oldSound = (Sound*)oldOutput->toModControllable();
```
m-m-adams, line 3324: Kits aren't sounds so you still shouldn't do this

The midiKnobArray comes from ModControllableAudio and not sound anyway

seangoodvibes: Updated Sound* to ModControllableAudio*

m-m-adams (approved):

PR #2853, opened 2024-10-31
Add SysEx protocol for file browsing and transfer between the Deluge & the 'vuefinder' web application.

jamiefaye:
The other side of this protocol is the modified version of Vuefinder presently residing at https://github.com/jamiefaye/vuefinder.

Simple Vuefinder intro:

How to run:
Build the DelugeFirmware branch called jffcasc.
Connect your Deluge to your computer using USB Midi.
Open the web application at:
https://www.fentonia.com/deluge/vuefinder/
When it loads you should see Midi In and Out displayed as "Deluge Port 1" and "Deluge Port 1". If you are calling the Deluge Midi ports something else, select the actual name using the popups.
Press the "Home" icon to refresh the file list. We don't (as of now) automatically load the top directory since we have to wait until the Midi In and Midi Out are set.
In the upper right hand corner is an icon made out of 4 rectangles. Press that to switch from icons to list file format. Clicking on a directory icon descends. Clicking on the up arrow ascends.The folder with a '+' in it will create a new directory. If you select a file, you can use the control key (or whatever) to bring up a small menu which includes "download" and "rename" (some of the other options do not work yet). If you select a directory instead of a file, you can download the directory and its nested contents. It may take awhile, with the current USB driver, the transfer rate is around 200K bytes/second.

The trash can deletes, and the "up-arrow in the tray" uploads and you will be prompted for the file/directory you want to upload.

Needless to say, test this with a backup SD card! We have not tested transferring while a Song is playing much.


seangoodvibes:
Hi Jamie,

Could you add a feature document for this here please :)? https://github.com/SynthstromAudible/DelugeFirmware/tree/community/docs/features

Thx!

jamiefaye:
> Hi Jamie,
> 
> Could you add a feature document for this here please :)? https://github.com/SynthstromAudible/DelugeFirmware/tree/community/docs/features
> 
> Thx!

Feature description for Vuefinder is at https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/features/Vuefinder.md

Some info on the underlying protocol is at: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/dev/SysExProtocolNotes.md

src/OSLikeStuff/task_scheduler.cpp
```diff
@@ -472,7 +472,7 @@ void TaskManager::resetStats() {
 }
 
 void TaskManager::printStats() {
-	D_PRINTLN("Dumping task manager stats: (min/ average/ max)");
```
m-m-adams, line 475: can we keep this?  it's helpful for finding issues since it can be grabbed by users using the midi logging

src/deluge/memory/general_memory_allocator.cpp
```diff
@@ -54,7 +54,8 @@ void GeneralMemoryAllocator::checkStack(char const* caller) {
 #if ALPHA_OR_BETA_VERSION
 
 	char a;
-
+	if (true)
```
m-m-adams, line 57: I think we should keep this

src/deluge/storage/storage_manager.cpp
```diff
@@ -809,61 +836,90 @@ void FileReader::readDone() {
...
-			uiTimerManager.routine();
+	writeClusterBuffer[fileWriteBufferCurrentPos] = b;
+	fileWriteBufferCurrentPos++;
 
-			if (display->haveOLED()) {
-				oledRoutine();
-			}
-			PIC::flush();
+	// Ensure we do some of the audio routine once in a while
+	if (callRoutines && !(fileWriteBufferCurrentPos & 0b11111111)) {
```
m-m-adams, line 910: this should probably call routineForSD instead of just the oled and pad stuff, dunno why it wasn't before

m-m-adams (commented):
Just minor nits, looks good in general!

src/deluge/memory/general_memory_allocator.cpp
jamiefaye: I removed the return.

src/deluge/storage/storage_manager.cpp
jamiefaye: I will put in routineForSD and hope nothing explodes.

m-m-adams (approved):

PR #2860, opened 2024-11-02
Handle first pad released when two pads are held. Fixes #2693

### LINKED ISSUE #2693: Bug with "Copy Clips" in Grid View Green Mode

Noticed that you can get stuck in UI_MODE_CLIP_PRESSED_IN_SONG_VIEW with "Copy Clips" pop-up stuck on the display

Steps to re-produce:
- Go into green mode in song grid view
- Press the grid to create a whole bunch of clips
- Try to create clips in multiple pads by pressing pads adjacent pads together
- You'll see that if you do it often enough you'll eventually get stuck with "copy clips" on the display

Proper handling of multiple pad presses so that you don't get stuck in a UI mode

_No response_

OLED

Built from community

_No response_

https://github.com/user-attachments/assets/f1790857-e801-418b-b919-e521265f7d83

RJ-Eckie:
Fixes #2693 where the "COPY CLIPS" popup would stay on the screen, when trying to copy a clip in grid mode and releasing the source-pad first.

This makes existing behavior more explicit: releasing the first pad of two held pads cancels copying. This should be added to the documentation.

https://github.com/user-attachments/assets/64fad287-4d88-48fa-847b-b4f78b524f0f

Bug existed only in launch-mode. Behavior in edit-mode does not match behavior in launch-mode yet. Code needs some restructuring to avoid duplicate code, but this could possibly be merged as is.

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -4128,7 +4128,7 @@ ActionResult SessionView::gridHandlePadsLaunch(int32_t x, int32_t y, int32_t on,
 	// Release first pad, while two pads are held
 	if (!on && x == gridFirstPressedX && y == gridFirstPressedY && gridSecondPressedX != -1
 	    && gridSecondPressedY != -1) {
-		display->popupTextTemporary("COPY CANCELLED");
+		display->popupTextTemporary("COPY CANCELED");
 	}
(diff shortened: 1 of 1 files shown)


RJ-Eckie:
Reorganizing branches - will re-submit

PR #2865, opened 2024-11-02
Handle first pad released when two pads are held. Fixes #2693 (re-submit)

### LINKED ISSUE #2693: Bug with "Copy Clips" in Grid View Green Mode

RJ-Eckie:
Fixes https://github.com/SynthstromAudible/DelugeFirmware/issues/2693 where the "COPY CLIPS" popup would stay on the screen when trying to copy a clip in grid mode and releasing the source-pad first, and in some instances when switching grid view modes while holding two pads. Also fixes inconsistent popups when copying.

Makes existing behavior more explicit: releasing the first pad of two held pads cancels copying. This should be added to the documentation.

https://github.com/user-attachments/assets/64fad287-4d88-48fa-847b-b4f78b524f0f

_spelling has been updated to "canceled" (common American spelling) since taking this video_

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -1079,6 +1079,10 @@ void SessionView::clipPressEnded() {
 	// the right model stack with param (otherwise midi follow mode will think you're still in a clip)
 	selectedClipYDisplay = 255;
 	clipWasSelectedWithShift = false;
+	// Cancel copying clip
+	if (gridFirstPressedX != -1 && gridFirstPressedY != -1 && gridSecondPressedX != -1 && gridSecondPressedY != -1) {
+		display->popupTextTemporary("COPY CANCELED");
+	}
 	gridResetPresses();
 
 	currentUIMode = UI_MODE_NONE;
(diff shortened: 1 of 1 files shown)


m-m-adams (approved):

m-m-adams:
Awesome thanks! With the popup it's pretty explicit so I don't see a real need to add it to the docs, at worst it'll be unclear the very first time

RJ-Eckie:
> Awesome thanks! With the popup it's pretty explicit so I don't see a real need to add it to the docs, at worst it'll be unclear the very first time

The value of having it in the docs would be to make people are aware of their option to cancel a copy after pressing two pads, instead of having to stumble upon it by accident first. But I'm fine either way

Issue #1909, opened 2024-05-12
It should be possible to launch and enter a clip at the same time in grid view


sapphire-arches:
### Please describe the problem:

In grid view it's very common to want to copy a clip and then edit it. Currently this requires 6 button presses:
- Switch to blue mode
- Copy the clip
- Switch to green mode
- Launch the new clip
- Switch to blue mode
- Edit the new clip

This is irritating and can break the flow.

### What is the expected behavior?

A new latch state should be added to grid view equivalent to launch+edit so the workflow above becomes:

- In blue mode, copy the clip
- Press green+blue to enter "clip launch + edit" state. The blue and green mode buttons should blink (or both be lit) in this state.
- Press the copied clip to both launch and enter it.

This mode would latch like the other display modes.

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

other

sdesign75:
Would also like the following behaviour added if possible…

In grid view, when Default Active Mode is set to SELECTION, could we make the green and blue pads behave like the Shift button? So if you press and hold it goes into the Blue mode and when you release it then it returns to Green mode and vice versa. If you tap it latches.

Exmpl-xe:
A similar idea excluding both green & blue modes as being separate. Lite hold clip to enter, hold + new to copy & tape to launch

PR #2868, opened 2024-11-03
Implement grid mode clip edit & launch - feedback appreciated. Implements #1909

RJ-Eckie:
This implements a way to edit and launch a clip in grid mode with one action.

### Changes to edit mode:
Edit mode is now either launching or non-launching:
- Blue: non-launching (unchanged)
- Cyan: identical to blue, but whenever a clip is entered that clip is also launched (quantized, or immediately when holding shift)

Launching/non-launching mode can be toggled by pressing green+blue at the same time
~~Launching is quantized. (Immediate launching with shift+cyan to be implemented.)~~

### Changes to copying in grid mode:
Copying now has three variants:
- Launch mode: after copying, remain in grid view (unchanged)
- Edit mode, non-launching: after copying, enters newly created clip
- Edit mode, launching: after copying, enters newly created clip _and_ launches it

--

This addresses all complaints in #1909 and even goes a step further. Especially when copying + launching it feels really smooth! Interested to hear what other people think.


RJ-Eckie:
Ready for review as far as I'm concerned. Happy to make changes if there's feedback.

src/deluge/gui/views/session_view.cpp
```diff
@@ -3188,7 +3190,7 @@ void SessionView::gridRenderActionModes(int32_t y, RGB image[][kDisplayWidth + k
 	}
 	case GridMode::BLUE: {
```
m-m-adams, line 3191: Cyan should be it's own mode

m-m-adams:
Are you still working on this? We've generally found that switching modes is not popular and are aiming to make it as unnecessary as possible, to do that it needs to be its own mode so it can tie into the momentary options

m-m-adams:
Feel free to reopen if you start working on it again

PR #2873, opened 2024-11-03
Feature/stereo warble

m-m-adams:
Code works, hoping for some user feedback before merging

Adds stereo warbling (happens automatically if rendering in stereo)

Uses constexpr ifs to choose between modfx functions, this works much faster since it removes the per sample if statements

### DIFF
--- a/src/deluge/gui/menu_item/mod_fx/type.h
+++ b/src/deluge/gui/menu_item/mod_fx/type.h
@@ -28,7 +28,7 @@ class Type : public Selection {
 public:
 	using Selection::Selection;
 
-	void readCurrentValue() override { this->setValue(soundEditor.currentModControllable->modFXType); }
+	void readCurrentValue() override { this->setValue(soundEditor.currentModControllable->modFXType_); }
 	void writeCurrentValue() override {
 		if (!soundEditor.currentModControllable->setModFXType(this->getValue<ModFXType>())) {
 			display->displayError(Error::INSUFFICIENT_RAM);
--- a/src/deluge/model/global_effectable/global_effectable.cpp
+++ b/src/deluge/model/global_effectable/global_effectable.cpp
@@ -41,7 +41,7 @@ GlobalEffectable::GlobalEffectable() {
 	lpfMode = FilterMode::TRANSISTOR_24DB;
 	filterSet.reset();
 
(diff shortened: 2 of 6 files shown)


RJ-Eckie:
I like it! What is this effect doing technically?
Could you write notes for what feedback and offset do? Rate and depth are self-explanatory I think.
Also I'm not really clear on what you mean by "happens automatically if rendering in stereo"

Findings so far:
- Offset has audible clicks when changing the value
- Effect gets noticably less intense when turning feedback all the way up - this may be intended behavior depending on what feedback does exactly
- Effect doesn't seem to turn off completely when depth is 0, output seems lower than chorus and stereo chorus on depth = 0. (This may not actually be a goal as phaser also doesn't adhere to this. I feel it's more user-friendly if you can change mod-fx type without hearing any change to the sound, when depth=0.)

m-m-adams:
Hey yeah there's docs already, this just renders the warbler in stereo if the deluge is rendering in stereo. Offset adjusts the delay line length and clicks the same way in chorus, stopping it would require much larger buffers and smoothing 

All the effects will sound different at 0 depth so I don't think that's a good goal, that only bypasses the chorus. Flanger and Warbler bypass when feedback is 0 but more generally you should just turn the effect off if you don't want to hear them

RJ-Eckie:
> Hey yeah there's docs already, this just renders the warbler in stereo if the deluge is rendering in stereo. Offset adjusts the delay line length and clicks the same way in chorus, stopping it would require much larger buffers and smoothing
> 
> All the effects will sound different at 0 depth so I don't think that's a good goal, that only bypasses the chorus. Flanger and Warbler bypass when feedback is 0 but more generally you should just turn the effect off if you don't want to hear them
> 
> Adjusting the offset when feedback is high has significant effects on the sound so play with that

Ah - did not realize this was adapting an existing effect! This makes more sense now

m-m-adams:
I mean it's a new effect I made for 1.3 so I'm also interested in feedback on controls and sound but yeah this PR is just adding stereo rendering to it

It's essentially a flangerish thing but with a random control process modulating it instead of a normal LFO

seangoodvibes:
Could you update the changelog entry for the Warbler FX to mention the stereoness :D? Thx

seangoodvibes (approved):

PR #2874, opened 2024-11-04
Align blinking cursor with decimal value. Fixes #2858

### LINKED ISSUE #2858: blinking bar doesn't always allign with the changed value, Transpose or Phase retrig

when adjusting Osc or Mod Transpose and also Osc Phase the blinking bar doesn't always allign with the value you change. It seems to not fit, when there are characters that don't use the full width like "1" or "."

Blinking bar alligns

_No response_

OLED

Nightly, Beta

beta and nightly as of 11-01-2024

1. Select "OSc Transpose" or "Phase retrig"
2. Use vertical encoder to move blinking bar
3. use main encoder to change values

RJ-Eckie:
This fixes #2858 where the indicator for the digit being edited did not line up with the digit itself
I also made the cursor not touch the digits, which increases legibility

This is definitely a bit of a dirty fix: the digits are manually positioned in monospacing - with the period being 1/2 of the width of the digits - which makes it possible for the cursor to align with them.

### There are some considerations here:
- This should probably live inside the` OLED::main.drawString` function
- The cursor should probably also be implemented there, so that it's reusable in other places
- In the current implementation the cursor appears under the minus sign when editing a leading zero on a negative value, which doesn't feel very neat (picture below)
- I personally feel that a quickly changing number on a display is much easier on the eyes when its digits are fixed in place, regardless of how many digits are visible. This would also make it possible to never have the cursor under the minus sign

Without going into all of those, this pr does fix the mentioned issue

IMG_2488
_cursor appearing under minus sign_

### DIFF
--- a/src/deluge/gui/menu_item/decimal.cpp
+++ b/src/deluge/gui/menu_item/decimal.cpp
@@ -151,11 +151,37 @@ void Decimal::drawPixelsForOled() {
 	int32_t digitWidth = kTextHugeSpacingX;
 	int32_t stringWidth = digitWidth * length;
 	int32_t stringStartX = (OLED_MAIN_WIDTH_PIXELS - stringWidth) >> 1;
-
-	hid::display::OLED::main.drawString(buffer, stringStartX, 20, digitWidth, kTextHugeSizeY);
-
 	int32_t ourDigitStartX = stringStartX + editingChar * digitWidth;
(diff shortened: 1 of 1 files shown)


RJ-Eckie:
Just found an existing monospaced drawing method in the code - looking at fixing this again by using that

PR #2875, opened 2024-11-04
Enter clip view after new clip on new track

### LINKED ISSUE #2872: Clip creation in grid view is not consistent: existing instrument vs new instrument

In grid view, when creating a new clip in an existing instrument (blue + empty pad) the new clip gets created and immediately switches to clip view of the new clip. When doing the same for an empty column, a new clip is created but we stay in grid view.

In 1.1 the behavior is consistent: when making a new clip in a new instrument, clip view is opened. From 1.2 behavior is no longer consistent. I'm assuming it is related to the new quick-choice for new clip type.

After creating a new clip in a new instrument (that is: after the NEW CLIP TYPE menu closes) the UI should enter clip view of the new clip.

#2492

OLED

Beta

0b4A7C5

Open song mode grid view
Press blue + empty pad in leftmost column (desired behavior)
Press blue + empty pad in empty column (inconsistent behavior)

RJ-Eckie:
Fix #2872

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -4074,7 +4074,6 @@ ActionResult SessionView::gridHandlePadsLaunch(int32_t x, int32_t y, int32_t on,
 				gridFirstPressedY = y;
 				// will create the track if it doesn't exist
 				clip = gridCreateClip(gridSectionFromY(y), track, nullptr);
-				transitionToViewForClip(clip);
 				// If playing and Rec enabled, selecting an empty clip creates a new clip and starts it playing
 				// (depending on setting)
 				if (clip != nullptr && playbackHandler.playbackState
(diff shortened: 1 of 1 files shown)


src/deluge/gui/views/session_view.cpp
```diff
@@ -4073,6 +4074,7 @@ ActionResult SessionView::gridHandlePadsLaunch(int32_t x, int32_t y, int32_t on,
 				gridFirstPressedY = y;
 				// will create the track if it doesn't exist
 				clip = gridCreateClip(gridSectionFromY(y), track, nullptr);
+				transitionToViewForClip(clip);
```
m-m-adams, line 4077: I don't think this is desirable in launch mode, it gets in the way of live looping

seangoodvibes:
The issue tagged in this PR doesn't seem to align with the fix.

The issue talks about inconsistency in blue mode (edit mode) but the fix touches green mode (launch mode).

m-m-adams (approved):

PR #2884, opened 2024-11-06
Cleanup decimal rendering

### LINKED ISSUE #2858: blinking bar doesn't always allign with the changed value, Transpose or Phase retrig

RJ-Eckie:
### Changes
- Fix #2858 (revert to c1.1 monospaced rendering for decimals)
- Render period at half-width
- Resize cursor, make it not touch digits

IMG_2496
_c1.1_

IMG_2494
_new_


m-m-adams (approved):

pssssssst:
Hi, thanks for fixing this.

PR #2886, opened 2024-11-06
Keep arranger clip instance length same after change

### LINKED ISSUE #2877: Arranger Mode clip instances longer than their clip length forget their length when changing section

In arranger mode, clip instances can be longer than their clip length. The clip will loop to fill up the length of the clip instane. When changing the clip section the instance reverts to the clip length (1 loop of the clip) as opposed to staying the same length.

White clips created from a section clip **do** stay the same length.

Length of a clip instance should stay the same when changing clips

_No response_

OLED

Built from community

1.3

- create two clips in song mode, from the same instrument, in different sections
- in arranger mode, place the first clip
- make the clip longer than 1 loop: hold the left-most pad and press an empty pad to the right of the clip
- change the clip instance: hold any pad of the clip and turn the select-knob to select the second clip

RJ-Eckie:
Fix #2877 

This deletes a bunch of safety-checks for the length of the new clip-instance
As we're taking the length from a valid clip-instance, we already know the length is valid

### DIFF
--- a/src/deluge/gui/views/arranger_view.cpp
+++ b/src/deluge/gui/views/arranger_view.cpp
@@ -1273,17 +1273,21 @@ void ArrangerView::editPadAction(int32_t x, int32_t y, bool on) {
 
 			// Already pressing - length edit
 			else if (currentUIMode == UI_MODE_HOLDING_ARRANGEMENT_ROW) {
-
+				// press on different row than currently holding pad
 				if (y != yPressedEffective) {
 					if (!pressedClipInstanceIsInValidPosition) {
 						return;
 					}
+					// create new clip instance on new press
 					return createNewClipInstance(output, x, y, squareStart, squareEnd, xScroll);
 				}
-
+				// press on same row, to the right of currently holding pad
 				if (x > xPressed) {
-					adjustClipInstanceLength(output, xPressed, y, squareStart, squareEnd);
+					// change length of selected clip instance
+					return adjustClipInstanceLength(output, xPressed, y, squareStart, squareEnd);
 				}
+				// press on same row, to the left of the currently holding pad, do nothing
+				return;
 			}
 		}
 
@@ -1377,7 +1381,8 @@ void ArrangerView::createNewClipInstance(Output* output, int32_t x, int32_t y, i
 	// Look for a clip instance to the left of, or including press position
 	int32_t i = output->clipInstances.search(squareEnd, LESS);
 	ClipInstance* clipInstance = output->clipInstances.getElement(i);
-
+	// default: not creating a clip instance
+	creatingNewClipInstance = false;
 	// If we did find a clip instance
 	if (clipInstance) {
 		Clip* clip = clipInstance->clip;
@@ -1392,7 +1397,9 @@ void ArrangerView::createNewClipInstance(Output* output, int32_t x, int32_t y, i
 				}
 				// To the right is allowed
 				else {
+					// create a new clip instance
 					clipInstance = createClipInstance(output, y, squareStart);
+					creatingNewClipInstance = true;
 				}
 			}
 		}
@@ -1401,7 +1408,9 @@ void ArrangerView::createNewClipInstance(Output* output, int32_t x, int32_t y, i
 		// ahead and make a new Instance here
 		int32_t instanceEnd = clipInstance->pos + clipInstance->length;
 		if (instanceEnd <= squareStart) {
+			// create a new clip instance
 			clipInstance = createClipInstance(output, y, squareStart);
+			creatingNewClipInstance = true;
 		}
 		else {
 			// If still here, the ClipInstance overlaps this square, so select it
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/arranger_view.cpp
```diff
@@ -2494,34 +2494,9 @@ void ArrangerView::selectEncoderAction(int8_t offset) {
...
-				newLength = maxLength;
-			}
-		}
-		if (newLength > kMaxSequenceLength - clipInstance->pos) {
-			newLength = kMaxSequenceLength - clipInstance->pos;
-		}
 		// log action
 		Action* action = actionLogger.getNewAction(ActionType::CLIP_INSTANCE_EDIT, ActionAddition::ALLOWED);
-		clipInstance->change(action, output, clipInstance->pos, newLength, newClip);
+		clipInstance->change(action, output, clipInstance->pos, desiredLength, newClip);
```
m-m-adams, line 2499: Should we preserve extending the length if the new clip is longer than the old one?

RJ-Eckie: When editing an existing clip instance it makes sense to me to always keep it the same length. In the context of an arrangement it makes sense for the intention to be: in this specific range of the arrangement, I want to change what happens, no matter the length of the source clip.

When creating a new clip instance, it does make more sense to have the result be the length of the chosen clip. This also gives the user a very quick choice between keeping an existing clip instance the same length (hold + turn select) and changing the instance length to clip length (tap to delete, hold + turn select to create new clip instance at clip length)

I'll implement this

RJ-Eckie:
New clip instances now adhere to the length of the clip that is chosen, while making the initial clip selection

Nice bonus behavior: when creating a new clip instance and the user makes a conscious choice to set the length of the clip instance with a second press, clip instance will now stay the chosen length regardless of chosen clip

src/deluge/gui/views/arranger_view.cpp
```diff
@@ -1273,17 +1273,21 @@ void ArrangerView::editPadAction(int32_t x, int32_t y, bool on) {
...
 				}
-
+				// press on same row, to the right of currently holding pad
 				if (x > xPressed) {
-					adjustClipInstanceLength(output, xPressed, y, squareStart, squareEnd);
+					// change length of selected clip instance
+					return adjustClipInstanceLength(output, xPressed, y, squareStart, squareEnd);
 				}
+				// press on same row, to the left of the currently holding pad, do nothing
+				return;
```
RJ-Eckie, line 1290: Return behavior implementation is a bit inconsistent - this is not absolutely needed here but doesn't hurt either

seangoodvibes:
I removed the cherry pick tag as I'm not sure whether this fix should go to beta or not

src/deluge/gui/views/arranger_view.cpp
m-m-adams: Yeah but I think this code also covers the initial clip selection, like if you put in a 4 bar blue clip (e.g. last clip you used) then try to change it to your 8 bar pink clip

RJ-Eckie:
> I removed the cherry pick tag as I'm not sure whether this fix should go to beta or not

Fair! I'm choosing to just implement stuff to my best ability from the perspective of getting things done, but by all means shoot at it if it doesn't seem ideal.

m-m-adams:
Yeah I think this is an improvement but better to avoid adding things to beta at this point, there might be something we haven't thought about

m-m-adams:
@RJ-Eckie are you still working on this? I'm not sure if it was clear but we do want to merge it, we just don't want to put it into the current beta (e.g. firmware 1.2). It is definitely a good add to 1.3

todd-gochenour:
+1 encountered this issue today.

Recorded a song to arranger. I missed launching a two bar drum phrase that was to repeat four times. There is a hit that is set to only play once out of four repeats. I remember last time (pre community firmware) that I could press the start and end pads to extend the length of the clip, same as extending a note length in clip view. Adding the clip four times consecutively causes the hit to fire four times.

I see RJ's bonus behavior above which sounds like a fix. Question: Can the length of an existing clip in arranger view be changed or is this behavior only when clip is first added?

todd-gochenour:
I answered my own question: **Yes,** this change does allow edits of an existing clip's length in arranger view.  I went and looked at pull request staged for release 1.3.  First time for me in the code.

Editing clip length in Arranger view is essential functionality.   I hope this change could make its way into the 1.2 release.

m-m-adams:
> +1 encountered this issue today.
> 
> Recorded a song to arranger. I missed launching a two bar drum phrase that was to repeat four times. There is a hit that is set to only play once out of four repeats. I tried to add the clip manualIy but can't extend the length. I remember last time (pre community firmware) that I could press the start pad (and hold) and then press the end pad to extend or truncate the length of the clip to cause it to repeat. Adding the clip four distinct times causes that initial hit to fire four times.
> 
> I see RJ's bonus behavior above which sounds like a fix. Question: Can the length of an existing clip in arranger view be changed or is this behavior only when clip is first added?

This sounds like an unrelated bug - you should definitely still be able to edit the length of clips in arranger with and without this change

todd-gochenour:
Retested my issue with new 12/04 build of community beta 1.2.   Existing clips in arranger mode are now properly extending their length when start/end pads are selected.

m-m-adams:
Closing since it seems to be dropped

PR #2903, opened 2024-11-11
[Draft] Rename Session to SongView, cleanup Song Layouts and Macros

RJ-Eckie:
Commits **refactor macros** and **Cleanup layout handling** are the ones that actually have fundamental code changes. Will try to do this a little neater next time.

Notes I kept while editing:
### Changes
1. Various renames:
   1. File names to `song_view.h` & `song_view.cpp`
   2. Global `sessionX` to `songViewX`
   4. `UIType::SESSION` to `UIType::SONG_VIEW`
   5. Used enums are now:
      - `SongViewLayout` (rows/grid)
      - `SongViewGridLayoutMode` (launch/edit)
      - `SongViewGridLayoutModeSelection` (select/defaultLaunch/defaultEdit)
      - `SongViewGridLayoutModifierPad` (defines right sidebar pad functions, renamed from colors to descriptive names)
2. Make all relevant enums scoped for safety and legibility, shorten element names
3. Roughly organize properties/methods, label in obnoxiously large comments (ongoing project)
4. Remove configuring macros as a `SongViewGridLayoutMode`
   - Configuring macros can be done while in edit mode, launch mode, or even in rows layout where those modes don't exist. This makes configuring macros not analogous to edit mode and launch mode. I've changed this as a preparation for further refactoring of GridLayout and its modes
5. Refactor macros enums
   - With "kind" and "type" being mostly synonymous, I feel like type is much more consistently used in coding and other technical environments, so I renamed to type
   - Used TitleCase
6. Cleanup Layout handling
   - Big restructure of the code. Custom staging blocks in vscode on mac seem to be broken so the diff is a bit of a mess. The result is pretty simple:
     - `selectLayout()` should be used wherever layout is changed
     - `selectTemporaryLayout()` saves current layout state before selecting (used by macros config)
     - `closeTemporaryLayout()` restores saved layout state
     - `cycleLayouts()` iterates through layouts (used by song button + turn select)
   - Some more descriptive renaming. Some names are getting pretty long, but that will probably resolve itself once rows & grid view are split into separate classes


seangoodvibes:
Hey @RJ-Eckie sorry to do this but we discussed these changes on the discord and the consensus was the following:

- Use Song instead of Project
- Use Session View instead of Song View
- Use Session Mode instead of Song Mode

I think you have some good refactorings here but given the user facing nature of "Song" and with how it's labeled on the Deluge we should keep that terminology to represent the "Project" and use "Session" where we're referring to the Session Mode or Session View.

RJ-Eckie:
Wait so this is basically throwing out everything we discussed before about Song View and Project?

PR #2904, opened 2024-11-11
Roadmap structure proposal

RJ-Eckie:
Here's a proposal for keeping track of decisions that have been made about changes to the codebase. While different people can have their own branches/repos with documentation in sketch-phase, these files are meant to represent _the truth_.

There are three main folders (code_structure, features and naming) with each a file that chronologically logs decisions, most recent first. I thought ordering by month would be granular enough. Those are accompanied by files that outline either existing code or the made decisions. Such as:
- Directory outline
  - Subfolder with outlines per class
- Style guide
- Naming convention
- Etc.

I added some of my own documentation plus some dummy files as examples. The goal is to have a central point where we can collect knowledge and decisions about the codebase. (Not to document every line of code that exists.) I'll be opening a thread where decisions can be discussed and made, before being added to these files.


docs/dev/roadmap/code_structure/class_outlines/main_c.md
```diff
@@ -0,0 +1,5 @@
+# main.c
```
m-m-adams, line 1: I think individual files might be too granular (unless the idea is to only create these files to document key decisions)?

My worry would be these getting left out of date with the actual files and turning into more misleading docs

docs/dev/roadmap/code_structure/directory_reference.md
```diff
@@ -0,0 +1,128 @@
...
+
+*User Interface. Many low-level and base methods of the user interface are defined here, and reused in higher level code*
+
+| Folder | Contents
+|-|-
+| browser | `Browser` is a `QwertyUI` which handles browsing files on the SD card. It has sub-classes for browsing DX7-presets (`DxSyxBrowser`,) audio samples (`SampleBrowser`,) and slots for instrument clips and songs (`SlotBrowser`)
+| keyboard | manages the keyboard overlay
+| load | `LoadUI` is a `SlotBrowser` with sub-classes `LoadInstrumentPresetUI`, `LoadMidiDeviceDefinitionUI` and `LoadSongUI`
+| rename | `RenameUI` is a `QwertyUI` with sub-classes `RenameClipUI`, `RenameDrumUI`, `RenameMidiCCUI` and `RenameOutputUI`
+| save | `SaveUI` is a `SlotBrowser` with sub-classes `SaveInstrumentPresetUI`, `SaveKitRowUI`, `SaveMidiDeviceDefinitionUI` and `SaveSongUI`
```
m-m-adams, line 90: probably needs a note about the slot browser itself? None of us really know what it's supposed to do since it was broken before the open source release but we think it's for managing old style deluge song files named like song123a, and the 'slots' are the number and the letter

m-m-adams (commented):
This is a great idea, thanks!

docs/dev/roadmap/code_structure/class_outlines/main_c.md
RJ-Eckie: I'd say that is a valid concern. My idea was to only create files for essential systems. Such as: absolute starting point (main,) file system, UI system, etc. A file would describe the broad, conceptual design choices for a subsystem which would allow for consistent implementation of subclasses, not necessarily detailed code explanations

docs/dev/roadmap/code_structure/directory_reference.md
nikodemus: How much is this aimed at being planning / consensus building for future, and how much descriptions of current state of affairs?

If the latter, I'm marginally concerned about offside (ie. not mixed with code) documentation going out of sync. In my experience current state of affairs is better described with high level comments, or README.md / DESIGN.md / SUBSYSTEM_NAME.md files in the same directories as the source files.

There's also risk of implying "this is the design for a good reason" when something is just "this is the way it's always been": there's a lot of things like that... _Documenting_ them is super valuable, as long as we take care not to accidentally enshrine them as explicit design decisions :)

For explicit decisions, consensus building and "future documentation" I love this - 🚀🔥

(I _really_ like when decision docs mention the goals and forces involved, so it's easier to figure out if there's reason to revisit or not, but depending on the case that can increase the workload to create and maintain the docs quite a lot.)

nikodemus:
How much is this aimed at being planning / consensus building for future, how much documenting explicit decisions, and how much simple descriptions of current state of affairs?

For the last, I'm marginally concerned about offside (ie. not mixed with code) documentation going out of sync. In my experience current state of affairs is better described with high level comments, or README.md / DESIGN.md / SUBSYSTEM_NAME.md files in the same directories as the source files.

There's also risk of implying "this is the design for a good reason" when something is just "this is the way it's always been": there's a lot of things like that... Documenting them is super valuable, as long as we take care not to accidentally enshrine them as explicit design decisions :)

For explicit decisions, consensus building and "future documentation" I love this - 🚀🔥

(I really like when decision docs mention the goals and forces involved, so it's easier to figure out if there's reason to revisit or not, but depending on the case that can increase the workload to create and maintain the docs quite a lot.)

PR #2934, opened 2024-11-19
Refactor MIDI engine a bit

sapphire-arches:
This PR is best reviewed 1 commit at a time.

The eventual goal is to make swapping the MIDI USB interaction to tinyusb easier, but this change should hopefully generally improve the readability of code that works with MIDI cables (not least by calling them what they are rather than MIDIDevice).


stellar-aria (approved):
Most of the churn is reorg and tool-assisted rename, which seems to be fine to me based on cursory checks. Breaking out the defaults to the channels also makes sense.

And I'm in the "one file should hold one class" camp, so breaking the MIDI bits into separate files is agreeable to me.

m-m-adams (commented):
I don't like that this has conceptually has us associating MPE with cables rather than controllers but I'm ok with merging it

stellar-aria:
> I don't like that this has conceptually has us associating MPE with cables rather than controllers but I'm ok with merging it

So the difference would be device-association vs communication-association?

m-m-adams:
> > I don't like that this has conceptually has us associating MPE with cables rather than controllers but I'm ok with merging it
> 
> So the difference would be device-association vs communication-association?

Yeah exactly. Definitely not a blocker though since the core issue is in the current structure and it's just being exposed by the rename

PR #2944, opened 2024-11-20
Voice cleanup

stellar-aria:
Just some preliminary cleanup in voice.cpp: making Sound usage `const` references where possible, pulling out the sine oscillator code into a new DSP block, etc.


src/deluge/dsp/oscillators/sine_osc.h
```diff
@@ -0,0 +1,101 @@
+#pragma once
+
+#include "deluge/util/misc.h"
+#include <array>
+#include <cstdint>
+
+namespace deluge::dsp {
+class SineOsc {
+	static constexpr size_t kSineTableSizeMagnitude = 8;
+	static constexpr std::array<int16_t, 512> sineWaveDiff __attribute__((aligned(CACHE_LINE_SIZE))) = {
```
sapphire-arches, line 10: I think we've had problems with the compiler duplicating initializations like this when they're in headers in the past, so this might need to stay in lookuptables.cpp.

src/deluge/model/voice/voice.h
```diff
@@ -140,7 +140,7 @@ class Voice final {
 	void renderFMWithFeedbackAdd(int32_t* thisSample, int32_t numSamples, int32_t* fmBuffer, uint32_t* phase,
 	                             int32_t amplitude, uint32_t phaseIncrement, int32_t feedbackAmount,
 	                             int32_t* lastFeedbackValue, int32_t amplitudeIncrement);
-	bool areAllUnisonPartsInactive(ModelStackWithVoice* modelStackWithVoice);
-	void setupPorta(Sound* sound);
-	int32_t combineExpressionValues(Sound* sound, int32_t expressionDimension);
+	bool areAllUnisonPartsInactive(ModelStackWithVoice& modelStackWithVoice) const;
+	void setupPorta(const Sound& sound);
+	int32_t combineExpressionValues(const Sound& sound, int32_t whichExpressionDimension) const;
```
sapphire-arches, line 145:
```suggestion
	int32_t combineExpressionValues(const Sound& sound, int32_t expressionDimension) const;
```

src/deluge/model/voice/voice_sample_playback_guide.h
```diff
@@ -32,10 +32,12 @@ class VoiceSamplePlaybackGuide final : public SamplePlaybackGuide {
 	bool shouldObeyLoopEndPointNow();
 	int32_t getBytePosToStartPlayback(bool justLooped);
 	int32_t getBytePosToEndOrLoopPlayback();
-	LoopType getLoopingType(Source* source);
+	LoopType getLoopingType(const Source& source) const;
```
sapphire-arches, line 35: I have a weak preference for east const rather than west as it makes the binding direction consistent, but I don't know if we ever actually agreed to a standard.

src/deluge/dsp/oscillators/sine_osc.h
stellar-aria: mmm. goodpoint. worst comes to worst I can move it to the cpp file.

stellar-aria: I'm not a fan of having them all grouped up like that tho

src/deluge/model/voice/voice_sample_playback_guide.h
stellar-aria: oh I think we did for pointers but I forgot about it for references lol

src/deluge/dsp/oscillators/sine_osc.h
stellar-aria: moved the table to the implementation file

src/deluge/model/voice/voice_sample_playback_guide.h
stellar-aria: consensus (in dev discord) seems to be that at least for references, west const is more readable

m-m-adams (approved):
Checked with the trace and this is a positive improvement all around

PR #2949, opened 2024-11-22
Refactor Sine Oscillator

stellar-aria:
- Upgrades Argon library and dependent code (Reverb's dual cosine lfo)
- Rewrites and simplifies core sine oscillator vector functions using Argon library.

### DIFF
--- a/src/deluge/dsp/oscillators/sine_osc.h
+++ b/src/deluge/dsp/oscillators/sine_osc.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include "deluge/util/misc.h"
-#include <arm_neon.h>
+#include <argon.hpp>
 #include <array>
 #include <cstdint>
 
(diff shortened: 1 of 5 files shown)


src/deluge/dsp/oscillators/sine_osc.h
```diff
@@ -25,45 +25,38 @@ class SineOsc {
...
 
-		int16x4_t strength2 =
-		    vreinterpret_s16_u16(vshr_n_u16(vshrn_n_u32(finalPhase, (32 - 16 - kSineTableSizeMagnitude)), 1));
+		// multiply by 2 due to interleave. i.e. sin(x) is actually table[x * 2]
+		Argon<uint32_t> indices = (phase >> (32 - kSineTableSizeMagnitude)) << 1;
 
-		int32x4_t enlargedValue1 = vreinterpretq_s32_u32(vshlq_n_u32(readValue, 16));
-		int16x4_t diffValue = vshrn_n_s32(vreinterpretq_s32_u32(readValue), 16);
+		//  load our two relevent table components
+		auto [sine, diff] = ArgonHalf<int16_t>::LoadGatherInterleaved<2>(sineWaveDiff.data(), indices);
```
sapphire-arches, line 55: This isn't a 1:1 refactor, but maybe the use of load2s makes up for the extra neon->CPU transfer for the phase vector? Otherwise LGTM

stellar-aria: Under the hood it uses the same `vld1q_lane_s32` intrinsic (tho in this case due to deinterleaving it's actually `vld2q_lane_s16`)

PR #2957, opened 2024-11-23
More MIDI refactor

sapphire-arches:
This increases code size by like 200 bytes and I have no idea why, so maybe it's not actually a good idea but the huge parameter lists were driving me insane.

### DIFF
--- a/src/deluge/io/midi/midi_engine.cpp
+++ b/src/deluge/io/midi/midi_engine.cpp
@@ -423,44 +423,47 @@ void MidiEngine::sendNote(MIDISource source, bool on, int32_t note, uint8_t velo
 	velocity = std::max((uint8_t)1, velocity);
 	velocity = std::min((uint8_t)127, velocity);
 
-	sendMidi(source, on ? 0x09 : 0x08, channel, note, velocity, filter);
+	if (on) {
+		sendMidi(source, MIDIMessage::noteOn(channel, note, velocity), filter);
+	}
+	else {
+		sendMidi(source, MIDIMessage::noteOff(channel, note, velocity), filter);
+	}
 }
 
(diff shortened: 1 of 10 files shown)


src/deluge/io/midi/midi_engine.cpp
```diff
@@ -557,9 +560,9 @@ void MidiEngine::sendUsbMidi(uint8_t statusType, uint8_t channel, uint8_t data1,
 				    && connectedDevice->cable[p] != &MIDIDeviceManager::upstreamUSBMIDICable3) {
 					// if it's a clock (or sysex technically but we don't send that to this function)
 					// or if it's a message that this channel wants
-					if ((statusType == 0x0F && connectedDevice->cable[p]->sendClock)
-					    || (statusType != 0x0F
-					        && connectedDevice->cable[p]->wantsToOutputMIDIOnChannel(channel, filter))) {
+					if ((message.statusType == 0x0F && connectedDevice->cable[p]->sendClock)
```
seangoodvibes, line 563: can we give 0x0F a name?

src/deluge/model/midi/status_event.h
```diff
@@ -0,0 +1,98 @@
...
+#include <cstdlib>
+
+/// Get the number of bytes associated with the provided status byte.
+size_t bytesPerStatusMessage(uint8_t status);
+
+/// Container for a MIDI status message.
+///
+/// See https://michd.me/jottings/midi-message-format-reference/ for a reference on the different status types and MIDI
+/// encoding in general.
+struct MIDIStatusMessage {
```
m-m-adams, line 30: bike shedding but maybe just midi message for brevity?

src/deluge/io/midi/midi_engine.cpp
sapphire-arches: Made checking this a method.

PR #2958, opened 2024-11-23
Feature/Reverse Stutter

markyounq:


soymonitus:
How does this work? is it by making it reverse when you past way the lowest speed possible? or directly reversing it the moment you press it?

markyounq:
> How does this work? is it by making it reverse when you past way the lowest speed possible? or directly reversing it the moment you press it?

currently it only is reverse or forward based on CF setting selection ("Reverse Stutter Rate"). The functionality is the exact same as normal stutter except reversed. 

Ideally this will get modified so you can go to the lowest speed then go into reverse as a continual motion, but my attempts at implementing it resulted in a confusing output that while technically switched between reversed & forward stutter,  had an extremely small range where it sounded anything like a normal stutter effect (something like 23-26).

m-m-adams:
@markyounq the build is failing since you edited the "wrong" string files - there's a json file to user edit that's used to generate the C files. Your strings should just be in english.json (oled) and seven_segment.json

markyounq:
> @markyounq the build is failing since you edited the "wrong" string files - there's a json file to user edit that's used to generate the C files. Your strings should just be in english.json (oled) and seven_segment.json

I removed it from the "src/deluge/gui/l10n/g_english.cpp" and its 7seg equiv.  

Is this one  "src\deluge\gui\l10n\strings.h" supposed to be edited or is it auto-generated? I didn't see anything in the header saying not to edit so I edited it

m-m-adams:
That one should be fine - I'm not sure what's happening with the actions, it says the _g files are edited but they're no longer part of your PR. Maybe try squashing everything together?

markyounq:
> @markyounq actually I think you now have the opposite problem - the strings are missing from the g_ files so they don't match before and after the compile. Try building locally and then committing the changes that get added to them

@m-m-adams  shame on me for trying to skip a step..thanks!

seangoodvibes:
@markyounq can you please document this feature in the following documents?

thanks!

Changelog: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/CHANGELOG.md

community features: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/community_features.md

menu hierarchies:
https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/menu_hierarchies.md

markyounq:
> @markyounq can you please document this feature in the following documents?
> 
> thanks!
> 
> Changelog: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/CHANGELOG.md
> 
> community features: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/community_features.md
> 
> menu hierarchies: https://github.com/SynthstromAudible/DelugeFirmware/blob/community/docs/menu_hierarchies.md

@seangoodvibes  added the relevant docs, thanks for the reminder

m-m-adams (approved):

seangoodvibes (approved):

PR #2961, opened 2024-11-23
Extract AmplitudeVector creation to own function

stellar-aria:
Moves the creation of amplitude vectors to its own function, using vector intrinsics.


src/deluge/processing/render_wave.h
```diff
@@ -113,10 +108,16 @@ startRenderingASyncLabel:
 		} while (writePos < bufferEndThisSyncRender);                                                                  \
 	}
 
-#define SETUP_FOR_APPLYING_AMPLITUDE_WITH_VECTORS()                                                                    \
-	int32x4_t amplitudeVector{0};                                                                                      \
-	setupAmplitudeVector(0) setupAmplitudeVector(1) setupAmplitudeVector(2) setupAmplitudeVector(3)                    \
-	    int32x4_t amplitudeIncrementVector = vdupq_n_s32(amplitudeIncrement << 1);
+inline int32x4_t createAmplitudeVector(int32_t amplitude, int32_t amplitudeIncrement) {
```
sapphire-arches, line 111: we should document the fact that amplitude and amplitudeIncrement are in "amplitude * 2 space", it took me awhile to figure out that's what all the shift by 1 was doing.

sapphire-arches (approved):

src/deluge/processing/render_wave.h
stellar-aria: yeah that's why I have the to-do at the return site

PR #2963, opened 2024-11-23
Refactor/clock type

m-m-adams:
Makes a clock type with auto conversion to/from double, and then replaces all internal times in the scheduler with it

This reduces the scheduler overhead by about 75% (it's small enough that it doesn't actually matter anyway but I feel good about it)


stellar-aria:
Can we just call this class "Time"? It'll conform to the naming conventions of all our other classes, and it looks like most of the variables using it are called `somethingTime` already.

stellar-aria (approved):

PR #2978, opened 2024-11-25
Feature / Arpeggiator "Note probability"

soymonitus:
This adds note probability to the arpeggiated notes (applied at the end of the chain, after rhythm and sequence length have been evaluated).

I had this PR in the past but was not happy with it because if first was not played, the sound entered an idle state and no more notes were played. I solved it now (you need to call reassessRenderSkippingStatus so it evaluates the note anyways)


docs/menu_hierarchies.md
```diff
@@ -715,6 +715,7 @@ The Sound menu contains the following menu hierarchy:
 		- Up & Down (UPDN)
 		- As Played (PLAY)
 		- Random (RAND)
+	- Note probability (NPRO)
```
seangoodvibes, line 718: Can we use PROB instead of NPRO?

thx!

src/deluge/gui/views/automation_view.cpp
```diff
@@ -121,7 +121,7 @@ const uint32_t mutePadActionUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDITION
 
 const uint32_t verticalScrollUIModes[] = {UI_MODE_NOTES_PRESSED, UI_MODE_AUDITIONING, UI_MODE_RECORD_COUNT_IN, 0};
 
-constexpr int32_t kNumNonGlobalParamsForAutomation = 63;
+constexpr int32_t kNumNonGlobalParamsForAutomation = 64;
```
seangoodvibes, line 124: Can you update the automation view documentation also please? The number and list of parameters is also in that document, thanks

soymonitus:
This one is ready for review

src/deluge/gui/menu_item/arpeggiator/midi_cv/note_probability.h
```diff
@@ -0,0 +1,40 @@
...
+class NoteProbability final : public Integer {
+public:
+	using Integer::Integer;
+	void readCurrentValue() override {
+		auto* current_clip = getCurrentInstrumentClip();
+		int64_t value = (int64_t)current_clip->arpeggiatorNoteProbability;
+		this->setValue((value * kMaxMenuValue + 2147483648) >> 32);
+	}
+	void writeCurrentValue() override {
+		getCurrentInstrumentClip()->arpeggiatorNoteProbability = (uint32_t)this->getValue() * 85899345;
```
stellar-aria, line 33: where did this number come from? can it be extracted to a constexpr definition somewhere for arpeggiatorNoteProbability stuff?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -367,6 +367,14 @@ bool ArpeggiatorBase::evaluateRhythm(bool isRatchet) {
 	return arpRhythmPatterns[rhythm].steps[rhythmPatternIndex];
 }
 
+// Returns if the arpeggiator should play the next note or not
+bool ArpeggiatorBase::evaluateNoteProbability(bool isRatchet) {
+	// If it is a rachet, use the last value, but it it is not a ratchet, calculate a new probability
+	int32_t randomChance = random(65535);
```
stellar-aria, line 373: @m-m-adams should we be using getNoise here for performance? (or even `getNoise() >> 16`)?

src/deluge/gui/menu_item/arpeggiator/midi_cv/note_probability.h
```diff
@@ -0,0 +1,40 @@
...
+#include "model/song/song.h"
+
+namespace deluge::gui::menu_item::arpeggiator::midi_cv {
+class NoteProbability final : public Integer {
+public:
+	using Integer::Integer;
+	void readCurrentValue() override {
+		auto* current_clip = getCurrentInstrumentClip();
+		int64_t value = (int64_t)current_clip->arpeggiatorNoteProbability;
+		this->setValue((value * kMaxMenuValue + 2147483648) >> 32);
```
stellar-aria, line 30:
```suggestion
		this->setValue((value * kMaxMenuValue + std::numeric_limits<int32_t>::max() + 1) >> 32);
```

stellar-aria (changes requested):

src/deluge/gui/menu_item/arpeggiator/midi_cv/note_probability.h
soymonitus: there is a comment in value_scaling.cpp that might explain it:
// (2147483648 / kMidMenuValue) == 85899345
If you search the code you will see it being used in several places. If I recall correctly, this was to scale the value somehow. I can't remember
image

int32_t computeFinalValueForStandardMenuItem(int32_t value) {
	if (value == kMaxMenuValue) {
		return 2147483647;
	}
	else if (value == kMinMenuValue) {
		return -2147483648;
	}
	else {
		// (2147483648 / kMidMenuValue) == 85899345
		return (uint32_t)value * 85899345 - 2147483648;
	}
}

src/deluge/modulation/arpeggiator.cpp
soymonitus: Since this PR is from long time ago, which I left abandoned due to a bug that I was just able to resolve this week after looking with different eyes, I have not a fresh memory of it, but I recall having used getNoise() and not getting truly random values. random() was the only way I got to get *real* probability

src/deluge/gui/menu_item/arpeggiator/midi_cv/note_probability.h
soymonitus: better with this. I'll make this change:

		this->setValue(computeCurrentValueForArpMidiCvRatchetsOrRhythm(value));

soymonitus:
@stellar-aria I made some changes and answered your questions. Let me know if any changes needed

src/deluge/gui/menu_item/arpeggiator/midi_cv/note_probability.h
soymonitus: Ok I found the refactor made with value_scaling and now I applied the same refactor to this parameter

src/deluge/modulation/arpeggiator.cpp
soymonitus: random is being used also inside  void ArpeggiatorBase::maybeSetupNewRatchet(ArpeggiatorSettings* settings)

stellar-aria: We probably just need to improve our local pseudo-random algo at some point then tbh. oh well 😅

PR #2984, opened 2024-11-27
Switch to a better memmove

stellar-aria:
This changes the (extremely naive) memmove implementation out for one that progressively copies in larger blocks until it reaches 8 quadwords.


stellar-aria:
Had to check for correctness after realizing that building with Debug disabled my test assertions in the benchmark/test. New implementation passed.

src/deluge/dsp/memmove.c
```diff
@@ -0,0 +1,48 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+static inline bool isAligned(const void* __restrict__ pointer, size_t byte_count) {
+	return (uintptr_t)pointer % byte_count == 0;
+}
+
+void* memmove(void* dst, const void* src, size_t len) {
+	if ((intptr_t)dst - (intptr_t)src >= (ptrdiff_t)len) {
```
m-m-adams, line 11: I'm like 90% sure this is UB and needs to be in assembly 

```
6.5.6 Additive operators
....
9 When two pointers are subtracted, both shall point to elements of the same array object, or one past the last element of the array object; the result is the difference of the subscripts of the two array elements.
```

src/deluge/dsp/memmove.c
```diff
@@ -0,0 +1,48 @@
...
+		const uint32_t* end = (uint32_t*)dst;
+		const uint32_t* s = (const uint32_t*)src + (len / 4);
+		uint32_t* d = (uint32_t*)dst + (len / 4);
+		while (d != end) {
+			*--d = *--s;
+		}
+		return dst;
+	}
+
+	if (isAligned(src, 2) && isAligned(dst, 2) && len % 2 == 0) {
```
m-m-adams, line 29: It should be faster to copy 2, then copy by 4s, then copy 2 again to end

src/deluge/dsp/memmove.c
```diff
@@ -0,0 +1,48 @@
...
+			*--d = *--s;
+		}
+		return dst;
+	}
+
+	/* copy backwards... */
+	const char* end = (char*)dst;
+	const char* s = (const char*)src + len;
+	char* d = (char*)dst + len;
+	while (d != end) {
```
m-m-adams, line 44: same here, better to copy until it's aligned, then go by 4s, then finish off

src/mem_functions.h
```diff
@@ -7,7 +7,7 @@ extern "C" {
 void* memset(void*, int, size_t);
 void* memcpy(void* dest, const void* src, size_t n);
 int strcmp(const char* str1, const char* str2);
-void* memmove(void* to, const void* from, size_t numBytes);
+void* memmove(void* dst, const void* src, size_t len);
```
m-m-adams, line 10: tiny nit but should it match memcpy if we're changing it anyway?

stellar-aria:
Apparently the multi-load/multi-store intrinsics are new to GCC 14.2 :(

nikodemus:
Idle thought: there's not a ton of memmove use-sites, I wonder if we could guarantee a healthy (like quadword) alignment for most?

stellar-aria:
> Idle thought: there's not a ton of memmove use-sites, I wonder if we could guarantee a healthy (like quadword) alignment for most?

We probably could, but better to just have an agnostic algorithm that's really fast anyways like this one and not accidentally get caught up when something in std:: uses it.

src/mem_functions.h
stellar-aria: Fixed

src/memmove.c
```diff
@@ -0,0 +1,107 @@
+#include "mem_functions.h"
+#include <arm_neon.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+[[gnu::always_inline]] inline uint8x16x2_t vld1q_u8_x2(unsigned char const* ptr) {
```
m-m-adams, line 8: would it be better to just update the toolchain and get these intrinsics?

m-m-adams (approved):
lgtm but also we should remember to update these when gcc is updated, this generates worse assembly than the vld1q_u8_x4 intrinsic

stellar-aria:
> lgtm but also we should remember to update these when gcc is updated, this generates worse assembly than the vld1q_u8_x4 intrinsic

Considering we'll likely get 14.2 this week, I'll probably just rip 'em out before the merge even happens. They'll collide and error otherwise

src/memmove.c
stellar-aria: yep! xpack hasn't repackaged 14.2 yet, so once that's done, I'll edit this and we can move forward.

stellar-aria:
Currently blocked by https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/issues/38

m-m-adams:
This seems to be broken, I'm getting a lot of signs of memory corruption when I test it

stellar-aria:
I'll retest it on some other platforms then, last I checked there wasn't any problems. It might be alignment stuff, which I thought about disabling the progressively larger copy sizes unless we hit half-word/word/doubleword/quadword alignment on both input and output

m-m-adams (approved):

PR #2985, opened 2024-11-27
Feature / Arpeggiator "Chord Type" for Kit Rows (and added back "Note Mode" to kit rows) 

soymonitus:
This feature allows you to chose an "emulated" chord for the arpeggiator in a kit row so you can use the Note Mode of the arpeggiator also on note rows. Before, it was useless to have a Note Mode as you could only have one note pressed, but with the new "Chord Type" parameter it is like if you were pressing a chord for the arpeggiator


soymonitus:
ready for review! @stellar-aria ? @seangoodvibes ? :)

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
```diff
@@ -53,7 +54,7 @@ void ColumnControlsKeyboard::enableNote(uint8_t note, uint8_t velocity) {
 
 	ColumnControlState& state = getState().columnControl;
 
-	for (int i = 0; i < MAX_CHORD_NOTES; i++) {
+	for (int i = 1; i < MAX_CHORD_NOTES; i++) {
```
m-m-adams, line 57: why isn't this starting from 0, won't that make it never play the root note?

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
```diff
@@ -27,6 +27,7 @@
 #include "processing/sound/sound_instrument.h"
 #include "storage/flash_storage.h"
 #include "util/functions.h"
+#include "util/lookuptables/lookuptables.h"
```
m-m-adams, line 30: is this used? I don't see an addition that touches any tables

m-m-adams (commented):
lgtm in general!

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
soymonitus: because this class was sending the root note 2 lines before that for loop
	currentNotesState.enableNote(note, velocity);
and I changed the definition of chords to contain the root note (for the arpeggiator to use), because they initially started in the second note.

src/deluge/gui/ui/keyboard/layout/column_controls.cpp
soymonitus: MAX_CHORD_NOTES is defined in lookuptables.h and used here

m-m-adams (approved):

PR #3001, opened 2024-12-01
[DRAFT] StemExport / Add more info to stem export file names

seangoodvibes:
Added BPM and Root Note and Scale Name to the Stem Export file names

image


m-m-adams (commented):
This could be done a lot more cleanly with sprintf! 

You really just need to grab the info and then use print formatting, all the concatenation seems hard to edit and likely to break if someone other than you tries to do it :)

PR #3004, opened 2024-12-02
setup dx7 when dx oscillator selected

m-m-adams:
Remove the need for the custom 1 plus synth shortcut by just setting up the dx7 engine if an oscillator is set to dx7

I think we should cherry pick this since it's tiny and makes the DX7 a lot more usable

### DIFF
--- a/src/deluge/gui/menu_item/osc/type.h
+++ b/src/deluge/gui/menu_item/osc/type.h
@@ -34,7 +34,7 @@ public:
 	Type(l10n::String name, l10n::String title_format_str) : Selection(name), FormattedTitle(title_format_str){};
 	void beginSession(MenuItem* navigatedBackwardFrom) override { Selection::beginSession(navigatedBackwardFrom); }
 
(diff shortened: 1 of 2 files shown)


seangoodvibes (approved):

bfredl:
While I agree with the UI change, this does unfortunately not include the steps to actually setup the synth for dx7:

- the ENV1 release should be set to 50 to let the DX7:s internal envelopes release phase play out (yes, we still do voice culling properly when the dx7 envelopes end)
-  The VEL-> master volume patch cable should not be created (or removed after the fact) as dx7 handles velocity per operator and this _additional_ global patch cable will throw off loaded patches which are properly velocity calibrated internally. 

I think we could do this before `ensureDX7Patch()` when `dxPatch == nullptr` (dx7 has not been set up yet) and the synth was not loaded from file (so it is a newly created synth).

Together with #2694 we can then completely remove the "custom 1 plus synth" hack.

PR #3010, opened 2024-12-03
setup dx7 synth properly when setup via osc type menu

bfredl:
follow up to #3004

- only setup dx7 on source1.
- the ENV1 release should be set to 50 to let the DX7:s internal envelopes release phase play out (yes, we still do voice culling properly when the dx7 envelopes end)
- The VEL-> master volume patch cable should not be used as dx7 handles velocity per operator and this additional global patch cable will throw off loaded patches which are properly velocity calibrated internally.

### DIFF
--- a/src/deluge/gui/menu_item/osc/type.h
+++ b/src/deluge/gui/menu_item/osc/type.h
@@ -34,7 +34,7 @@ public:
 	Type(l10n::String name, l10n::String title_format_str) : Selection(name), FormattedTitle(title_format_str){};
 	void beginSession(MenuItem* navigatedBackwardFrom) override { Selection::beginSession(navigatedBackwardFrom); }
 
-	bool mayUseDx() { return !soundEditor.editingKit(); }
+	bool mayUseDx() { return !soundEditor.editingKit() && soundEditor.currentSourceIndex == 0; }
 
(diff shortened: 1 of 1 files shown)


src/deluge/gui/menu_item/osc/type.h
```diff
@@ -66,6 +68,28 @@ class Type final : public Selection, public FormattedTitle {
 			soundEditor.currentSound->unassignAllVoices();
 		}
 
+		if (newValue == OscType::DX7 && soundEditor.currentSource->dxPatch == nullptr
```
seangoodvibes, line 71: My only comment here is that we should probably check that all the pointers are not null before using them

seangoodvibes (approved):

m-m-adams:
If you want a DX7 patch loaded exactly as it sounds then the previous shortcut still exists. The osc type menu shouldn't change anything about the existing patch! 

Setting up a DX7 patch to sound exactly as designed should be a synth mode and not an oscillator type, the oscillator type should just slot it into the existing patch. I've been playing with this for a while and it feels better integrated as an oscillator this way

bfredl:
closed too soon, the "only setup dx7 on source1." part still applies (for now, we might want to lift it later)

PR #3029, opened 2024-12-08
Feature / Enable Arpeggiator for Midi Drums and CV drums

soymonitus:
As a comment I found in the source code, "For now, only SoundDrums have Arps, but that's actually a kinda pointless restriction...". So this PR is to remove that restriction.

- [X] Base code for rendering the arps
- [x] Writing / reading of params in XML
- [x] Create menus in kit row where you can configure arp settings
- [x] Testing MIDI
- [x] Testing CV


soymonitus:
This is ready for review, although I was not able to test it for Gate. Hope someone can test it

src/deluge/model/instrument/kit.cpp
```diff
@@ -972,22 +1006,53 @@ int32_t Kit::doTickForwardForArp(ModelStack* modelStack, int32_t currentPos) {
...
 			bool reversed = (clipIsActive && modelStackWithNoteRow->isCurrentlyPlayingReversed());
 
-			int32_t ticksTilNextArpEventThisDrum = soundDrum->arpeggiator.doTickForward(
-			    &soundDrum->arpSettings, &instruction, currentPosThisRow, reversed);
+			if (thisNoteRow->drum->type == DrumType::MIDI || thisNoteRow->drum->type == DrumType::GATE) {
+				NonAudioDrum* nonAudioDrum = (NonAudioDrum*)thisNoteRow->drum;
 
-			ModelStackWithSoundFlags* modelStackWithSoundFlags =
-			    modelStackWithNoteRow->addOtherTwoThings(soundDrum, &thisNoteRow->paramManager)->addSoundFlags();
+				uint32_t sequenceLength = (uint32_t)nonAudioDrum->arpeggiatorSequenceLength;
```
stellar-aria, line 1012: please avoid the use of C-style casts and use C++ casts like static_cast or reinterpret_cast where needed

soymonitus: WHy is this needed and this type of cast is not ok? If i want to do math on several values and need to cast it first so the math operation uses unsigned, how do I do it?
uint32_t gateThreshold = (uint32_t)nonAudioDrum->arpeggiatorGate + 2147483648;
iis this ok?

stellar-aria: They exist in the current codebase because we haven't gotten around to refactoring them. For assignment, you generally don't need to cast primitives. Also, for binary operations ints are promoted to uint.

C-style casting isn't good because the compiler has to try and figure out which kind of cast to use instead of it being specified by the compiler, and it doesn't convey the intent behind the cast to any readers. For details see <https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting>

soymonitus: I have removed the cast you mentioned because it was not even needed

PR #3031, opened 2024-12-08
[DRAFT] browser: try to keep selected item on the middle row

### LINKED ISSUE #2767: Some OLED menus do not keep the highlighted section in the 2nd row when scrolling

A newer feature was implemented for OLED menus where, when scrolling down menu options, the highlighted section would remain the 2nd row. The highlighted section only reaches the bottom third row when it reaches the end of that menu's list.

Currently, any SONG or PRESET browser and SAMPLE browser does not respect this change.

To be consistent, while scrolling down/up menu items, the highlighted section should remain as the 2nd/middle row for the SONG, SYNTH, KIT, MIDI, and SAMPLE browser until it reaches the first or last item in the menu.

_No response_

OLED

Beta

1.2-beta

_No response_

nikodemus:
* Highlight should now be on the the first line for the first file, and on the last line for the last file.

* This is done by getting rid of scrollPosVertical and inferring it based on the fileIndexSelected. ...but browser.cpp is messy, and the assumptions that make this work are probably brittle.

* Fixes #2767, hopefully.

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -44,7 +44,6 @@ String Browser::currentDir{};
 bool Browser::qwertyVisible;
 
 CStringArray Browser::fileItems{sizeof(FileItem)};
-int32_t Browser::scrollPosVertical;
 int32_t Browser::fileIndexSelected;
 int32_t Browser::numCharsInPrefix;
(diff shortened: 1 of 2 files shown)


nikodemus:
Needs to deal with fileItems with more care: currently the wraparound can skip the last item.

nikodemus:
This is not a good approach.

PR #3033, opened 2024-12-09
[HOLD] check max times and avoid blocking in the same pass through task list

m-m-adams:
By going in priority order in the first pass (instead of inverse priority) the blocking logic can be made much simpler. Now it just walks through the tasks, if the task is ready and finishes before the latestFinishTime it's set as the next one to run and the latestFinishTime is updated to the last finish time that won't interfere with its start time


seangoodvibes (approved):

m-m-adams:
Beating my head against the wall here, tests pass locally on Linux and Mac

m-m-adams:
This has some rough knock on effects - it makes high priority tasks run better but low priority ones run worse, I'm gonna put a hold on it for now and try to use this logic for real time tasks and the previous logic for non real time ones

m-m-adams:
Now yields mid audio render (usually twice) to allow other tasks to run

PR #3034, opened 2024-12-09
[Draft] Developer Documentation Refactor

todd-gochenour:
Added 'Additional Developer Resouces' footer to each of the developer specific documentation.  The link to this page reveals a list of other developer documents and relevant references to the WIKI.


docs/dev/menu_hierarchies.md
```diff
@@ -1788,3 +1788,7 @@ The Audio Clip menu contains the following menu hierarchy:
 </details>
 
 </details>
+
+## Additional Information
```
seangoodvibes, line 1792: This should not move to dev. This is a user facing document to help them understand the menu hierarchies.

docs/dev/bug_fixes.md
```diff
@@ -1,13 +1,13 @@
 # Bug Fixes
```
seangoodvibes, line 1: I think we can just delete this document as it's not being maintained.

It would be too much effort to log every PR in here.

seangoodvibes (changes requested):

docs/dev/menu_hierarchies.md
todd-gochenour: Right.  So I'm not familiar with the process of reverting this change from this PR.  Should I just move the file back to root?

seangoodvibes: > Right. So I'm not familiar with the process of reverting this change from this PR. Should I just move the file back to root?

yes you would just move the file back

I would also remove the developer note that was added at the bottom

then you can commit the changes

docs/CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ which allows both creating discussion, ideas, polls and more as well as voting o
 * For more in-depth technical discussion you are welcome to join the #dev-chat channel in
   the [Deluge User Discord](https://discord.com/invite/gWaJEh6pPP).
 
-# Code review and testing
+## Code review and testing
```
seangoodvibes, line 44: This section feels a bit out of place. It feels like it should have come after the Repository contributions section which was removed.

I would keep a header Repository contributions and then refer dev contributors to the dev documentation

todd-gochenour: Changed header and added a distinct link to 'Guidelines...' before the 'Additional Developer Resources' footer.

seangoodvibes: > Changed header and added a distinct link to 'Guidelines...' before the 'Additional Developer Resources' footer.

Hmm I don't see any change

seangoodvibes (approved):

Issue #2936, opened 2024-11-19
MIDI Transpose causes sequences to jump around


seangoodvibes:
### Please describe the problem:

When you use the midi transpose feature while you're in a scale mode instrument clip, the notes currently visible shift out of view make it difficult to edit existing sequences because they aren't constantly shifting out of view

### What is the expected behavior?

Notes get transposed but stay visible so that you can edit them while transposing is happening

### What hardware did you reproduce it with?

OLED, 7SEG

### What firmware did you reproduce it with?

Release 1.1/Beethoven

nikodemus:
Possibly a bad interaction with https://github.com/SynthstromAudible/DelugeFirmware/pull/2364

PR #3054, opened 2024-12-11
fix sharpening/flattening scale notes

### LINKED ISSUE #3038: Editing note row note by press-turning select changing octave as well as note.

Was setting the iterance for a row with audition + turning select right when I inadvertently pressed select while turning and the audio glitched out. The root notes disappear from the audition column though scale mode is still engaged, and the audition pads jumped down to the -2 octave. Pressing “Scale” twice fixes it. 

It was an accident so not sure what would be expected, probably nothing. If scale mode is off nothing happens besides blinking the scale button. 

_No response_

OLED

Beta

1.2.0 Beta DD19D1F

1. With scale mode on, in a synth clip, press and hold an audition pad. 
2. Then press the select encoder and turn right.

nikodemus:
- ScaleChange was missing the target scale, used to figure out a possible y-adjustment. Fixes #3038.

- Same issue in rotateMusicalScale(), used from Song::transposeAllScaleModeClips, which might be causing #2936.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -643,6 +643,7 @@ void Song::rotateMusicalMode(int8_t change) {
 		}
 	}
 
+	changes.target = changes.applyTo(changes.source);
 	replaceMusicalMode(changes, false);
 }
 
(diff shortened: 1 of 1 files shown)


nikodemus:
Note: I didn't have time to test the MIDI transpose yet!

seangoodvibes:
Thanks @nikodemus I'll test this evening but the code looks safe!

m-m-adams (approved):

PR #3073, opened 2024-12-16
Update gh-pages home page with 1.2RC links.

todd-gochenour:


todd-gochenour:
Still haven't figured out github push/pull.

PR #3076, opened 2024-12-16
Updating cross-links in developer documentation

todd-gochenour:
So that links work in doxygen documentation.


sapphire-arches:
Navigation in the GH native UI still works which is nice.

todd-gochenour:
This change is now unnecessary as the /docs markdown documents are excluded from the API documentation build.

PR #3079, opened 2024-12-16
Feature / Arpeggiator 3.0

soymonitus:
- Added arpeggiator for all types of Kit rows (Sound, Midi and Gate).
- Menus reorganized in a more logical way now that we have a lot more parameters.
- New parameters:
    - Chord Probability: evaluates in a note step to play a chord instead of a single note
    - Chord Polyphony: this is the companion parameter of the previous one, and defines how many notes the chord will have
    - Step Repeat: defines how many times each step is repeated in the pattern created by Octave Mode & Note Mode
    - Bass Probability: evaluates a note to be replaced by the lowest (bass) note of the available set of notes.
    
TO DO:
- [x] Implement Chord Polyphony + Probability
- [x] Update docs
- [x] Testing


docs/features/automation_view.md
```diff
@@ -31,7 +31,7 @@ Automatable Parameters are broken down into four categories for Automation View
 > - **LFO 1** Rate
 > - **LFO 2** Rate
 > - **Mod FX** Offset, Feedback, Depth, Rate
-> - **Arp** Rate, Gate, Note Probability, Ratchet Probability, Ratchet Amount, Sequence Length, Rhythm, Spread Velocity, Spread Gate, Spread Octave
+> - **Arp** Rate, Gate, Note Probability, Ratchet Probability, Ratchet Amount, Sequence Length, Bass Probability, ChordProbability, Rhythm, Chord Polyphony, Spread Velocity, Spread Gate, Spread Octave
```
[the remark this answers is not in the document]
soymonitus: i changed it to Bass Probability but probably left some "bass chance"s over the place

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -2393,9 +2399,89 @@ void InstrumentClip::writeDataToFile(Serializer& writer, Song* song) {
...
+		writer.writeAttribute("randomizerLock", arpSettings.randomizerLock);
+
+		// Write locked params
 		char buffer[9];
+		// Note probability
+		writer.insertCommaIfNeeded();
+		writer.write("\n");
+		writer.printIndents();
+		writer.writeTagNameAndSeperator("lockedNoteProb");
+		writer.write("\"0x");
```
sapphire-arches, line 2438: Can you use `writeAttributeHexBytes` here instead of the more custom serialization code?

It might require splitting `lockedNoteProb` (and `locked{Base,Chord,Ratchet}Prob`) into 2 separate entries but I think that's likely a good idea anyway and it will make evolving the serialization code less painful if the higher level functions are in use.

src/deluge/model/drum/non_audio_drum.cpp
```diff
@@ -110,12 +125,381 @@ void NonAudioDrum::modChange(ModelStackWithThreeMainThings* modelStack, int32_t
...
+
+	// Write locked params
+	char buffer[9];
+	// Note probability
+	writer.insertCommaIfNeeded();
+	writer.write("\n");
+	writer.printIndents();
+	writer.writeTagNameAndSeperator("lockedNoteProb");
+	writer.write("\"0x");
+	intToHex(arpSettings.lastLockedNoteProbabilityParameterValue, buffer);
```
sapphire-arches, line 165: (same comment as above for using writeAttributeHexBytes)

src/deluge/model/instrument/kit.cpp
```diff
@@ -662,6 +663,60 @@ void Kit::renderOutput(ModelStack* modelStack, StereoSample* outputBuffer, Stere
 	GlobalEffectableForClip::renderOutput(modelStackWithTimelineCounter, paramManager, outputBuffer, numSamples,
 	                                      reverbBuffer, reverbAmountAdjust, sideChainHitPending,
 	                                      shouldLimitDelayFeedback, isClipActive, OutputType::KIT, recorder);
+
+	for (int32_t i = 0; i < ((InstrumentClip*)activeClip)->noteRows.getNumElements(); i++) {
+		NoteRow* thisNoteRow = ((InstrumentClip*)activeClip)->noteRows.getElement(i);
+		// For Midi and Gate rows, we need to call the render method of the arpeggiator
+		if (thisNoteRow->drum
+		    && (thisNoteRow->drum->type == DrumType::MIDI || thisNoteRow->drum->type == DrumType::GATE)) {
```
sapphire-arches, line 671:
```suggestion
		if (thisNoteRow->drum == nullptr) { continue; }
		if (thisNoteRow->drum->type != DrumType::MIDI && thisNoteRow->drum->type != DrumType::GATE) { continue; }
```

Since there is no `else` case here, this can be used as a "filter" to help with loop readability.

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -34,35 +34,50 @@ void NonAudioInstrument::renderOutput(ModelStack* modelStack, StereoSample* star
...
+		uint32_t chordProbability = activeInstrumentClip->arpeggiatorChordProbability;
+		uint32_t ratchetProbability = activeInstrumentClip->arpeggiatorRatchetProbability;
+		uint32_t spreadVelocity = activeInstrumentClip->arpeggiatorSpreadVelocity;
+		uint32_t spreadGate = activeInstrumentClip->arpeggiatorSpreadGate;
+		uint32_t spreadOctave = activeInstrumentClip->arpeggiatorSpreadOctave;
+
+		arpeggiator.updateParams(rhythm, sequenceLength, chordPolyphony, ratchetAmount, noteProbability,
+		                         bassProbability, chordProbability, ratchetProbability, spreadVelocity, spreadGate,
+		                         spreadOctave);
+
```
sapphire-arches, line 52: Seeing this block copied-but-not a few times makes me think that there should be a class like `ArpeggiatorConfiguration` that contains all these state variables, so it can be shared between `InstrumentClip`, `NonAudioDrum`, etc.

That would let us unify the serialization/deserialization code, avoid these 15-line-long copy blocks (since the signature for `Arpeggiator::updateParams` could just be `updateParams(ArpeggiatorConfiguration const & config)`, and make adding new config/params even easier.

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -85,8 +100,13 @@ void NonAudioInstrument::sendNote(ModelStackWithThreeMainThings* modelStack, boo
 		// Run everything by the Arp...
 		arpeggiator.noteOn(arpSettings, noteCodePreArp, velocity, &instruction, fromMIDIChannel, mpeValues);
 
-		if (instruction.noteCodeOnPostArp != ARP_NOTE_NONE) {
-			noteOnPostArp(instruction.noteCodeOnPostArp, instruction.arpNoteOn);
+		for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
+			if (instruction.arpNoteOn != nullptr && instruction.arpNoteOn->noteCodeOnPostArp[n] != ARP_NOTE_NONE) {
```
sapphire-arches, line 104:
```suggestion
		if (instruction.arpNoteOn != nullptr) {
			for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
				if (instruction.arpNoteOn->noteCodeOnPostArp[n] == ARP_NOTE_NONE) { break; }
```

The optimizer is probably already doing this transform, but IMO this improves readability.

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -96,11 +116,24 @@ void NonAudioInstrument::sendNote(ModelStackWithThreeMainThings* modelStack, boo
...
+			else {
+				break;
+			}
 		}
-		if (instruction.noteCodeOnPostArp != ARP_NOTE_NONE && type == OutputType::CV) {
-			noteOnPostArp(instruction.noteCodeOnPostArp, instruction.arpNoteOn);
+		// CV instruments could switch on a note to do a glide
+		if (type == OutputType::CV) {
+			for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
+				if (instruction.arpNoteOn != nullptr && instruction.arpNoteOn->noteCodeOnPostArp[n] != ARP_NOTE_NONE) {
```
sapphire-arches, line 130: Suggest same transform as above (move `instruction.arpNoteOn` nullptr check out of the loop and make `ARP_NOTE_NONE` check a filter).

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -137,57 +170,13 @@ void NonAudioInstrument::polyphonicExpressionEventOnChannelOrNote(int32_t newVal
...
-			// If there's actual arpeggiation happening right now and noteMode is AS_PLAYED...
-			if ((settings != nullptr) && settings->mode != ArpMode::OFF
-			    && settings->noteMode == ArpNoteMode::AS_PLAYED) {
-				// If it's not this noteCode's turn, then do nothing with it
-				if (arpeggiator.whichNoteCurrentlyOnPostArp != n) {
-					continue;
+			for (int32_t i = 0; i < ARP_MAX_INSTRUCTION_NOTES; i++) {
+				if (arpNote->noteCodeOnPostArp[i] != ARP_NOTE_NONE
+				    && arpNote->outputMemberChannel[i] != MIDI_CHANNEL_NONE) {
+					polyphonicExpressionEventPostArpeggiator(newValue, arpNote->noteCodeOnPostArp[i],
```
sapphire-arches, line 177: Should we `break` if the `noteCodeOnPostArp` is `ARP_NOTE_NONE` here? If so,

```suggestion
				if (arpNote->noteCodeOnPostArp[i] == ARP_NOTE_NONE) { break; }
				if (arpNote->outputMemberChannel[i] == MIDI_CHANNEL_NONE) { continue; }
				polyphonicExpressionEventPostArpeggiator(newValue, arpNote->noteCodeOnPostArp[i],
```

src/deluge/modulation/arpeggiator.h
```diff
@@ -28,7 +28,9 @@
 class PostArpTriggerable;
 class ParamManagerForTimeline;
 
-#define SPREAD_LOCK_MAX_SAVED_VALUES 16
+#define RANDOMIZER_LOCK_MAX_SAVED_VALUES 16
+#define ARP_NOTE_NONE 32767
+#define ARP_MAX_INSTRUCTION_NOTES 4
```
sapphire-arches, line 33: Can these be `constexpr uint32_t` values instead?

src/deluge/processing/sound/sound_instrument.cpp
```diff
@@ -395,20 +388,21 @@ void SoundInstrument::sendNote(ModelStackWithThreeMainThings* modelStack, bool i
 
 		arpeggiator.noteOff(arpSettings, noteCode, &instruction);
 
-		if (instruction.noteCodeOffPostArp != ARP_NOTE_NONE) {
-
+		for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
+			if (instruction.noteCodeOffPostArp[n] != ARP_NOTE_NONE) {
```
sapphire-arches, line 392:
```suggestion
			if (instruction.noteCodeOffPostArp[n] == ARP_NOTE_NONE) { break; }
```

Should avoid re-indenting everything at least.

src/deluge/modulation/arpeggiator.cpp
sapphire-arches, line 1: I'm basically assuming that everything in here works fine :sweat_smile:

sapphire-arches (commented):
Caveat: I basically just assumed that all the menu items and the `arpeggiator.cpp` code are working fine.

Otherwise, the change looks fine architecturally, with the possible exception of some cleanup around all the configuration variables.  A few minor coding style nits that shouldn't matter *too* much.

soymonitus:
Thanks for all the feedback @sapphire-arches i will address all of the comments and let you know for a second review ;)

src/deluge/modulation/arpeggiator.h
sapphire-arches: Actually the `*MAX*` items should be `size_t` rather than `uint32_t` (it's the same on our platform but one day the difference will be important).

soymonitus:
@sapphire-arches all feedback addressed!!

src/deluge/gui/menu_item/arpeggiator/midi_cv/bass_probability.h
```diff
@@ -0,0 +1,68 @@
...
+			Drum* currentDrum = ((Kit*)currentClip->output)->selectedDrum;
+			if (currentDrum != nullptr
+			    && (currentDrum->type == DrumType::MIDI || currentDrum->type == DrumType::GATE)) {
+				auto* nonAudioDrum = (NonAudioDrum*)currentDrum;
+				nonAudioDrum->arpSettings.bassProbability = value;
+			}
+		}
+		else if (currentClip->type == ClipType::INSTRUMENT) {
+			((InstrumentClip*)currentClip)->arpSettings.bassProbability = value;
+		}
```
sapphire-arches, line 61: For the menu items, could all ARP menu items share a `beginSession` implementation that grabs the arp settings for the current context (so basically this logic), and stuffs that into `soundEdtior` somewhere? I think it would dramatically simplify the `writeCurrentValue`/`readCurrentValue` for the majority of these new menu items.

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -2393,62 +2373,23 @@ void InstrumentClip::writeDataToFile(Serializer& writer, Song* song) {
...
+			writer.writeAttribute("bassProbability", arpSettings.bassProbability);
+			writer.writeAttribute("chordProbability", arpSettings.chordProbability);
+			writer.writeAttribute("ratchetProbability", arpSettings.ratchetProbability);
+			writer.writeAttribute("ratchetAmount", arpSettings.ratchetAmount);
+			writer.writeAttribute("sequenceLength", arpSettings.sequenceLength);
+			writer.writeAttribute("chordPolyphony", arpSettings.chordPolyphony);
+			writer.writeAttribute("rhythm", arpSettings.rhythm);
+			writer.writeAttribute("spreadVelocity", arpSettings.spreadVelocity);
+			writer.writeAttribute("spreadGate", arpSettings.spreadGate);
+			writer.writeAttribute("spreadOctave", arpSettings.spreadOctave);
```
sapphire-arches, line 2392: I think this should be done with a method on `ArpSettings`, so it can be shared. Something like `ArpSettings::writeMidiSettingsToFile(Serializer& writer)`.

src/deluge/modulation/arpeggiator.h
```diff
@@ -114,87 +133,124 @@ class ArpeggiatorSettings {
...
-	int8_t lockedSpreadGateValues[SPREAD_LOCK_MAX_SAVED_VALUES]{};
-	int8_t lockedSpreadOctaveValues[SPREAD_LOCK_MAX_SAVED_VALUES]{};
+	// Up to 16 pre-calculated randomized values for each parameter
+	int8_t lockedNoteProbabilityValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedBassProbabilityValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedChordProbabilityValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedRatchetProbabilityValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedSpreadVelocityValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedSpreadGateValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
+	int8_t lockedSpreadOctaveValues[RANDOMIZER_LOCK_MAX_SAVED_VALUES]{};
```
sapphire-arches, line 168: Did `std::array` cause problems here?

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -2458,7 +2399,38 @@ void InstrumentClip::writeDataToFile(Serializer& writer, Song* song) {
...
+		writer.writeAttributeHexBytes("lockedVelocitySpreadArray", (uint8_t*)arpSettings.lockedSpreadVelocityValues,
+		                              RANDOMIZER_LOCK_MAX_SAVED_VALUES);
+		// Spread gate
+		writer.writeAttribute("lastLockedGateSpread", arpSettings.lastLockedSpreadGateParameterValue);
+		writer.writeAttributeHexBytes("lockedGateSpreadArray", (uint8_t*)arpSettings.lockedSpreadGateValues,
+		                              RANDOMIZER_LOCK_MAX_SAVED_VALUES);
+		// Spread octave
+		writer.writeAttribute("lastLockedOctaveSpread", arpSettings.lastLockedSpreadOctaveParameterValue);
+		writer.writeAttributeHexBytes("lockedOctaveSpreadArray", (uint8_t*)arpSettings.lockedSpreadOctaveValues,
+		                              RANDOMIZER_LOCK_MAX_SAVED_VALUES);
```
sapphire-arches, line 2433: Similar to above, these should be written via a shared implementation in `ArpeggiatorSettings`.

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -2729,99 +2701,127 @@ Error InstrumentClip::readFromFile(Deserializer& reader, Song* song) {
 			while (*(tagName = reader.readNextTagOrAttributeName())) {
 
 				if (!strcmp(tagName, "rate")) {
-					arpeggiatorRate = reader.readTagOrAttributeValueInt();
+					arpSettings.rate = reader.readTagOrAttributeValueInt();
```
sapphire-arches, line 2704: Reading can be shared `ArpeggiatorSettings::readTagFromFile(Deserializer&  reader, char const * tagName)`.`

src/deluge/model/clip/instrument_clip.cpp
sapphire-arches: Similarly, the non-MIDI stuff would be like `ArpSettings::writeCommonSettingsToFile` or something like that.

src/deluge/modulation/arpeggiator.h
soymonitus: By using writeHexBytes that you suggested i needed a plain C array

sapphire-arches: `std::array` provides `.data` which gives a pointer to the first element.

sapphire-arches: similarly `.size` should be used so the constant isn't referenced directly and instead stays in sync via the array type. This could even be extended by adding a method on Serializer which takes a templated std::array or iterator begin/end pair if you really want to get in to the c++ weeds.

soymonitus: Ohh i was looking for something like that but autocompletion was not working. Ok i will revert it back to array

soymonitus:
@sapphire-arches implemented all changes!

sapphire-arches (approved):

PR #3092, opened 2024-12-18
Rewrite ClusterPriorityQueue to use STL containers instead of custom ResizeableArray

stellar-aria:
This replaces the current implementation with a much simpler and easier to reason about implementation based on a std::map and a std::set

Impact:
- Switching from newlib-nano to newlib and enabling exceptions increases size by about ~~70k~~ 20k
- Performance impact is negligible
- Attempting to link against sbrk/malloc now no longer results in a link-time error, instead a runtime error. I don't think this is a problem in practice, since the errors should get reported and tracked down before each release.

### DIFF
--- a/src/deluge/storage/cluster/cluster.cpp
+++ b/src/deluge/storage/cluster/cluster.cpp
@@ -234,7 +234,8 @@ bool Cluster::mayBeStolen(void* thingNotToStealFrom) {
 }
 
 void Cluster::addReason() {
-	// If it's going to cease to be zero, it's become unavailable
+	// If it's going to cease to be zero, it's become unavailable,
+	// so remove it from the stealables queue
 	if (this->numReasonsToBeLoaded == 0) {
 		this->remove();
 	}
--- a/src/deluge/storage/cluster/cluster.cpp
+++ b/src/deluge/storage/cluster/cluster.cpp
@@ -232,3 +232,12 @@ bool Cluster::mayBeStolen(void* thingNotToStealFrom) {
 	}
 	return true;
 }
+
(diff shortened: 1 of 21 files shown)


src/deluge/storage/cluster/cluster.cpp
```diff
@@ -232,3 +232,12 @@ bool Cluster::mayBeStolen(void* thingNotToStealFrom) {
 	}
 	return true;
 }
+
+void Cluster::addReason() {
+	// If it's going to cease to be zero, it's become unavailable
```
m-m-adams, line 237: since this is no longer used to hackily remove things from the loading queue should we clarify the comment to specify that it's being removed from its stealable queue?

src/deluge/storage/cluster/cluster_priority_queue.h
```diff
@@ -17,21 +17,43 @@
...
+			queued_clusters_.erase(&cluster);
+			return 1;
+		}
+		return 0;
+	}
+
+	/* This is currently a consequence of how priorities are calculated (using full 32bits) next-gen getPriorityRating
+	 * should return an int32_t */
+	constexpr bool hasAnyLowestPriority() const {
+		return !priority_map_.empty() && priority_map_.rbegin()->first == static_cast<uint32_t>(-1);
```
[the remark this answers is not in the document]
stellar-aria: fixed

src/deluge/storage/cluster/cluster.cpp
stellar-aria: done!

m-m-adams (approved):

PR #3096, opened 2024-12-19
PingPong Stutter

markyounq:
a stutter that pingpongs between forward & reverse, if reverse stutter is also on, it will start by  stuttering in reverse then forward


soymonitus:
Do you think it would be better now that we have more options to have it per instrument? It will free from community settings and allow you to configure each instrument separate

soymonitus:
Thanks Mark a lot for your work on this. Your code served as inspiration and guide for implement it the way it is now in nightly/beta. Closing this as it is no longer needed

PR #3097, opened 2024-12-19
Refactor AudioFileManager and Sample::RawDataType

stellar-aria:
Moves most of AudioFileManager's member functions to using references for non-nullable values, and migrates some code to Cluster where it operates (Cluster::alloc, Cluster::dealloc, Cluster::size, etc.). Also convert's sample RAW_DATA values to an enum class.

### DIFF
--- a/src/deluge/model/sample/sample.h
+++ b/src/deluge/model/sample/sample.h
@@ -74,22 +74,22 @@ public:
 	int32_t getValueSpan();
 	void finalizeAfterLoad(uint32_t fileSize);
 
-	inline void convertOneData(int32_t* value) {
+	inline q31_t convertOneData(int32_t value) {
 		// Floating point
 		if (rawDataFormat == RAW_DATA_FLOAT)
-			*value = q31_from_float(std::bit_cast<float>(value));
+			return q31_from_float(std::bit_cast<float>(value));
 
 		// Or endianness swap
 		else if (rawDataFormat == RAW_DATA_ENDIANNESS_WRONG_32) {
-			*value = swapEndianness32(*value);
+			return swapEndianness32(value);
 		}
 
 		else if (rawDataFormat == RAW_DATA_ENDIANNESS_WRONG_16) {
-			*value = swapEndianness2x16(*value);
+			return swapEndianness2x16(value);
 		}
 
 		else if (rawDataFormat == RAW_DATA_UNSIGNED_8) {
-			*value ^= 0x80808080;
+			return value ^ 0x80808080;
 		}
 	}
 
--- a/src/deluge/model/sample/sample.h
+++ b/src/deluge/model/sample/sample.h
@@ -26,10 +26,11 @@
 #include "util/fixedpoint.h"
 #include "util/functions.h"
 #include <bit>
+#include <cstdint>
 
 #define SAMPLE_DO_LOCKS (ALPHA_OR_BETA_VERSION)
 
-enum class RawDataFormat {
+enum class RawDataFormat : uint8_t {
 	NATIVE = 0,
 	FLOAT = 1,
 	UNSIGNED_8 = 2,
--- a/src/deluge/model/sample/sample.h
+++ b/src/deluge/model/sample/sample.h
@@ -25,15 +25,18 @@
 #include "util/container/array/ordered_resizeable_array_with_multi_word_key.h"
 #include "util/fixedpoint.h"
 #include "util/functions.h"
+#include <bit>
 
 #define SAMPLE_DO_LOCKS (ALPHA_OR_BETA_VERSION)
 
-#define RAW_DATA_FINE 0
-#define RAW_DATA_FLOAT 1
-#define RAW_DATA_UNSIGNED_8 2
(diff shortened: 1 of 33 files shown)


src/deluge/storage/wave_table/wave_table.cpp
```diff
@@ -514,10 +514,10 @@ Error WaveTable::setup(Sample* sample, int32_t rawFileCycleSize, uint32_t audioD
 			}
 
 			// If we're less than one sample from the end of the cluster...
-			if (byteIndexWithinCluster > audioFileManager.clusterSize - byteDepth) {
+			if (byteIndexWithinCluster > Cluster::size - byteDepth) {
 				bytesOverlappingFromLastCluster = *(uint32_t*)source;
-				byteIndexWithinCluster -= audioFileManager.clusterSize; // Might end up negative, indicating that we've
-				                                                        // got a sample overlapping the cluster boundary
+				byteIndexWithinCluster -= Cluster::size; // Might end up negative, indicating that we've
+				                                         // got a sample overlapping the cluster boundary
```
sapphire-arches, line 520: Should reformat/rewrite comment

src/deluge/storage/cluster/cluster.h
```diff
@@ -29,25 +32,51 @@ class SampleCache;
```
sapphire-arches, line 32:
```suggestion
/// Header data for a cluster. The actual cluster data is expected to be in the same allocation, after this class member. To correctly allocate an instance of this class, you must also allocate Cluster::size bytes with enough padding to safely absorb an offset of at least CACHE_LINE_SIZE.
```

src/deluge/storage/cluster/cluster.h
```diff
@@ -29,25 +32,51 @@ class SampleCache;
...
+		SAMPLE_CACHE,
+		PERC_CACHE_FORWARDS,
+		PERC_CACHE_REVERSED,
+		OTHER,
+	};
+
+	static Cluster* alloc(Cluster::Type type = Cluster::Type::SAMPLE, bool shouldAddReasons = true,
+	                      void* dontStealFromThing = nullptr);
+
+	static void dealloc(Cluster& cluster);
```
sapphire-arches, line 49: Could this be a non-static method operating on `this` instead? It feels silly to need to call it using `Cluster` when 95% of the time the variable name is `cluster` anyway (and the type information for the variable is easy to find).

src/deluge/storage/audio/audio_file_manager.cpp
```diff
@@ -1008,38 +936,38 @@ bool AudioFileManager::loadCluster(Cluster* cluster, int32_t minNumReasonsAfter)
 	}
 #endif
 
-	cluster->addReason(); // So that it can't accidentally hit 0 reasons while we're loading it, cos then it
-	                      // might get deallocated.
+	cluster.addReason(); // So that it can't accidentally hit 0 reasons while we're loading it, cos then it
+	                     // might get deallocated.
```
sapphire-arches, line 940: Comment could be moved up.

src/deluge/model/sample/sample.h
```diff
@@ -25,15 +25,18 @@
...
 
 #define SAMPLE_DO_LOCKS (ALPHA_OR_BETA_VERSION)
 
-#define RAW_DATA_FINE 0
-#define RAW_DATA_FLOAT 1
-#define RAW_DATA_UNSIGNED_8 2
-#define RAW_DATA_ENDIANNESS_WRONG_16 3
-#define RAW_DATA_ENDIANNESS_WRONG_24 4
-#define RAW_DATA_ENDIANNESS_WRONG_32 5
+enum class RawDataFormat {
```
sapphire-arches, line 32:
```suggestion
enum class RawDataFormat : uint8_t {
```

src/deluge/storage/cluster/cluster.h
stellar-aria: Mmmmmaybe? It's generally seen as bad practice to call a destructor on `this` from inside a member function, just because of all the caveats. https://isocpp.org/wiki/faq/freestore-mgmt#delete-this

src/deluge/model/sample/sample.h
stellar-aria: Fixed

src/deluge/storage/audio/audio_file_manager.cpp
stellar-aria: Fixed

src/deluge/storage/cluster/cluster.h
stellar-aria: Added

src/deluge/storage/cluster/cluster.h
stellar-aria: Eh, yeah I can do that.

src/deluge/storage/wave_table/wave_table.cpp
stellar-aria: Done

sapphire-arches (approved):

PR #3118, opened 2024-12-22
minimal infrastructure for horizontal menus

nikodemus:
- Visual idea for OLED, works transparetnly on 7-seg:
  ```
  MENU TITLE     [1]2
  -------------------
  A    B    C    D
  10   5    0    12
  ```
  If there's > 4 items some of them will flow to the next page, as
  indicated on the title row. If there's less than 4 the space is
  divided evenly between them.

- Currently only the envelope menus are rendered horizontally. Others to
  be added in separate PRs.

- Submenus for which `supportsHorizontalRendering()` is true get it,
  unless the user has disabled horizontal menus entirely.

- Active value is displayed by inverting the column, and editable by
  holding `SHIFT` and turning `SELECT`. Shortcut pad for the menu-item
  is highlighted, as are active modulations.

- Since `SHIFT` is used to edit the values, it doesn't edit by steps of
  5 in the horizontal menus, instead a simple encoder acceleration has
  been implemented there: by turning the knob faster the value will
  start increasing faster, by upto 5 per click.

- Clicking the `SELECT` encoder in horizontal menu will no open the "big
  value editor", but instead goes directly to the modulations. Similarly
  `SHIFT + SELECT` will delete automation for the currently
  selected value, and `SHIFT + PAD` for a modulation pad will patch that.

- When a shortcut (like envelope 1 attack) is opened via pads, on OLED
  it opens the parent horizontal menu instead, with the right item
  activated. If you select an inactive shortcut, the relevant menu is
  still opened, but with first item in focus instead.

- Names for horizontal menu items use `getShortName()` which defaults to
  `getName()`; envelope segments return just the first letter of the name
  instead, so the columns are labelled A, D, S, R.

- User preference for horizontal menus provided, defaults to ON.

- This is based on earlier draft PR
  https://github.com/SynthstromAudible/DelugeFirmware/pull/2446, which
  remains a good source for implementing more features for
  horizontal menus.


nikodemus:
Almost done: I was about to toggle the draft status when I noticed community features menu crashing if you scroll to the bottom. 😓

nikodemus:
once this is in adding other "integer" menus for sounds should be super easy, and i highly encourage others-not-me to get in the action :)

1. add `supportsHorizontalRendering() -> true` for the menu
2. add the parent menu links for shortcuts in `parentsForSoundShortcuts`
3. (optional, if the default is ugly) define getShortName() that provides a nice short label for the horizontal menu column

for non-sound items, you'll also need to add a new "parents" table to menus.cpp, and fetch the parent from it in parallel with the shortcut in sound_editor.cpp

if the menu contains things which aren't `patched_param::Integer`s, you'll probably need to steal some code to handle them from the earlier PR, but like 90% of it's there ready to go

i think some menus benefit from re-arranging in this setup, though, eg. moving oscillator & noise volumes to a new mixer menu - but that's not strictly necessary. just better UX to have the values you may want to adjust at the same time in the same menu when possible. earlier PR did a lot of this, sometimes a bit indiscriminately

mweigt:
Great Idea! Anything that reduces menu diving and allows easier sound editing is good!

Just an Idea: 
There happen to be exactly four buttons right below the display (synth,kit,midi,cv), which currently have no function while being inside a menu - one *could* think of using them to select the active parameter to edit?

nikodemus:
> Great Idea! Anything that reduces menu diving and allows easier sound editing is good!
> 
> Just an Idea: There happen to be exactly four buttons right below the display (synth,kit,midi,cv), which currently have no function while being inside a menu - one _could_ think of using them to select the active parameter to edit?

Not all menus will have exactly four elements. Some will have two or three, and in those cases IMO it is better to use the space available.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@
     - `HORIZONTAL ENCODER ◀︎▶︎` + `PLAY` is changed to `CROSS SCREEN` + `PLAY`
   - `Menu Highlighting (HIGH)` changes how menu highlighting is rendered on `OLED` displays by drawing a vertical bar `|` on the left edge of the display beside the selected menu item instead of highlighting the area of the selected menu item by inverting the text.
 
+#### <ins>Horizontal Menus</ins>
+- Following items display in horizontal menus on OLED, with multiple values visible and editable at the same time. Hold `SHIFT` and turn `SELECT` to edit them (horizontal menus can be disabled via Community Features):
```
seangoodvibes, line 23: "Following items" is confusing wording to me. Could we update this to be more clear?

src/deluge/gui/ui/menus.cpp
```diff
@@ -1508,7 +1508,23 @@ MenuItem* paramShortcutsForSounds[][8] = {
     {&lfo2RateMenu,           &lfo2SyncMenu,           &lfo2TypeMenu,                  &reverbAmountMenu,              &reverbPanMenu,       &reverbWidthMenu,       &reverbDampingMenu,       &reverbRoomSizeMenu                },
     {&delayRateMenu,          &delaySyncMenu,          &delayAnalogMenu,               &delayFeedbackMenu,             &delayPingPongMenu,   nullptr,                nullptr,                  nullptr                            },
 };
-
+Submenu* parentsForSoundShortcuts[][8] = {
```
seangoodvibes, line 1511: just a nit, I know you were copying the other arrays, but can we replace "8" with kDisplayHeight?

src/deluge/gui/ui/menus.h
```diff
@@ -85,6 +85,8 @@ extern MenuItem* paramShortcutsForAudioClips[15][8];
 extern MenuItem* paramShortcutsForSongView[15][8];
 extern MenuItem* paramShortcutsForKitGlobalFX[15][8];
 
+extern deluge::gui::menu_item::Submenu* parentsForSoundShortcuts[15][8];
```
seangoodvibes, line 88: same nit here about trying to replace [8] and [15] with kDisplayHeight and kDisplayWidth respectively

seangoodvibes:
@nikodemus just testing this now.

first thing I wanted to check was that you would still be able to enter automation view from the horizontal menu for the parameter selected by pressing "c"

unfortunately it's not working

if you turn off the horizontal menu and enter any parameter menu, including ADSR, you can press clip to open automation view as the "root UI" for you to interact while you remain in the menu

could you try to get that working?

thanks

nikodemus:
> if you turn off the horizontal menu and enter any parameter menu, including ADSR, you can press clip to open automation view as the "root UI" for you to interact while you remain in the menu

@seangoodvibes thanks - I never new you could do that :) Should be working now...

nikodemus:
Automation view should be ok now.

Two "can be handled later" things I would like to solve, though:

- When looking at automation from horizontal menu by hitting CLIP, right now turning SELECT (no shift) adjusts the value instead of changing the column, because I could not figure out how to tell automation view to display a different parameter yet. `selectAutomationViewParameter()` seems promising, but I'd need to add something that redirects to it to `MenuItem` -- and if there's something better I would prefer to avoid that.

- If you edit the automation (and hence enter automation view proper, I guess?) after entering automation view from a horizontal menu, hitting CLIP to exit automation view takes you entirely out of the menu. I think this makes sense for a single value editor, but in case of horizontal menu I think it should return there. Haven't had the time to dig in enough to figure out how tricky that would be. I'm vaguely guessing `swapOutRootUILowLevel()` is what's doing the work here.

nikodemus:
@seangoodvibes the automation view interaction should be fine now. 🚀

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@
     - `HORIZONTAL ENCODER ◀︎▶︎` + `PLAY` is changed to `CROSS SCREEN` + `PLAY`
   - `Menu Highlighting (HIGH)` changes how menu highlighting is rendered on `OLED` displays by drawing a vertical bar `|` on the left edge of the display beside the selected menu item instead of highlighting the area of the selected menu item by inverting the text.
 
+#### <ins>Horizontal Menus</ins>
+- The menus for following items have been updated on OLED, with multiple values visible and editable at the same time. Hold `SHIFT` and turn `SELECT` to edit them. This feature is on by default, and can be disabled via `SETTINGS > COMMUNITY FEATURES`.
```
seangoodvibes, line 23:
```suggestion
- The menus for the following items have been updated on OLED, with multiple values visible and editable at the same time. Hold `SHIFT` and turn `SELECT` to edit them. This feature is on by default, and can be disabled via `SETTINGS > COMMUNITY FEATURES`.
```

docs/community_features.md
```diff
@@ -31,6 +31,9 @@ back up your SD card!
 now stands out from the tail in proportion to its velocity. At velocity 127 it is identical to official,
 at velocity 0 it would look the same as its tail (but you can't have 0 velocity).
 
+#### 2.2 Horizontal Menus
+- The menus for following items have been updated on OLED, with multiple values visible and editable at the same time. Hold `SHIFT` and turn `SELECT` to edit them. This feature is on by default, and can be disabled via `SETTINGS > COMMUNITY FEATURES`.
```
seangoodvibes, line 35:
```suggestion
- The menus for the following items have been updated on OLED, with multiple values visible and editable at the same time. Hold `SHIFT` and turn `SELECT` to edit them. This feature is on by default, and can be disabled via `SETTINGS > COMMUNITY FEATURES`.
```

seangoodvibes:
@nikodemus reviewed it, looks great! just a couple documentation nits and then I'm ready to approve and get this in!

seangoodvibes (approved):

PR #3123, opened 2024-12-23
RWCL Fix #2: Replace RWCL with Yield in Preset scrolling functions

seangoodvibes:
Replaced call to RWCL with Yield in a few spots to help with preset scrolling which would previously drop voices:

- View::navigateThroughPresetsForInstrumentClip
- Song::ensureAllInstrumentsHaveAClipOrBackedUpParamManager
- Browser::emptyFileItems
- Browser::deleteSomeFileItems
- ArrangerView::navigateThroughPresets


stellar-aria (approved):
I have no context for how well these work, but as a stopgap fix until we get preemptive scheduling, this lgtm

PR #3124, opened 2024-12-23
RWCL Fix #3: Replace RWCL with Yield in UI rendering functions

seangoodvibes:
Replaced RWCL with Yield in several UI rendering spots

Should help with drop outs happening in Song Row View


stellar-aria (approved):
Same as previous review: No idea how these affect things, but they're a decent stopgap until preemption

PR #3130, opened 2024-12-24
Convert MIDI labels to STL map

stellar-aria:
Also converts the backing label storage of a String to a std::string_view, since only the raw C-string was ever stored/retrieved


src/deluge/model/instrument/midi_instrument.cpp
```diff
@@ -1237,19 +1237,25 @@ void MIDIInstrument::sendNoteToInternal(bool on, int32_t note, uint8_t velocity,
 	}
 }
 
-String* MIDIInstrument::getNameFromCC(int32_t cc) {
+std::string_view MIDIInstrument::getNameFromCC(int32_t cc) {
 	if (cc >= 0 && cc < kNumRealCCNumbers) {
-		MIDILabel* midiLabel = midiLabelCollection.labels.getLabelFromCC(cc);
```
seangoodvibes, line 1242: Looking at this on my phone and this change seems wrong here as if it's >=0 and < num real cc's then it's in range, not out of range

Will double check when I get in front of computer

src/deluge/model/instrument/midi_instrument.h
```diff
@@ -137,5 +137,6 @@ class MIDIInstrument final : public NonAudioInstrument {
 	void outputAllMPEValuesOnMemberChannel(int16_t const* mpeValuesToUse, int32_t outputMemberChannel);
 	Error readMIDIParamFromFile(Deserializer& reader, int32_t readAutomationUpToPos,
 	                            MIDIParamCollection* midiParamCollection, int8_t* getCC = NULL);
-	MIDILabelCollection midiLabelCollection;
+
+	deluge::fast_map<uint8_t, std::string_view> labels;
```
seangoodvibes, line 141: Just want to confirm, does this change have any impact on the weight of the label strings in memory?

I know it was one of the reasons Mark pushed me to use a resizable array.

stellar-aria: No if anything, it's lighter.

src/deluge/model/instrument/midi_instrument.cpp
stellar-aria: Fixed

seangoodvibes (approved):

PR #3131, opened 2024-12-24
Convert MIDIParamVector to STL Map

stellar-aria:


src/deluge/modulation/midi/midi_param_collection.h
```diff
@@ -67,8 +67,11 @@ class MIDIParamCollection final : public ParamCollection {
 
 	deluge::modulation::params::Kind getParamKind() override { return deluge::modulation::params::Kind::MIDI; }
 
-	MIDIParamVector params;
+	/// A map between the CC value and the Automatable Param
+	deluge::fast_map<int32_t, AutoParam> params;
```
seangoodvibes, line 71: Do we need the cc to take up int32_t? (I saw you used int8 in the midi labels PR

stellar-aria: Ah, yeah, we can probably switch that out, I was initially following the key parameter input, but it was internally stored as a uint8_t

stellar-aria: Fixed

seangoodvibes (approved):

seangoodvibes (approved):

PR #3132, opened 2024-12-24
Update the kit velocity keyboard view's zoom levels behavior

Metamere:
Instead of only having square pads for the velocity keyboard view, allow for rectangular pads, and choose dimensions for those rectangles (and squares) to maximize the used control surface area. I made an interactive demo using p5.js to prototype the behavior, so you can test out my proposed scheme and compare it to the current one here: https://openprocessing.org/sketch/2492402
In my scheme, the number of pads in each zoom level has a more gradual curve, with each carefully set up to maximize usability. It is functional in its current state, but I would like to add a couple of refinements related to the pad brightness and color selection.


Metamere:
Refinements completed. Please review and test at your convenience.

src/deluge/gui/ui/keyboard/keyboard_screen.cpp
```diff
@@ -638,7 +638,7 @@ ActionResult KeyboardScreen::buttonAction(deluge::hid::Button b, bool on, bool i
 	}
 
 	else {
-		requestRendering();
+		// requestRendering();
```
sapphire-arches, line 641:
```suggestion
```

src/deluge/gui/ui/keyboard/keyboard_screen.cpp
```diff
@@ -153,7 +153,7 @@ ActionResult KeyboardScreen::padAction(int32_t x, int32_t y, int32_t velocity) {
 				pressedPads[idx].active = false;
 				markDead = idx;
 				if ((AudioEngine::audioSampleTimer - pressedPads[idx].timeLastPadPress) > FlashStorage::holdTime) {
-					pressedPads[idx].padPressHeld = true;
+					pressedPads[idx].padPressHeld = true; // why is this needed if the pad has been released?
```
sapphire-arches, line 156:
```suggestion
					pressedPads[idx].padPressHeld = true;
```

This logic detects when a pad in the `pressedPads` list has been held for long enough to be considered "held". That is, it detects the difference between a short tap and a long press.

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
```diff
@@ -24,121 +24,191 @@
...
-					// this press is more recent, use its velocity.
-					currentNotesState.notes[noteOnIdx].velocity = velocity;
-					noteOnTimes[noteOnIdx] = thisOnTime;
+	uint32_t active_notes = 0;
+	std::array<int32_t, kMaxNumActiveNotes> note_on_times{};
+	uint8_t edge_size_x = (uint32_t)getState().drums.edge_size_x;
+	uint8_t edge_size_y = (uint32_t)getState().drums.edge_size_y;
+	uint8_t offset = getState().drums.scroll_offset;
+	uint8_t pads_per_row = kDisplayWidth / edge_size_x;
+	uint8_t highest_clip_note = getHighestClipNote();
```
sapphire-arches, line 36:
```suggestion
	uint32_t edge_size_x = (uint32_t)getState().drums.edge_size_x;
	uint32_t edge_size_y = (uint32_t)getState().drums.edge_size_y;
	uint32_t offset = getState().drums.scroll_offset;
	uint32_t pads_per_row = kDisplayWidth / edge_size_x;
	uint32_t highest_clip_note = getHighestClipNote();
```

Using smaller integer types can help reduce storage requirements in RAM, but stack space is essentially free and using smaller integers can force the compiler to insert sign-extension/high-bit-clear instructions.

This applies to `x` in this function, and the similar variables in `renderPads`.

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
```diff
@@ -24,121 +24,191 @@
...
+	bool odd_pad = (edge_size_x % 2 == 1 && edge_size_x > 1);
+	for (int32_t idx_press = 0; idx_press < kMaxNumKeyboardPadPresses; ++idx_press) {
+		if (presses[idx_press].active) {
+			uint8_t x = presses[idx_press].x;
+			if (x < kDisplayWidth) {
```
sapphire-arches, line 41: These checks might be easier to read if they're written like below:

```suggestion
		if (!presses[idx_press].active) { continue; }
		uint8_t x = presses[idx_press].x;
		if (x >= kkDisplayWidth) { continue; }
```

That way, anyone reading the loop knows that these checks are more like filters on the presses and not something where interesting things happen off-screen (below the end of the then-block).

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
```diff
@@ -24,121 +24,191 @@
...
+			zoom_level_changed = true;
+		}
+		else
+			return;
+
+		state.edge_size_x = zoom_arr[state.zoom_level][0];
+		state.edge_size_y = zoom_arr[state.zoom_level][1];
+		char buffer[14] = "Pad Area:   ";
+		auto display_offset = (display->haveOLED() ? 9 : 0);
+		intToString(state.edge_size_x * state.edge_size_y, buffer + display_offset, 1);
```
sapphire-arches, line 105:
```suggestion
		DEF_STACK_STRING_BUF(buffer, 16);
		if (display->haveOLED()) {
			buffer.append("Pad Area: ");
		}
		buffer.appendInt(state.edge_size_x * state.edge_size_y, buffer + display_offset);
```

src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -2375,8 +2375,10 @@ void InstrumentClip::writeDataToFile(Serializer& writer, Song* song) {
 	// Community Firmware parameters (always write them after the official ones, just before closing the parent tag)
 	writer.writeAttribute("keyboardLayout", keyboardState.currentLayout);
 	writer.writeAttribute("keyboardRowInterval", keyboardState.isomorphic.rowInterval);
-	writer.writeAttribute("drumsScrollOffset", keyboardState.drums.scrollOffset);
-	writer.writeAttribute("drumsEdgeSize", keyboardState.drums.edgeSize);
+	writer.writeAttribute("drumsScrollOffset", keyboardState.drums.scroll_offset);
+	writer.writeAttribute("drumsEdgeSizeX", keyboardState.drums.edge_size_x);
+	writer.writeAttribute("drumsEdgeSizeY", keyboardState.drums.edge_size_y);
```
sapphire-arches, line 2380: These can be inferred from the zoom level, no?

src/deluge/gui/ui/keyboard/keyboard_screen.cpp
Metamere: But isn't that already covered by the section above, in lines 126-133? If the pad is released, then it's not being held anymore, right?

src/deluge/model/clip/instrument_clip.cpp
Metamere: Yes they can, though it would require looking up the values from the array every time. Is that the more efficient way to go? Is it better to minimize these state variables?

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
Metamere: The display_offset variable is not defined then, but if I remove that portion, it no longer displays to the screen.

src/deluge/model/clip/instrument_clip.cpp
sapphire-arches: The code can stay the same, but I don't think they need to be stored in the save file. Instead, the X/Y edge sizes should be computed based on just the saved zoom level.

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
sapphire-arches: oh, the `display->displayPopup(buffer)` line just below this should use `buffer.c_str()` instead as well.

src/deluge/gui/ui/keyboard/keyboard_screen.cpp
sapphire-arches: Oh I completely missed that this is in the release branch -- I think this is a copy/paste error and the check should be removed to unconditionally set `padPressHeld` to false here.

src/deluge/gui/ui/keyboard/layout/velocity_drums.cpp
Metamere: I don't see that line in the code. Did you possibly miss pasting a line into your suggestion?

Metamere: Oh, I pasted over that line, I see the issue. Thanks.

src/deluge/model/clip/instrument_clip.cpp
Metamere: Can those variables be declared as public in the .h file so the values don't need to be looked up in every function, or is that also bad practice here? Sorry, my C++ skills are pretty rusty and out-of-date.

Metamere:
I'm not very experienced with using github Do I want to do the sync changes, or pull origin? 
image

src/deluge/model/clip/instrument_clip.cpp
sapphire-arches: those should already be public, since KeyboardStateDrums is a struct.

If we want to be object oriented about it, everything in KeyboardStateDrums should be private with a `setZoomLevel` function which maintains the invariant that the zoom level and edge sizes align.

src/deluge/gui/ui/keyboard/state_data.h
```diff
@@ -30,8 +30,10 @@ struct KeyboardStateIsomorphic {
 };
 
 struct KeyboardStateDrums {
-	int32_t scrollOffset = 0;
-	int32_t edgeSize = 4;
+	int32_t scroll_offset = 0;
+	int32_t edge_size_x = 4;
+	int32_t edge_size_y = 4;
```
sapphire-arches, line 35:
```suggestion
```

No longer used, I think.

src/deluge/gui/ui/keyboard/layout/velocity_drums.h
```diff
@@ -42,25 +46,36 @@ class KeyboardLayoutVelocityDrums : KeyboardLayout {
 	bool supportsKit() override { return true; }
 
 private:
-	inline uint8_t noteFromCoords(int32_t x, int32_t y) {
-		uint8_t edgeSize = (uint32_t)getState().drums.edgeSize;
-		uint8_t padsPerRow = kDisplayWidth / edgeSize;
-		return (x / edgeSize) + ((y / edgeSize) * padsPerRow) + getState().drums.scrollOffset;
-	}
+	RGB note_colours[128];
```
sapphire-arches, line 49: Sorry I missed this in the first review, but is there a performance reason to precompute all these colors? Otherwise, I think we get a better binary size tradeoff by just computing the color using `::fromHue` in the rendering function.

Metamere: That was in the original version, so I thought that was the best way to do it. I can try changing it, sure. I'm also testing out setting up private variables for edge_size_x and edge_size_y in the .h, and using getter and setter functions in the .cpp. I think that should make things cleaner there. Would that be good?

sapphire-arches: > That was in the original version, so I thought that was the best way to do it. I can try changing it, sure

If you use a relwithdebinfo build, the debug logging over sysex should have information about how much time UI work is taking. If you mash on the keys to force a bunch of rerenders that should give you some sense of the performance impact.

> I'm also testing out setting up private variables for edge_size_x and edge_size_y in the .h, and using getter and setter functions in the .cpp. I think that should make things cleaner there. Would that be good?

Yeah, I like using getters/settters there.

sapphire-arches:
If you want to write your own changelog entry, I think that's the last thing this PR needs and then it's good to merge. Thanks!

Metamere:
Um, sorry, but I'm not sure what I need to do from this point. I hope I didn't screw it up. Do I need to edit the file at https://github.com/SynthstromAudible/DelugeFirmware/blob/community/CHANGELOG.md to include that with this PR? Do I add my changes to the 1.3.0 section there?

Metamere:
Alright, I updated the changelog entry. I covered everything I could think of that the user could notice, but please let me know if that is more than is needed.

sapphire-arches (approved):

Metamere:
It looks like the build checks fail because I took out the precalculate function from the .cpp. It didn't build right when I took it out the override in of the .h, so do I need to leave it in the .cpp but empty? Or do something else?

sapphire-arches:
Deleting it from both `.h` and `.cpp` should fix the build, hopefully.

Metamere:
> Deleting it from both `.h` and `.cpp` should fix the build, hopefully.

When I do that it gives this error when building:
C:/Users/metam/Documents/DelugeFirmware/src/deluge/gui/ui/keyboard/keyboard_screen.cpp:61:37: error: cannot declare variable 'deluge::gui::ui::keyboard::keyboardLayoutVelocityDrums' to be of abstract type 'deluge::gui::ui::keyboard::layout::KeyboardLayoutVelocityDrums'
   61 | layout::KeyboardLayoutVelocityDrums keyboardLayoutVelocityDrums{};
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from C:/Users/metam/Documents/DelugeFirmware/src/deluge/gui/ui/keyboard/keyboard_screen.cpp:54:
C:/Users/metam/Documents/DelugeFirmware/src/deluge/gui/ui/keyboard/layout/velocity_drums.h:31:7: note:   because the following virtual functions are pure within 'deluge::gui::ui::keyboard::layout::KeyboardLayoutVelocityDrums':
   31 | class KeyboardLayoutVelocityDrums : KeyboardLayout {
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from C:/Users/metam/Documents/DelugeFirmware/src/deluge/gui/ui/keyboard/keyboard_screen.cpp:46:
C:/Users/metam/Documents/DelugeFirmware/src/deluge/gui/ui/keyboard/layout.h:59:22: note:     'virtual void deluge::gui::ui::keyboard::KeyboardLayout::precalculate()'
   59 |         virtual void precalculate() = 0;
      |                      ^~~~~~~~~~~~
[130/369  35% :: 20.818] Building CXX object CMakeFiles/deluge.dir/RelWithDebInfo/src/deluge/gui/ui/menus.cpp.obj
ninja: build stopped: subcommand failed.

stellar-aria:
@Metamere just put the empty body `{};` at the declaration in the header so we know it's stubbed there, and it should be fine to merge

Metamere:
Okay, that all makes sense now, and I got it fixed. Thanks for your patience!

stellar-aria (approved):

sapphire-arches (approved):

PR #3144, opened 2024-12-25
Removed high cpu usage indicator from docs

seangoodvibes:


nikodemus (approved):

stellar-aria:
Shouldn't this be based against release/1.2?

PR #3136, opened 2024-12-25
Minor MIDI system refactor

sapphire-arches:
Each commit here cleans up some small thing that caused me problems during the big root complex refactor.

Best reviewed one commit at a time.

PR #3157, opened 2024-12-26
Create MIDI "root complex" abstraction.

sapphire-arches:
This is basically here as an intermediate step for the tinyusb integration, where the guts of all the *USB* classes will be ripped out and replaced with stuff that calls in to tinyusb.

Draft until #3136 is merged, as this PR builds on that one (only the last 3 commits here are new).


m-m-adams:
@sapphire-arches I'm trying to test this but it's got an issue with GSL markers, do you know where that should be included from?

m-m-adams (approved):

stellar-aria (approved):

PR #3159, opened 2024-12-26
Add Midi devices templates

stosem:
Hello, dear Deluge community!

Since we provide such a great feature as MIDI device templates, I thought it would be nice to have some popular synths templates out of the box (like we did for midi_follow).
If we will share our device MIDI templates, we can create a huge collection of synths right in Deluge (more than Oxi One has :).

Here is my small contribution, I've added 4 of my synths, which I can test. I'm still not sure about the CC parameter naming convention, and XML file location. So I would like to receive any feedback from you.
What do you think about this idea?

Thanks! And have a nice holidays!


seangoodvibes (approved):

seangoodvibes:
Hi @stosem thanks for contributing! I just moved your files to an sd_card folder so it's more clear you're providing files that can be copied onto the sd card. Thanks a lot!

m-m-adams (approved):
Awesome!! I've only used the hydrasynth one but that's super convenient

PR #3162, opened 2024-12-26
Refactor out staticDIR

stellar-aria:
Removes global staticDIR in favor of local FatFS::Directory instances

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -18,6 +18,7 @@
 #include "gui/ui/browser/browser.h"
 #include "definitions_cxx.hpp"
 #include "extern.h"
+#include "fatfs.hpp"
 #include "gui/context_menu/delete_file.h"
 #include "gui/l10n/l10n.h"
 #include "gui/ui_timer_manager.h"
@@ -35,6 +36,7 @@
 #include "storage/file_item.h"
 #include "storage/storage_manager.h"
 #include "util/functions.h"
+#include "util/try.h"
 #include <cstring>
 #include <new>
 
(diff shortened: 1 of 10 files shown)


src/deluge/gui/ui/browser/browser.cpp
```diff
@@ -304,10 +295,12 @@ Error Browser::readFileItemsForFolder(char const* filePrefixHere, bool allowFold
 		audioFileManager.loadAnyEnqueuedClusters();
 		FilePointer thisFilePointer;
 
-		result = f_readdir_get_filepointer(&staticDIR, &staticFNO, &thisFilePointer); /* Read a directory item */
+		std::tie(staticFNO, thisFilePointer) = D_TRY_CATCH(dir.read_and_get_filepointer(), error, {
```
m-m-adams, line 298: Could we use an actual try/catch now?

stellar-aria: nope, the FatFS interface uses std::expected, which is what D_TRY/D_TRY_CATCH work with. I honestly prefer std::expected to exceptions when possible anyways, closer to Rust's Result and it doesn't rely on the weird type matching

m-m-adams (commented):
I think we also need to configure fatfs as re entrant to do this safely, this change means multiple different dirs can be held but fatfs still can't access anything other than the last one it opened. To do that I think we need mutex support (and blocking) as well

jamiefaye (commented):
The reason we have staticDIR is because iterating through a large directory is quite slow. Since there is a limit on how many directory entries can go into a SysEx reply, we limit each request to a small number of line items. If we iterated from the beginning to get to the subrange requested each time, it really does get slow. So I came up with a cache scheme.

stellar-aria:
> I think we also need to configure fatfs as re entrant to do this safely, this change means multiple different dirs can be held but fatfs still can't access anything other than the last one it opened. To do that I think we need mutex support (and blocking) as well.
Looks like FatFS needs to be provided some ff_mutex_take, ff_mutex_give, ff_mutex_create and ff_mutex_delete.

> The reason we have staticDIR is because iterating through a large directory is quite slow. Since there is a limit on how many directory entries can go into a SysEx reply, we limit each request to a small number of line items. If we iterated from the beginning to get to the subrange requested each time, it really does get slow. So I came up with a cache scheme.

What do you mean? Git blame for staticDIR and staticFNO point to Rohan's original commit to the codebase. And staticDIR is never used in a function that doesn't call f_opendir() first, so it's never used contextually without a relevant creation the way that I would expected piecemeal loading works

stellar-aria:
@jamiefaye if you're referring to sxDIR in https://github.com/SynthstromAudible/DelugeFirmware/blob/d4dda6825904b8ccb3450588b5feee592d7c8029/src/deluge/storage/smsysex.cpp#L26 this PR does not touch that

jamiefaye:
I am wrong about the staticDIR. The DIR iteration caching stuff is over in
smsysex.cpp.

– Jamie

PR #3172, opened 2024-12-28
replace UI::getName() virtual functions with a lookup table for names

nikodemus:
- conditionalize name storage on ENABLE_MATRIX_DEBUG, the only current user of the name

- make sure all concrete UIs actually have an UIType, and no abstract ones have one. minor exception with ContextMenus, which all same the same type.

- shrinks the binary by ~1.2kB. storing the info in members is marginally better, but leads to uglier code. (earlier 52kB claim was just bogus measurement, *blush*)

### DIFF
--- a/src/definitions_cxx.hpp
+++ b/src/definitions_cxx.hpp
@@ -194,29 +194,37 @@ constexpr int32_t kNumInstrumentSlots = 1000;
 // Don't ever make this less! The zoom rendering code uses this buffer for its stuff
 constexpr size_t kFilenameBufferSize = 256;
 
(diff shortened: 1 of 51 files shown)


seangoodvibes (approved):

seangoodvibes:
Looks good to me but it's not building, looks like something was using getName after all lol

nikodemus:
> Looks good to me but it's not building, looks like something was using getName after all lol

ironically, that came in the horizontal menu PR 😅

seangoodvibes (approved):

PR #3183, opened 2024-12-30
refactor rename-uis, allow naming CV clips as well

nikodemus:
- allow deleting names (by replacing them with empty) for all but drums

- consolidate shared code in RenameUI class: subclasses implement only a couple of methods to specialize their behaviour

- ad-hoc dispatch in sound_editor.cpp replaced with normal shortcut machinery

- shrinks the binary by just a hair, but *shrinks*, not grows!

### DIFF
--- a/src/deluge/gui/ui/rename/rename_drum_ui.h
+++ b/src/deluge/gui/ui/rename/rename_drum_ui.h
@@ -24,21 +24,15 @@ class SoundDrum;
 
 class RenameDrumUI final : public RenameUI {
 public:
-	RenameDrumUI();
-	bool opened() override;
-	ActionResult buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) override;
-	ActionResult padAction(int32_t x, int32_t y, int32_t velocity) override;
-	ActionResult verticalEncoderAction(int32_t offset, bool inCardRoutine) override;
-	bool getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) override;
-
-	// ui
-	bool exitUI() override;
+	RenameDrumUI(const char* title_) : RenameUI(title){};
 
 protected:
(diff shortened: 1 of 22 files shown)


src/deluge/gui/ui/rename/rename_drum_ui.h
```diff
@@ -24,21 +24,15 @@ class SoundDrum;
...
-	RenameDrumUI();
-	bool opened() override;
-	ActionResult buttonAction(deluge::hid::Button b, bool on, bool inCardRoutine) override;
-	ActionResult padAction(int32_t x, int32_t y, int32_t velocity) override;
-	ActionResult verticalEncoderAction(int32_t offset, bool inCardRoutine) override;
-	bool getGreyoutColsAndRows(uint32_t* cols, uint32_t* rows) override;
-
-	// ui
-	bool exitUI() override;
+	RenameDrumUI(const char* title_) : RenameUI(title){};
```
sapphire-arches, line 27:
```suggestion
	RenameDrumUI(const char* title_) : RenameUI(title_){};
```

sapphire-arches: If you're feeling brave, making the member in QwertyUI a `string_view` and renaming it to `title_` so it matches the convention we're trying to use might be better.

nikodemus: > If you're feeling brave, making the member in QwertyUI a `string_view` and renaming it to `title_` so it matches the convention we're trying to use might be better.

I'm a bit confused. Is the convention to name initializer arguments `foo_` and members `foo`, or the other way around?

nikodemus:
Updated, and found the same title_ -> title bug twice more. I've no idea how those passed my own testing... 😅

sapphire-arches (approved):

PR #3184, opened 2024-12-31
Convert staticDIR to FatFS interface

stellar-aria:
A halfway-compromise to #3162 until we have re-entrancy. This converts staticDIR to from DIR to FatFS::Directory

### DIFF
--- a/src/fatfs/fatfs.hpp
+++ b/src/fatfs/fatfs.hpp
@@ -126,8 +126,6 @@ using FileInfo = FILINFO;
 
 class Directory {
 public:
-  Directory(Directory &) = default;  // Copy constructor
-  Directory(Directory &&) = default; // Move constructor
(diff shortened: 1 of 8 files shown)


src/fatfs/fatfs.hpp
```diff
@@ -161,7 +159,7 @@ class Directory {
   constexpr DIR &inner() { return dir_; }
 
 private:
-  Directory() = default;
+  //Directory() = default;
```
seangoodvibes, line 163: why commenting this out instead of removing?

stellar-aria: So ideally it would be private, to prevent creating an empty directory handler, but that's not possible with us using a global version we have to initialize as empty

seangoodvibes (approved):

PR #3188, opened 2024-12-31
Refactor DX7 openFile, general memory system improvements

stellar-aria:
Refactors tryLoad/openFile to take a std::string_view, and moves the file operations to the C++ FatFS interface. 

Tested on OLED.


src/deluge/storage/DX7Cartridge.h
```diff
@@ -87,31 +88,31 @@ class DX7Cartridge {
...
 		}
 
-		if (pos[0] != 0xF0) {
+		if (pos[0] != std::byte{0xF0}) {
 			// it is not, just copy the first 4096 bytes
 			memcpy(voiceData + 6, pos, 4096);
 			return String::STRING_FOR_DX_ERROR_NO_SYSEX_START;
 		}
 
 		int i;
```
sapphire-arches, line 108: `i` should be `size_t` (or `std::span<std::byte>::size_type` if you want to be verbose about it)

src/deluge/gui/menu_item/dx/cartridge.cpp
```diff
@@ -17,64 +17,65 @@
...
+	using enum deluge::l10n::String;
 	const int minSize = SMALL_SYSEX_SIZE;
 
-	FILINFO fno;
-	int result = f_stat(path, &fno);
-	FSIZE_t fileSize = fno.fsize;
-	if (fileSize < minSize) {
-		display->displayPopup(get(String::STRING_FOR_DX_ERROR_FILE_TOO_SMALL));
+	FatFS::FileInfo fno = D_TRY_CATCH(FatFS::stat(path), error, {
+		return false; //
```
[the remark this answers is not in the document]
stellar-aria: That collapsed the return onto the previous line, and I wanted to keep them separate to make it obvious where control flow could occur

src/deluge/storage/DX7Cartridge.cpp
```diff
@@ -99,74 +108,115 @@ void DX7Cartridge::packProgram(uint8_t* src, int idx, char* name, char* opSwitch
...
-		unpackPgm[op * 21 + 16] = bulk[op * 17 + 14] & 0x7F; // output level
-		char fcoarse_mode = bulk[op * 17 + 15] & 0x3F;       // bits 6-7 don't care per sysex spec
-		unpackPgm[op * 21 + 17] = fcoarse_mode & 1;
-		unpackPgm[op * 21 + 18] = (fcoarse_mode >> 1) & 0x1F;
-		unpackPgm[op * 21 + 19] = bulk[op * 17 + 16] & 0x7F; // fine freq
-		unpackPgm[op * 21 + 20] = (detune_rs >> 3) & 0x7F;
+		// FIXME(@stellar-aria): This looks like it's overwriting the previous for loop??? Is this correct?
+		// See original source here:
+		// https://github.com/asb2m10/dexed/blob/a95b61d5196ac018a57ede34bfb037c886551d63/Source/PluginData.cpp#L126
+		std::copy_n(bulk_op, 11, unpack_op);
```
bfredl, line 176: yeah it is weird. The DX7 manual indeed does say that 99 is the maximum possible value, so I think the `normparm()` variant is the correct one and the `memcpy()` should be removed.

bfredl (approved):

seangoodvibes (approved):

PR #3189, opened 2025-01-01
enable some warnings

nikodemus:
adds:

    -Warray-bounds=1
    -Wnull-dereference
    -Wparentheses
    -Wsequence-point
    -Wswitch

Parentheses catches a few apparent buglets. Switch shows a couple of _possibly_ missing cases, but not obvious.

### DIFF
--- a/src/deluge/model/global_effectable/global_effectable.cpp
+++ b/src/deluge/model/global_effectable/global_effectable.cpp
@@ -490,6 +490,9 @@ int32_t GlobalEffectable::getKnobPosForNonExistentParam(int32_t whichModEncoder,
 			case CompParam::BLEND:
 				current = compressor.getBlend() >> 24;
 				break;
+
+			// explicit fallthrough case
+			case CompParam::LAST:;
 			}
 		}
 	}
@@ -573,7 +576,7 @@ ActionResult GlobalEffectable::modEncoderActionForNonExistentParam(int32_t offse
 				unit = " HZ";
 				break;
 
-			case CompParam::BLEND:
+			case CompParam::BLEND: {
 				if (display->haveOLED()) {
 					popupMsg.append(deluge::l10n::get(deluge::l10n::String::STRING_FOR_BLEND));
 				}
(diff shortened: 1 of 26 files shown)


src/deluge/model/global_effectable/global_effectable.cpp
```diff
@@ -573,7 +576,7 @@ ActionResult GlobalEffectable::modEncoderActionForNonExistentParam(int32_t offse
 				unit = " HZ";
 				break;
 
-			case CompParam::BLEND:
+			case CompParam::BLEND: {
```
stellar-aria, line 579:
```suggestion
			case CompParam::BLEND:
```

src/deluge/model/global_effectable/global_effectable.cpp
```diff
@@ -586,6 +589,10 @@ ActionResult GlobalEffectable::modEncoderActionForNonExistentParam(int32_t offse
 				unit = " %";
 				break;
 			}
```
stellar-aria, line 591:
```suggestion
```

stellar-aria (changes requested):

stellar-aria:
Lets remove these just to match the others, unless they cause problems? I don't _think_ there's a scoping issue with the body of the switch arm, but better to be on the safe side.

src/deluge/model/global_effectable/global_effectable.cpp
nikodemus: @stellar-aria, the

```
q31_t level = current == 64 ? ONE_Q31 : lshiftAndSaturate<24>(current + 64);
```

on line 584 makes this brace-pair required: otherwise the added case becomes a jump over a declaration, which isn't allowed.

nikodemus: The alternative to adding is moving that declaration, or inlining that expression next line where `level` is used.

stellar-aria: Alright, that makes sense

stellar-aria (approved):

PR #3193, opened 2025-01-02
simplify SoundEditor::beginScreen() and updatePadLightsFor() logic

### LINKED ISSUE #3081: Accessing a parameter shortcut in a Clip then using the NAME shortcut will have shortcut pads blinking while keyboard is visible.

Having a parameter shortcut active (eg. SHIFT+LPF FREQUENCY) then pressing SHIFT or AUDITION+NAME to rename the synth will bring up the alpha numeric keyboard as expected.

Meanwhile, whatever parameter shortcut you entered previously, those pads will blink while the keyboard is up. This means that extraneous pads outside of the lit up alpha numeric keyboard will blink, yet serve no purpose, or pads within the alphanumeric keyboard will blink, but just be confusing/distracting.

Instead:
1) Once the keyboard is open, the last param shortcut previously open should be exited to prevent any other pads to blink on the main pad grid. 
2) After submitting a new name or backing out of the naming process, the last param shortcut will remain exited.

Currently, 1) isn't implemented. However, 2) partially implemented in that this is how Kit (Affect Entire disabled) currently behaves.
In Synth, Kit (Affect Entire enabled) and Audio clips, after submitting a name or exiting the renaming prompt, the shortcut entered previously would still be active. So 2) would need to be applied to those cases.

_No response_

OLED

Beta

1.2-rc1

_No response_

nikodemus:
- beginScreen() used to return false & skip part of the setup if the UI was one of a list of specific UIs. Do the full setup always, and return false if the UI isn't SoundEditor itself.

- updatePadLightsFor() had a similar list of exceptions, replace most of that list with a simple check of "is the current UI the SoundEditor"; unlike before clear the existing pad lights to start with anyhow.

- Fixes #3081

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -584,13 +584,19 @@ void SoundEditor::setupShortcutsBlinkFromTable(MenuItem const* const currentItem
 }
 
 void SoundEditor::updatePadLightsFor(MenuItem* currentItem) {
+	// First we clear everything.
+	memset(sourceShortcutBlinkFrequencies, 255, sizeof(sourceShortcutBlinkFrequencies));
+	memset(sourceShortcutBlinkColours, 0, sizeof(sourceShortcutBlinkColours));
+	uiTimerManager.unsetTimer(TimerName::SHORTCUT_BLINK);
 
-	if (!inSettingsMenu() && !inNoteEditor() && currentItem != &sampleStartMenu && currentItem != &sampleEndMenu
-	    && currentItem != &audioClipSampleMarkerEditorMenuStart && currentItem != &audioClipSampleMarkerEditorMenuEnd
-	    && currentItem != &fileSelectorMenu && currentItem != static_cast<void*>(&nameEditMenu)) {
+	// In these cases that's all we need to do.
+	if (inSettingsMenu() || inNoteEditor() || getCurrentUI() != &soundEditor) {
+		return;
+	}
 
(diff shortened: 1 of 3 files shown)


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -584,13 +584,19 @@ void SoundEditor::setupShortcutsBlinkFromTable(MenuItem const* const currentItem
 }
 
 void SoundEditor::updatePadLightsFor(MenuItem* currentItem) {
+	// First we clear everything.
+	memset(sourceShortcutBlinkFrequencies, 255, sizeof(sourceShortcutBlinkFrequencies));
```
stellar-aria, line 588: std::fill here, please
If you want you could also turn `sourceShortcutBlinkFrequencies` into a std::array and then just use `.fill(255)`

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -584,13 +584,19 @@ void SoundEditor::setupShortcutsBlinkFromTable(MenuItem const* const currentItem
...
+	memset(sourceShortcutBlinkColours, 0, sizeof(sourceShortcutBlinkColours));
```
stellar-aria, line 589: same here

stellar-aria (changes requested):

nikodemus:
Can't use array::fill or std::fill because they're 2D arrays, but I think brace-initializers should work.

stellar-aria (approved):

PR #3196, opened 2025-01-02
Refactor StringBuf

stellar-aria:
Changes StringBuf to StackString for better semantics.
Allocate inside StackString to limit the ownership lifetime strictly to the containing object.


stellar-aria:
Yeah I'll need to merge it in and do those two new instances

PR #3199, opened 2025-01-03
change NoteRow status to a boolean

stellar-aria:


src/deluge/model/note/note_row.cpp
```diff
@@ -91,7 +91,7 @@ Error NoteRow::beenCloned(ModelStackWithNoteRow* modelStack, bool shouldFlattenR
 
 	firstOldDrumName = nullptr;
 	ignoreNoteOnsBefore_ = 0;
-	// soundingStatus = STATUS_OFF;
+	// sequenced = false;
```
nikodemus, line 94: How old is this comment-out? Should the line be just deleted?

nikodemus (approved):

src/deluge/model/note/note_row.cpp
stellar-aria: pretty old, but I did the replacement mechanically, not by hand. I could remove it. TBH the beenCloned system is on my shortlist for refactoring anyways, so it's probably not worth it rn.

PR #3204, opened 2025-01-04
StringBuf niceties

stellar-aria:
Pared down from #3196 after having it pointed out to me by @sapphire-arches that alloca would not do what I wanted the way I was using it...

### DIFF
--- a/src/deluge/util/d_string.h
+++ b/src/deluge/util/d_string.h
@@ -110,14 +110,14 @@ private:
 class StringBuf {
 	// Not templated to optimize binary size.
 public:
-	StringBuf(char* buf, size_t capacity) : capacity_(capacity), buf_(buf) { buf_[0] = 0; }
+	StringBuf(char* buf, size_t capacity) : capacity_(capacity), buf_(buf) { memset(buf_, '\0', capacity_); }
 
 	void append(const char* str) { ::strncat(buf_, str, capacity_ - size() - 1); }
 	void append(char c) { ::strncat(buf_, &c, 1); }
 	void clear() { buf_[0] = 0; }
-	void truncate(std::size_t newSize) {
+	void truncate(size_t newSize) {
 		if (newSize < capacity_) {
-			buf_[newSize] = 0;
+			buf_[newSize] = '\0';
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/deluge/util/d_string.h
```diff
@@ -132,13 +132,16 @@ class StringBuf {
 	[[nodiscard]] const char* data() const { return buf_; }
 	[[nodiscard]] const char* c_str() const { return buf_; }
 
-	[[nodiscard]] std::size_t capacity() const { return capacity_; }
-	[[nodiscard]] std::size_t size() const { return ::strlen(buf_); }
+	[[nodiscard]] size_t capacity() const { return capacity_; }
+	[[nodiscard]] size_t size() const { return std::strlen(buf_); }
+	[[nodiscard]] size_t length() const { return std::strlen(buf_); }
```
nikodemus, line 137: Why both? All the std stuff tends to `.size()` and the deluge specials tend to `.getLength()` or `.getNumWhatevs()`, so `.lenght()` seems a bit inconsistent?

Not asking changes, just wondering.

nikodemus (approved):

nikodemus:
> Pared down from #3196 after having it pointed out to me by @sapphire-arches that alloca would not do what I wanted the way I was using it...

I wonder, if we had something like:

```
class StackStringBase {
    virtual char* buf() = 0;
    virtual size_t capacitu() = 0;
    ...
}:

template <int SZ>
class StackString final : public StackStringBase {
protected:
    inline char* buf() final { return buf_; }
    inline size_t capacity() final { return SZ; }
private:
    char buf_[SZ];
};
```

How much extra crap per allocated size would we get? Could we force the size to a power of 2 to keep tha down to something sensible?

src/deluge/util/d_string.h
stellar-aria: it's conformance to the string interface, which has length() because of strlen and size() to act as a standard container

stellar-aria:
in-line definitions are always inline linkage, but we may get some benefit from making them constexpr if possible. I don't think there are constexpr definitions of the c-string stuff, but there are for std::string_view. there may be some benefit to rewriting these as constexpr since then the compiler could reason about and catch out-of-bounds errors at compile time

PR #3209, opened 2025-01-04
[DRAFT] LayeredMenu: allows multiple presses on a single shortcut to cycle

nikodemus:
- replaces the SELECT press changing unison count to unison stereo spread: does the same thing, but works by pressing the same shortcut again.

- the same mechanism should be usable as-is for eg. providing fast access to new arp parameters.

---

the second commit adds envelopes 3 & 4 using this mechanism. these commits are best reviewed separately


ok-reza:
nice ! feels good
im noticing that the shortcut's pad doesn't blink when the 1st or 2nd shortcut is in focus.  i was also thinking that shortcuts with multiple pages will either:
1) blink a different color to tell user there's multiple layers (eg. yellow instead of white).
2) blink white but add another color (eg. yellow) within the sequence of blinking. to leave a breadcrumb that another layer exists, without completely departing from white pad blinks with shortcuts.

not sure how necessary or QoL improving this would be, but maybe entering the 2nd shortcut should occur on release instead of on press.
it's honestly just a bad habit/unnecessary muscle memory of mine, but if i'm already on say LPF freq shortcut, sometimes i still press and hold shift+lpf (despite already being in focus) then add env1 to ensure i'm modulating lpf with env1. it's not necessary as all you need to do is hit shift+env1 to make the modulation, but i guess my logic is that it feels ike im making a connection between two shortcuts in one action.

nikodemus:
> not sure how necessary or QoL improving this would be, but maybe entering the 2nd shortcut should occur on release instead of on press. it's honestly just a bad habit/unnecessary muscle memory of mine, but if i'm already on say LPF freq shortcut, sometimes i still press and hold shift+lpf (despite already being in focus) then add env1 to ensure i'm modulating lpf with env1. it's not necessary as all you need to do is hit shift+env1 to make the modulation, but i guess my logic is that it feels ike im making a connection between two shortcuts in one action.

Hm. relase vs press would be quite tricky right now, esp. if we don't want to enter normal shortcuts on release.

Another option would be to eg. make TRIPLET VIEW change the layer when in shortcut.

nikodemus:
As, forgot to fix the blink...

seangoodvibes:
Need to add some docs for those extra envelopes!

nikodemus:
Missing bits in addition to the blink missing:
- env 3 & 4 aren't available as patch source shortcuts yet
- didn't test how the regular menu access to them works

Additionally, the layering might be confusing on 7-seg. Maybe making it so that holding shift in menu shows the item name on 7-seg (or need to hold shift to see the value & edit?) That could be combined with allowing horiz menu - like access to neighbouring params when the name is shown? @m-m-adams and other 7-seg users?

m-m-adams (approved):

sapphire-arches:
I've added documentation and fixed the merge conflicts, but there's a bug on 7seg (at least in emulation) -- it doesn't appear to be possible to switch between envelope layers on 7seg.

sapphire-arches:
also pad flashing for selected parameters is broken :cry:

seangoodvibes:
Closing as this is being implemented by another dev :)!

PR #3218, opened 2025-01-06
Fix rename-UIs PR

stellar-aria:


m-m-adams (approved):
I haven't reviewed in a ton of detail but lgtm and fixes the freezes!

PR #3227, opened 2025-01-06
Fix: When using the backspace in the keyboard when searching samples, it doesn't re-search

### LINKED ISSUE #2931: When using the backspace in the keyboard when searching samples, it doesn't re-search

When you start typing, using the grid pads, a name to search for a sample within a folder when the sample browser is open, if you type something that produces zero results, then even if you use the backspace key to delete a character, it doesn't research with the new text

When you type KIC and starts showing the samples that start with KIC... (such as all the kicks) if you type KICC it shows no result, but if you press the backspace to go back to KIC, it show show again the same samples that when you initially searched

_No response_

OLED

Nightly

_No response_

_No response_

mweigt:
First commit fixes the bug - search is updated on backspace.

But I did not like the behavior when deleting the last character, namely showing an empty prompt. IMO when no characters are to be matched, all files should be shown.

Second commit re-activates logic that was already in the code (comment did not say why it was disabled, but maybe watch out for bugs in this context), now:
- backspacing correctly re-searches
- typing characters that lead to zero matches is ignored (i.e. starting with KIC, if there are no files starting with KICC, typing "C" is ignored)
- when deleting the last character, all files are shown, as makes sense 

Fixes #2931


stellar-aria:
Hey! Thanks for fixing this. Instead of just removing the `false` in your second commit, can you swap it out for `display->haveOLED()`? I think the false was from when Rohan was making the 7SEG to OLED transition due to how confusing it can be having stuff still appearing with the tiny four-character display flashing.

src/deluge/gui/ui/browser/browser.cpp
```diff
@@ -1220,8 +1220,7 @@ bool Browser::predictExtendedText() {
 
 		// Otherwise if we already tried that, then our whole search is fruitless.
 notFound:
-		if (false && !mayDefaultToBrandNewNameOnEntry) { // Disabled - now you're again always allowed to type
-			                                             // characters even if no such file exists.
+		if (!mayDefaultToBrandNewNameOnEntry) {
```
stellar-aria, line 1223:
```suggestion
		if (display->haveOLED() && !mayDefaultToBrandNewNameOnEntry) {
```

stellar-aria (changes requested):

m-m-adams (approved):

stellar-aria (approved):

PR #3238, opened 2025-01-08
Horizontal menus for arpeggiator's randomizer submenu

soymonitus:
- First, I modified the horizontal menus class to prefer the 7SEG strings instead of the OLED strings for the menu titles. This will increase the probability of menus to look better by default.
- Second, I updated the arpeggiator docs as they had some typos and errors
- Third, I moved the Ratchet Amount and Chord Polyphony inside the Randomizer menu so they are close to their related parameters Ratchet Probability and Chord Probability
- Fourth, I set the Randomizer submenu as horizontal menu so dialing probabilities is much quicker

Next step (maybe in a different PR): once this  PR https://github.com/SynthstromAudible/DelugeFirmware/pull/3209 is merged I will add the Randomizer submenu as a layered shortcut behind any of the arp params (likely to be "Octaves" or "Gate")


soymonitus:
Ready for review!

src/deluge/gui/menu_item/menu_item.h
```diff
@@ -243,8 +243,10 @@ class MenuItem {
 	[[nodiscard]] virtual std::string_view getName() const { return deluge::l10n::getView(name); }
 	/// @brief Get the name for use on horizontal menus.
 	///
-	/// By default this redirects to getName(), but can be overriden.
-	virtual void getColumnLabel(StringBuf& label) { label.append(getName().data()); }
+	/// By default this redirects to 7SEG's name, but can be overriden.
+	virtual void getColumnLabel(StringBuf& label) {
+		label.append(deluge::l10n::getView(deluge::l10n::built_in::seven_segment, name).data());
+	}
```
sapphire-arches, line 249: Can this be done in a separate change? The rest of this change is non controversial and can be merged as-is, but using 7seg strings can make some things more opaque.

soymonitus: The horizontal menus for arp settings depend on this change. If I don't use 7Seg labels then i would have two menu items with the same name: RATCHet amount and RATCHet probability, same for CHORd polyphony and CHORd probability. WHen you use horizontal menus you actually only have 4 characters of space for each menu item so I think using the 7seg labels will give much more adequate labels for those menus. The current horizontal menus don't have this problem as the labels are the same in OLED and 7SEG, that is, LPF, DRIVE, MODE, RESO, and ADSR

soymonitus: The other alternative would be to create custom menu items for the arp params which use a custom label instead of the normal OLed label. But don't you think we will end up with lots of custom implementations when converting most menus to horizontal menus? While if we make the 7seg label the default, it will be more probable than the label fits as is without custom implementations

soymonitus: @nikodemus what you think about this?

soymonitus: @sapphire-arches I made some commits to revert the default column label to original state, and then added overrides for the arpeggiator params so they use the 7Seg version. I also added the horizontal menus to the grid

soymonitus:
@sapphire-arches I got all the menus working now with custom implementations for the labels. Also, fixed the horizontal menus so they allow to fill the entire width for all pages. Rhythm parameter now shows the encoded rhythm as in vertical mode.
Ready for review

sapphire-arches (approved):

PR #3240, opened 2025-01-08
Scale-Aware-Accidentals

todd-gochenour:
1. Move noteCodeToString() from functions.cpp to preset_scales.cpp because of new dependency on Scale.
2. Move codeIsSharp[] and noteCodetoLetter[] from lookuptables.cpp to preset_scales.cpp and renamed codeIsAltered[] and noteLetter[].  Public usage of
these tables in key_range and keyboard_screen still need review as they are still displaying sharps.
3. Add getAccidental() to preset_scales.cpp. This function determines if the relative major key of the current song is sharp or flat.
4. Modify song.cpp and instrument_clip_minder.cpp to send the key and scale from the current song to noteCodeToString() so it can determine which accidental to use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -245,6 +245,7 @@ also affect normal sequenced notes while arpeggiator is Off.
 - A white playhead is now rendered in Song Grid and Performance Views that let's you know when a clip or section launch event is scheduled to occur. The playhead only renders the last 16 notes before a launch event.
   - Note: this playhead can be turned off in the Community Features submenu titled: `Enable Launch Event Playhead (PLAY)`
 - The display now shows the number of Bars (or Quarter Notes for the last bar) remaining until a clip or section launch event in all Song views (Grid, Row, Performance).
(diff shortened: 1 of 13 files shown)


todd-gochenour:
Test Script:

1) On startup with synth Rich Saw Bass and a default key in C and scale is Major.  Note that the root note is 0 with a new song and not 60 (C3).   C3 and C4 are top and bottom audition pads lit to show they are the root.
2) Press SCALE quickly and now the notes are chromatic.  Press the note above C3 and it will read D&3.  '&' is used because there isn't a flat character available.
3) Press SCALE quickly to return to diatonic C major.  Step up to the 3rd pad above C3, the F note, and press SCALE-PAD to change key to F Major.
4) Step up to the 3rd pad above F, notice that it reads B&3, and press SCALE-PAD to change to the key of Bb.  Step up to the 3rd above that and SCALE-PAD the key of Eb.
5) Step down one pad from Eb and SCALE-PAD the key of D.  Step up to the 3rd pad above D, notice that it reads F#3, and press SCALE-PAD to change the key to F#.  The key of F# has the most accidentals of any key at 6. Step down one pad and it shows F3 instead of E#3, enharmonically correct but not technically accurate.
6) Step down 2 steps down from F# to D# and SCALE-PAD to change the key to Eb. Key signatures will always favor the the fewest accidentals, so Eb with three flats is favored over D# with 9 sharps.

todd-gochenour:
Closing Github pulled in code from my previous PR.

PR #3242, opened 2025-01-08
Scale-Aware-Accidentals

todd-gochenour:
preset_scales.cpp because of new dependency on Scale.
2. Move codeIsSharp[] and noteCodetoLetter[] from
lookuptables.cpp to preset_scales.cpp and renamed
codeIsAltered[] and noteLetter[].  Public usage of
these tables in key_range and keyboard_screen
still need review as they are still displaying sharps.
3. Add getAccidental() to preset_scales.cpp. This
function determines if the relative major key of
the current song is sharp or flat.
4. Modify song.cpp and instrument_clip_minder.cpp
to send the key and scale from the current song
to noteCodeToString() so it can determine which
accidental to use.


todd-gochenour:
I don't understand.

Issue #3168, opened 2024-12-28
In the velocity drum keyboard view, rapid pressing of many pads can cause notes to get stuck, and temporarily lock you into that view


Metamere:
### Please describe the problem:

pressing many keys rapidly causes things to lag badly, and during this, something goes wrong and one or more of the pressed notes (I have seen seven) can get stuck. While stuck, this prevents exiting the clip or the keyboard screen until they are manually unstuck by pressing the pads on the stuck drum pad until the specific one that is stuck is found. If you change the zoom level or scroll while notes are stuck, different sounds will trigger on each screen update.

### What is the expected behavior?

When you remove all fingers from the pads, they should be signaled as released in the code, and if the samples are set to cut, they should stop playing. With all notes released, you should be free to move to different views, and updating the zoom level or scroll offset should not trigger notes to play.

### When did you start noticing the issue?

This week while working on an update to the velocity drum keyboard view. I detected the issue in 1.3 nightly and confirmed it was present in 1.2.

### Is there a relevant Pull request?

possibly the one I am working on #3132, as it at least reduces the symptoms of the issue and makes them easier to deal with, but it will likely take others to fix the root issue.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Release 1.2/Chopin

### What is the firmware name:

1.2.0-DEV-3EACBC3B

### If possible provide the steps to reproduce the issue and upload additional media:

Open a kit with at least a couple of dozen samples, press the keyboard button, then press shift and turn horizontal encoder CCW until pad size is set to 2 x 2, in order to make pressing many different samples' pads easier, and then use all fingers to rapidly press many different pads. Observe that some pads stay lit, and that exiting the view is not possible.

Metamere:
I just tested it with mashing on a midi keyboard rather than the screen, and it didn't have any issues. No lag, no stuck notes. Since that is the case, then I think the issue is likely coming from also having to update the screen so many times. It tries to update the screen each time a pad is pressed or released. I think we should eliminate unnecessary updates, and also limit how quickly we can update that view. One unnecessary update happens in the following example: if you press a drum pad on two of the pads while holding down the first, then release one finger, then the other. it will update the screen four times, even though nothing will change in the middle two updates, since the pad stays highlighted from after the first press to until the last finger is removed from it.

Metamere:
[...]

PR #3243, opened 2025-01-08
Scale-Aware-Accidentals

todd-gochenour:
- Key scales can display either sharps or flats. The major keys of G, D, A, E, B and F# all display sharps and C, F, Db, Ab, Eb and Bb all display flats.  Scale modes and exotic minor scales will use the accidental of its relative major key.  Accidentals only apply to black keys on a piano.  Db, Eb, F#, Ab and Bb are the five black keys with their accidentals.  In the key of E the third degree will display G#.  In the key of Eb the fourth degree will display Ab.  The are enharmonically equivalent.
- With the 7SEG display, where before a sharp was represented with the decimal point now a sharp is represented using the upper left vertical segment and a flat is represented using the lower left vertical segment.
- A refactor of the function noteCodeToString() will consolidate logic and lookup tables into preset_scales.cpp where it can utilize scale information to determine if the relative major key of the current song is sharp or flat.  The function signature takes parameters currentSong->key.rootNote and currentSong->getCurrentScale() when making this decision.
- Areas affected by this change include InstrumentClipMinder, AutomationView, ChordLibrary, ChordKeyboard, KeyboardScreen, MultiRange and KeyRange.
- Notes are shown as a popup when pressing the audition pad of a synth instrument clip and in the KeyboardScreen when pressing any note pad.   Press a pad within the ChordKeyboard and ChordLibrary views and the chord name will popup with the appropriate accidental based on key.   Notes are shown in the OLED screen layout when scrolling vertically in AutomationView>Velocity and when scrolling horizontally defining a default key range in Settings>Defaults>Key.  Create a multisample synth and the SYNTH-BROWSE shortcut will show a Note Range list for each sample loaded.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.9.2
+    rev: v0.9.4
     hooks:
       # Run the linter.
       - id: ruff
--- a/contrib/debug/heap_state.py
+++ b/contrib/debug/heap_state.py
@@ -47,7 +47,7 @@ try:
         GMA_START_OFFSET = gdb.lookup_global_symbol(
             "GeneralMemoryAllocator::get()::generalMemoryAllocator"
         ).address.cast(TY_UINT32)
-    except:
+    except (gdb.error, gdb.GdbError):
         print("falling back to string parsing")
         GMA_START_OFFSET = int(
             gdb.execute(
@@ -59,7 +59,7 @@ try:
             16,
         )
 
-except ModuleNotFoundError as e:
+except ModuleNotFoundError:
     gdb = None
 
     class Field:
@@ -295,7 +295,9 @@ def walk_linked_list_from_node(inferior, offset, end):
 
     while offset != end:
         try:
-            memory = inferior.read_memory(offset, TY_BidirectionalLinkedListNode.sizeof)
+            memory = inferior.read_memory(  # noqa: F841
+                offset, TY_BidirectionalLinkedListNode.sizeof
+            )  # noqa: F841
             offset = bytes_as_uint32(inferior.read_memory(offset + next_offset, 4))
             list_addr = bytes_as_uint32(inferior.read_memory(offset + list_offset, 4))
             print(f"Stealable node at: {offset:08x} for list {list_addr:08x}")
@@ -306,7 +308,7 @@ def walk_linked_list_from_node(inferior, offset, end):
 
 
 def parse_cache_manager(inferior, offset):
-    memory = inferior.read_memory(offset, TY_CacheManager.sizeof)
+    memory = inferior.read_memory(offset, TY_CacheManager.sizeof)  # noqa: F841
     # 10 is NUM_STEALABLE_QUEUES
     for i in range(10):
         print("---- Stealables for stealable queue", i)
@@ -340,13 +342,13 @@ def parse_memory_region(inferior, offset, mem_region_idx):
     num_elements = bytes_as_uint32(
         inferior.read_memory(empty_offset + num_elements_offset, 4)
     )
-    element_size = bytes_as_uint32(
+    element_size = bytes_as_uint32(  # noqa: F841
         inferior.read_memory(empty_offset + element_size_offset, 4)
     )
     words_in_key = bytes_as_uint32(
         inferior.read_memory(empty_offset + words_in_key_offset, 4)
     )
-    memory = inferior.read_memory(
+    memory = inferior.read_memory(  # noqa: F841
         empty_offset, TY_OrderedResizeableArrayWithMultiWordKey.sizeof
     )
 
@@ -456,8 +458,6 @@ def parse_args():
 
 
 def main():
-    import sys
-
     parser = parse_args()
     with open(parser.memory_dump, "rb") as inf:
         inf = FakeInferior(inf.read(), 0x20020000)
--- a/contrib/doc/pinmap/generate.py
+++ b/contrib/doc/pinmap/generate.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import io
 import xml.etree.ElementTree as ET
 
 
@@ -23,20 +22,20 @@ class Pin:
     def render_as_output(self, g, clazz):
         path = ET.SubElement(g, "path")
         if self.facing_left:
-            p = f"M 0,0"
+            p = "M 0,0"
             p += f"H {self.width} "
             p += f"V {UNIT_HEIGHT} "
-            p += f"H 0 "
+            p += "H 0 "
             p += f"l -{HALF_HEIGHT},-{HALF_HEIGHT} "
             p += "Z"
             self.tip_x = self.left - HALF_HEIGHT
         else:
-            p = f"M 0,0 "
+            p = "M 0,0 "
             p += f"H {self.width} "
             p += f"l {HALF_HEIGHT},{HALF_HEIGHT} "
             p += f"l -{HALF_HEIGHT},{HALF_HEIGHT} "
-            p += f"H 0 "
-            p += f"Z"
+            p += "H 0 "
+            p += "Z"
             self.tip_x = self.left + self.width + HALF_HEIGHT
         path.attrib["d"] = p
         self.tip_y = self.top + HALF_HEIGHT
@@ -70,14 +69,14 @@ class Pin:
             d += "H 0 "
             d += f"A 5 5, 0, 0, 0, 0, {UNIT_HEIGHT}"
             d += f"H {self.width} "
-            d += f"Z"
+            d += "Z"
             self.tip_x = self.left - HALF_HEIGHT
         else:
-            d = f"M 0,0 "
+            d = "M 0,0 "
             d += f"H {self.width} "
             d += f"a 5 5, 0, 0, 1, 0, {UNIT_HEIGHT}"
-            d += f"H 0 "
-            d += f"Z"
+            d += "H 0 "
+            d += "Z"
             self.tip_x = self.left + self.width + HALF_HEIGHT
         path.attrib["d"] = d
         self.tip_y = self.top + HALF_HEIGHT
--- a/scripts/tasks/task-build.py
+++ b/scripts/tasks/task-build.py
@@ -2,7 +2,6 @@
 import argparse
 import importlib
 import subprocess
-import sys
 from typing import Sequence
 import util
 import os
--- a/scripts/tasks/task-clean.py
+++ b/scripts/tasks/task-clean.py
@@ -1,7 +1,6 @@
 #! /usr/bin/env python3
 import argparse
 import subprocess
-import sys
 import util
 import os
 
--- a/scripts/tasks/task-configure.py
+++ b/scripts/tasks/task-configure.py
@@ -2,7 +2,6 @@
 import argparse
 import importlib
 import subprocess
-import sys
 from typing import Sequence
 import util
 import os
@@ -18,7 +17,9 @@ class CondensedChoiceFormatter(argparse.ArgumentDefaultsHelpFormatter):
 
 
 def argparser() -> argparse.ArgumentParser:
-    fmt = lambda prog: CondensedChoiceFormatter(prog)
+    def fmt(prog):
+        return CondensedChoiceFormatter(prog)
+
     parser = argparse.ArgumentParser(
         prog="configure",
         description="Configure a CMake build (automatically called by build)",
--- a/scripts/tasks/task-docs.py
+++ b/scripts/tasks/task-docs.py
@@ -29,7 +29,7 @@ def main(argv: Sequence[str] = sys.argv) -> int:
 
     project_root = util.get_git_root()
     build_dir = project_root.absolute() / "build"
-    source_dir = project_root.absolute()
+    # source_dir = project_root.absolute()
     index_page = build_dir / "html/index.html"
 
     if not os.path.exists("build"):
--- a/scripts/tasks/task-license.py
+++ b/scripts/tasks/task-license.py
@@ -2,7 +2,6 @@
 
 # Copyright 2023 Kate Whitlock
 import argparse
-import os
 from pathlib import Path
 import util
 from functools import partial
--- a/scripts/tasks/task-loadfw.py
+++ b/scripts/tasks/task-loadfw.py
@@ -161,7 +161,7 @@ def try_read_hex_key():
     try:
         with open(name, "r") as f:
             return f.read().strip()
-    except:
+    except FileNotFoundError:
         hex_key = input(
             "Enter hex key from COMMUNITY FEATURES > ALLOW INSECURE DEVELOP SYSEX MESSAGES: "
         )
--- a/scripts/tasks/task-midi-clock.py
+++ b/scripts/tasks/task-midi-clock.py
@@ -2,7 +2,6 @@
 import sys
 import time
 import argparse
-import sys
 import util
 
 
--- a/scripts/tasks/task-nuke.py
+++ b/scripts/tasks/task-nuke.py
@@ -1,6 +1,5 @@
 #! /usr/bin/env python3
 import argparse
-from pathlib import Path
 import shutil
 import util
 import os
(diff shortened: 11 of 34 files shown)


todd-gochenour:
Test Script:

1) On startup with synth Rich Saw Bass and a default key in C and scale is Major.  Note that the root note is 0 with a new song and not 60 (C3).   C3 and C4 are top and bottom audition pads lit to show they are the root.
2) Press SCALE quickly and now the notes are chromatic.  Press the note above C3 and it will read D&3.  '&' is used because there isn't a flat character available.
3) Press SCALE quickly to return to diatonic C major.  Step up to the 3rd pad above C3, the F note, and press SCALE-PAD to change key to F Major.
4) Step up to the 3rd pad above F, notice that it reads B&3, and press SCALE-PAD to change to the key of Bb.  Step up to the 3rd above that and SCALE-PAD the key of Eb.
5) Step down one pad from Eb and SCALE-PAD the key of D.  Step up to the 3rd pad above D, notice that it reads F#3, and press SCALE-PAD to change the key to F#.  The key of F# has the most accidentals of any key at 6. Step down one pad and it shows F3 instead of E#3, enharmonically correct but not technically accurate.
6) Step down 2 steps down from F# to D# and SCALE-PAD to change the key to Eb. Key signatures will always favor the the fewest accidentals, so Eb with three flats is favored over D# with 9 sharps.

mweigt:
> note above C3 and it will read D&3. '&' is used because there isn't a flat character available.

Why not use "b", since that is very much what the flat sign looks like, e.g.  "Db3"?

todd-gochenour:
Test runthrough video: https://www.youtube.com/watch?v=fTYfCT8OS74

todd-gochenour:
There are no lower case letters on the Deluge.  Only upper case.  '&' is the closest shape and is temporary until a new glyph is generated.

m-m-adams:
Oh man this has been one of my most wanted deluge features forever. Does it change the song scale or is the accidental ignored for scale determination?

todd-gochenour:
It was a surprise to me when I first picked D# (from key of E) as the new key and it popped up 'E& Major' for a second and then every altered note (black key) was displayed in flats.   I picked A# (from the key of B) for the new key and it popped up 'B& Major'.   I got this behavior for free.   

Today I'm testing mode and alternate scales.  I just found a defect when displaying the 4th degree of C Lydian which is displaying note G instead of F#.

todd-gochenour:
Isn't it a convention for parameter changes to first show the current value before changing to the next?   SHIFT-SCALE is not doing this.

seangoodvibes:
> Isn't it a convention for parameter changes to first show the current value before changing to the next?   SHIFT-SCALE is not doing this.

It depends. It's applied on some things but not all things (eg note probability, iteration shows current value before changing it).

todd-gochenour:
Can I apply it to SHIFT-SCALE ?

seangoodvibes:
> Can I apply it to SHIFT-SCALE ?

Sure, we can see how it feels

todd-gochenour:
Tested C Lydian for augmented 4th F# in relative major of G.
Tested C Mixolydian for 7 degree of B& in relative major of F.
Tested C Locrian for everything flat except C and F in relative major of D&.
Tested C Melodic Minor for flat 3rd of E& treated like a dorian minor
Tested C Harmonic Minor for flat 3rd, flat 6th of A& and natural 7th of B.
Test Failed C Hungarian Minor for flat 3rd, sharp 4 of G& (should be F#), flat 6th and natural 7th.  
*A case where a scale has both flats and sharps. *
Test Failed C Marva treated as a major with flat 2, G& (should be F#).  Ditto.
Tested C Arabian treated as a major with flat 5, flat 6 and flat 7
Tested C Whole Tone treated as a major with flat 5, flat 6 and flat 7
Tested C Blues with E&, G& and B&
Tested C Pentatonic Minor with E& and B&.
Tested C Hirajosh with E& and A&.

Not sure yet how to handle the mixture of sharps and flats in the Hungarian Minor and Marva scales.  There are also the cases E#, F&, B# and C& which show the natural enharmonic note because accidentals in this code only apply to the black keys of a piano.

todd-gochenour:
Tested D& Major with B, C#, D#, E, F#, G#, A#
Tested D& Minor B, C#, D, E, F#, G, A
Test Failed B Lydian for augmented 4th F (should be E#).
Tested D& Mixolydian for 7 degree of A in relative major of E.
Tested D& Locrian with no sharps or flats.
Tested D& Melodic Minor B, C#, D, E, F#, G#, A#
Tested D& Harmonic Minor B, C#, D, E, F#, G, A#
Test Failed B Hungarian Minor B, C#, D, F (Should be E#), F#, G, A#
Tested D& Marva B, C, D#, F, F#, G#, A#
Tested D& Arabian B, C#, D#, E, F, G, A
Tested D& Whole Tone B, C#, D#, F, G, A
Tested D& Blues B, D, E, F, F#, A (F is added flat 5 blue note)
Tested D& Pentatonic Minor B, D, E, F#, A
Tested D& Hirajosh B, C#, D, F#, G

The Deluge strategy did not handle displaying E# before this change either. 

Tested D& Major with D&, E&, F, G&, A&, B&, C
Tested D& Minor Became C# minor with C#, D#, E, F#, G#, A, B with relative major of F#. 
Tested D& Dorian Became C# minor with relative major of B
Tested D& Lydian D&, E&, F, G, A&, B&, C with relative major of A&.
Tested D& Mixolydian became C# minor D#, D#, F, F#, G#, A#, B
Tested D& Locrian C#, D, E, F#, G, A, B
Test failed D& Melodic Minor C#, D#, E, F#, G#, A#, C (should be B#)
Test failed D& Harmonic Minor C#, D#, E, F#, G#, A, C (should be B#)
Test failed D& Hungarian Minor C#, D#, E, G(as augmented 4 is F##?), G#, A, C (should be B#)
Tested D& Marva D&, D (maybe E&&?), F, G, A&, B&, C
Tested D& Arabian D&, E&, F, G&, G, A, B
Tested D& Whole Tone D&, E&, F, G, A, B
Tested D& Blues C#, E, F#, G, G#, B 
Tested D& Pentatonic Minor C#, E, F#, G#, B
Tested D& Hirajosh C#, D#, E, G#, A

The Deluge strategy did not handle displaying B# before this change either. There are cases where double accidentals like F## or E&& may be technically correct, but the enharmonic equivalent of G and D are easier to comprehend.  There will be cases where a scale contains two of the same note letter, one with an accidental and one without.

todd-gochenour:
Added logic so that SHIFT-SCALE shows current scale on first press before cycling scales on subsequent presses.  This logic applied to instrument clip view and to keyboard screen.  Release of the SHIFT key resets the test for first press.

seangoodvibes:
> Added logic so that SHIFT-SCALE shows current scale on first press before cycling scales on subsequent presses.  This logic applied to instrument clip view and to keyboard screen.  Release of the SHIFT key resets the test for first press.

Instead of creating a new function "shift has changed" I think you should leverage the logic that is used for other "previews" which basically shows a pop up of a "type" 

And then before changing scale you check if the popup of that type is present. If yes, change scale.

Look for the code "display->hasPopupOfType"

For examples of this behaviour

todd-gochenour:
Pull Request Build Release is failing because intToString() is undefined in preset_scales.cpp yet I checked this file in this branch and it does possess #include "util/d_string.h" at the top of the file.  I'm not sure how to resolve this.

Pull Request Build Debug and Run tests both fail because something is wrong with the toolchain.

Run docker run --rm \
fatal: No url found for submodule path 'python-rtmidi' in .gitmodules
Error: Process completed with exit code 128.

I see this same error reported now when I run loadfw and sysex-logging even though these tools are now working properly.

todd-gochenour:
Look for the code "display->hasPopupOfType"

I did look at this first thing.  There is no persistent 'popup' used in the cycling of scales and there is no select encoder involved in changing the value.  With SHIFT-SCALE the action is immediate.  The release of the SHIFT key resets the test.  This was the best option based on the current implementation.

todd-gochenour:
Changes to SHIFT-SCALE have been reverted.

sapphire-arches:
> Pull Request Build Release is failing because intToString() is undefined in preset_scales.cpp yet I checked this file in this branch and it does possess #include "util/d_string.h" at the top of the file.  I'm not sure how to resolve this.

 That only declares the symbol; the definition is in `cfunctions.c` which is not included in the test build. You'll want to add it to `tests/unit/function_mocks.h`

todd-gochenour:
I think I understand that the final Run Tests issue is in building/linking the unit test.  Preset_scales.cpp now has the function noteCodeToString() which uses intToString().  This function was originally in functions.cpp.  Is the build of the unit test not including cfunctions.c and that's why the linker is failing?

todd-gochenour:
I see the cMakeLists.txt defining deluge_SOURCES and functions.cpp is not in the list.  lookuptables.cpp is in the list.  I pulled code from both into preset_scales.cpp, which is in the list.   So do I just add cfunctions.c to this list to get this to work?

todd-gochenour:
Testing chord keyboards that is currently displaying all sharps unaware of the current key and scale.  Looking now to change this.

todd-gochenour:
Testing keyboards.  
1) Key of E& (with A& and B&).  Flats are visible in Isomorphic, In-Key, piano, chord and chord library.  Once we get the glpyh for 'flat' we can replace some chord names that say 'flat5flat9' to '&5&9'.
2) Key of D ( with F# and C#). Sharps are visible in all keyboard views.

todd-gochenour:
I am fixing noteCodeToString() to detect OLED and displaying '.' for the accidental when its 7SEG

todd-gochenour:
Tested KeyboardScreen::selectEncoderAction to change the key.   Scale is pressed and held while encoder is rotated.  This behavior doesn't occur when not in keyboard mode.  This behavior is also not a popup and could be a candidate for refactor, or this same behavior can be used in instrument clip.

The encoder

todd-gochenour:
Testing AutomationView::renderAutomationEditorDisplayOLED() and AutomationView::renderNoteEditorDisplayOLED()

I was unable to see where the noteName was displayed.  Pressing note and SELECT encoder together brought up Note Editor but I don't see where the note name is displayed.  

I reverted the change to use current key and scale to determine accidentals.  These views will always show sharps.

seangoodvibes:
> Testing AutomationView::renderAutomationEditorDisplayOLED() and AutomationView::renderNoteEditorDisplayOLED()
> 
> 
> 
> I was unable to see where the noteName was displayed.  Pressing note and SELECT encoder together brought up Note Editor but I don't see where the note name is displayed.  
> 
> 
> 
> I reverted the change to use current key and scale to determine accidentals.  These views will always show sharps.

Note editor display is referring to the velocity view display

If you're in a synth / midi / cv clip, when in velocity view the selected note corresponding to the selected "note row" should be shown on the display

todd-gochenour:
When I searched for Note Editor, I found that PAD-SELECT brought up the note editor menu and ROW-SELECT brought up the row editor.  These were new functions I had not seen before.   I also interacted with the horizontal menu layout yesterday for the first time. and I like it.   Perhaps the note editor should be refactored to use horizontal format.  Out of scope.

todd-gochenour:
and AutomationView::renderAutomationEditorDisplayOLED()  ?  There are two references in that class.

seangoodvibes:
> and AutomationView::renderAutomationEditorDisplayOLED()  ?  There are two references in that class.
> 
> 
> 
> and  MultiRange::getText()?   Where is that?

AutomationView::renderAutomationEditorDisplayOLED() is for rendering the display for regular parameter automation (eg lpf frequency)

I don't know what multi range get text is

todd-gochenour:
renderAutomationEditorDisplayOLED() doesn't print a note.  My bad.  The second reference is renderNoteEditorDisplay7SEG().

Open question about how to distinquish flat from sharp on 7SEG.  Perhaps an 'b' for flat and '.' for sharp?

seangoodvibes:
> renderAutomationEditorDisplayOLED() doesn't print a note.  My bad.  The second reference is renderNoteEditorDisplay7SEG().
> 
> 
> 
> Open question about how to distinquish flat from sharp on 7SEG.  Perhaps an '_' for flat and '.' for sharp?
> 
> 
> 
> Open question about how to create the flat glyph, also.

To create new fonts, check fonts.c

With that file open if you do a control + f and search for "1" and then turn your head sideways you'll see how the fonts are created.

Re 7seg im not the best person for that, id ask @m-m-adams what they think as they have a 7seg

image

todd-gochenour:
In implementing the flat glyph, I first tried adding 'b' one step after '~' in the font description table and wrote special handling in Canvas::getCharIndex() to deal with this character placed out of ASCII order.   Unfortunately, there are synth names that use lowercase letters like 'Double' and these b characters aren't becoming uppercase.

I have to use some ascii character to represent flat.   Tilde, backtick, dollar sign are all options.  I'm already using '&' as it looks the closest to 'b' of all the various options.  I didn't find the use of '&' in any file names.  The simple solution is to change the '&' glyph to look like a 'b'.

'#' and '&' represent sharp and flat.

sapphire-arches:
You could add it in the "high bit" ASCII range (i.e. chars >127) -- these should never appear in filenames that we support today. You'd need to add logic in `Canvas::getCharIndex` to handle it.

If you want to test 7seg behavior on an OLED, you can press SHIFT+LEARN+AFFECT ENTIRE with 7seg emulation mode set to Toggle (see community_features.md, search for "Emulated Display").

todd-gochenour:
Flat glyph now represented by uint8_t value of 129.   Glyph added after tilde and index will convert char 129 to ('~' + 1).  The chord names restored to 'FLAT5' and 'FLAT9'.  Logic in KeyboardLayoutChord::drawChordName() will detect 'FLAT' in chord name and convert it to char 129 before display.  Chord name will read Ab-7b5b9.

todd-gochenour:
7SEG Displays sharp as upper left vertical segment (segment 2) and flat as lower left vertical segment (segment 3).  They appear as ticks next to the note name.  The decimal dot is no longer used to signify an accidental.

jkschwartz:
Do we have any performance testing tools to see if adding this calculation to each key press has a performance impact? 1.2 already has some performance issues on the isometric keyboard view: https://github.com/SynthstromAudible/DelugeFirmware/issues/3168

sapphire-arches:
> Do we have any performance testing tools to see if adding this calculation to each key press has a performance impact? 1.2 already has some performance issues on the isometric keyboard view: #3168

The calculations added here are near trivial, I doubt they will have performance impact. We don't have great tools for that though, if you have ideas please join us on Discord or leave something in the discussions tab.

#3168 doesn't look like a performance bug, it seems like something is going wrong in the keyboard state reconciliation logic.

todd-gochenour:
* Tested MultiRange menu option.  Procedure is:  New Song, New Synth, SHIFT-BROWSE, locate folder with multisamples, press SELECT and hold to get LOAD FILES(S) and press SELECT again to choose MULTISAMPLES.  Press SHIFT-BROWSE again and MultiRange menu appears.  Db, Eb, F#, Ab and Bb appear in ranges.
* Removed &isNatural parameter from noteCodeToString() as it is obsolete as 7SEG doesn't showDot for sharps anymore, and MultiRange will instead calculate left/right string length using strlen().
* When rootNote is omitted from call noteCodeToString(), it will use noteCode as the root.  This allows simple calls to noteCodeToString(note, buffer)
* Adjusted KeyRange highlight right side of range (C-F#) to line up correctly with note display.  The dash is half the width of a digit.

todd-gochenour:
My development and testing are complete.  This PR is ready for review.

jkschwartz:
I started testing this change tonight. So far it is working great both with new and old songs. One item / observation- some scales are not possible with this change, right? If someone wanted put something together in D#, they would just use Eb?

docs/community_features.md
```diff
@@ -1036,6 +1036,11 @@ to each individual note onset. ([#1978])
     - `DEFAULTS > SCALE > ACTIVE SCALES` sets the active scales for new songs. When `RANDOM` is set as
       `DEFAULTS > SCALE > INIT SCALE`, the random scale is selected from default active scales.
 
+-  ([#3243]) Scale Aware Accidentals
```
seangoodvibes, line 1039: you'll need to add a link to this PR at the bottom of the community features document as well in order for the #3243 reference to work.

todd-gochenour:
> If someone wanted put something together in D#, they would just use Eb?

Yes, as both keys are enharmonically equivalent.   Eb has 3 flats, Bb, Eb and Ab.  D# has 9 sharps, and as there are only 7 notes in a scale, 2 of these sharps are double sharps, D#, E#, F##, G#, A#, B#, C##.   The scale with the least number of accidentals is favored.

PR #3246, opened 2025-01-08
Update community_features.md

baymud:
Added a note that "chords" is not available as an option in sidebar controls when you're in either of the chord keyboard layouts (spent too much time troubleshooting thinking I should see it, so trying to save others any confusion. Oops)


sapphire-arches (approved):

soymonitus:
@baymud I am seeing the green Chords sidebar on the isometric keyboard. Why are you saying that it is only available in the chord keyboards?

PR #3257, opened 2025-01-10
Refactor Midi Follow CC assignments to use lookup tables and add rest of automateable parameters

soymonitus:
- Refactor Midi Follow to use lookup tables (quicker to resolve the values and not tied to grid)
- Add the rest of parameters to be controlled by midi follow
    - New arpeggiator parameters
    - Stutter rate
    - Compressor threshold
    - New envelopes

TO DO:

- [x] Upload the new Loopy Pro template
- [x] Fix bug with loading midi follow xml

https://github.com/user-attachments/assets/b40a9ab3-d4ef-4b6f-af85-0594ccc9fcf2

    
shortcuts

reserved

list


soymonitus:
@seangoodvibes this is ready for review

src/deluge/io/midi/midi_follow.cpp
```diff
@@ -100,12 +79,257 @@ void MidiFollow::init() {
...
+	globalParamToCC.fill(MIDI_CC_NONE);
+}
+
+void MidiFollow::initDefaultMappings() {
+	clearMappings();
+
+	// PATCHED SOUND
+
+	// Noise
+	ccToSoundParam[41] = params::LOCAL_NOISE_VOLUME;
```
m-m-adams, line 96: We should probably sort these by CC number? 

Or stick a comment asking the next person to add a CC to sort them, not like it matters until then

soymonitus: I thought that it was better to short them by param type/name, as cc number doesn't mean anything. The next person adding another parameter would have to use the midi cc documentation and our "list of reserved CCs" to pick the right CC number from the available

soymonitus: By the way, we should be locking some CC numbers for the upcoming macro params :)

m-m-adams: > I thought that it was better to short them by param type/name, as cc number doesn't mean anything. The next person adding another parameter would have to use the midi cc documentation and our "list of reserved CCs" to pick the right CC number from the available

Yeah but from the standpoint of "which CCs are available" it would be a lot clearer if they were in order

soymonitus: Ok i will reorder then

soymonitus: @m-m-adams order changed

soymonitus:
@m-m-adams did you see anything else to change or is it good to go? :)

PR #3259, opened 2025-01-12
Add to String API

stellar-aria:
Adds new functions to String match std::string API, adds new std::string-emitting deluge-specific character conversion functions to util/string

### DIFF
--- a/src/deluge/util/string.cpp
+++ b/src/deluge/util/string.cpp
@@ -46,7 +46,7 @@ std::string fromInt(int32_t number, size_t min_num_digits) {
 
 std::string fromFloat(float number, int32_t precision) {
 	std::array<char, 64> buffer;
-	char* ptr = D_TRY_CATCH(to_chars(buffer.begin(), buffer.end(), number, precision), error, {
+	char* ptr = D_TRY_CATCH(deluge::to_chars(buffer.data(), buffer.data() + buffer.size(), number, precision), error, {
 		return std::string{}; //
 	});
 
--- a/src/deluge/util/string.cpp
+++ b/src/deluge/util/string.cpp
@@ -33,6 +33,10 @@ std::string fromInt(int32_t number, size_t min_num_digits) {
 	// Convert number to string
 	std::string result = std::to_string(number);
 
+	if (result.size() < min_num_digits) {
+		result.reserve(min_num_digits);
+	}
+
 	// Add leading zeros if needed
 	if (result[0] != '-' && result.size() < min_num_digits) {
 		result.insert(result.begin(), min_num_digits - result.size(), '0');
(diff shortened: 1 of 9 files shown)


src/deluge/util/string.cpp
```diff
@@ -0,0 +1,83 @@
...
+std::string fromFloat(float number, int32_t precision) {
+	std::array<char, 64> buffer;
+	char* ptr = D_TRY_CATCH(deluge::to_chars(buffer.data(), buffer.data() + buffer.size(), number, precision), error, {
+		return std::string{}; //
+	});
+
+	return std::string{buffer.data(), ptr};
+}
+
+std::string fromSlot(int32_t slot, int32_t subSlot, size_t minNumDigits) {
```
m-m-adams, line 56: What's this for?

src/deluge/util/string.cpp
```diff
@@ -0,0 +1,83 @@
...
+	else if (result[0] == '-' && result.size() - 1 < min_num_digits) {
+		result.insert(result.begin() + 1, min_num_digits - (result.size() - 1), '0');
+	}
+
+	return result;
+}
+
+std::string fromFloat(float number, int32_t precision) {
+	std::array<char, 64> buffer;
+	char* ptr = D_TRY_CATCH(deluge::to_chars(buffer.data(), buffer.data() + buffer.size(), number, precision), error, {
```
m-m-adams, line 49: Should this return an error of some sort instead of an empty string?

src/deluge/util/string.cpp
```diff
@@ -0,0 +1,83 @@
...
+		std::unexpected{std::errc::no_buffer_space};
+	}
+	return first + written;
+}
+} // namespace deluge
+
+namespace deluge::string {
+std::string fromInt(int32_t number, size_t min_num_digits) {
+	// Convert number to string
+	std::string result = std::to_string(number);
```
m-m-adams, line 34: This should reserve the min number of digits to avoid needing a second allocation

stellar-aria: This becomes a move operation, so reserving is only relevant if the result's size is less than min_num digits after assignment, I think

src/deluge/util/string.cpp
stellar-aria: slot browser stuff :/

src/deluge/util/string.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+std::string fromSlot(int32_t slot, int32_t subSlot, size_t minNumDigits) {
+	std::string buffer = fromInt(slot, minNumDigits);
+	if (subSlot != -1) {
+		buffer.push_back(static_cast<char>(subSlot + 'A'));
+	}
+	return buffer;
+}
+
+std::string fromNoteCode(int32_t noteCode, size_t* getLengthWithoutDot, bool appendOctaveNo) {
```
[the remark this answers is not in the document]
stellar-aria: It's guaranteed to be four characters or less, so it will never allocate.

src/deluge/util/string.cpp
stellar-aria: Added comment to clarify reason for return, per discord discussion

src/deluge/util/string.cpp
stellar-aria: Added a reserve below

m-m-adams (approved):

PR #3260, opened 2025-01-12
[DRAFT] Switch to a quadratic curve for note brightness based on note velocity

Metamere:
Switched to a quadratic curve to make the brightness gradient based on note velocity much more apparent. Float calculations made this much simpler, and means that the adjustFractional() function is not needed anymore. This is similar to what was done for PR #3132 for the kit drum velocity keyboard view updates. See comparison photos below, which were taken with the same exposure settings.

Comparing before/after changes for green, blue, and red.
G-linear
G-Quadratic
B-linear
B-Quadratic
R-linear
R-Quadratic

And here it is at the lowest system brightness setting possible, with a drastically longer exposure time needed to capture the images.
R-low-linear
R-low-quadratic
It is still well visible at the lowest brightness settings in a dark room. Not that those super low velocities and settings are commonly used, but it still makes for a good baseline check.


src/deluge/model/note/note_row.cpp
```diff
@@ -1959,12 +1959,15 @@ void NoteRow::renderRow(TimelineView* editorScreen, RGB rowColour, RGB rowTailCo
...
 			}
 
-			// Or if Note starts exactly on square...
+			// Or if note starts exactly on square...
 			else if (note && note->pos == squareStartPos) {
 				drewNote = true;
-				// this is just colour * (0.25 + 0.75 * velocity) but in fixed point, i.e. it'll go linearly from 0.25
-				// to max brightness with velocity
-				pixel = rowColour.adjustFractional((65 + note->velocity + (note->velocity / 2)) << 8, 255 << 8);
+				float velocity_ratio = (float)(note->velocity) / 127; // velocity range is 1-127
```
stellar-aria, line 1965: Please don't use a C-style cast, use either `static_cast<float>(value)` or `std::bit_cast<float>(value)` whether you are casting the float via conversion or preserving bit positions, respectively
```suggestion
				float velocity_ratio = static_cast<float>(note->velocity) / 127.f; // velocity range is 1-127
```

src/deluge/model/note/note_row.cpp
```diff
@@ -1959,12 +1959,15 @@ void NoteRow::renderRow(TimelineView* editorScreen, RGB rowColour, RGB rowTailCo
...
+				float colour_intensity = 0.03 + 0.97 * velocity_ratio * velocity_ratio;
+				// The brightness is reduced as necessary to be proportional to the note velocity.
+				// below .03 (~4/127) it is not visible enough, so we set that as the minimum
+				// A quadratic curve makes the brightness gradient appear more linear.
+				pixel = rowColour.transform([colour_intensity](uint8_t chan) { return (chan * colour_intensity); });
```
stellar-aria, line 1970: If you want to, instead of adding the small transform here, you can add an
```c++
[[nodiscard]] constexpr RGB operator*(float value) { return RGB(r * value, g * value, b * value); }
``` 
to the RGB class which simply returns a new RGB with each of the channels multiplied by the input value.
It would turn this line into
```c++
pixel = rowColour * colour_intensity;
```
which I think is much easier to understand

stellar-aria (changes requested):
LGTM pending the small cast nit

seangoodvibes:
Do we know how this looks with blurred notes and tails?

I think in the previous implementation the note head was still distinguishable from the note tail.

Also would be curious what happens with blurred notes.

Metamere:
> Do we know how this looks with blurred notes and tails?
> I think in the previous implementation the note head was still distinguishable from the note tail.

Fine with blurred notes. Not great with tails, as the dimmer notes become difficult to distinguish from the tails. After Sean and I experimented with different methods of addressing that issue, it seems it will need some more consideration as to how to address this issue. It will probably be possible, but if no satisfactory solution can be found, we may need to discuss whether we want to incorporate this change or even revert this feature.

seangoodvibes:
> > Do we know how this looks with blurred notes and tails?
> 
> > I think in the previous implementation the note head was still distinguishable from the note tail.
> 
> 
> 
> Fine with blurred notes. Not great with tails, as the dimmer notes become difficult to distinguish from the tails. After Sean and I experimented with different methods of addressing that issue, it seems it will need some more consideration as to how to address this issue. It will probably be possible, but if no satisfactory solution can be found, we may need to discuss whether we want to incorporate this change or even whether we should eliminate this feature.

Hey so after speaking with a couple folks on the discord, let's try out this change with just the quadratic curve and then go from there. We'll see what people think.

stellar-aria (approved):

seangoodvibes:
Switching PR to draft just to prevent accidentally merging this until @Metamere gives the go ahead

Metamere:
The consensus was that this change was not beneficial.

PR #3274, opened 2025-01-12
Bugfix/only reallocate if were saving a reasonable amount of space

m-m-adams:
The delay reallocates every time it's rate reduces to "improve quality" which seems dicey, and if delay rate is patched or automated results  in continuously reallocating and filling ram. Instead lets only do that if we're off by a factor of 2


stellar-aria:
Closed as #3275 includes this commit

PR #3279, opened 2025-01-13
Added 2 more envelopes

soymonitus:
Full credit to @nikodemus 

This PR adds two more envelopes, which can be accessed from the sound editor menu.

In a future PR (Layered Menus), they could be made accessible from double clicking the ADSR shortcuts on the grid, but this is not in scope in this PR


nikodemus:
I would appreciate if this was rebased to preserve my authorship of the envelope commit. (Amending with ``--author "New Author Name <email@address.example>"`` does it.)

Not a huge deal, but, eh.

soymonitus:
I think this PR needs to be merged without squashing commits, so authorship of "Added 2 envelopes" by Nikodemus is preserved?

PR #3284, opened 2025-01-15
Add Start Marker Editing in Audio Clip View

m0nkmaster:
Adds the ability to edit the starting point of audio clips directly from the clip view, similar to how end points are currently edited. This makes manipulating audio clip trimming more intuitive and efficient and removes the need for the "reverse the clip then crop the end, revert reverse" workaround. See: https://www.youtube.com/watch?v=JoLCEb-pv-A&t=1304s. Here's another workaround using Arranger view: https://www.youtube.com/watch?v=iWhVUsx40Mg&t=76s .

### Changes

* Added startMarkerVisible state to track when start marker is being edited
*  Implemented start marker showing in blue (end marker stays red)
*  Modified waveform rendering to respect both start and end markers
*  Added proper cleanup of clip playback state during marker edits

### Usage

**To edit the start point:**
1. Press any pad in the first column where your audio clip starts
2. The start marker will appear in blue
3. Press another pad to move the start point
4. Press the same pad again to hide the marker

**To edit the end point (unchanged):**
1. Press any pad in the last column where your audio clip ends
2. The end marker will appear in red
3. Press another pad to move the end point
4. Press the same pad again to hide the marker

### DIFF
--- a/src/deluge/gui/views/audio_clip_view.cpp
+++ b/src/deluge/gui/views/audio_clip_view.cpp
@@ -144,11 +144,11 @@ bool AudioClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplayWidth
 		}
 
 		int32_t xEnd = std::min(kDisplayWidth, visibleWaveformXEnd);
-		int32_t xStart = std::max(0, visibleWaveformXStart);
+		int32_t xStart = (visibleWaveformXStart < 0) ? 0 : visibleWaveformXStart;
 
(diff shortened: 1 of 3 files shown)


src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -1,9 +1,9 @@
 /*
- * Copyright © 2019-2023 Synthstrom Audible Limited
+ * Copyright 2019-2023 Synthstrom Audable Limited
```
seangoodvibes, line 2: the company is called Synthstrom Audible. This should not be changed.

PR #3285, opened 2025-01-15
Feature / Arpeggiator new Walk & Pattern note mode

soymonitus:
Added two new note modes: Walk and Pattern
As in the docs:
    - `Walk1 (WALK1)` is the "slow" walk and next note is selected this way:
        - 30% of probability to walk the sequence a step in reverse.
        - 30% of probability to just repeat the same step of the sequence.
        - 40% of probability to walk the sequence a step forward (as normal).
    - `Walk2 (WALK2)` is the "normal" walk and next note is selected this way:
        - 25% of probability to walk the sequence a step in reverse.
        - 25% of probability to just repeat the same step of the sequence.
        - 50% of probability to walk the sequence a step forward (as normal).
    - `Walk3 (WALK3)` is the "fast" walk and next note is selected this way:
        - 20% of probability to walk the sequence a step in reverse.
        - 20% of probability to just repeat the same step of the sequence.
        - 60% of probability to walk the sequence a step forward (as normal).
    - `Pattern (PATT)` each time you select this `Note Mode` on the menu it will generate a new pattern for you.
      The pattern is like if you had selected `Played Order` and the Deluge had decided the order for you.
      Tip: as this mode is the last selectable one, if you try to scroll past the end of the menu, the Deluge
      will interpret it as a re-selection of the `Pattern` mode and so it will generate a new pattern for you.
      Note: the pattern is saved to the song so you can recreate it after reloading the song.
      Note: this mode is not available for Kit Rows.
 
This PR also does some refactoring and cleanup of the arpeggiator code, removing dead code that was doing nothing, and repacking duplicated code into common functions.

The tests for how long it takes a sequence of notes to cycle:

I did several test passes with a 4-note chord on 1-bar, with an arpeggio of 4 octaves (16th notes) (With notemode=up it takes 1 bar to reach the highest note)
- It took ~10 bars for the Walk1 (slow) to reach the highest note
- It took ~6 bars for the Walk2 (normal) to reach the highest note
- It took ~3 bars for the Walk3 (fast) to reach the highest note


soymonitus:
This is ready for review!

m-m-adams (approved):

PR #3288, opened 2025-01-16
Maybe fix ghost notes issue?

stellar-aria:
It definitely fixes an oversight where .insertElementIfNonePresent() was replaced with emplace() instead of operator[]/at(), as it does assignment in addition to insertion

### DIFF
--- a/src/deluge/model/instrument/instrument.h
+++ b/src/deluge/model/instrument/instrument.h
@@ -21,6 +21,7 @@
 #include "io/midi/learned_midi.h"
 #include "model/clip/clip_instance_vector.h"
 #include "model/output.h"
+#include "storage/flash_storage.h"
 
 class StereoSample;
 class ModControllable;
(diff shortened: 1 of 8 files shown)


src/deluge/model/instrument/instrument.h
```diff
@@ -86,7 +87,7 @@ class Instrument : public Output {
 
 	virtual bool isAnyAuditioningHappening() = 0;
 
-	uint8_t defaultVelocity;
+	uint8_t defaultVelocity = FlashStorage::defaultVelocity;
```
seangoodvibes, line 90: Just want to make sure, the flash storage is already read before this default variable is initialized?

stellar-aria: Good question. The timing hasn't changed by moving it here (initialization still happens at instantiation), so I presume it is fully set up.

seangoodvibes (approved):

PR #3291, opened 2025-01-16
Add ability to trim audio clips from start position

m0nkmaster:
# Add ability to trim audio clips from their start position

## Feature Description
Adds the ability to trim audio clips from their start position in Audio Clip View, similar to the existing end-position trimming functionality. This gives users more control over audio sample editing directly from the grid interface.

Currently to trim the beginning of an audio clip you need a workaround by reversing the clip, see:
https://youtu.be/57g6LKpI_Dw?si=NRJbwa36cfgHx8og&t=58

## Implementation Details
- Added start marker functionality in AudioClipView that mirrors existing end marker behaviour
- Added `changeUnderlyingSampleStart()` method to handle start position adjustments
- Modified pad action handling to support start marker selection and adjustment
- Added runtime feature flag `TrimFromStartOfAudioClip` to control availability of this feature

## Changes
- Added start marker visual feedback (green blinking column) when selecting start position
- Start marker can be toggled by tapping the first column of the audio clip
- Changing the start marker adjusts the clip's start position while maintaining end position
- Sample length is automatically recalculated based on new start position
- Changes are tracked in action history for undo/redo support

## Acceptance Criteria
- [ ] Start marker should be toggleable by tapping first column of clip
- [ ] Start marker should blink green when active
- [ ] Pressing a pad to the right should trim from start of clip
- [ ] Changes should be undoable/redoable
- [ ] Reversed samples should maintain correct behavior
- [ ] Minimum clip length of 1 tick should be enforced
- [ ] Feature should be toggleable via community settings
- [ ] Visual feedback should be clear and consistent with existing UI patterns7
- [ ] Is non destructive - reversing audio will allow you to re-introduce trimmed start audio

## Testing Notes
- Test with both normal and reversed samples
- Check interaction with existing end marker functionality
- Verify undo/redo behaviour
- Test with various sample lengths and clip settings spanning multiple bars
- Test on both OLED and 7SEG display devices
- Reversing audio will allows you to re-introduce trimmed start audio

## Demo
https://youtu.be/9S2BkhaB50c


m0nkmaster:
This is the first iteration. Ideally I would like to change the behaviour slightly so that when you trigger Start Trimming Mode and click a new column to the right, the view doesn't jump to make that new start column one. Instead you'd see the columns to the left greyed out - the same behaviour as when editing the end. But it's a little bit tricky to work out how that would affect playback. It introduces the concept of columns before the start of playback i.e. minus (<0) columns. I need to think about that. I know that the green column always jumping to be col one is a bit of a jaunting experience. 

However, as undo is implemented, it's pretty natural to quickly step back if you've cropped too much audio.

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -230,57 +276,43 @@ void AudioClipView::graphicsRoutine() {
...
-
 		uint8_t tickSquares[kDisplayHeight];
 		memset(tickSquares, newTickSquare, kDisplayHeight);
 
-		const uint8_t* colours = getCurrentClip()->getCurrentlyRecordingLinearly() ? twos : zeroes;
-		PadLEDs::setTickSquares(tickSquares, colours);
+		const uint8_t* coloursPtr = getCurrentClip()->getCurrentlyRecordingLinearly()
+		                                ? (const uint8_t*)"\2\2\2\2\2\2\2\2"
+		                                : (const uint8_t*)"\0\0\0\0\0\0\0\0";
+		PadLEDs::setTickSquares(tickSquares, coloursPtr);
```
stellar-aria, line 297: Please don't do type-casts like this, it's extremely unclear what the data should be and why. If it should be an array, please use an array:

```suggestion
		const std::array<uint8_t, 8> coloursArray = {0};
		if (getCurrentClip()->getCurrentlyRecordingLinearly) {
		    coloursArray.fill(2);
		}
		PadLEDs::setTickSquares(tickSquares, coloursArray.data());
```

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -439,184 +471,141 @@ ActionResult AudioClipView::buttonAction(deluge::hid::Button b, bool on, bool in
...
 void AudioClipView::changeUnderlyingSampleLength(AudioClip* clip, const Sample* sample, int32_t newLength,
                                                  int32_t oldLength, uint64_t oldLengthSamples) const {
 	uint64_t* valueToChange;
 	int64_t newEndPosSamples;
 
 	uint64_t newLengthSamples =
-	    (uint64_t)(oldLengthSamples * newLength + (oldLength >> 1)) / (uint32_t)oldLength; // Rounded
-	// AudioClip reversed
-	if (clip->sampleControls.reversed) {
+	    (uint64_t)(oldLengthSamples * newLength + (oldLength >> 1)) / (uint32_t)oldLength; // rounding
```
stellar-aria, line 584: please use C++ style casting and division over C casting and shifting
```suggestion
	    static_cast<uint64_t>(oldLengthSamples * newLength + (oldLength / 2)) / static_cast<uint32_t>(oldLength); // rounding
```

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -439,184 +471,141 @@ ActionResult AudioClipView::buttonAction(deluge::hid::Button b, bool on, bool in
...
-		// rounding happened along the way and just go with the original
-		if (sample->fileLoopEndSamples) {
-			int64_t distanceFromFileEndMarker = newEndPosSamples - (uint64_t)sample->fileLoopEndSamples;
-			if (distanceFromFileEndMarker < 0) {
-				distanceFromFileEndMarker = -distanceFromFileEndMarker; // abs
-			}
-			if (distanceFromFileEndMarker < 10) {
-				newEndPosSamples = sample->fileLoopEndSamples;
-			}
+		if ((int64_t)newEndPosSamples > (int64_t)sample->lengthInSamples) {
```
stellar-aria, line 595:
```suggestion
		if (newEndPosSamples > static_cast<int64_t>(sample->lengthInSamples)) {
```

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -627,19 +616,67 @@ void AudioClipView::changeUnderlyingSampleLength(AudioClip* clip, const Sample*
...
-void AudioClipView::playbackEnded() {
+// ----------- "Start" pointer logic -----------
+void AudioClipView::changeUnderlyingSampleStart(AudioClip* clip, const Sample* sample, int32_t newStartTicks,
+                                                int32_t oldLength, uint64_t oldLengthSamples) const {
+	int32_t oldEndTick = oldLength;
+	int32_t newLengthTicks = oldEndTick - newStartTicks;
+	if (newLengthTicks < 1) {
+		newLengthTicks = 1;
+	}
+	uint64_t newLengthSamples = (uint64_t)(oldLengthSamples * newLengthTicks + (oldLength >> 1)) / (uint32_t)oldLength;
```
stellar-aria, line 627:
```suggestion
	uint64_t newLengthSamples = static_cast<uint64_t>(oldLengthSamples * newLengthTicks + (oldLength / 2)) / static_cast<uint32_t>(oldLength);
```

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -682,25 +719,18 @@ void AudioClipView::adjustLoopLength(int32_t newLength) {
...
 			if (newLength > 0) {
-
 				action = shortenClip(newLength);
-
-				// Scroll / zoom as needed
 				if (!scrollLeftIfTooFarRight(newLength)) {
-					// If this zoom level no longer valid...
 					if (zoomToMax(true)) {
-						// editor.displayZoomLevel(true);
+						// ...
```
stellar-aria, line 733: Feel free to invert the condition and remove the `{} else {`

stellar-aria (commented):
Not a true review, just some little style and QoL things I saw on a quick browse

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -105,72 +107,121 @@ bool AudioClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplayWidth
...
-
 		for (int32_t y = 0; y < kDisplayHeight; y++) {
 			memset(image[y], 0, kDisplayWidth * 3);
 		}
+		return true;
 	}
 
-	// Or if yes Sample...
-	else {
+	AudioClip* clip = getCurrentAudioClip();
```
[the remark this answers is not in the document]
m-m-adams: I think it needs to be checked before making a reference - we shouldn't end up here without the current clip being an audio clip but it's probably possible, in which case getCurrentAudioClip returns a nullptr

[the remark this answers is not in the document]
m0nkmaster: I'm not super familiar with C++, I've given it a go updating to use references where possible and ensure we have checked for null, see what you think.

m0nkmaster:
Recommended changes implemented

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -159,14 +159,14 @@ bool AudioClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplayWidth
 					// If endMarkerVisible, show red (bright vs. dim).
 					if (endMarkerVisible) {
 						if (blinkOn) {
-							image[y][endSquareDisplay][0] = 255; // bright red
-							image[y][endSquareDisplay][1] = 0;
-							image[y][endSquareDisplay][2] = 0;
+							image[y][endSquareDisplay][0] = colours::red.r; // bright red
```
stellar-aria, line 161: Note, you should just be able to do
`image[y][endSquareDisplay] = colours::red;` here (and for the others), as image is just a 2D array of RGBs

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -105,72 +106,107 @@ bool AudioClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplayWidth
...
-
 		for (int32_t y = 0; y < kDisplayHeight; y++) {
 			memset(image[y], 0, kDisplayWidth * 3);
 		}
+		return true;
 	}
 
-	// Or if yes Sample...
-	else {
+	AudioClip& clip = *getCurrentAudioClip();
```
m-m-adams, line 117: this can fail if the current clip isn't an audio clip, you have to check for a null pointer before converting it to a reference

auto clipptr = getCurrentAudioClip(); 
if (clipptr == nullptr) {some error return or whatever} 
else {clip = *clipptr}

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -180,10 +216,9 @@ bool AudioClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplayWidth
 
 ActionResult AudioClipView::timerCallback() {
 	blinkOn = !blinkOn;
-	uiNeedsRendering(this, 0xFFFFFFFF, 0); // Very inefficient!
+	uiNeedsRendering(this, 0xFFFFFFFF, 0);
```
m-m-adams, line 219: I think some note is still worthwhile since this re renders the whole UI when we probably only need to render the column at the sample marker

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -230,57 +261,45 @@ void AudioClipView::graphicsRoutine() {
...
 	}
-
-	// Tempoless or arranger recording
 	else if (!playbackHandler.isEitherClockActive()
 	         || (currentPlaybackMode == &arrangement && getCurrentClip()->getCurrentlyRecordingLinearly())) {
 		newTickSquare = kDisplayWidth - 1;
-
-		// Linearly recording
-		if (getCurrentClip()->getCurrentlyRecordingLinearly()) { // This would have to be true if we got here, I think?
-			getCurrentAudioClip()->renderData.xScroll = -1;      // Make sure values are recalculated
```
m-m-adams, line 241: why is this removed?

src/deluge/gui/views/audio_clip_view.cpp
```diff
@@ -439,184 +458,143 @@ ActionResult AudioClipView::buttonAction(deluge::hid::Button b, bool on, bool in
...
 
-					// If tapped on the marker itself again, make it invisible
-					if (x == endSquareDisplay) {
-						if (blinkOn) {
-							uiNeedsRendering(this, 0xFFFFFFFF, 0);
-						}
-						uiTimerManager.unsetTimer(TimerName::UI_SPECIFIC);
+				// =========== Handling END marker =============
+				if (endMarkerVisible) {
+					// If user taps the same or adjacent end marker column => toggle off
```
m-m-adams, line 499: why is it including the start as well?

m-m-adams (commented):
I think this generally looks good, just some minor things I'm not clear on

src/deluge/gui/views/audio_clip_view.cpp
m0nkmaster: Check added which follows the pattern of `getSample()` which is just above.

src/deluge/gui/views/audio_clip_view.cpp
m0nkmaster: Reinstated, accidental removal whilst refactoring.

src/deluge/gui/views/audio_clip_view.cpp
m0nkmaster: I'm checking for presses on both markers because if you're clicking the start marker while trimming the end, it suggests you want to focus on the start. 

Trimming the end back to the start doesn't make much sense. 

This approach simplifies switching between trimming the start and end.

m0nkmaster (commented):
Responded to comments on audio_clip_view.cpp changes.

m-m-adams (approved):
Awesome, Lgtm!

seangoodvibes:
@m0nkmaster can you please document this feature addition in our community features document + our change log?

Thank you

m0nkmaster-disney:
> @m0nkmaster can you please document this feature addition in the following documents.
> 
> * [ x ]  community features
> * [  x ]  change log
> * [ x ]  menu hierarchies

Ready to roll!!!

docs/community_features.md
```diff
@@ -1172,6 +1172,17 @@ for the Lumi Keys Studio Edition, described below.
     - While Lumi has limited options for MPE separation, it will be configured to align with the dominant MPE range
       defined on the Deluge (upper or lower dominant).
 
+### 4.11.0 - Audio Clip View - Trimming Clips
+
+- ([#3291]) Added a new `STRING_FOR_COMMUNITY_FEATURE_TRIM_FROM_START_OF_AUDIO_CLIP` feature to allow users to trim from the start of a clip without needing to reverse it.
```
seangoodvibes, line 1177: For this PR reference to work you need to add a link at the bottom of the community features document :)

docs/community_features.md
```diff
@@ -1172,6 +1172,17 @@ for the Lumi Keys Studio Edition, described below.
...
+  - You can revert to an earlier portion by pressing undo (or reverse the clip and altering as before).
+  - This lets you alter the start of an audio clip without time stretching.
+  - Previously, this was only possible by reversing the audio clip and trimming the start as if it were the end.
+
+See this demo for more details:
+[Audio Clip View - Trimming Tips](https://www.youtube.com/watch?v=iWhVUsx40Mg&t=45s&ab_channel=RonCavagnaro).
+
 ## 5. Community Features Menu (aka Runtime Settings)
 
 In the main menu of the Deluge (accessed by pressing both "SHIFT" + the "SELECT" encoder) there is
```
seangoodvibes, line 1188: The community features menu setting should also be referenced in this section

seangoodvibes: See 1252 where an entry should be added

seangoodvibes (approved):

seangoodvibes (approved):

m-m-adams (approved):

PR #3294, opened 2025-01-18
Change to using an animated SD card icon instead of a loading or saving popup

Metamere:
The loading animation was not working as intended, because a timer was only set once for the standard popup time. It needed to be set for a short duration, then set again once it runs out and an update is made. During the discovering of this, I created a better way to indicate loading, that doesn't block the name of the thing you are loading. I made a small SD card icon that shows up in the upper right corner. It only displays when loading or saving time exceeds 350ms. Then it will display a static icon for 750ms. Then it will load an alternate SD card icon that has the empty inner portion filled with a checkerboard. Then it will alternate between the two images every 250 ms, causing a blink to occur every 500ms (2Hz). 

Also reduced text scrolling and light blinking speeds slightly to make things easier to track and less distracting.

### DIFF
--- a/src/deluge/hid/display/oled.cpp
+++ b/src/deluge/hid/display/oled.cpp
@@ -791,69 +791,66 @@ void OLED::popupText(char const* text, bool persistent, PopupType type) {
 
 void updateWorkingAnimation() {
 	// D_PRINTLN("working animation count: %d", working_animation_count);
-	int32_t y_pos = OLED_MAIN_TOPMOST_PIXEL + 2;
 	deluge::hid::display::oled_canvas::Canvas& image = deluge::hid::display::OLED::main;
 
-	int32_t w = 15;
-	int32_t w2 = 5;
-	int32_t h = 7;
+	int32_t w1 = 5; // spacing between rectangles
+	int32_t w2 = 5; // width of animated portion of rectangle
+	int32_t h = 8; // height of rectangle
+	int32_t offset = w2 - 2; // causes shifting lines to overlap by 2 pixels
 	int32_t x_max = OLED_MAIN_WIDTH_PIXELS - 1;
-	int32_t x_min = x_max - w;
-	int32_t t_reset = w + w2 * 2;
-
-	if (!started_animation) {
-		// clear space and draw outer borders that will not change during the animation
+	int32_t x_min = x_max - (w1 + w2 * 2);
+	int32_t x2 = x_max - w2; // starting position of right rectangle
+	int32_t y1 = OLED_MAIN_TOPMOST_PIXEL + 2; // top of rectangles
+	int32_t y2 = y1 + h - 1; // bottom of rectangles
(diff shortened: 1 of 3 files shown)


src/deluge/hid/display/oled.cpp
```diff
@@ -56,8 +56,8 @@ oled_canvas::Canvas OLED::console;
 
 bool OLED::needsSending;
 
-int32_t workingAnimationCount;
-char const* workingAnimationText; // NULL means animation not active
+int32_t working_animation_count;
+// char const* workingAnimationText; // NULL means animation not active
```
stellar-aria, line 60:
```suggestion
```

src/deluge/hid/display/oled.cpp
```diff
@@ -413,7 +433,7 @@ void OLED::removePopup() {
 
 	oledPopupWidth = 0;
 	popupType = PopupType::NONE;
-	workingAnimationText = nullptr;
+	// workingAnimationText = nullptr;
```
stellar-aria, line 436:
```suggestion
```

src/deluge/hid/display/oled.cpp
```diff
@@ -979,11 +1007,13 @@ void OLED::stopScrollingAnimation() {
 
 void OLED::timerRoutine() {
 
-	if (workingAnimationText) {
-		workingAnimationCount = (workingAnimationCount + 1) & 3;
+	if (working_animation_count) {
+		// working_animation_count = (working_animation_count + 1) & 1;
```
stellar-aria, line 1011:
```suggestion
```

src/deluge/hid/display/oled.cpp
```diff
@@ -979,11 +1007,13 @@ void OLED::stopScrollingAnimation() {
...
+		working_animation_count++;
 		updateWorkingAnimation();
+		markChanged();
 	}
-
+	// if(workingAnimationText) return;
```
stellar-aria, line 1016:
```suggestion
```

src/deluge/hid/display/oled.cpp
```diff
@@ -778,36 +798,44 @@ void OLED::popupText(char const* text, bool persistent, PopupType type) {
...
+			image.drawGraphicMultiLine(deluge::hid::display::OLED::SD_icon1, OLED_MAIN_WIDTH_PIXELS - 9, y_pos, 7);
+		}
+		uiTimerManager.setTimer(TimerName::DISPLAY, 250);
 	}
-
-	error = textNow.concatenate(buffer);
-	OLED::popupText(textNow.get(), true, PopupType::LOADING);
 }
 
 void OLED::displayWorkingAnimation(char const* word) {
```
stellar-aria, line 823: This signature will probably need to be changed if it's no longer valid...

stellar-aria (changes requested):
Pending

PR #3297, opened 2025-01-18
Check resources prior to running tasks

m-m-adams:
Make sure that all of a tasks resources are available before trying to start it

This avoids deadlock from two tasks waiting on each other (it's essentially priority ceiling if all tasks had the same priority) as well as preventing multiple access to single resources

### DIFF
--- a/src/OSLikeStuff/scheduler_api.h
+++ b/src/OSLikeStuff/scheduler_api.h
@@ -29,7 +29,13 @@ extern "C" {
 typedef void (*TaskHandle)();
 typedef bool (*RunCondition)();
 typedef int8_t TaskID;
-
+typedef uint32_t ResourceID;
+#define NO_RESOURCE 0
+// for actually reading the SD
+#define SD 1
+#define USB 2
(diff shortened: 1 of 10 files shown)


src/OSLikeStuff/scheduler_api.h
```diff
@@ -29,7 +29,13 @@ extern "C" {
 typedef void (*TaskHandle)();
 typedef bool (*RunCondition)();
 typedef int8_t TaskID;
-
+typedef uint32_t ResourceID;
+#define NO_RESOURCE 0
```
stellar-aria, line 33: can these get prefixed if they're going to be global defines?

stellar-aria (changes requested):
LGTM pending prefixing those defines so we don't accidentally end up with collisions

stellar-aria (approved):

stellar-aria:
@m-m-adams looks like the tests need to be updated now

stellar-aria (approved):

PR #3301, opened 2025-01-19
Refactor/small obj allocator

m-m-adams:
Made the alignment info into class members instead of global constants in order to allow creating an extra region for small objects (e.g. strings, maybe other stuff) that don't fit the same allocation pattern as non stealable song and audio data

Also  make the cache manager optional since stealables are only allocated in one region. I think branching on whether there's a cache manager is better than using a virtual reclaim memory function but that's just gut feel

todo: actually creating the extra region


m-m-adams:
Putting this back to draft, I've got a weird crash popping up

seangoodvibes (approved):

PR #3303, opened 2025-01-21
Diatonic Tick Marks

todd-gochenour:
My favorite keyboard is the In-Key layout for its muted monotonic color composition.  I find it easier to visualize and comprehend relative notes in a scale without the wash of color. I have made this keyboard layout my default.  

Problem with this default is that there is no chromatic equivalent.  Pressing the SCALE button to change between diatonic and chromatic layouts does nothing.  What I want displayed is the Isomorphic keyboard lit with the same muted color composition.  This layout makes obvious which notes are diatonic and which notes lie outside of scale.  Diatonic notes are lit and notes outside the scale are black. When changing from a major to a minor scale, the shifting diatonic notes are visually identifiable. 

This design language should also be used by the synth kit's audition pads.  Currently audition pads show a colored root note and a stack of either 6 or 10 black notes depending on SCALE before the next root.  Lighting the diatonic notes and having the Perfect 5th degree slightly brighter helps visual comprehension by dividing the lower three diatonic notes from the upper two. Note rows become identifiable at a glance.  No more pressing buttons to see popups.

The KEYBOARD and SCALE buttons should operate independent of each other.  Pressing the KEYBOARD button shouldn't affect the state of SCALE.  There are 4 states that the layout could be in:
- If SCALE is on and KEYBOARD is on, then the In-Key keyboard layout is used.  
- If SCALE is off and KEYBOARD is on, then the Isomorphic keyboard layout is used.  
- If SCALE is on and KEYBOARD is off, then the audition pads show lit diatonic notes same as the In-Key keyboard.
- If SCALE is off and KEYBOARD is off, then the audition pads show lit diatonic notes and black outside notes, same as the Isomorphic keyboard.


m-m-adams:
Could we just update the out-of-scale isomorphic layout to show fifths? It's a lot nicer to orient this way and I don't see a reason it should be confined to people who use the in-key layout as a default

todd-gochenour:
> Could we just update the out-of-scale isomorphic layout to show fifths? It's a lot nicer to orient this way and I don't see a reason it should be confined to people who use the in-key layout as a default

The diatonic ticks are fundamental to the usability of the keyboard.  These ticks communicate whether there is a major or minor 3rd, major or dominant 7th, whether the scale is harmonic or melodic minor.   These ticks are akin to black keys on a piano or fret markers on a guitar neck.  How does suppressing this information improve the user's orientation?

I find the isomorphic out-of-scale (chromatic) keyboard with its row of 10 black pads between colored root notes to be disorienting.  Highlighting the fifth alone does not communicate anything useful.  In the isomorphic layout the fifth is always the pad directly below the colored root for all scales accept Locrian (and the fourth is always the pad above the root except for Lydian).  With this design optimization, the colored root and highlighted fifth serve as anchors to visually assess intervals of the 2nd, 3rd, 4th, 6th and 7th notes in the scale.  

I've been using this visual device now for a couple of weeks and my confidence knowing at a glance which notes are in or out of key has improved considerably.  The benefit really shines when vertically scrolling out-of-scale note rows as the diatonic ticks animate the movement such that scrolling is more precise.   I hope that one day playing out-of-scale grace notes will no longer force the well-defined scale to become a non-descript user scale.

m-m-adams:
Yeah dude I tried it and liked it, I don't think it should be limited to the in key keyboard.

todd-gochenour:
Oh, OK.  Sorry for misunderstanding your first request.  Good to know you approve of the design.  I will remove use of the function isInKeyLayoutDefault() from the code.  

One issue still to resolve.  When the SCALE button is off, from the common chromatic keyboard layout with diatonic tick marks, I need to develop a mechanism to decide which In-Scale keyboard (Isomorphic or In-Key) should display once the SCALE button is turned back on.

seangoodvibes:
I'm a bit puzzled by the problem this is trying to solve

The isomorphic keyboard is both useable with scale on or off. Turning scale on or off only changes what notes are highlighted so you know what notes are in scale, etc, but it still lets you play out of scale notes.

The in key keyboard is only usable with scale on because it's designed to only allow you to play notes "in key" - there are never any out of scale notes

The isomorphic keyboard is the original deluge keyboard implementation and I do not think we should be modifying how it works

todd-gochenour:
I have removed the constraint requiring an In-Key Keyboard default set to enable diatonic tick marks in the Isomorphic keyboard and on Note Row Audition pads.    

The only new functionality added gives the In-Key keyboard the same function as the Isomorphic keyboard, that is the ability to toggle between diatonic and chromatic layouts using the SCALE button.   Other than the presence of diatonic tick marks there is no other change in behavior or functionality.    

Does this change require a user preference switch to turn on Diatonic Tick Marks?   I guess that was the purpose of the In-Key keyboard default constraint which I just removed.

todd-gochenour:
I withdraw this PR.

PR #3310, opened 2025-01-23
[DRAFT] Add Battery level pop-up 🔋🪫

m0nkmaster:
### Add Battery Health Pop-up and Power Management Improvements

Adds a new feature to display battery health pop-up on the OLED display and refactors the power management system for improved functionality and maintainability.

#### Key Changes:
- **Battery Health Pop-up**: 
  - Added a pop-up to the OLED display that shows the current battery level as a percentage and the remaining millivolts (mV).
  - The pop-up can be accessed by pressing `SHIFT` + `CROSS SCREEN`.
  - Charging is recommended when the battery voltage drops below 2900mV.

- **Power Management Refactor**:
  - Introduced a new `PowerManager` class to handle power-related functionalities.
  - Moved battery voltage reading and LED indication logic to `PowerManager`.
  - Improved battery status detection with defined thresholds for `CRITICAL`, `WARNING`, `HEALTHY`, and `FULL` states.
  - Implemented a method to display power status on the OLED display.

- **Code Cleanup**:
  - Removed redundant code related to battery voltage reading and LED control from `deluge.cpp`.

- **Documentation Updates**:
  - Updated `CHANGELOG.md` and `community_features.md` to reflect the new battery health pop-up feature.
---
Follows this discussion:
https://github.com/SynthstromAudible/DelugeFirmware/discussions/2991


src/deluge/deluge.cpp
```diff
@@ -165,11 +165,15 @@ void inputRoutine() {
...
 		batteryMV = (voltageReadingLastTime) >> 15;
-		// D_PRINT("batt mV: ");
-		// D_PRINTLN(batteryMV);
 
-		// See if we've reached threshold to change verdict on battery level
+		int32_t newBatteryMV = batteryMV;
+
+		// Update battery display if voltage has changed significantly
+		// TODO: Do we need this, can it be self-contained in session_view?
+		sessionView.displayPotentialBatteryChange(newBatteryMV);
```
seangoodvibes, line 173: I think if you store new battery MV in a global variable then sessionview::graphicsroutine can use it to call display potential battery change

Graphics routine is used to update the display for other things like the bpm

src/deluge/deluge.cpp
```diff
@@ -713,7 +718,7 @@ extern "C" int32_t deluge_main(void) {
 	setPinMux(ANALOG_CLOCK_IN.port, ANALOG_CLOCK_IN.pin, 2);
 
 	// Line out detect pins
-	setPinAsInput(LINE_OUT_DETECT_L.port, LINE_OUT_DETECT_L.pin);
+	setPinAsInput(LINE_OUT_DETECT_L.port, LINE_OUT_DETECT_R.pin);
```
m-m-adams, line 721: Was this change included by mistake?

m0nkmaster: Yes, reverted.

src/deluge/gui/views/session_view.cpp
```diff
@@ -2430,16 +2431,31 @@ void SessionView::displayPotentialTempoChange(UI* ui) {
...
-		}
-	}
+    // Only update display if we're in the session view
+    if (getCurrentUI() == this) {
+        // Always update if:
+        // 1. Voltage has changed significantly (50mV)
+        // 2. Last displayed value is 0 (first time or just entered view)
+        // 3. Every ~1 second
+        static uint32_t lastUpdateTime = 0;
+        uint32_t currentTime = *TCNT[TIMER_SYSTEM_SLOW];
```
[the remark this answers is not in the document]
m0nkmaster: Ok, good idea!

m0nkmaster:
It's now working as I would want; as a pop-up. The trigger for this is shift and cross-screen, which I don't think is used elsewhere. Just got to add documentation and do a bit more testing. 

Happy for anyone else to give it a test drive.

m-m-adams:
Could you put it in the menu somewhere as well? We want to make everything findable without using shortcuts

src/deluge/deluge.cpp
```diff
@@ -16,6 +16,27 @@
...
+#include "hid/matrix/matrix_driver.h"
+#include "io/debug/print.h"
+#include "io/midi/midi_engine.h"
+#include "model/settings/runtime_feature_settings.h"
+#include "processing/engines/audio_engine.h"
+#include "storage/file_item.h"
+#include "storage/storage_manager.h"
+#include "system/power_manager.h"
+#include "util/cfunctions.h"
+#include "util/functions.h"
```
stellar-aria, line 40: No newline here please, it's affecting how the includes are sorted and making the change bigger than it should be

stellar-aria (changes requested):
There's a bunch of stuff from clang-tidy that needs to be addressed, including naming conventions for new code and narrowing conversions

jtdaugherty:
Hi, folks - I was looking around to see whether there was a way to see what the Deluge's charge level is, and this is the closest thing I could find. Is this work still going to be completed and merged? If so, I'm looking forward to it!

src/deluge/deluge.cpp
m0nkmaster: Fixed in d71ac508: removed the extra blank line so the include block remains clean and the diff stays minimal.

m0nkmaster:
Implemented in ba547f72: Battery status is now available from the Settings menu (beside Firmware version), while the Shift + Cross Screen shortcut remains. Added the OLED and seven-segment menu labels too.

PR #3313, opened 2025-01-24
Feature / Enable a new menu in Sound menu to make synths and sounddrums also send midi at the same time as they play a sound

soymonitus:
The reason for this feature is for your sample drums or synths to trigger external synths, devices, or effects.
There is a new entry in the SOUND menu called MIDI with options:
- Channel: OFF, 1, 2, ...16
- Note: 0, 1, ...127 (defaults to 60, that is, C3) (shown only for kit rows). The note you set is the "base" note sent by the kit row, but if you have the arpeggiator enabled it will send the correct note, based on 'Chord Simulator' parameter and based on 'Octaves' parameter, depending on the post-arp note that triggers.

My use case:
In my case I will use this feature to send a midi note in kick drum rows, that will trigger an external sidechain compressor, so external synths that don't go through the Deluge can also be sidechain compressed individually apart from the internal Deluge sounds.

For synths:
It currently supports regular MIDI with channel or poly aftertouch. MPE is not supported yet and can be done in a different PR


m-m-adams (commented):
I think this should be on all sounds and not just drums - the code is already shared with synths so it would just be moving  the menus from drum to sound

soymonitus:
> I think this should be on all sounds and not just drums - the code is already shared with synths so it would just be moving  the menus from drum to sound 

But this aimed to be a simple functionality. Is it ok then to only support regular midi and not MPE? I wanted to make drums still light, like the mididrum class is, which is as simple as a channel and a note.

Is there any way to make synths and drums behave differently? Kit rows have been always simpler as much as they can, but i agree that eventually a synth should be able to also output midi. But synths are a different beast cause MPE comes into play

m-m-adams:
MPE comes into play for drums as well!

soymonitus:
> MPE comes into play for drums as well! 

Not for the existing midi drum rows though

soymonitus:
I support my argument in for example the CV drums, they don't even support pitch or velocity, they are just triggers, if tou want that you have the CV synth clips. Same goes for midi drums, just a note. I think they do have polyAT though (but not MPE) so i should maybe add that too to this PR

soymonitus:
I think I can move all the implementation from Sound to SoundDrum so each type of sound has their own implementation. I think putting everything into the sound class is not the right move as the midi_instrument class has a LOT of code for midi, so it might be better to refactor it into a MidiProcessor class (or something like that), so midiinstruments and soundinstruments both can have a MidiProcessor in charge of doing all the midi and mpe stuff.

For soundDrums, as the goal is to make them super simple and light on resources, I can move all my implementation from Sound to SoundDrum, so it feels similar to MidiDrum in terms of simplicity.

soymonitus:
@m-m-adams I moved all my code to SoundDrum class as per the explanation from my previous messages. Now the midi out for SounDrums is kept damn simple (light on resources), while in the future another implementation can be done for SoundInstruments, which will be far more complicated (will require a big refactoring of MidiInstrument to decouple MIDI related code from it, or even make SoundInstrument inherit from MidiInstrument

m-m-adams:
> > MPE comes into play for drums as well! 
> 
> Not for the existing midi drum rows though.
> 
> That's why i expected this feature just as if you merge a sound row with a midi row into one. I don't expect anything else from it

Existing midi drum rows can have mpe data recorded in them, it's been around since official 4.0.

Since the implementation is based on sound and not drum it already supports synths. It basically just needs the menus moved from drum to sound

m-m-adams:
I don't think moving the code to drum was a good idea. It was fine in sound and reduces code duplication

soymonitus:
@m-m-adams I had to do several changes to avoid sending duplicated noteOffs, and to send the correct polyphonic expression events. Now it is fully working for every case (sequenced notes, auditioned notes, arpeggiator off and on, allNoteOff events (when changing arp preset for example), etc. Meaning there is always a note off for every note on.

m-m-adams (approved):

m-m-adams (approved):

PR #3314, opened 2025-01-25
Feature / New "Reverse" probability (for sequenced and arpeggiated notes)

soymonitus:
**Reverse Probability (RVRS):** It applies a chance of inverting the Reverse sample setting just for the current note to be played. This probability only affects the oscillators whose type is set to Sample. Note: you can find this parameter also at the root level of the sound menu, under Randomizer, because this parameter affects both sequenced notes and arpeggiated notes.

This PR also contains this:
- Fixed midi follow XML generation, where Sidechain Level parameter was not correctly mapped
- Even more cleanup of arpeggiator code
- Expose these params for non-arp notes: Velocity Spread, Note Probability, Reverse Probability, Lock randomizer
- Now, when Lock Randomizer is ON, the sequence is reseted when clip starts from the beginning (or when it loops), rather than restarting on every chord change (which was very limiting) 
- Removed outdated strings
- Reordered arpeggiator params in automation view menu so they have the same order as in the sound menu
- Make the Global Randomizer menu be an "horizontal" menu too, accessible by doing shift+Velocity shortcut

list


soymonitus:
Found the crash and fixed it. This is ready for review!

soymonitus:
The person that reviews this: please take special attention to the places where i have replaced "reversed" with isReversed() AND the places where i have not replaced it. I think i made the right choices but maybe some more experimented dev can find something else

soymonitus:
Updated loopy pro template and docs. Ready for review

contrib/midi_follow/loopy_pro/Deluge Midi Follow.lpproj.zip
stellar-aria, line 1: If this isn't a binary, and is just a weird custom text file or folder, please commit the file/folder directly instead of through a zip. git already uses gzip compression

src/deluge/gui/menu_item/arpeggiator/octaves.h
```diff
@@ -29,5 +29,8 @@ class Octaves final : public Integer {
 	bool isRelevant(ModControllableAudio* modControllable, int32_t whichThing) override {
 		return !soundEditor.editingGateDrumRow();
 	}
+	void getColumnLabel(StringBuf& label) override {
+		label.append(deluge::l10n::getView(deluge::l10n::built_in::seven_segment, this->name).data());
```
stellar-aria, line 33: clang-tidy is correct, please use `l10n::get` here instead of `l10n::getView().data()`
```suggestion
		label.append(deluge::l10n::get(deluge::l10n::built_in::seven_segment, this->name));
```

src/deluge/gui/menu_item/arpeggiator/randomizer_lock.h
```diff
@@ -40,6 +40,10 @@ class RandomizerLock final : public Selection {
 		};
 	}
 
+	void getColumnLabel(StringBuf& label) override {
+		label.append(deluge::l10n::getView(deluge::l10n::built_in::seven_segment, this->name).data());
```
stellar-aria, line 44: same here
```suggestion
		label.append(deluge::l10n::get(deluge::l10n::built_in::seven_segment, this->name));
```

src/deluge/gui/menu_item/arpeggiator/octaves.h
m-m-adams, line 33: You can't do this, string view data is an array of characters and not a null terminated string

More realistically you could refactor string buf append to take a string view pretty easily - it just needs the signature changed and then an update to use strncat with the view length instead of strcat

src/deluge/model/sample/sample_controls.h
```diff
@@ -27,4 +27,7 @@ class SampleControls {
 	InterpolationMode interpolationMode;
 	bool pitchAndSpeedAreIndependent;
 	bool reversed;
```
m-m-adams, line 29: this should become private now

tests/spec/string_spec.cpp
```diff
@@ -5,58 +5,50 @@ extern "C" void putchar_(char c) {
...
-		});
-
-		it("returns the length without the dot", _ {
-			size_t length;
-			expect(string::fromNoteCode(63, &length)).to_equal("D.3"s);
-			expect(length).to_equal(2);
-		});
-	});
-});
+describe string(
```
stellar-aria, line 8: this will need reformatting after this change:
```suggestion
// clang-format off
describe string(
```
to preserve the manual formatting

src/deluge/gui/menu_item/arpeggiator/octaves.h
soymonitus: Not sure how to implement what you mean? Could you show an example?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -679,13 +704,17 @@ void ArpeggiatorForDrum::switchNoteOn(ArpeggiatorSettings* settings, ArpReturnIn
...
 		instruction->arpNoteOn = &arpNote;
 	}
 }
 
 bool ArpeggiatorBase::getRandomProbabilityResult(uint32_t value) {
 	if (value == 0) {
-		return false;
+		return false; // If value at min probability, return always false
+	}
+	if (value == 4294967295u) {
```
stellar-aria, line 716:
```suggestion
	if (value == std::numeric_limits<uint32_t>::max()) {
```

src/deluge/gui/menu_item/arpeggiator/octaves.h
soymonitus: I will try to implement that suggestion

stellar-aria (dismissed):
Minor nits, overall logic seems sound, although I didn't do a deep-dive

src/deluge/gui/menu_item/arpeggiator/octaves.h
stellar-aria: See review below, tldr just use `l10n::get`

src/deluge/model/sample/sample_controls.h
soymonitus: I will now then implement a setter method for the places where it needs that

contrib/midi_follow/loopy_pro/Deluge Midi Follow.lpproj.zip
soymonitus: ok I will unzip

src/deluge/gui/menu_item/arpeggiator/octaves.h
soymonitus: done

src/deluge/gui/menu_item/arpeggiator/randomizer_lock.h
soymonitus: done

src/deluge/model/sample/sample_controls.h
soymonitus: changed the original method to isCurrentlyReversed cause it is a different usage. now it makes it clear we are not accessing the "reversed" boolean but a computed field

m-m-adams (approved):

PR #3319, opened 2025-01-29
Resized and reformatted the Audio Source context menu

Metamere:
- The Audio source context menu now covers the whole screen. Does not affect other context menus.
- It displays three rows of menu items instead of two, and due to the wider window and the shortening of one of the strings (output became out), everything fits and horizontal scrolling text is no longer needed.
- Removed abbreviation, Bal. to Balanced since it fits fine now.
- Deleted the strings for options that are no longer used (monitoring options, track FX option)


Metamere:
My format tool isn't working for some reason. I tried nuking and configuring, deleting the toolchain and redownloading, same issue. I posted about that in the support-for-devs channel.

Metamere:
It looks like V21 works fine, but V22 formatting is broken.

m-m-adams (approved):

PR #3332, opened 2025-02-04
LFO 3 & 4

soymonitus:


src/deluge/processing/sound/sound.h
```diff
@@ -84,6 +84,7 @@ class Sound : public ModControllableAudio {
 	uint32_t sourcesChanged; // Applies from first source up to FIRST_UNCHANGEABLE_SOURCE
 
 	LFO globalLFO;
+	LFO globalLFO2;
```
seangoodvibes, line 87: You'll have to really test to confirm adding a line here doesn't reintroduce any sound bugs as per the comment below.

soymonitus: Tested removing the padding and the release issue re-appears, so the padding is still needed. I see no other appreciable changes in sound, at least with the bass sound mentioned in the issue ticket.

soymonitus:
@seangoodvibes Ready for review

m-m-adams (approved):

m-m-adams (approved):

PR #3339, opened 2025-02-06
Enable clang-tidy and Ruff linting in Github Actions

stellar-aria:


stellar-aria:
Well that's going to get annoying real fast

sapphire-arches (approved):

PR #3383, opened 2025-02-15
always update velocity and octave spread

m-m-adams:


src/deluge/processing/sound/sound.cpp
```diff
@@ -2553,53 +2553,57 @@ void Sound::render(ModelStackWithThreeMainThings* modelStack, std::span<StereoSa
...
-		    (uint32_t)unpatchedParams->getValue(params::UNPATCHED_ARP_RATCHET_AMOUNT) + 2147483648;
-		arpSettings->noteProbability =
-		    (uint32_t)unpatchedParams->getValue(params::UNPATCHED_ARP_NOTE_PROBABILITY) + 2147483648;
-		arpSettings->bassProbability =
-		    (uint32_t)unpatchedParams->getValue(params::UNPATCHED_ARP_BASS_PROBABILITY) + 2147483648;
-		arpSettings->chordProbability =
-		    (uint32_t)unpatchedParams->getValue(params::UNPATCHED_ARP_CHORD_PROBABILITY) + 2147483648;
-		arpSettings->ratchetProbability =
-		    (uint32_t)unpatchedParams->getValue(params::UNPATCHED_ARP_RATCHET_PROBABILITY) + 2147483648;
+	if (arpSettings) {
```
stellar-aria, line 2556:
```suggestion
	if (arpSettings != nullptr) {
```

stellar-aria (approved):
My only nit, I've got another PR to reduce the magic numbers errors and cognitive complexity warnings

soymonitus:
I am not a reviewer and can't request changes but this change is not completely correct. SpreadOctave is not used outside of the arpeggiator. Only velocity is taken into account as a parameter for both arpeggiated and sequenced notes.

All the code in this part is already cleaned up and refactored in the "Reverse probability" PR. I also moved all that code inside a function of ArpeggiatorSettings to avoid code duplication (this piece chunk code is done in several places).

My suggestion is that this PR is not merged and instead, the Reverse Probability PR is reviewed

seangoodvibes (changes requested):
To avoid accidentally merging this:

Please see @soymonitus comment above

PR #3385, opened 2025-02-15
Align mod knob array to stop weird patch cable bug

m-m-adams:
Still don't know the cause but this seems to be a solution


stellar-aria:
Do we still need the temporary there too then?

stellar-aria (approved):

m-m-adams:
> Do we still need the temporary there too then?

Oops, I messed up resolving conflicts and accidentally added it back...

stellar-aria (approved):

PR #3387, opened 2025-02-15
Bugfix: Update midi follow feedback to check param context

seangoodvibes:
When midi feedback is sent while editing params with gold knobs or with select encoder while in the menu, it should check the context to ensure that feedback is not sent for params that do not match the current midi follow controllable context.

Fixes a bug in kits where when affect entire is enabled, it would send feedback if you were editing kit row params.

### DIFF
--- a/src/deluge/io/midi/midi_follow.cpp
+++ b/src/deluge/io/midi/midi_follow.cpp
@@ -552,21 +552,49 @@ void MidiFollow::displayParamControlError(int32_t soundParamId, int32_t globalPa
 	}
 }
 
-/// a parameter can be learned t one cc at a time
+/// a parameter can be learned to one cc at a time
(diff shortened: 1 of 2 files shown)


stellar-aria (changes requested):
Can you please amend the commit to not include the string_spec.cpp file? This will be `git reset HEAD~1` on this branch, and then `git reset -- tests/spec/string_spec.cpp`, `git commit -m "message"`

seangoodvibes:
> Can you please amend the commit to not include the string_spec.cpp file? This will be `git reset HEAD~1` on this branch, and then `git reset -- tests/spec/string_spec.cpp`, `git commit -m "message"`

Done :)

stellar-aria (approved):

PR #3391, opened 2025-02-16
Favourites Feature

busa-projects:
A Favourites-Feature has been added to the Load-UIs for most File-Types. The Favourites are displayed above the Keyboard and are only visible when the keyboard is shown. Favourites can be configured to either offer 16 Favourites (default), 16 Banks with 16 Favourites or be completely disabled via `SETTINGS > DEFAULTS -> UT -> KEYBOARD -> FAVOURITES`.
- In the `Favourites Only (16)`-Mode , 16 Favourite Slots are available at the Toprow above the Keyboard.
- In the `Favourites and Banks (256)`-Mode, the Toprow contains 16 Banks, each with 16 Favourites displayed on the second row.
Usage:
- Save a Favourite
  - Select the Bank (Only available in `Favourites and Banks (256)`-Mode)
  - Browse to the File you want to save
  - Hold `SHIFT` and select a empty Favourite Slot (light blue)
- Delete a Favourite
  - Select the Bank (Only available in `Favourites and Banks (256)`-Mode)
  - Hold `SHIFT` and select the Favourite Slot you want to delete
- Recall a Favourite
  - Select the Bank (Only available in `Favourites and Banks (256)`-Mode)
  - Select the Favourite
- Change the Colour of a Favourite
  - Select the Favourite you want to modify
  - Hold `SHIFT` and turn `VERTICAL ENCODER`

Tip:
If you are Browsing Songs or Samples an dont want to have the Preview to hide the Keyboard on Scrolling, just press the `KEYBOARD`-Button to pin the Keyboard and the Favourites.


alifeinbinary:
I think something worth considering is whether to make the favourites recallable through Program Change messages. I could see a purpose for it but I might be alone in that. Each group of 16 would be a sub-bank, each class would be a bank i.e. Songs on Bank 1, Synths on Bank 2, Kits on Bank 3 etc.

busa-projects:
Thanks for the suggestion. It is definitely worth considering. However, it is something that should be well thought through as programme change messages could also be used in other places. I am happy to take it on board as an idea for possible extensions.

src/deluge/model/favourite/favourite_manager.h
```diff
@@ -0,0 +1,68 @@
...
+	static constexpr uint8_t favouriteDefaultColor = 4;
+	void registerCallback(std::function<void()> callback);
+
+	uint8_t currentBankNumber;
+	std::optional<uint8_t> currentFavouriteNumber;
+
+private:
+	void loadFavouritesBank();
+	void saveFavouriteBank() const;
+	std::function<void()> onUpdateCallback;
```
m-m-adams, line 59: Have you considered any non-callback based implementations? It's a good architecture but it does make the code less beginner friendly. If you think this is the best way to go then I'm fine with it, just curious about alternatives

src/deluge/model/favourite/favourite_manager.h
```diff
@@ -0,0 +1,68 @@
...
+	void setFavorite(uint8_t position, uint8_t colour, const std::string& filename);
+	void unsetFavorite(uint8_t position);
+	bool isEmpty(uint8_t position) const;
+	Error loadFavouritesFromFile(Deserializer& reader);
+	void close();
+	std::array<std::optional<uint8_t>, 16> getFavouriteColours() const;
+	void changeColour(uint8_t position, int32_t offset);
+	const std::string& getFavoriteFilename(uint8_t position);
+	static constexpr uint8_t favouriteDefaultColor = 4;
+	void registerCallback(std::function<void()> callback);
```
m-m-adams, line 51: Can you add a doc comment on what the function should do?

m-m-adams (commented):
I generally think this is good to go. If you think we should keep the callback it just needs some docs explaining what the callback function should be implementing

I see a bunch of UI discussion in discord but that can be iterated on after merging

src/deluge/model/favourite/favourite_manager.h
busa-projects: i thought it was the cleanest way to update the UI, if the Model changes. But its doable without, as i now from user Actions, when the Model change and can trigger a update in the same user action. Refactored accrordingly

src/deluge/model/favourite/favourite_manager.h
busa-projects: remove Callback

m-m-adams (approved):

m-m-adams (approved):

PR #3394, opened 2025-02-18
Write old format note data [ONLY ON ALPHA/BETA], so songs can be loaded in 1.2 and below

soymonitus:
Do here an attempt to keep the most significant information, in order of priority
- 1st - Iterance preset
- 2nd - Fill/NotFill
- 3rd - Probability
- 4th - Custom iterance (since this does not exist in 1.2 it will be always discarded)


m-m-adams:
Could we put this behind the alpha/beta define and bump the minimum version? I'm ok with using it to make nightly testing easier but I really don't want this to get in the final release, we should be striving to make sure people don't need to downgrade and ensuring the documentation is clear that will be the case. If we bump the minimum version people can manually change their songs if needed and it'll be more explicit that we don't support it

soymonitus:
@m-m-adams I added the alpha macro tag for both key pieces of code that could be removed once we get 1.3 out

m-m-adams (approved):

PR #3404, opened 2025-02-21
[DRAFT] Feature/ Implement sustain and sostenuto pedals

topisani:
Both pedals are added as unpatched params, and then handled in the individual voices. I am not entirely certain how this will interact with MPE, but i assume they arent often used much together anyway. If someone can help test this combo, please do.

Currently, the pedal states are only accessible in the automation view, and through midi, where the CCs 64 and 66 are used as per midi spec. 

Note:
Currently, there is no existing mechanism to disable automation interpolation for some parameters. This may cause issues when automating the pedals.

Its been a while since ive looked at the deluge firmware, so I may very likely have missed something.


seangoodvibes:
@topisani could you add documentation for this feature please to our community features and change log documents?

thanks!

src/deluge/model/voice/voice.cpp
```diff
@@ -569,13 +572,27 @@ void Voice::calculatePhaseIncrements(ModelStackWithVoice* modelStack) {
 
 void Voice::noteOff(ModelStackWithVoice* modelStack, bool allowReleaseStage) {
 
+	ParamManagerForTimeline* paramManager = (ParamManagerForTimeline*)modelStack->paramManager;
+	Sound& sound = *static_cast<Sound*>(modelStack->modControllable);
+
+	bool sustainPedal = paramManager->getUnpatchedParamSet()->getValue(params::UNPATCHED_PEDAL_SUSTAIN) >= 0;
+	bool sostenutoPedal = paramManager->getUnpatchedParamSet()->getValue(params::UNPATCHED_PEDAL_SOSTENUTO) >= 0;
```
soymonitus, line 579: Are you sure this is right? I think deluge params have the full integer range, meaning that a value of 0 is in fact -2147483648 (search for this magic number in the code and you will see what I mean

src/deluge/model/voice/voice.cpp
```diff
@@ -725,6 +742,14 @@ uint32_t Voice::getLocalLFOPhaseIncrement(LFO_ID lfoId, deluge::modulation::para
 
 	bool didStereoTempBuffer = false;
 
+	if (sustainedByPedal) {
+		// noteOff does nothing if the relevant pedal is still held, so it is safe to call here.
+		noteOff(modelStack);
+	}
+	if (paramManager->getUnpatchedParamSet()->getValue(params::UNPATCHED_PEDAL_SOSTENUTO) < 0) {
```
soymonitus, line 749: Same as my other comment. And why < 0?

soymonitus (commented):
Left a couple of comments

src/deluge/model/voice/voice.cpp
topisani: Yep, that's what this is for - according to the midi spec,  everything below 64 counts as off, so that's from -2^31 to 0

topisani:
I've been busy with other things, but I will get back to this PR. I've found an issue with the way it interacts with legato mode, but that should be a fairly easy fix.

topisani:
> What is the exact problem you are having with the legato mode? Maybe some dev can continue your work if you don't have time at the moment, so it would be good to have a list of known issues to address

Right, If someone has time to take this the final step i would really appreciate it. My deluge is currently packed away for moving.

IIRC, the issue with legato is that noteOn/noteOff are not called. This at least breaks sostenuto, and possibly also sustain, i dont fully remember. With sostenuto it would be hard to get working, but the usecase is probably also not that likely, so it is mostly about making sure it breaks in a well-defined way (i.e. without leaving stuck notes).

Testing should be fairly easy, use a legato synth and control the sustain/sostenuto CCs, testing the various orderings of note/pedal on/off events.

src/deluge/model/voice/voice.cpp
dctucker:
```suggestion
	if (paramManager->getUnpatchedParamSet()->getValue(params::UNPATCHED_PEDAL_SOSTENUTO) <= 0) {
```

PR #3411, opened 2025-02-23
Remove AlignedBuf and replace with alignas(n) std::array

stellar-aria:


src/deluge/dsp/dx/fm_core.h
```diff
@@ -59,5 +59,5 @@ class FmCore {
 	bool neon = false;
 
 protected:
-	AlignedBuf<int32_t, DX_MAX_N> buf_[2];
+	alignas(16) std::array<int32_t, DX_MAX_N> buf_[2];
 };
```
sapphire-arches, line 63: is this like cache line size /4?

stellar-aria: /2 actually, since AlignedBuf's default for the template argument for alignment was 16

m-m-adams (approved):

PR #3415, opened 2025-02-24
Remove intrusive lists from Voice, VoiceSample, and TimeStretcher

stellar-aria:


m-m-adams (commented):
Should we just go straight to the object pool instead?

stellar-aria:
Potentially, yeah

m-m-adams (approved):

PR #3416, opened 2025-02-25
[Draft] Slicer Multisample (HELP NEEDED!)

soymonitus:
This new feature is the ability to slice a single sample and use its slices as input for the multisample code to create a new synth.

A new menu item is added to the context menu when long-pressing select on a sample for a new synth. You will see options Multisample, **Slice Multi**, Basic, Wavetable, SCW.
When "Slice Multi" is selected, the  "slicer" UI is opened where you can select the number of same-length slices (auto) or push <> to do the slicing manually. After you confirm pressing Select I call the function importSampleAsSlicedMultisample from SampleBrowser, passing the list of slice startPos.

**HELP NEEDED:**
at this point my code is not working properly, as I get memory bad-access crashes, related to using the same sample. The Sample class has a midiNote property used by the detectPitch function, but as we only have one sample I have several approaches, one was copying in memory the same Sample numSlices times, another would be to create a "sortArea" as an array of a new class that could be called SampleSlice, that contains a ref to the Sample, midiNote property and the startPos of the slicer.
For the pitch detection I discovered that there is a property called beginningOffsetForPitchDetection in Sample that might be useful for telling it to start detecting pitch at a certain position.

So I left my work at this point as I was getting crashes that I cannot debug as I don't have a debugger. Hopefully someone can take my work and continue it


src/deluge/gui/ui/browser/sample_browser.cpp
```diff
@@ -1837,6 +1826,92 @@ bool SampleBrowser::importFolderAsMultisamples() {
 	return true;
 }
 
+bool SampleBrowser::importSampleAsSlicedMultisample(int32_t numClips, SliceItem* slicePoints) {
```
soymonitus, line 1829: This is where I left. This entry point is where, given the slice points, you need to take the currently selected sample from the browser and create the MultisampleRange objects for the synth.

PR #3425, opened 2025-02-28
Feature / Add support for affecting-all kit rows while in the sound editor by holding Affect-Entire button

soymonitus:
Just like what happens with Sample Mode, Reverse, Pitch/Speed... we want to be able to batch change all kit rows at the same time. This works by holding the affect-entire button (it will flash indicating that this action is available for this parameter), and when you edit the parameter, it will apply the same value to all kit rows within the current clip.
All Sound parameters are implemented except for Oscillator Type and patch cables.
Additional features: batch editing a param for all kit rows while record engaged will record automation for all rows!

Considerations:
- Although adding support for all patched and unpatched params at once is just a handful of lines of code (I hooked my code in the same place as midi follow feedback does), I could not come up with a way to reduce code for the other non-automatable params, as each has its own custom logic.
- Here is a comparison in binary size of current community vs this PR:
Community: 1,581,892 bytes
This PR: 1,585,772 bytes

Things this feature will enable:
- Quickly change any parameter for all rows like Delay Sync, because we dub fans all love dotted eights
- Come up with happy accidents by batch changing some parameter more aimed at sound design
- Use it as a performance tool: suddenly enable/disable arpeggiator, or increase some arp randomizer amount for all rows, or tighten the sound by decreasing the decay of all envelopes 1's.
- Transpose all rows by semitones (different than Master Transpose)


soymonitus:
Ready for review

src/deluge/gui/menu_item/arpeggiator/chord_type.h
```diff
@@ -20,21 +20,45 @@
...
+#include "processing/sound/sound.h"
 #include <cstdint>
 
 namespace deluge::gui::menu_item::arpeggiator {
 class ChordType : public Selection {
 public:
 	using Selection::Selection;
 	void readCurrentValue() override { this->setValue(soundEditor.currentArpSettings->chordTypeIndex); }
+
+	bool usesAffectEntire() override { return true; }
```
soymonitus, line 35: This is the virtual method used to tell the sound editor that this parameter supports holding affect-entire button when in kit to do batch edits. The line
`bool usesAffectEntire() override { return true; }`
has been added to all parameters that have been modified in this PR, to make them batch-edit-enabled

src/deluge/gui/menu_item/arpeggiator/chord_type.h
```diff
@@ -20,21 +20,45 @@
...
-		int32_t value = this->getValue();
-		if (value >= 0 && value < MAX_CHORD_TYPES) {
-			soundEditor.currentArpSettings->chordTypeIndex = value;
+		int32_t current_value = this->getValue();
+		if (current_value < 0 || current_value >= MAX_CHORD_TYPES) {
+			return;
+		}
+
+		// If affect-entire button held, do whole kit
+		if (currentUIMode == UI_MODE_HOLDING_AFFECT_ENTIRE_IN_SOUND_EDITOR && soundEditor.editingKitRow()) {
```
soymonitus, line 43: You will find in all the parameters that have been edited in this PR that a new if-else condition is added, mimicking the ones which were already there in the code for sample/reverse.h and sample/repeat.h for example. Basically this is:
```
if (currentUIMode == UI_MODE_HOLDING_AFFECT_ENTIRE_IN_SOUND_EDITOR && soundEditor.editingKitRow()) {
    // Loop through all kit rows and apply the change to each one (only if they meet the requirements stated by isRelevant)
} else {
    // Just edit the current selected row
}

```

src/deluge/model/model_stack.cpp
```diff
@@ -108,6 +109,14 @@ int32_t ModelStackWithNoteRow::getRepeatCount() const {
 	}
 }
 
+ModelStackWithThreeMainThings* getModelStackFromSoundDrum(void* memory, SoundDrum* soundDrum) {
```
soymonitus, line 112: A new utility method has been added to be able to get a modelstack from a sounddrum kit row, to avoid code repetition

src/deluge/gui/menu_item/arpeggiator/mode.h
```diff
@@ -31,36 +31,60 @@ class Mode final : public Selection {
...
-					}
-				}
-				else if (soundEditor.editingCVOrMIDIClip()) {
-					getCurrentInstrumentClip()->stopAllNotesForMIDIOrCV(modelStack->toWithTimelineCounter());
+			Kit* kit = getCurrentKit();
+
+			// If was off, or is now becoming off...
+			if (soundEditor.currentArpSettings->mode == ArpMode::OFF || current_value == ArpMode::OFF) {
+				if (getCurrentClip()->isActiveOnOutput() && !soundEditor.editingKitAffectEntire()) {
+					kit->cutAllSound();
```
soymonitus, line 47: Some parameters execute additional logic when edited. In this case, when the arp mode was changed just for one row, it called `currentSound->unassignAllVoices()` , so here when batch editing all arp modes of all kit rows, we have to do the same. Luckily there is already a method in kit that cuts all voices for all rows so we don't have to call unassignAllVoices per row.

Same logic applies to other parameters, if there is some extra work to do before or after editing the parameter, we have to do the equivalent but for all the rows

src/deluge/gui/menu_item/audio_compressor/compressor_values.h
```diff
@@ -21,10 +27,27 @@ class CompressorValue : public DecimalWithoutScrolling {
...
+			}
+		}
+		// Or, the normal case of just one sound
+		else {
+			setCompressorValue(knobPos, &soundEditor.currentModControllable->compressor);
+		}
 	}
 	virtual uint64_t getCompressorValue() = 0;
-	virtual void setCompressorValue(q31_t value) = 0;
+	virtual void setCompressorValue(q31_t value, RMSFeedbackCompressor* compressor) = 0;
```
soymonitus, line 50: Use a specific compressor instead of the one from selected kit row, so this method is more flexible and we can use it for the batch edit of other rows

src/deluge/gui/menu_item/filter/info.h
```diff
@@ -55,17 +56,18 @@ class FilterInfo {
 			return std::min(selection, kNumLPFModes);
 		}
 	}
-	void setMode(int32_t value) const {
+	void setMode(int32_t value) const { setModeForModControllable(value, soundEditor.currentModControllable); }
+	void setModeForModControllable(int32_t value, ModControllableAudio* modControllable) const {
```
soymonitus, line 60: Same here as with compressor, modify it so we can edit any modControllable and not only the one from selected kit row

src/deluge/gui/menu_item/sample/transpose.h
```diff
@@ -44,25 +50,58 @@ class Transpose final : public source::Transpose, public FormattedTitle {
...
+	bool usesAffectEntire() override { return true; }
 	void writeCurrentValue() override {
 		int32_t transpose, cents;
 		computeFinalValuesForTranspose(this->getValue(), &transpose, &cents);
 
-		if ((soundEditor.currentMultiRange != nullptr) && soundEditor.currentSound->getSynthMode() != SynthMode::FM
-		    && soundEditor.currentSource->oscType == OscType::SAMPLE) {
-			(static_cast<MultisampleRange*>(soundEditor.currentMultiRange))->sampleHolder.transpose = transpose;
-			(static_cast<MultisampleRange*>(soundEditor.currentMultiRange))->sampleHolder.setCents(cents);
+		// If affect-entire button held, do whole kit
```
soymonitus, line 58: The transpose parameters were actually the most complex parameters to make compatible with batch-edit, because you have to carefully take the multsample range and whatnot. But anyway not a big deal after all.

soymonitus (commented):
@stellar-aria Added comments to my own PR to explain some things. Basically what this huge PR is about is to take all parameters of a sound and add to them the same if-else that sample reverse, or sample repeat mode already had, that is, check if we are in a kit and holding affect-entire, then do the edit to all rows, and if not, then just edit the selected row. The only complexity here is that some parameters executed some additional logic before or after editing the parameter (for example LFOs: they resync the lfos after editing them), so I had to carefully do the same pre/post logic for all kit rows

PR #3431, opened 2025-03-02
Begin FixedPoint conversion

stellar-aria:
Starts with the RMSCompressor

### DIFF
--- a/src/deluge/dsp/compressor/rms_feedback.h
+++ b/src/deluge/dsp/compressor/rms_feedback.h
@@ -55,7 +55,7 @@ public:
 	/// @param volAdjustL Linear gain to apply to the left channel as a 4.27 signed fixed point number.
 	/// @param volAdjustL Linear gain to apply to the right channel as a 4.27 signed fixed point number.
 	/// @param finalVolume Linear peak-to-peak volume scale, as a 3.29 fixed-point integer.
-	void render(std::span<StereoSample> buffer, q31_t volAdjustL, q31_t volAdjustR, q31_t finalVolume);
+	void render(std::span<StereoSampleFixed> buffer, q31_t volAdjustL, q31_t volAdjustR, q31_t finalVolume);
 
 	/// Render the compressor with neutral left/right gain and with the finalVolume tweaked so the compressor applies
 	/// 0db gain change at theshold zero. Used by the per-clip compressors because the clip volume is applied without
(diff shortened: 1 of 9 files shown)


src/deluge/dsp/compressor/rms_feedback.h
```diff
@@ -198,10 +204,10 @@ class [[gnu::hot]] RMSFeedbackCompressor {
 
 	/// State for the internal envelope follower
 	float state = 0;
-	/// Current left channel volume as a 5.26 signed fixed-point number
-	q31_t currentVolumeL = 0;
-	/// Current right channel volume as a 5.26 signed fixed-point number
-	q31_t currentVolumeR = 0;
+	/// Current left channel volume
+	FixedPoint<28> currentVolumeL = 0.f;
```
m-m-adams, line 208: This is a change from the comment above - I dunno if the comment is right though

src/deluge/gui/menu_item/audio_compressor/compressor_values.h
```diff
@@ -75,11 +75,10 @@ class Blend final : public Integer {
 	void writeCurrentValue() override {
 		auto value = this->getValue();
 		if (value < kMaxKnobPos) {
-			q31_t knobPos = lshiftAndSaturate<24>(value);
-			soundEditor.currentModControllable->compressor.setBlend(knobPos);
+			soundEditor.currentModControllable->compressor.setBlend(FixedPoint<7>::from_raw(value));
```
m-m-adams, line 78: I guess this automatically converts up to 31?

src/deluge/util/fixedpoint.h
```diff
@@ -578,9 +597,48 @@ class FixedPoint {
...
+
+	[[nodiscard]] constexpr int32_t integral() const { return static_cast<int32_t>(value_ >> fractional_bits); }
+
 private:
 	BaseType value_{0};
 };
+
+/// @brief Multiply by an integral type
+template <std::integral T, std::size_t FractionalBits, bool Rounded, bool FastApproximation>
+constexpr FixedPoint<FractionalBits, Rounded, FastApproximation>
```
m-m-adams, line 623: Do these need to be templated on the fractional bits?

stellar-aria: yeah, it allows compile-time calculation of bitshifts and stuff, which are required for some operations like ssat<bit>()

m-m-adams: Don't these templates  all do the same thing regardless of the number of  bits? I saw some of the ones above are only templated on the LHS and not the specific FixedPoint type and I think that would be better

stellar-aria: the ones above are not global. These match only a FixedPoint on the RHS, which is why all the template parameters are required

stellar-aria: the ones above are not global. These match only a FixedPoint on the RHS, which is why all the template parameters are required

stellar-aria: The only other way to do this is extremely messy: a set of template template parameters, with a same_as_v restriction on the template instance type

stellar-aria: in truth, the ones in the class declaration are also overloaded on those template parameters, it's just at a class level instead of a function level, and then aliased so they don't appear in the signature

PR #3445, opened 2025-03-04
Remove aligned buf (take 2)

stellar-aria:


m-m-adams (approved):
Looks fine as long as you're sure dx7 patches still sound the same

PR #3458, opened 2025-03-10
Move Song default initializers to the member declaration

stellar-aria:


sapphire-arches (approved):
This could benefit from a pass over the comments to clean them up (make the proper doc comments, delete things like the old "0 = patched,..."note on `lastSelectedParamKind`. Otherwise LGTM.

stellar-aria:
I'll go through and do that when I touch song next, which will be soon with the output refactorings

PR #3464, opened 2025-03-10
Set stutter default to quantized

seangoodvibes:
Updated stutter config so that it is set to quantized by default


soymonitus (approved):
Looks good to me!

PR #3489, opened 2025-03-20
Adding option to choose sharps or flats for note display

jkschwartz:
Add Default UI setting to allow for notes to be displayed with either sharps or flats

with the setting enabled: on oled displays C, DB, D, EB, E, F etc
on 7SEG: C, D., D, E., F, etc


jkschwartz:
Still need to make a few changes. Most importantly, I'd like to incorporate the proper symbol for flats from https://github.com/SynthstromAudible/DelugeFirmware/pull/3243. I also should do some clean up so I don't repeat myself so often.

soymonitus:
I think the default option now is different from stock, isn't it? If so, Could you invert the flash storage option?

stellar-aria:
@jkschwartz I'm going to mark this as draft, when you're ready for review you can switch it back

jkschwartz:
borked this up, closing this PR and will reopen once fixed.

PR #3495, opened 2025-03-21
Fix select encoder infinite loop in audio source menu

seangoodvibes:
Fixed an infinite menu loop that would occur when releasing select encoder in the audio clip input source menu


m-m-adams:
This doesn't actually fix the crash, it fixes an infinite loop that occurs if the crash doesn't occur. Release builds still crash

seangoodvibes:
> This doesn't actually fix the crash, it fixes an infinite loop that occurs if the crash doesn't occur. Release builds still crash

Weird...my release build doesn't crash.

m-m-adams (approved):

PR #3503, opened 2025-03-21
Fix culling priority selection

stellar-aria:


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -300,15 +301,21 @@ void terminateOneVoice(size_t numSamples) {
...
 
-	auto& voice = *std::ranges::min_element(all_voices, [](const auto& best, const auto& voice) {
+	const Sound::ActiveVoice* best = nullptr;
+	for (const auto& voice : all_voices) {
 		// if we're not skipping releasing voices, or if we are and this one isn't in fast release
-		if (voice->envelopes[0].state <= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement < SOFT_CULL_INCREMENT) {
-			return voice->getPriorityRating() > best->getPriorityRating();
+		if (voice->envelopes[0].state > EnvelopeStage::FAST_RELEASE
+		    || voice->envelopes[0].fastReleaseIncrement >= SOFT_CULL_INCREMENT) {
```
m-m-adams, line 308: This should be &&, fast release increment is set when the note enters fast release. The constructor initializes it below the fast release increment but that's probably not good to rely on

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -325,13 +332,21 @@ void forceReleaseOneVoice(size_t num_samples) {
...
 
-	auto& voice = *std::ranges::min_element(all_voices, [](const auto& best, const auto& voice) {
-		if (voice->envelopes[0].state <= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement < 4096) {
-			return voice->getPriorityRating() > best->getPriorityRating();
+	const Sound::ActiveVoice* best = nullptr;
+	for (const auto& voice : all_voices) {
+		// if we're not skipping releasing voices, or if we are and this one isn't in fast release
+		if (voice->envelopes[0].state > EnvelopeStage::FAST_RELEASE
+		    || voice->envelopes[0].fastReleaseIncrement >= 4096) {
```
m-m-adams, line 339: Same here but I think this is the initial increment so all voices meet this criteria

src/deluge/processing/sound/sound.cpp
```diff
@@ -4855,14 +4857,21 @@ void Sound::terminateOneActiveVoice() {
...
 
-	const ActiveVoice& voice = *std::ranges::min_element(voices_, [](const auto& best, const auto& voice) {
+	ActiveVoice* best = nullptr;
+	for (ActiveVoice& voice : voices_) {
 		// if we're not skipping releasing voices, or if we are and this one isn't in fast release
-		if (voice->envelopes[0].state <= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement < SOFT_CULL_INCREMENT) {
-			return voice->getPriorityRating() > best->getPriorityRating();
+		if (voice->envelopes[0].state > EnvelopeStage::FAST_RELEASE
+		    || voice->envelopes[0].fastReleaseIncrement >= SOFT_CULL_INCREMENT) {
```
m-m-adams, line 4864: Ditto

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -300,15 +301,21 @@ void terminateOneVoice(size_t numSamples) {
...
 	}
 
-	auto& voice = *std::ranges::min_element(all_voices, [](const auto& best, const auto& voice) {
+	const Sound::ActiveVoice* best = nullptr;
+	for (const auto& voice : all_voices) {
 		// if we're not skipping releasing voices, or if we are and this one isn't in fast release
-		if (voice->envelopes[0].state <= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement < SOFT_CULL_INCREMENT) {
-			return voice->getPriorityRating() > best->getPriorityRating();
+		if (voice->envelopes[0].state > EnvelopeStage::FAST_RELEASE
```
m-m-adams, line 307: it'll never be > since the next stage is off 

I think the logic change here is wrong and the previous behaviour is correct. We're looking for a voice which is either 

1) not fast releasing, or 
2) fast releasing slower than we want 

This functionally removes the first check and relies on the initializer value for fast release increment being low

src/deluge/processing/sound/sound.cpp
```diff
@@ -4876,14 +4885,21 @@ void Sound::forceReleaseOneActiveVoice() {
...
 
-	const ActiveVoice& voice = *std::ranges::min_element(voices_, [](const auto& best, const auto& voice) {
+	ActiveVoice* best = nullptr;
+	for (ActiveVoice& voice : voices_) {
 		// if we're not skipping releasing voices, or if we are and this one isn't in fast release
-		if (voice->envelopes[0].state <= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement < 4096) {
-			return voice->getPriorityRating() > best->getPriorityRating();
+		if (voice->envelopes[0].state > EnvelopeStage::FAST_RELEASE
+		    || voice->envelopes[0].fastReleaseIncrement >= 4096) {
```
m-m-adams, line 4892: Ditto

m-m-adams (approved):

soymonitus:
Crashes I was having when notes were culled are now gone. I found another crash when loading songs but found that it reproduces also in nightly so not related to this PR

PR #3506, opened 2025-03-23
Updated horizontal menu controls

seangoodvibes:
Updated horizontal menu controls to remove "holding shift" as a modifier for select encoder behavioural changes

Updated description:

- Horizontal Menu controls. There are two different behaviours that can be toggled between to edit values and select menu items within the horizontal menu. You can toggle between them by going to the `SETTINGS > DEFAULTS -> UI -> HORIZONTAL MENU (HORZ)` and toggling `Alternative Select Behaviour (SELE)` on or off.
  - Note: regardless of the behaviour toggle selected, you will always be able to change the selected horizontal menu's item's value by turning the select encoder when you're holding audition pad or sticky shift is enabled.
    - Whenever you're holding audition pad or sticky shift is enabled, you can change the selected horizontal menu by `Pressing + Turning the select encoder`
  - If you're not holding an audition pad or sticky shift is disabled, then the following select behaviours apply.
  - With `Alternative Select Behaviour (SELE)` DISABLED:
    - `Turn select encoder` to change the selected menu item
    - `Press + Turn select encoder` to change the value of the selected menu item
  - With `Alternative Select Behaviour (SELE)` ENABLED:
    - `Turn select encoder` to change the value of the selected menu item
    - `Press + Turn select encoder` to change the selected menu item

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,17 +34,15 @@ set the stutter configuration independently of the song configuration if you wis
   - LFO 1-4.
   - Arpegiattor
 - Horizontal Menu controls. There are two different behaviours that can be toggled between to edit values and select menu items within the horizontal menu. You can toggle between them by going to the `SETTINGS > DEFAULTS -> UI -> HORIZONTAL MENU (HORZ)` and toggling `Alternative Select Behaviour (SELE)` on or off.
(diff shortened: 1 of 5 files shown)


soymonitus (approved):
looks good to me

PR #3519, opened 2025-03-26
Oscillator Refactor (1/?)

stellar-aria:
Begins the process of refactoring the Oscillator system.

- All oscillator types split to individual render functions.
- Sync versions of oscillators also split out.
- Introduces the new `deluge::dsp` Core-API which includes:
  - Generators (no input, one output)
  - Processors (one input, one output)
  - Mixers (two inputs, one output)
  - Pipelines (static composition of multiple processors and/or a generator)


m-m-adams:
I haven't gotten into the code yet but I'm noticing some artifacts just playing through my songs and the problem songs folder:

1. Basic waves and wavetable oscillators click when they start  playing and get a "tearing" sound when the deluge is at a high load
2. There's noticeably increased aliasing noise in sample playback - check Moulfrit/Melolpartage or Glitchy 2. Seems to be related to pitch shifting
3. Performance is  reduced for sample and wavetable oscillators - Glitchy 2 now frequenty falls to 2 voices, MS Noise Bug needs 7 notes removed to get rid of the tearing artifacts 
4. Audio clips have a click whenever they start or stop timestretching

PR #3521, opened 2025-03-26
Migrate Stealables to etl::intrusive_list

stellar-aria:
Removes custom BidirectionalLinkedList in favor of etl::intrusive_list.

SampleCache logic is the fragile point, I think I migrated it properly, but it needs a double check.


stellar-aria:
looks like we'll need to add ETL to the unit testing suite

stellar-aria:
Well that's not a good sign

m-m-adams (approved):

seangoodvibes (approved):

PR #3535, opened 2025-03-31
Create new DSP core API

stellar-aria:
Extracts all the added files from #3519 into their own PR. That will get rebased on top of this once merged. This does not change any generated code in the codebase


src/deluge/dsp/CMakeLists.txt
```diff
@@ -24,13 +24,20 @@ set_target_properties(deluge_dsp
         CXX_STANDARD ${DELUGE_CXX_STANDARD}
         CXX_STANDARD_REQUIRED ON
         CXX_EXTENSIONS ON
-        UNITY_BUILD true
+        UNITY_BUILD ON
 )
 
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  set_target_properties(deluge_dsp PROPERTIES 
+    UNITY_BUILD OFF
```
m-m-adams, line 32: why the change? It's got a big impact on debug performance without affecting debugging as far as I've noticed

src/deluge/dsp/core/chain.h
```diff
@@ -0,0 +1,39 @@
...
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+#include "dsp/core/generator.h"
+#include "dsp/core/types.h"
+
+namespace deluge::dsp {
+template <typename T>
+class Chain : public BlockProcessor<T> {
```
m-m-adams, line 24: For this one I understand what it's doing but a doc comment on why would be helpful

src/deluge/dsp/core/concepts.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+template <typename T>
+concept sample_converter = requires(T t, typename T::value_type sample) {
+	{ t.render(sample) };
+};
+
+/// @brief A concept for processors that render a block of samples using SIMD operations.
+template <typename T>
+concept block_adapter =
```
m-m-adams, line 54: what makes this one different from block_processor? 

Actually the ones below also look the same to me so I might be missing something

src/deluge/dsp/core/conditional.h
```diff
@@ -0,0 +1,176 @@
...
+	using value_type = typename std::remove_pointer_t<ProcessorType>::value_type;
+
+	Conditional(CondType condition, ProcessorType processor) : condition_(condition), processor_(processor) {}
+	Conditional(CondType condition, ProcessorType processor, ProcessorType elseProcessor)
+	    : condition_(condition), processor_(processor), elseProcessor_(elseProcessor) {}
+
+	/// @brief Process a block of samples using SIMD operations, conditionally applying processors.
+	/// @param sample The input sample to process.
+	/// @return The processed sample.
+	[[gnu::always_inline]] value_type render(value_type sample) final {
```
m-m-adams, line 54: I think this should be on the render block function instead of render. Per sample ifs get hoisted out of the loop inconsistently. (same for remainder of this file)

src/deluge/dsp/core/converter.h
```diff
@@ -0,0 +1,128 @@
...
+	/// @param sample The input sample of type U to convert.
+	/// @return The converted sample of type T.
+	virtual T render(U sample) = 0;
+
+	/// @brief Convert a block of type U to type T by calling render() for each sample.
+	/// @param input The input buffer of type U to convert.
+	/// @param output The output buffer of type T to fill with converted samples.
+	void renderBlock(Buffer<U> input, Buffer<T> output) final {
+		for (size_t i = 0; i < input.size(); ++i) {
+			output[i] = render(input[i]); // Call the process function for each sample
```
m-m-adams, line 70: Because of the conditional class I don't think this will work well - it requires them to check their condition on each sample

src/deluge/dsp/core/generator.h
```diff
@@ -0,0 +1,74 @@
...
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+#pragma once
+#include "processor.h"
+#include <algorithm>
+#include <argon.hpp>
+#include <span>
+
+namespace deluge::dsp {
+/// @brief A base class for generators that process a stream of samples.
```
m-m-adams, line 24: generate not process right?

src/deluge/dsp/CMakeLists.txt
stellar-aria: clangd can't do semantic analysis or intellisense due to the way that compile commands shoves all the files together: they don't show up as actually compiled, so tag parsers and semantic analyzers don't know they're used in the binary.

src/deluge/dsp/core/concepts.h
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+#include <span>
+
+namespace deluge::dsp {
+/// @brief A concept for generators that render a block of samples.
```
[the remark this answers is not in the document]
stellar-aria: It's not specified because you can pass the same range/span to both input and output and it'll process in-place.

src/deluge/dsp/core/concepts.h
stellar-aria: It's not now, you're right

src/deluge/dsp/core/conditional.h
stellar-aria: That would be a different class then, a ConditionalBlockProcessor.

src/deluge/dsp/core/adapter.h
```diff
@@ -0,0 +1,85 @@
...
+/// "inner" type and an "outer" type. This allows for floating-point processors in fixed-point pipelines
+/// and vice versa.
+
+namespace deluge::dsp {
+/// @brief A class that adapts a processor to work in a pipeline of a different type.
+/// @tparam T The type of the input samples.
+/// @tparam ProcessorType The type of the processor to be wrapped by the adapter.
+template <typename T, typename ProcessorType>
+class Adapter : public Converter<T, typename std::remove_pointer_t<ProcessorType>::value_type>, public Processor<T> {
+public:
```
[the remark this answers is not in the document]
stellar-aria: remove_pointer_t removes the pointer qualification so that you can pass either a pointer or a non-pointer type and still get the underlying class for accessing static members

src/deluge/dsp/core/converter.h
stellar-aria: See previous comment about ConditionalBlockProcessor

src/deluge/dsp/core/conditional.h
```diff
@@ -0,0 +1,176 @@
...
+		if (condition_()) {                               // Check the condition
+			util::dereference(processor_).render(sample); // Apply the processor
+		}
+		if (elseProcessor_) {                                  // Check if else processor exists
```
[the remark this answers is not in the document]
stellar-aria: This will double the number of specializations, but I'll do it lol

src/deluge/dsp/core/pipeline.h
```diff
@@ -0,0 +1,33 @@
...
+// You should have received a copy of the GNU General Public License along with this program.
+// If not, see <https://www.gnu.org/licenses/>.
+
+#pragma once
+#include <dsp/core/processor.h>
+#include <queue>
+
+/// A DSP pipeline is a collection of processors that are executed concurrently (whether threaded or SIMD). This is done
+/// by introducing a single sample delay. For N processors there are N stored samples. The first processor is fed the
+/// input sample, the second processor is fed the previous output sample of the first processor, and so on.
```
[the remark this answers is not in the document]
stellar-aria: Pipeline is different from chain, which is different from sequence. 

Sequence is what the previous pipeline model was: a composition of sequential single-sample processors into an efficient block processor.
Chain is a composition of multiple block processors
Pipeline is a way to take a single parallel block processor (which would normally operate on multiple blocks simultaneously) and make it operate on a single block

src/deluge/dsp/core/converter.h
stellar-aria: ConditionalProcessor has been removed

m-m-adams (commented):
Could you leave the filters out? This is getting too big to review again

src/deluge/dsp/core/pipeline.h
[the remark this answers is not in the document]
m-m-adams: Ok - could you test that on our hardware and include benchmarks to confirm? Typically performing a single operation repeatedly will cache better than performing several different ones. I suppose it matters what you're imaging a single render call to be - I'm picturing something like a full filter or oscillator render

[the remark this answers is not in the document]
m-m-adams: good to go - misunderstanding of how big a single render function was going to be in this case

src/deluge/dsp/core/processor.h
```diff
@@ -0,0 +1,108 @@
...
+	/// @param output The output buffer to fill with processed samples.
+	virtual void renderBlock(Signal<T> input, Buffer<T> output) = 0;
+};
+
+/// @brief A base class for processors that process a single sample at a time.
+/// @note Only inherit from this class if you plan on also inheriting from BlockProcessor and providing a custom
+/// implementation of renderBlock().
+/// @tparam T The type of the samples to process.
+template <typename T>
+struct SampleProcessor {
```
m-m-adams, line 45: Should this just be merged with BlockProcessor if they need to be used together?

stellar-aria: They don't, you can have a BlockProcessor that cannot be a SampleProcessor.

m-m-adams: But you can't have a SampleProcessor that's not a BlockProcessor so why not inherit?

stellar-aria: It comes down to overload resolution and compile-time optimizations: renderBlock is final so that it can be inlined, but some processors might want to define their own renderBlock, but can't due to its state as final

stellar-aria: oh wait I think I see what you're saying, SampleProcessor as a refinement of BlockProcessor. Hm.

stellar-aria: Yeah I can probably do that, let me give it a try

src/deluge/util/fixedpoint.h
```diff
@@ -123,24 +125,23 @@ class FixedPoint {
 	/// @brief Convert from a float to a fixed point number
 	/// @note VFP instruction - 1 cycle for issue, 4 cycles result latency
 	[[gnu::always_inline]] constexpr FixedPoint(float value) noexcept {
+#if __ARM_ARCH_7A__ && !defined(__clang__)
```
m-m-adams, line 128: Why the change back to defines?

stellar-aria: GCC x86_64 can't even properly parse the inline asm with ARM-specific register constraints (namely w)

m-m-adams (approved):

PR #3567, opened 2025-04-07
Change horizontal menu default select encoder behaviour

seangoodvibes:
Changed default horizontal menu select encoder behaviour.

`Alternative Select Behaviour (SELE)` in `SETTINGS > DEFAULTS -> UI -> HORIZONTAL MENU (HORZ)` is now enabled by default.


soymonitus (approved):
LGTM

PR #3569, opened 2025-04-08
Bugfix/unreachable perc cache code

m-m-adams:
I'm not certain what was wrong with the ternary but it lead the compiler to believe that reversed is always false and optimize out huge parts of the perc cache code. With this fix it now seems to properly find and mark percussive points and pitch shifting is a lot better

### DIFF
--- a/src/deluge/model/sample/sample.cpp
+++ b/src/deluge/model/sample/sample.cpp
@@ -274,8 +274,7 @@ Error Sample::fillPercCache(TimeStretcher* timeStretcher, int32_t startPosSample
 	uint16_t startTime = MTU2.TCNT_0;
 #endif
 
-	int32_t reversed = (playDirection == 1) ? 0 : 1;
-
+	auto reversed = static_cast<int32_t>(playDirection != 1);
 	// If the start pos is already beyond the waveform, we can get out right now!
 	if (!reversed) {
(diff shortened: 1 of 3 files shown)


src/deluge/model/sample/sample.cpp
```diff
@@ -274,8 +274,7 @@ Error Sample::fillPercCache(TimeStretcher* timeStretcher, int32_t startPosSample
 	uint16_t startTime = MTU2.TCNT_0;
 #endif
 
-	int32_t reversed = (playDirection == 1) ? 0 : 1;
-
+	auto reversed = static_cast<int32_t>(playDirection != 1);
```
stellar-aria, line 277: can this be size_t? it looks like it's used as an index...

src/deluge/model/sample/sample.cpp
```diff
@@ -274,8 +274,7 @@ Error Sample::fillPercCache(TimeStretcher* timeStretcher, int32_t startPosSample
...
 	// If the start pos is already beyond the waveform, we can get out right now!
 	if (!reversed) {
```
stellar-aria, line 279: and then this can just be `if (playDirection)`

stellar-aria (approved):

PR #3606, opened 2025-04-15
Q31 mult, various

stellar-aria:
Some small QoL

- replace Multiply32x32... << 1 with q31_mult and q31_mult_rounded
- fix MultiplyAdd in FixedPoint
- Argonize the bits of interp in WaveTable


src/deluge/storage/wave_table/wave_table.cpp
```diff
@@ -937,34 +937,31 @@ WaveTable::doRenderingLoop(int32_t* __restrict__ thisSample, int32_t const* buff
 		}
 
 		// Grab the actual waveform data from memory, for both cycles that we need for this sample
-		int16x8x2_t interpolationBuffer[2];
```
m-m-adams, line 940: This should all be in its own pr

src/deluge/util/fixedpoint.h
```diff
@@ -399,7 +400,7 @@ class FixedPoint {
 	[[nodiscard]] constexpr FixedPoint MultiplyAdd(const FixedPoint& a, const FixedPoint& b) const {
 		if constexpr (fast_approximation && (((fractional_bits * 2) - 32) == (fractional_bits - 1))) {
 			// fractional_bits - 1 is due to the left shift
-			return from_raw(signed_most_significant_word_multiply_add(value_, a.raw(), b.raw()) << 1);
+			return from_raw(signed_most_significant_word_multiply_add(value_ >> 1, a.raw(), b.raw()) << 1);
```
m-m-adams, line 403: Why?

stellar-aria: Before the fix:
(Q31 + (Q31 * Q31 >> 32)) << 1
(Q31 + Q30) << 1
Err

After the fix:
((Q31 >> 1) + (Q31 * Q31 >> 32)) << 1
(Q30 + Q30) << 1
Q31

src/deluge/dsp/compressor/rms_feedback.cpp
```diff
@@ -97,25 +99,22 @@ void RMSFeedbackCompressor::render(std::span<StereoSample> buffer, q31_t volAdju
...
-			sample.l = multiply_32x32_rshift32(sample.l, wet);
-			sample.l = multiply_accumulate_32x32_rshift32_rounded(sample.l, dry_it->l, dry);
-			sample.l <<= 1; // correct for the two multiplications
-			// same for r because StereoSample is a dumb class
-			sample.r = multiply_32x32_rshift32(sample.r, wet);
-			sample.r = multiply_accumulate_32x32_rshift32_rounded(sample.r, dry_it->r, dry);
-			sample.r <<= 1;
+		if (wet != 1.f) {
+			// sample = (sample * wet) + (dry_it * dry);
+			sample.l = (sample.l * wet).MultiplyAdd(dry_it->l, dry);
```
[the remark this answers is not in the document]
stellar-aria: no, the whole point of the float conversion being potentially constexpr is for cases like this where it's a constant that can happen at compile-time. `wet` and `dry` are FixedPoint<31>, so the only conversion would be at the compare.

src/deluge/storage/wave_table/wave_table.cpp
stellar-aria: done

src/deluge/dsp/compressor/rms_feedback.cpp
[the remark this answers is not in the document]
m-m-adams: Check the code gen, it certainly looks like it's happening at runtime

image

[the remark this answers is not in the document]
stellar-aria: that big red block at the right is for the amplitudeIncrement line at the top (vshr 8, etc etc), not the for loop. The body of the for loop is after .L132. the amplitudeIncrement line is unchanged in this patch, so that was already happening.

m-m-adams (approved):

PR #3611, opened 2025-04-16
Argonize Wavetable interp

stellar-aria:


src/deluge/storage/wave_table/wave_table.cpp
```diff
@@ -937,34 +937,31 @@ WaveTable::doRenderingLoop(int32_t* __restrict__ thisSample, int32_t const* buff
 		}
 
 		// Grab the actual waveform data from memory, for both cycles that we need for this sample
-		int16x8x2_t interpolationBuffer[2];
+		std::array<std::array<Argon<int16_t>, 2>, 2> interpolationBuffer;
```
m-m-adams, line 940: This size seems wrong, the original is 8x2 and I think this makes 8x2x2 if I'm reading it correctly

src/deluge/storage/wave_table/wave_table.cpp
```diff
@@ -937,34 +937,31 @@ WaveTable::doRenderingLoop(int32_t* __restrict__ thisSample, int32_t const* buff
...
 		for (int32_t i = 0; i < kInterpolationMaxNumSamples >> 3; i++) {
-			interpolationBuffer[0].val[i] = vld1q_s16(&table1[whichValueStored[i]]);
-		}
-		for (int32_t i = 0; i < kInterpolationMaxNumSamples >> 3; i++) {
-			interpolationBuffer[1].val[i] = vld1q_s16(&table2[whichValueStored[i]]);
+			interpolationBuffer[0][i] = Argon<int16_t>::Load(&table1[whichValueStored[i]]);
+			interpolationBuffer[1][i] = Argon<int16_t>::Load(&table2[whichValueStored[i]]);
```
m-m-adams, line 943: If you want to index up to kInterpolationMaxNumSamples>>3 then that should be used to set the size as well, this is going to break if that ever changes

src/deluge/storage/wave_table/wave_table.cpp
```diff
@@ -976,45 +973,34 @@ WaveTable::doRenderingLoop(int32_t* __restrict__ thisSample, int32_t const* buff
...
+			auto multipliedDifference = difference.MultiplyFixedPoint(strength2);
+			kernelVector[i] = value1 + multipliedDifference;
 		}
 
 		// Apply the windowed sinc kernel to the waveform data
-		int32x2_t twosies[2];
+		std::array<ArgonHalf<int32_t>, 2> twosies;
 		for (int32_t c = 0; c < 2; c++) {
-			int32x4_t multiplied;
+			Argon<q31_t> multiplied = 0;
```
m-m-adams, line 988: these aren't quite equivalent, it should initialize to Multiply(kernelVector[i].GetLow(), interpolationBuffer[c][i].GetLow()) and then loop starting from 1

src/deluge/storage/wave_table/wave_table.cpp
stellar-aria: An Argon<int16_t> is an int16x8_t. There's then a 2-element array of a 2-element multivector of those.
`int16x8x2_t[2] = int16x8_t[2][2] = array<array<Argon<int16_t>, 2>, 2>`

src/deluge/storage/wave_table/wave_table.cpp
stellar-aria: It is the same, it goes from 
```
multiplied = kernelVector[i].GetLow() * interpolationBuffer[c][i].GetLow();`
```
to
```
multiplied = 0;
multiplied += kernelVector[i].GetLow() * interpolationBuffer[c][i].GetLow();
```
this is the same code that is already in interpolate.cpp

m-m-adams: They're the same result in a different number of steps, the previous one skips initialization

stellar-aria: It introduces a single cycle VMOV.S32 from an immediate at no optimization, and at -O2 GCC removes it on its own: https://godbolt.org/z/YWovK9EPE

src/deluge/storage/wave_table/wave_table.cpp
stellar-aria: done

m-m-adams (approved):

PR #3616, opened 2025-04-17
Refactor/ CONVERT_AND_STORE macro to lambda

stellar-aria:


m-m-adams:
This should probably be an inline function rather than a lambda

m-m-adams (approved):

PR #3617, opened 2025-04-17
Refactor/ Remove SampleClusterArray in favor of std::vector

stellar-aria:


src/deluge/model/sample/sample.cpp
```diff
@@ -98,14 +99,19 @@ Error Sample::initialize(int32_t newNumClusters) {
 	waveTableCycleSize = 2048; // Default
 	fileExplicitlySpecifiesSelfAsWaveTable = false;
 
-	return clusters.insertSampleClustersAtEnd(newNumClusters);
+	try {
```
m-m-adams, line 102: reserve the number of new clusters first - adding 1000 clusters will take 10 allocations otherwise. It's one of the more important optimizations in the resizeable array

src/deluge/model/sample/sample.h
```diff
@@ -152,7 +151,7 @@ class Sample final : public AudioFile {
 
 	uint32_t waveTableCycleSize{0}; // In case this later gets used for a WaveTable
 
-	SampleClusterArray clusters;
+	std::vector<SampleCluster> clusters;
```
m-m-adams, line 154: A deque might be more appropriate - the only time we append is sampling/recording but vectors need to copy and reconstruct their entire contents on memory expansion and dequeues don't

src/deluge/storage/cluster/cluster.cpp
```diff
@@ -198,7 +198,7 @@ void Cluster::steal(char const* errorCode) {
 		if (ALPHA_OR_BETA_VERSION && sample == nullptr) {
 			FREEZE_WITH_ERROR("E181");
 		}
-		sample->clusters.getElement(clusterIndex)->cluster = nullptr;
+		sample->clusters[clusterIndex].cluster = nullptr;
```
m-m-adams, line 201: if the sample is a nullptr we should probably return instead of continuing to de reference it, even if just for alpha/beta

src/deluge/model/sample/sample.h
stellar-aria: done

src/deluge/storage/cluster/cluster.cpp
stellar-aria: Fixed

src/deluge/model/sample/sample.cpp
stellar-aria: done

m-m-adams (approved):

PR #3645, opened 2025-04-26
Fix horz menu selection using synth-kit-cv-midi buttons for envelopes

### LINKED ISSUE #3644: Select horizontal menus with synth-kit-midi-cv buttons doesn't work for envelope 1-4 menus (1.3 alpha)

Recently added ability to select horz menu item using synth-kit-midi-cv buttons works for all menus except envelope 1-4 menus. If you enter envelope menu and press any of synth-kit-midi-cv buttons, it will do nothing.

This feature should work consistently for all horz menus

Today, on the latest alpha

No AFAIK

OLED

Beta

1.3.0-BETA-E222803

1. enable horizontal menus feature in the community settings
2. make a new song
3. make a new synth clip
4. enter the synth clip
5. enter env1 menu using the env1 attack shortcut
6. try to change menu item by pressing kit/midi/cv button

lburygin:
Fixes #3644


seangoodvibes:
Thanks for fixing but i'm not sure  this is the right solution

The fact that it works for all other horizontal menu's points to another problem in my opinion

I think what we're missing is a horizontal menu definition for the envelope menu in menus.cpp

In menus.cpp you can see how horizontal menus are defined for the other horizontal menus

I think we definitely need the horizontal menu class because it identifies menus declared in menus.cpp as horizontal

For example, in the current horizontal menu implementation for envelopes, are you able to see 4 pages for all 4 envelopes?

seangoodvibes:
I think we just need something like this:

image

seangoodvibes:
Or this as it's in the same file as the envelope menu definitions

image

seangoodvibes (approved):

seangoodvibes:
Looks and works good thanks!

PR #3679, opened 2025-05-10
Feature/ Remove alternative select behavior for horizontal menus

lburygin:
The alternative select behavior for horizontal menus is now default and the only behavior.


seangoodvibes (approved):

seangoodvibes:
Thank you!

PR #3687, opened 2025-05-11
BugFix/ routineWithClusterLoading Audio Clicks (#3680)

seangoodvibes:
Fixed some routineWithClusterLoading clicks specifically around audio recording and singleRow / waveform rendering by yielding until the audio routine has been called.

Cherry picked to 1.2 branch so we can release a 1.2.1 bug fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 > To find a detailed list of how to use each feature, check here: [Community Features](docs/community_features.md)
 
+## c1.2.1 Chopin
+
+### Sound Engine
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
+repos:
```
m-m-adams, line 1: This file shouldn't be included in 1.2

seangoodvibes: Deleted, it required me to add it when I was committing earlier

PR #3695, opened 2025-05-13
Program menu

dctucker:
This pull request adds two features:

- Program menu
- Learned MIDI channel updates clip program when recording

## Program menu

Consolidates three Program, Bank, and Sub-Bank items into one. Works similar to Decimal, where moving the horizontal encoder moves the blinking cursor.

IMG_4062

IMG_4061

Implemented this separately from horizontal menus to maximize compatibility with prior versions, useful for those of us who like to run custom firmware based on earlier releases.

Compatible with 7-segment display

IMG_4063
IMG_4064
IMG_4065
IMG_4066

## Learned MIDI channel updates clip program when recording

MIDI tracks that are learned to a channel will listen for program and bank changes (CC 0 and 32), and assign incoming values into the active clip when `RECORD` is illuminated. This is useful for toggling between scenario "A", when you want to try different sounds on the instrument without the clip forgetting the previous values, and scenario "B", when you want the clip to remember the most recently selected sound.


seangoodvibes:
Re: horizontal menu comment, I'm not sure I understand what you mean about maximizing compatibility with other firmwares. If you mean so you can cherry pick the feature to older firmwares, I don't think that is reason enough to add a new menu UI.

Ultimately we are really trying hard to reduce menu duplication because new menu items are heavy memory users.

I would advocate strongly here for this menu to use the new horizontal menu system.

Thanks

dctucker:
I will follow up in the future with menu improvements in a separate PR to keep things simple.

PR #3698, opened 2025-05-14
Learned MIDI channel updates clip PGM/BANK/SUB values

dctucker:
This pull request is a trimmed-down version of #3695.

## Learned MIDI channel updates clip program values when recording

MIDI tracks that are learned to a channel will listen for program and bank changes (CC 0 and 32), and assign incoming values into the active clip when `RECORD` is illuminated.

This is useful for toggling between the following scenarios:

- `RECORD` off, when you want to try different sounds on the instrument without the clip forgetting the previous values.
- `RECORD` on, when you want the clip to remember the most recently selected sound.


seangoodvibes:
@dctucker could you add documentation for this in the following pages:

https://github.com/SynthstromAudible/DelugeFirmware/blob/community/website/src/content/docs/features/community_features.md

https://github.com/SynthstromAudible/DelugeFirmware/blob/community/website/src/content/docs/changelogs/CHANGELOG.mdx

thanks!

dctucker:
Ok, documentation added. I'm a bit out of the loop on the release requirements and timing, so please let me know if I need to move the doc updates into 1.4.

dctucker:
I'm testing this now after merging community, and it seems program change isn't updating yet banks are. I'll mark this as ready for review once I get that working again.

dctucker:
:facepalm: Ok I figured out what was going on. My new SD card somehow ended up with a `SETTINGS/MIDIDevices.XML` containing `<mpeUpperZone numMemberChannels="13" />`, either by way of restoring from backups (likely), or being set as a default (unlikely but if so I'd like to understand why). I was able to deduce this since the problem went away when starting with the old SD card or no card inserted.

That does raise a concern though; shouldn't it behave the same regardless of MPE input settings?

seangoodvibes:
> 🤦 Ok I figured out what was going on. My new SD card somehow ended up with a `SETTINGS/MIDIDevices.XML` containing `<mpeUpperZone numMemberChannels="13" />`, either by way of restoring from backups (likely), or being set as a default (unlikely but if so I'd like to understand why). I was able to deduce this since the problem went away when starting with the old SD card or no card inserted.
> 
> That does raise a concern though; shouldn't it behave the same regardless of MPE input settings?

I believe in the previous versions of the firmware the deluge would try to automatically detect a device as mpe and set it up as such. but that created other problems, so i think in 1.2 (or maybe 1.3) that feature was removed. 

So it's possible your SD card has a MIDIDevices file with the MPE configuration of a past firmware.

dctucker:
That makes sense; thank you for the explanation. I confirmed my backup does not contain this file, so it might have been auto-detected when loading one of the earlier firmware during testing. Maybe including a file in my backup containing explicit values to disable MPE would make sense.

Other than the MPE behavior being a little unusual, this is working as expected, and since I don't have MPE it's a little harder for me to confirm expected behavior, but I should at least be able to eliminate the unexpected behavior. I'll try to figure that out and push a commit. I'm pretty sure it just has to do with the difference in function calls between CC and PGM messages.

seangoodvibes:
@dctucker could you repoint your Pr to our main branch?

I'll ask the team if we want to include this in 1.3 or not

Thanks!

PR #3700, opened 2025-05-15
Horizontal MIDI program menu

dctucker:
This pull request implements horizontal menus for Bank / Sub-bank / PGM on MIDI tracks.

Reworked version of #3695

IMG_4068
IMG_4072

- PGM is the rightmost item, but is selected first.

PR #3701, opened 2025-05-15
Horizontal program change midi input

dctucker:
Merges and integrates

- https://github.com/SynthstromAudible/DelugeFirmware/pull/3698
- https://github.com/SynthstromAudible/DelugeFirmware/pull/3700


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -1452,6 +1465,20 @@ bool SoundEditor::pitchBendReceived(MIDICable& cable, uint8_t channel, uint8_t d
 	return false;
 }
 
+bool SoundEditor::midiPCReceived(MIDICable& cable, uint8_t channel, uint8_t program) {
+
+	auto currentItem = getCurrentMenuItem();
+	if (currentItem == &midiBankMenu || currentItem == &midiSubMenu || currentItem == &midiPGMMenu) {
+		midiPGMMenu.readValueAgain();
+		if (display->haveOLED() && runtimeFeatureSettings.isOn(RuntimeFeatureSettingType::HorizontalMenus)) {
+			midiProgramMenu.drawPixelsForOled();
```
lburygin, line 1474: Seems like there is no need to call drawPixelsForOled, when you call readValueAgain it already should rerender the HorizontalMenu

Maybe the problem with rerendering is that the value in getCurrentInstrumentClip()->midiPGM does not update properly? Maybe you should call writeCurrentValue() before readValueAgain() to update it

lburygin: I see you added bool SoundEditor::midiPCReceived(MIDICable& cable, uint8_t channel, uint8_t program) overload but where does it get called? 
In melodic_instrument.cpp:402 there is another overload get called, not this one:
`soundEditor.midiCCReceived(cable, channel, ccNumber, value);`

lburygin: Confirmed: readValueAgain() updates the value in horizontal menu.
Sadly I don't have any midi controller to test bank / sub / PGM parameter changes, but could you verify that value is really get changed? Does it get changed when you leave the horizontal menu and enter it again?

dctucker: It seems to update only when more than one of the fields changes, and always updates when moving the cursor by holding down the SELECT knob, and of course also by exiting and re-entering the h-menu.

PR #3719, opened 2025-05-18
[DRAFT] Program Change automation lane

dctucker:
Adds a way to automate program changes within a clip.

### TODO

- [x] View updates upon receiving MIDI input.
- [x] Try to increase input precision.


dctucker:
To increase precision, we could borrow the approach from the velocity automation lane where the exact value is remembered and can be modified by holding and turning the knob.

seangoodvibes:
@dctucker could you add documentation for this in the following pages:

https://github.com/SynthstromAudible/DelugeFirmware/blob/community/website/src/content/docs/features/community_features.md

https://github.com/SynthstromAudible/DelugeFirmware/blob/community/website/src/content/docs/changelogs/CHANGELOG.mdx

thanks!

abstractrealism:
Just giving this a try and loading old songs with this build is freezing my Deluge. Works fine seemingly in the blank startup song, but loading anything else freezes without an error code. Scrolling through the songs folder the pads update to show the preview of the songs, it's upon pressing Select that it freezes, right as the Loading animation pops up. 


In a new song though the feature seems to work great! 
One thing that might make it a little smoother to use (and maybe the comment above about borrowing from velocity view might also solve this, I'm not quite picturing what that'd be like) would be if any values you place are held automatically, rather than having to press the start and end point, hold them both, and adjust both gold knobs to the same value. Being able to sweep through programs could be cool and not saying to remove that ability, but I would imagine a more frequent use-case would be to toggle back and forth between maybe two or three programs, and being able to just set the start point for each might make that smoother. I think Ableton's automation lanes might have a functionality like this but can't remember for sure.

dctucker:
Ah, thanks for letting me know about the crashes. I'll try to pick this back up at some point.

[This related functionality](https://github.com/SynthstromAudible/DelugeFirmware/pull/3698) should be ready to go in the meantime, although I'm not sure how soon it will merge.

abstractrealism:
@dctucker I think someone let me know on Discord that those were actually just nightly bugs, unrelated to your PR, though I'd have to search back to confirm. Also happy to test again sometime!

Issue #3707, opened 2025-05-16
Level parameter setting from midifollow.xml not being saved with songs


davidvandenbor:
### Please describe the problem:

In the midifollow.xml file there are 2 parameters that deal with volume levels of a clip, one is pre-FX and one is post-FX. When you turn the golden rotator, in the OLED display ```<volumePostFX>``` popup shows up as **"MASTER LEVEL"**, and when you turn a rotator on the Faderfox ```<volume>``` shows up as **"LEVEL"** in the OLED display. In the default midifollow XML file the parameters show up as:

```
  <volumePostFX>7</volumePostFX>
  <volume>7</volume>
```
I noticed that The "master level" (the first orange button with the top-left gold rotator) was unreachable for my midi controller (faderfox ec4) but the moment I renamed one of the CC's to another number, the Deluge was able to distinguish between the two, so that worked:

```
  <volumePostFX>7</volumePostFX>
  <volume>114</volume>
```
But here's where the problem occurs: when I save songs, the ```<volume>``` setting doesn't get saved with the songs! When I re-open songs, the ```<volume>``` CC level gets set back to its default level again (25) So during the editing or performance of the song, the ```<volume>``` levels retain their positions within each clip, but the moment you save the song, all the ```<volume>``` levels get reset back again at 25.

So in a nutshell, only ```<volumePostFX>``` gets saved with a song, but ```<volume>``` doesn't.

### What is the expected behavior?

```<volume>```of clips should be saved with a song, just as ```<volumePostFX>``` levels are.

### When did you start noticing the issue?

After updating to the latest nightly, Beta 6 (deluge-v1_3_0-beta+2025_05_14-b0e6c3d.bin)

### Is there a relevant Pull request?

Not that I know of, I searched the open issues

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

deluge-v1_3_0-beta+2025_05_14-b0e6c3d.bin

### If possible provide the steps to reproduce the issue and upload additional media:

Steps are provided above, it should be pretty easy to stumble on the issue.

seangoodvibes:
@davidvandenbor where did you get volume from? I don't recall that ever being in the midi follow xml, I don't know what that parameter is but I don't think you should be using it, I think you found a backdoor into a deluge parameter that was not meant to be publicly accessible

seangoodvibes:
[...]

PR #3722, opened 2025-05-18
Fix midi follow: CC value 7 was written twice to midi follow xml file

soymonitus:
Fixes: https://github.com/SynthstromAudible/DelugeFirmware/issues/3707


seangoodvibes (approved):

seangoodvibes:
Thanks for fixing @soymonitus newly generated file looks good now!

PR #3729, opened 2025-05-18
Bugfix/ Fixed incorrect diffusion label for digital & mutable reverb in the horizontal menu

lburygin:
For digital & mutable reverb models width param was displayed as WDTH instead of DIFF

### DIFF
--- a/src/deluge/gui/menu_item/reverb/width.h
+++ b/src/deluge/gui/menu_item/reverb/width.h
@@ -45,6 +45,7 @@ public:
 		using enum l10n::String;
 		switch (AudioEngine::reverb.getModel()) {
 		case dsp::Reverb::Model::DIGITAL:
+			[[fallthrough]];
 		case dsp::Reverb::Model::MUTABLE:
 			label.append(deluge::l10n::get(l10n::String::STRING_FOR_DIFFUSION));
 			break;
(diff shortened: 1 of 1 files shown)


src/deluge/gui/menu_item/reverb/width.h
```diff
@@ -42,7 +42,14 @@ class Width final : public Integer {
 	[[nodiscard]] std::string_view getTitle() const override { return getName(); }
 
 	void getColumnLabel(StringBuf& label) override {
-		label.append(deluge::l10n::get(l10n::String::STRING_FOR_WIDTH_SHORT));
+		using enum l10n::String;
+		switch (AudioEngine::reverb.getModel()) {
+		case dsp::Reverb::Model::DIGITAL:
```
seangoodvibes, line 47: So is this going to display width for digital and diffusion for mutable?

The switch statement is a bit unclear to me compared to other switch statements I've seen in the deluge codebase

Eg usually after a condition is met ive seen "break;" and in other cases I've seen "[[fallthrough]]"

Can you double check just to ensure consistency with how the switch is presented?

Im away from deluge for most of the day so cant test either

Thanks!

lburygin: Oh yeah thank you for pointing out, late night coding mistakes, of course "break" is needed here; fixed it

seangoodvibes (approved):

PR #3766, opened 2025-05-24
Feature / Arp Glide Probability

soymonitus:
- Added `GLIDE PROBABILITY (GLID)` paramater that will allow you to control the chance of delaying a note's note-off event to be executed at the same time as the following arp note. If your sound has some `Portamento` applied, then it will produce a glide effect.


soymonitus:
Ready for review!

PR #3773, opened 2025-05-26
Bugfix/ Fixed crash when using any of mod-fx shortcuts the first time right after booting up

lburygin:
Steps to reproduce:

1. Enable the horizontal menus feature
2. Turn off the deluge
3. Turn on the deluge
4. Open a synth clip if not opened already
5. Press any of mod-fx shortcuts except mod-fx type shortcut
6. Crash

The issue was that we're focusing the child item in the parent horizontal menu before modControllable gets initialized properly, so it was crashing on the isRelevant() checks

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -1267,14 +1267,6 @@ doSetup:
 						return ActionResult::DEALT_WITH;
 					}
 
-					// If we're on OLED, a parent menu & horizontal menus are in play,
-					// then we swap the parent in place of the child.
-					if (parent != nullptr && parent->renderingStyle() == Submenu::RenderingStyle::HORIZONTAL) {
-						if (parent->focusChild(item)) {
-							item = parent;
-						}
-					}
-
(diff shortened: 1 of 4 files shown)


src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -1567,7 +1559,7 @@ void SoundEditor::modEncoderButtonAction(uint8_t whichModEncoder, bool on) {
 	}
 }
 
-bool SoundEditor::setup(Clip* clip, const MenuItem* item, int32_t sourceIndex) {
+bool SoundEditor::setup(Clip* clip, const MenuItem* item, const MenuItem* parent, int32_t sourceIndex) {
```
nikodemus, line 1562: should be `const Submenu*` so there's no need for the cast. Probably just need to forward declarare `class Submenu;` in the header (no need to include submenu.h there.)

nikodemus (changes requested):
I would prefer if the "open menu and show first applicable param" behaviour was kept for non-applicable pads: they're probably just accidental presses of the wrong pad, so opening the menu was an intentional choice.

PR #3777, opened 2025-05-27
[DRAFT] Tuning System

dctucker:
This PR will add functionality to...

- [x] Allow re-tuning the existing 12-note octave
- [x] Load Scala files
- [x] Implement MIDI Tuning Standard
- [x] Tuning bank per song
  - [x] Clips can select a tuning from the bank

Likely not implemented in this PR:

- [ ] Allow redefining the size of an octave

## Allow re-tuning the existing 12-note octave

The firmware in its current implementation is very heavily weighted in favor of generating tones from a set of 12 pitches that double in frequency at each octave, and there are features present which depend on these assumptions. For now this will likely remain unaltered, and the focus will be on offset tuning of the 12 intervals, while leaving open the possibility of a future implementation that will allow retuning across the entire range of MIDI notes.

## Load Scala files

The [Scala scale file format](https://www.huygens-fokker.org/scala/scl_format.html) will be used for importing and exporting tuning systems on the SD card. Since we are sticking with a 12-note octave (for now), we will refuse to load scales with more than 12 pitches. Pitches are specified in cents or ratios.

## Implement MIDI Tuning Standard

This will allow integration with software more specialized in exploring various tunings. There is published [documentation](https://midi.org/midi-tuning-updated-specification) for this standard, which provides a framework to reason about how to support tuning systems, in addition to specifying how to form MIDI messages with varying degrees of precision. The manual for the Novation Bass Station II demonstrates one such implementation. Our implementation will primarily focus on fulfilling the goal of retuning the octave.

## Tuning bank per song

A tuning bank will be a collection of tunings. The first and default tuning will always be equal temperament. When loading a scale file, it will become part of the song's tuning bank.

### Clips can select a tuning from the bank

To facilitate changing between tuning systems as needed during performance, a clip will store the numerical index of the desired tuning from the bank. This will feel familiar to setting a MIDI program on a clip, and will be available via the menu item displaying the name of the selected tuning.


dctucker:
Storage format within the song

image

dctucker:
This is getting close, but there are a few things I need to test before I can give it my stamp of approval. I've had to pause working on this, but I anticipate resuming soon.

PR #3786, opened 2025-05-29
[DRAFT] display identified sample pitch, allow editing it

nikodemus:
- In multirange display, display the identified pitch of the sample after the range as (NOTE+CENTS), both editable.

- The column buttons can be used to select the editable value in addition to the horizontal encoder, making the UI more consistent on OLED.

- 7-segment display of multisample ranges has been overhauled. Only the low end of each note range is displayed initially: use horizontal encoder or the column buttons to view & edit other columns.

- BOTTOM and TOP in the note range renamed to LOW and HIGH (LO and HI for 7-seg) so that NOTE-NOTE (NOTE+CT) information fits onto OLED screen.

- MenuItem::drawItemsForOled has an additional startX parameter, allowing the default 1-character wide left-padding to be removed, necessary to make the expanded MultiRange fit without scrolling.

- SampleHolderForVoice::cents made protected, so finetuning cannot be accidentally be forgotten.

- TODO: it's not 100% clear to me yet how the different transpose controls relate to different sample types - figure it out and make sure it works correctly. (It probably does, but I want to know.)


darwiniandude:
This is an excellent idea, really essential for working with imported samples.

nikodemus:
Closing this one, reopening one for main.

PR #3806, opened 2025-06-02
Add shortcut for the note probability

lburygin:
Added "NOTE" pad shortcut for the note probability menu

Other:
- Fixed second layer shortcut toggle behavior in case you want to set a patch cable and go back to the original parameter using shift + shortcut


seangoodvibes:
I'm not sure we want to do this

I think the reason we did it for the random pad was because it was a "global randomizer" menu that I think also works outside of arp usage?

If the note probability menu is just for the arp, I'm not sure we'd want this shortcut to go there

seangoodvibes:
Need more info on this to understand.

Are you referring to the note probability in this menu which is supposed to work for arp and non arp notes?

Edit: I guess it makes sense as it's like the velocity spread shortcut

image

lburygin:
Oh it was requested by Monitus in the discord and we landed that it's ok to add:

lburygin:
> Are you referring to the note probability in this menu which is supposed to work for arp and non arp notes?
> Edit: I guess it makes sense as it's like the velocity spread shortcut

Yep, it is

seangoodvibes:
> > Are you referring to the note probability in this menu which is supposed to work for arp and non arp notes?
> > Edit: I guess it makes sense as it's like the velocity spread shortcut
> 
> Yep, it is
> 
> 

sounds good

We should probably add both velocity spread and note probability shortcuts to that new shortcut section in the changelog

lburygin:
Updated the documentation in the #3829 & #3830 PRs

seangoodvibes (approved):

PR #3824, opened 2025-06-05
[DRAFT] Feature : Ability to randomize automation of any automatable parameter

joxihan:


joxihan:
Automation View : Add Randomization feature for any automatable parameter by pressing Shift + Random pad.

src/deluge/gui/views/automation_view.cpp
```diff
@@ -4925,6 +4933,21 @@ void AutomationView::updateAutomationModPosition(ModelStackWithAutoParam* modelS
 	}
 }
 
+ActionResult AutomationView::handleRandomizedParamChange(ModelStackWithAutoParam* modelStackWithParam, Clip* clip,
+                                                         int32_t effectiveLength) {
+	Output* output = clip->output;
+	OutputType outputType = output->type;
+	int32_t xScroll = currentSong->xScroll[navSysId];
+	int32_t xZoom = currentSong->xZoom[navSysId];
+	seedRandom();
```
m-m-adams, line 4944: This shouldn't be needed

src/deluge/gui/views/automation_view.cpp
```diff
@@ -4925,6 +4933,21 @@ void AutomationView::updateAutomationModPosition(ModelStackWithAutoParam* modelS
...
+	for (int x = 0; x <= kRandomShortcutX; x++) {
```
m-m-adams, line 4945: Use the pad width/height constants instead of the two pads

src/deluge/gui/views/automation_view.cpp
```diff
@@ -2400,8 +2402,14 @@ ActionResult AutomationView::padAction(int32_t x, int32_t y, int32_t velocity) {
...
 	}
 
+	if (velocity > 0 && inAutomationEditor() && Buttons::isShiftButtonPressed() && x == kRandomShortcutX
+	    && y == kRandomShortcutY) {
+		return handleRandomizedParamChange(modelStackWithParam, clip, effectiveLength);
+	}
+
 	// Edit pad action...
 	if (x < kDisplayWidth) {
+
```
m-m-adams, line 2414: Not needed here

m-m-adams (commented):
Thanks for the PR! Just a couple things to consider:

1. This only randomizes the current page of automation, should it do the whole clip? 

2. That shortcut should arguably open a randomization parameter. We're trying to avoid putting actions on shortcuts 

It also needs a menu entry to trigger it - we need those to help people discover functionality and as a way to find things without memorizing the shortcuts

PR #3921, opened 2025-07-07
[DRAFT] Loop in arrangement

wargoth:


seangoodvibes:
Thanks for opening a PR for this long awaited feature! I assume this is still a work in progress so I'm going to change the PR to draft for now.

When it's ready for review, you can switch it from draft.

Couple early observations:
- the change history is very difficult to follow (e.g. it seems to be mix your changes in with other functions so it looks like you're deleting code and stuff). Also as much as possible, if you are changing existing code, please add comments to explain the change so we can follow the history.
- the looping feature documentation should be saved here: https://github.com/SynthstromAudible/DelugeFirmware/tree/community/website/src/content/docs/features
- what are the patch.patch / song.patch files? Were those included by mistake?

Thanks for contributing!

wargoth:
Oops, didn't mean to open a PR in the main repo yet, this is very early WIP.

PR #3942, opened 2025-07-10
Feature/ Add stem export for kit rows

seangoodvibes:
Cherry picked separately from nightly PR #3927 

Note: I think this should be ok to include in c1.3 as it's pretty self-contained. If you don't agree please let me know and we can close this.

Added Stem Export for Kit Rows using the same shortcut as in Song and Arranger (Hold Save + Press Record)
- Exports each kit row individually.
- Kit global FX (except for Pitch adjustment) are excluded by default.
- Muted kit rows and kit rows with no notes in them are excluded from the export.
- Stems are labelled as a "Drum" and include the Drum Name. Saved to EXPORTS/DRUMS folder.
- A separate normalization setting has been created for Drums and enabled by default.
- A new Actions/Audio Export menu has been created in the Kit Global FX menu which allows for configuration of Drum Normalization and whether to include Kit global FX.


lburygin:
It works but it exports only N-1 of N rows, e.g. I have kick, snare and hat rows, all unmuted, if I export the kit, it will generate only two files: one for kick and one for snare

PR #3963, opened 2025-07-22
Bugfix/ Fix skipping of the select encoder's turns

### LINKED ISSUE #3947: Select encoder can skip turns

A turn or series of turns of the select encoder may be skipped, which creates feeling that UI is not responsive.
The problem always been there (since 1.0 community release), but it's got worse on nightly when the #3889 PR merged. 
The most noticeable case is when you turn the encoder quickly with a good amplitude. On the nightly it's easy to reproduce, you can just open a random parameter menu and try turning the encoder fast back and forth, then try turning it slowly to see the difference. On the older fws you can start playback on some heavy song and then do the same steps. 
Turns also can be skipped in trivial cases like just scrolling through songs or samples menu

As the scheduler optimizations probably on the way to 1.3 release, we could tag this issue with 1.3 milestone

All select encoder turns should be registered

Today

No

OLED

Beta, Nightly

23da77d

On the nightly:
- Open any song or create a new one
- Open random parameter menu, e.g. LPF frequency
- Turn encoder fast a several times, then turn encoder slowly and see the difference

On the beta:
- Open some of cpu-heavy songs
- Open random parameter menu, e.g. LPF frequency
- Turn encoder fast a several times, then turn encoder slowly and see the difference

lburygin:
Fixes #3947


seangoodvibes (approved):

src/deluge/deluge.cpp
```diff
@@ -581,7 +581,7 @@ void registerTasks() {
 	addRepeatingTask(&(AudioEngine::routine_task), p++, 8 / 44100., 64 / 44100., 128 / 44100., "audio  routine",
 	                 RESOURCE_NONE);
 	// this one runs quickly and frequently to check for encoder changes
-	addRepeatingTask([]() { encoders::readEncoders(); }, p++, 0.0005, 0.001, 0.002, "read encoders", RESOURCE_NONE);
+	addRepeatingTask([]() { encoders::readEncoders(); }, p++, 0.0001, 0.0001, 0.0002, "read encoders", RESOURCE_NONE);
```
m-m-adams, line 584: can the time go higher? 10000x a second seems like overkill

Did you check how often it actually gets ran? (it'll be printed if SCHEDULER_DETAILED_STATS is true)

lburygin: Hi! Sorry for the long answer

I'm running ./dbt sysex-logging command and my deluge has relwithdebinfo build installed. Also SCHEDULER_DETAILED_STATS is set to 1. 

But I don't see any logs over usb, what I'm doing wrong? Thanks for help in advance!

image

lburygin: image
here is how it looks on DEx

lburygin: re: can it go higher, you're right 0.0001 is overkill, tested closely, it works fine at 0.0004-0.0005, starts to skip at 0.0006

PR #3972, opened 2025-07-25
Kit velocity drum popup display bug fix

Metamere:
There was a bug where in a kit in velocity drum keyboard view, if a pad for one drum sound was held and then a pad for a different sound was pressed and released, for certain pairs, it sent a null pointer or other invalid drum object, which resulted in it displaying incorrect information (either midi info, "no sound", empty, or a different drum name from the kit), and if the second pad was pressed and released enough times, it would inevitably cause a crash. It was not obvious which pairs of kit sounds would cause issues or why, but it was always the same ones for a given kit, so it likely had something to do with the specific combination of memory addresses.

To remedy this issue, the drum object was replaced with the selectedDrum object, with a check for a nullptr. This required the drawNoteCode function call to be moved below the setSelectedDrum function call for when it is not in the velocity drum view.

This also has the effect of making the popup display stay stable when holding a pad for one sound down and then pressing and and releasing another sound's pad rapidly, rather than having it flicker back and forth rapidly between the two. This makes it easier to read and more pleasant to view.


Metamere:
Oops, I based this on the 1.3 beta instead of the community branch, so I had to redo it.

PR #3976, opened 2025-07-25
[DRAFT] Added live unpitched input source for subtractive engine

seangoodvibes:
Added new osc types to the subtractive engine for live unpitched input


nikodemus:
I've been thinking of adding support for unpitched samples, for things like piano hammer transients.

Could we expose the NOTE -> OSC PITCH mapping in the mod matrix?

That way UI doesn't need to grow multiple new oscillator types? (...besides, it'll be just a matter of time before someone wants to have one FM oscillator not follow the pitch...)

nikodemus:
I'll try a simple `Source->isTracking` with a toggle in the oscillator menu first.

seangoodvibes:
> I'll try a simple `Source->isTracking` with a toggle in the oscillator menu first.

If your solution works better and is more sustainable go for it!

This PR is still buggy anyway.

Issue #3943, opened 2025-07-10
Crash changing kit row to midi


macbutch:
A few notes. This seemed fairly easy to reproduce:

1. Add a kit row below the other rows
2. Change it to midi without loading a sample first

When it crashed, I was able to save and then reload. It had the new kit row (with no sample) and then crashed again when I tried the same operation. 

Creating a kit row at the top seems to work fine. 

![image](https://github.com/user-attachments/assets/07763fb0-ba57-4fb5-a6ea-1682b013ea81)

Metamere:
~~I can't seem to reproduce this issue. Is there a kit you can send that this happens on? I am creating the new row by holding the audition pad and then pressing midi. I don't think there is a way to add a kit row without a sound, is there?
Oh, I see how yeah. It switches to midi just fine for me though.~~
Oh, actually yeah, I see how it freezes things until you press the right combination of other things to unfreeze it. Annoying, yeah.

Metamere:
Alright, with my PR #3977 merged, this issue should be fixed and can be closed.

nikodemus:
@macbutch can you confirm this is no longer reproducible, or provide exact step-by-step reproduction?

PR #3977, opened 2025-07-26
Prevent the creation of "NO SOUND" kit rows

Metamere:
Prevent the creation of a kit row upon pressing a pad in the main screen outside of the bounds of the current kit. Instead, now it will simply deselect any selected rows.

That kind of empty row with "NO SOUND" does not act like the other types of kit row, cannot be moved up and down within the kit, and when you try to change it to another type, it doesn't properly take, and instead either creates an additional row, or causes things to glitch and often eventually crash. Glitches included freezing certain inputs, and the sticky shift getting stuck on without the indicator light on the shift button. This type of glitchy kit row has been an issue since at least v1.1. 

You can still create an empty row by pressing audition+kit and then backing out without selecting anything, and it will display U1, U2, U3, etc. even though it doesn't have a sound associated with it. That type of placeholder kit row doesn't seem to have any of the issues associated with a "NO SOUND" kit row.

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -4151,7 +4151,7 @@ doDisplayError:
 		}
 	}
 
-	// If a kit, add scrolling limits to keep at least the top row or bottom row of the kit on the screen
+	// If a kit, prevents creating a new kit row beyond the adjacent empty rows
 	else {
 		// If it's more than one row below, we can't do it
 		if (yDisplay < -1 - clip->yScroll) {
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -4151,7 +4151,7 @@ doDisplayError:
 		}
 	}
 
-	// Or, if a kit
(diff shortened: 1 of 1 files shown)


Metamere:
This PR is to fix the issue brought to attention in #3943.

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -4151,7 +4151,7 @@ ModelStackWithNoteRow* InstrumentClipView::createNoteRowForYDisplay(ModelStackWi
 		}
 	}
 
-	// Or, if a kit
+	// If a kit, add scrolling limits to keep at least the top row or bottom row of the kit on the screen
```
seangoodvibes: Is this comment correct? It says add scrolling limits but this is not a vertical scroll function

Metamere: That was my interpretation, but I'll try commenting those lines out to test.

Metamere: You're right, that's handled in the scrollVertical_limit function. I think those lines just limit adding a row to the row immediately above or below the kit.

PR #3981, opened 2025-07-28
Horizontal Menus/ Oscillators menu support

lburygin:
Added Horizontal Menu support and graphics for the osc1/2, modulator1/2, sample1/2 menus.

I had to do important refactoring: previously the current source was determined by global soundEditor.currentSource field. Now every source menu item knows itself to which source it belongs to. This makes possible things like the oscillator mixer page and a combined horizontal menu for all source menus.

TODO:
- Update the documentation in a separate PR


docs/menus/oscillator/volume.md
```diff
@@ -1,3 +1 @@
-Oscillator output volume.
-
-Does not apply to RINGMOD oscillators.
+Modulator depth.
```
seangoodvibes, line 1: Is this correct? The menu item is volume?

seangoodvibes (approved):

docs/menus/oscillator/volume.md
lburygin: Yeah that's a mistake, will fix in follow up PR :)

PR #3989, opened 2025-08-03
turning off pitch tracking for oscillators

nikodemus:
- Main use cases are unpitched transient samples and unpitched
  inputs, but doing this in a more general way allows eg. a secondary
  fixed pitch oscillator. The fully general approach would be to represent
  NOTE -> OSC PITCH in the mod matrix - this is a halfway step there, maybe.

  Turns out turning off pitches momentarily is also a fun performance effect!

- Source::isTracking specifies if it tracks the current note, or used C3 as pitch
  instead. Controlled from the oscillator menu.

- calculatePhaseIncrementForSource() uses C3 instead of the post-arp pitch if the
  oscillator isn't tracking.

- There's a new menu "OSC TRACKING", which contains both SYNC and the tracking toggles for both oscillators.

- Currently works for all oscillator types, but in case of FM the the modulators are always pitched. Maybe there's some sound design potential there, or should we disallow for FM -- or allow disabling tracking for modulators as well? (FM sounds terrible on community right now, so it's hard to gauge what's right leaving this as-is for now.)


src/deluge/gui/menu_item/osc/tracking.h
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2014-2023 Synthstrom Audible Limited
```
seangoodvibes: You can put your name here Niko instead of Synthstrom

seangoodvibes:
@nikodemus i see this is ready for review but i also see the to do list. Can you confirm what it is and isnt completed?

Thanks

nikodemus:
> @nikodemus i see this is ready for review but i also see the to do list. Can you confirm what it is and isnt completed?

Oops, forgot to update the PR comment. Done now. It's IMO complete, the behaviour with FM needs to be considered, but I would leave that until sine/FM/Argon is fixed.

m-m-adams (approved):

PR #4001, opened 2025-08-06
fix assigning a drum to a no-sound kit row

### LINKED ISSUE #4000: cannot convert a no-sound kit row to midi or other types
If there's no-sound row in a kit, and you attempt to assign a sound to it or convert it to eg. a midi drum, the new drum appears on a different row and the original no-sound row becomes wholly unresponsive, with double-illuminated audition pads.

Song file demonstrating this attached.

[Bug-Nosound-Assign.zip](https://github.com/user-attachments/files/21622234/Bug-Nosound-Assign.zip)

nikodemus:
When creating new kit rows, check if the auditioned row already exists as a no-sound row, in which case we can use it directly.

Fixes #4000.


nikodemus:
(I think this bug is in release/1.3 as well, need to check.)

Issue #3861, opened 2025-06-21
Crash after duplicating MIDI track and adding a note and pressing play


dctucker:
### Please describe the problem:

Deluge crashes when pressing play after duplicating a MIDI track, clearing all notes, ~and adding a drone note~.

Two images showing the crash info:
- ![Image](https://github.com/user-attachments/assets/d2031cfd-14b5-47aa-9198-c7e6bb9802e2)
- ![Image](https://github.com/user-attachments/assets/d5c95315-7217-4b3d-b822-120868f864c4)

Apparent stack traces:

```
$ arm-none-eabi-addr2line -Capife build/Release/deluge.elf 0x200e6265 0x200e7049 0x200e7ae9 0x200e7abd 
0x200e6265: AutoParam::getValueAtPos(unsigned long, ModelStackWithAutoParam const*, bool) at ??:?
0x200e7049: OrderedResizeableArray::search(long, long, long) at /home/casey/Development/DelugeFirmware/src/deluge/util/container/array/ordered_resizeable_array.h:28
 (inlined by) AutoParam::setPlayPos(unsigned long, ModelStackWithAutoParam const*, bool) at /home/casey/Development/DelugeFirmware/src/deluge/modulation/automation/auto_param.cpp:1401
 (inlined by) AutoParam::setPlayPos(unsigned long, ModelStackWithAutoParam const*, bool) at /home/casey/Development/DelugeFirmware/src/deluge/modulation/automation/auto_param.cpp:1392
0x200e7ae9: std::_Rb_tree_iterator<std::pair<unsigned char const, AutoParam> >::operator++() at /home/casey/Development/DelugeFirmware/toolchain/v22/linux-x86_64/arm-none-eabi-gcc/arm-none-eabi/include/c++/14.2.1/bits/stl_tree.h:283
 (inlined by) MIDIParamCollection::setPlayPos(unsigned long, ModelStackWithParamCollection*, bool) at /home/casey/Development/DelugeFirmware/src/deluge/modulation/midi/midi_param_collection.cpp:86
0x200e7abd: MIDIParamCollection::setPlayPos(unsigned long, ModelStackWithParamCollection*, bool) at /home/casey/Development/DelugeFirmware/src/deluge/modulation/midi/midi_param_collection.cpp:81
```

```
$ arm-none-eabi-addr2line -Capife build/Release/deluge.elf 0x200e7aef 0x200e7abd 0x200e9863 0x200bbf63 
[...]

PR #4030, opened 2025-08-13
Revert "Convert MIDIParamVector to STL Map (#3131)"

dctucker:
This reverts commit a4e8d2449f61d39dec2259eb1c53e26aff094b9d.

As much as I want to prefer leveraging STL, and I appreciate the effort that went into committing this change, I think there's a stronger case to be made for minimizing crashes by prioritizing bug-fixes. We can revisit the usage of STL Map at a later time, or I'm also open to sagely advice in this PR for how to resolve the problem described in the issue without reverting the entire commit.

Resolves https://github.com/SynthstromAudible/DelugeFirmware/issues/3861.

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -44,6 +44,7 @@
 #include "model/scale/utils.h"
 #include "model/song/song.h"
 #include "modulation/arpeggiator.h"
+#include "modulation/midi/midi_param.h"
 #include "modulation/midi/midi_param_collection.h"
 #include "modulation/patch/patch_cable_set.h"
 #include "processing/engines/audio_engine.h"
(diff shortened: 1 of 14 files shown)


dctucker:
@seangoodvibes This pull request effectively resolves the bug, and I'm not sure what other attempts are being made to resolve the issue. If I push changes that cause something to break, while I may not like seeing my changes reverted, I still realize that it's the prudent thing to do in the interest of keeping developers and users of the software confident and engaged.

I'm not sure how to proceed. If there is a proposed alternative change that resolves the bug, please link it to this PR and the original issue.

seangoodvibes:
> @seangoodvibes This pull request effectively resolves the bug, and I'm not sure what other attempts are being made to resolve the issue. If I push changes that cause something to break, while I may not like seeing my changes reverted, I still realize that it's the prudent thing to do in the interest of keeping developers and users of the software confident and engaged.
> 
> 
> 
> I'm not sure how to proceed. If there is a proposed alternative change that resolves the bug, please link it to this PR and the original issue.

I changed it to draft because @stellar-aria was adamant on discord that we should not revert the changes. I didn't want anyone accidentally merging this without knowing that.

If Katherine wants to merge this she can.

m-m-adams (approved):
Some notes on the crash:
1. The midi param crashing bug seems very fundamental. The automation system relies on param vectors being trivially copyable (e.g. able to be memcopied). 
2. The map attempts to fix by copy constructing the map after the memcpy but it seems to be too late at that point 
3. I don't see a way to fix this while using a map and I also don't see a way to remove this requirement from the param manager without a much larger rewrite

PR #4040, opened 2025-08-18
Centered scrolling in file browsers

Metamere:
Building on my browser refactor with fast scrolling that was added to the nightly, this refines the regular scrolling behavior to be more consistent with other menus.
- During regular scrolling, keep the selected file in the center of the screen, except at the beginning and ending of the folder to indicate when it wraps and to prevent empty spaces on the screen. Follow those rules when doing the initial display as well (before the current file would always be shown selected at the top of the screen).
- Adjust the fast scrolling to move four places when jumping from the middle of the screen to move it out to the edge and show a full new set of files on the screen.


src/deluge/gui/ui/browser/browser.cpp
```diff
@@ -1324,11 +1331,23 @@ void Browser::updateUIState() {
...
 	}
 
-	if (scrollPosVertical > fileIndexSelected) {
-		scrollPosVertical = fileIndexSelected;
+	if (Buttons::isButtonPressed(deluge::hid::button::SHIFT)) { // fast scrolling
+		// Fast scroll logic - only adjust if selection goes off-screen
+		if (scrollPosVertical > fileIndexSelected) {
+			scrollPosVertical = fileIndexSelected;
+		}
+		else if (scrollPosVertical < fileIndexSelected - NUM_FILES_ON_SCREEN + 1) {
```
lburygin: could std::clamp be used here instead of branching?

e.g.:
std::clamp(scrollPosVertical, fileIndexSelected - NUM_FILES_ON_SCREEN + 1, fileIndexSelected)

Metamere: Ooh, that's good. I could probably use that in a few places. Thanks.

PR #4051, opened 2025-08-23
resolve `-Wodr` warnings in menus.cpp

dctucker:
This resolves compiler warnings about "previous declaration" and "the C++ One Definition Rule" by removing the `const` from the `extern` definitions in `menus.h`.

~@soymonitus:~ @lburygin: just want to make sure this change is fine.


soymonitus:
This code is not mine so not sure

dctucker:
My bad! I think I got the names confused :man_facepalming:

PR #4055, opened 2025-08-24
Improve performance with more yields

m-m-adams:
Replaces all usages of the sd routine and directly calling audio rendering with yields

### DIFF
--- a/src/OSLikeStuff/scheduler_api.h
+++ b/src/OSLikeStuff/scheduler_api.h
@@ -71,6 +71,9 @@ void removeTask(TaskID id);
 void yield(RunCondition until);
 /// timeout in seconds, returns whether the condition was met
 bool yieldWithTimeout(RunCondition until, double timeout);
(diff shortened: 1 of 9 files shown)


seangoodvibes:
Need to test more and confirm if it is coming from this PR but in a kit when I hold audition pad and press record it starts recording and then when I press record to stop I get "sd card error". It doesn't happen every time but it happens frequent enough that I noticed it.

I cherry picked the 4 commits to 1.3 branch cause resampling is broken in nightly.

m-m-adams:
> Need to test more and confirm if it is coming from this PR but in a kit when I hold audition pad and press record it starts recording and then when I press record to stop I get "sd card error". It doesn't happen every time but it happens frequent enough that I noticed it.
> 
> I cherry picked the 4 commits to 1.3 branch cause resampling is broken in nightly.

Have you been able to confirm that? I don't see an obvious reason to think it's related

Issue #3946, opened 2025-07-12
Stem export skips a clip/kit row, if I start exporting while playback is on


lburygin:
### Please describe the problem:

If I start exporting when playback is on, the first clip (or kit row) gets skipped, it says
"Exported 1 of N" on the beginning, and there is one file missing in the exported files

When playback is off it says "Exported 0 of N" on the beginning and everything gets exported as expected

### What is the expected behavior?

Stem export shouldn't skip a clip (or a kit row)

### When did you start noticing the issue?

Today

### Is there a relevant Pull request?

no

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

23da77d8cf86d1c24515944787aa1182359d7c4c

### If possible provide the steps to reproduce the issue and upload additional media:

1. Create a new song
2. Add a kit clip, use any of your existing kits
3. Fill the kit clip with some notes: add notes for a kick, snare and hat for example
4. Press the play button (start playback)
3. Press the affect-entire button
4. Open the KIT FX menu by pressing the select knob
5. Go to actions -> export audio -> start export
6. Check the number of exported audio files after export completes

PR #4061, opened 2025-08-27
Bugfix/ Fix stem export skipping

seangoodvibes:
Fixed issue where export would skip a stem if playback was enabled when you started the stem export process

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/3946


seangoodvibes:
@lburygin can you try this and see if it fixes things? might need to cherry pick to 1.3 in case exporting is broken on nightly. thanks!

lburygin:
Yes it does fix the issue :)

PR #4064, opened 2025-08-27
Manage display notification updates for parameter value changes from gold knob encoders

Metamere:
- Added a throttling routine to reduce the maximum frequency of updates to something that the OLED display and CPU can handle, and that our eyes can actually have a chance of perceiving. In addition, it checks for whether what it wants to display has changed in any way that warrants an update. So if the value is already at 50 for example, it just extends the timer instead of asking for another update every tick. So the number of updates attempted in a fast turn from 0-50 in a second could be reduced from 150 (or 300 with two knobs) to only about 22. In testing, this appeared plenty smooth enough.
- Added an arbitration routine to pass the display notification ownership between the two encoders if they are being operated in tandem, like when adjusting LPF frequency and resonance, or any two custom parameters. This prevents them from rapidly alternating updates and making it unintelligible. Now, it will trade off every second, or 250 ms after the last input from the encoder that last had control, and it's great to use. 
- Because of the high resolution of the encoders, and it trying to send a screen update every tick, this throttling drastically reduces screen flickering that came from overloading of the system. This should also help avoid some potential performance issues during playback.
- also changed several parameter names involved in the function from camelCase to snake_case as per the new style guide.
- Recommended for cherry pick to 1.3 beta branch.

### DIFF
--- a/src/deluge/gui/views/view.cpp
+++ b/src/deluge/gui/views/view.cpp
@@ -1090,9 +1090,6 @@ void View::displayModEncoderValuePopup(params::Kind kind, int32_t paramID, int32
 	static uint32_t last_actual_display_time = 0; // Used for frequency throttling
 
 	// Timing constants for display arbitration (in AudioEngine sample units)
-	const uint32_t MIN_DISPLAY_OWNERSHIP_TIME = 44100; // 1000ms at 44.1kHz (minimum juggling time)
-	const uint32_t DISPLAY_TIMEOUT = 11025;            // 250ms at 44.1kHz (ball drop timeout)
-	const uint32_t MIN_UPDATE_INTERVAL = 2000;         // ~45ms at 44.1kHz (minimum perceptible update frequency)
 
 	uint32_t current_time = AudioEngine::audioSampleTimer;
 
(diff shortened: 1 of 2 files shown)


src/deluge/gui/views/view.cpp
```diff
@@ -1072,96 +1072,200 @@ void View::potentiallyMakeItHarderToTurnKnob(int32_t whichModEncoder, ModelStack
...
+	static params::Kind display_owner_kind = params::Kind::NONE;
+	static int32_t display_owner_param_id = -1;
+	static PatchSource display_owner_source1 = PatchSource::NONE;
+	static PatchSource display_owner_source2 = PatchSource::NONE;
+	static uint32_t display_ownership_start_time = 0;
+	static uint32_t last_display_update_time = 0; // Used for timeout detection in arbitration
+	static uint32_t last_actual_display_time = 0; // Used for frequency throttling
+
+	// Timing constants for display arbitration (in AudioEngine sample units)
+	const uint32_t MIN_DISPLAY_OWNERSHIP_TIME = 44100; // 1000ms at 44.1kHz (minimum juggling time)
```
seangoodvibes: Would be a good idea to remove the magic numbers as we like to call them and replace it with:

Instead of = 44100, do
= kSampleRate

Instead of = 11xxx, do
= kSampleRate / 4

Instead of = 2000, do
= kSampleRate / 22

src/deluge/gui/views/view.cpp
```diff
@@ -1981,9 +2085,9 @@ void View::drawOutputNameFromDetails(OutputType outputType, int32_t channel, int
 oledDrawString:
 			deluge::hid::display::oled_canvas::Canvas& canvas = hid::display::OLED::main;
 #if OLED_MAIN_HEIGHT_PIXELS == 64
-			int32_t yPos = OLED_MAIN_TOPMOST_PIXEL + 30;
+			int32_t yPos = OLED_MAIN_TOPMOST_PIXEL + 32;
```
seangoodvibes: What are these pixel adjustments for?

src/deluge/gui/views/view.cpp
Metamere: That makes sense.Should I add them in the private variables in the header so they don't have to be calculated each time?

seangoodvibes: Yes I would do that

Also while I'm at it, what are those static variables? I've never seen / used that before. Are they just variables that are local to the function but persist?

Eg I see you're storing previous values there.

PR #4092, opened 2025-09-23
Fix for scrolling-in-folders-that-have-2-items bug

Metamere:
When doing regular scrolling with the new centered scrolling behavior in folders that have exactly two items, it was looking for a third file that wasn't there, and it would somehow pull another item from a higher level directory, though it would only display it, not allow it to be selected. This change prevents that issue. Also does some proactive edge case protection.


Metamere:
I choose you @lburygin! It doesn't look like I can specifically add you as a reviewer. Maybe I don't have suitable permission levels.

PR #4097, opened 2025-09-27
[DRAFT] Fix text search unable to find last file in folder (#4039)

### LINKED ISSUE #4039: Text search does not work for the last item in a folder

When you use the alphanumeric text input search method, it is unable to find the last file in any folder. If the last files in a folder are Vibraphone, Wurlitzer, and Xylophone, then you get results for V or W, but not X. If a folder is then added in the last position of the folder's contents alphabetically (for example a folder named Zeta), then it can find the last file that was previously unsearchable, but now the new folder will be unsearchable.

It should be able to find any file or folder within a folder by typing in its filename using the text search feature.

I discovered the bug yesterday, while testing to make sure that the text search feature still worked properly with my refactor of the file browser code. I thought it must have been something I caused, but it turned out to have been a bug for years, in all of the releases since 1.0. It was not an issue in the first commit in the repository fresh from Synthstrom, so sometime in the development work towards 1.0 the bug was introduced.

_No response_

OLED

Release 1.0/Amadeus

_No response_

- Go into a file browser menu for a synth, kit, etc. 
- Find the last file in the folder, then scroll back up. Try typing the first letters of the last file's name, and it will not show up. Compare by typing the first letters of a different file.

gtanner:
### *WIP: currently not working, needs some more investigation*

**Bug:** The alphanumeric text search could not find the last file in any folder. When typing the first letters of the last file's name (e.g., 'X' for 'Xylophone'), it would not appear in search results.

**Root cause:** In predictExtendedText(), when searching for a prefix like 'X', the code appends `~` to create `X~` and uses binary search to find where this would be inserted. For the last file, this correctly returns `index == numElements`. However, the condition `i >= numElements` incorrectly treated this valid case as 'not found'.

**Fix:** Changed the condition from `i >= numElements` to `i > numElements` at line 1412. This allows the valid case where i == numElements, which occurs when the last item matches the search prefix. The subsequent `i--` then correctly selects the last file.

Also fixed a similar bounds check issue in setFileByFullPath() at line 506, changing from `i > numElements` to `i >= numElements` for proper validation of exact file searches.

Fixes #4039

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -503,7 +503,7 @@ Error Browser::setFileByFullPath(OutputType outputType, char const* fullPath) {
 
 	//  Get the File Index
 	fileIndexSelected = fileItems.search(fileName);
-	if (fileIndexSelected > fileItems.getNumElements()) {
+	if (fileIndexSelected >= fileItems.getNumElements()) {
 		return Error::FILE_NOT_FOUND;
 	}
 
@@ -1415,7 +1415,8 @@ doSearch:
 	int32_t i = fileItems.search(searchString.get());
 
 	// If that search takes us off the right-hand end of the list...
-	if (i >= fileItems.getNumElements()) {
+	// Note: i == numElements is valid when the last item matches our search prefix
+	if (i > fileItems.getNumElements()) {
 
 		// If we haven't yet done a whole new read from the SD card etc, from within this function, do that now.
 		if (!doneNewRead) {


gtanner:
NOTE: I wanted create and run the unit tests for this but ran into problems since I am on a 64 bit OS.

I will build and load this on my deluge later to manually test and update the PR with the results.

gtanner:
Did some testing and this does not really fix the problem, will need to take a better look at this

gtanner:
seems fixed by https://github.com/SynthstromAudible/DelugeFirmware/pull/4422

PR #4104, opened 2025-10-01
midi.guide csv to Deluge xml MIDI template converter

stosem:
We have a large database of MIDI device templates in the following resources:
* https://midi.guide
* https://github.com/pencilresearch/midi

But unfortunately they are not supported Deluge yet. The next small python script allows us to converts a CSV file from these resources into the Deluge's XML file for the sd_card/MIDI/ device templates.

Just download the desired CSV file and pass it as a parameter to this python script.
This script can be a part of Deluge's contrib, but it would be nice to have a native Deluge integration in the midi.guide website ;)


seangoodvibes:
Thanks for this, would it be possible to run the script and generate all the midi device definitions files and include them in your PR for upload onto the website? We have a definition folder under the contrib folder as well which might also be a better spot for the Python script.

https://github.com/SynthstromAudible/DelugeFirmware/tree/community/contrib/sd_card/MIDI_DEVICES/DEFINITION

contrib/sd_card/MIDI_DEVICES/DEFINITION/Elektron/Digitone II.xml
seangoodvibes, line 1: Looks like there's duplicate CC's in here

contrib/sd_card/MIDI_DEVICES/DEFINITION/Elektron/Digitone.xml
```diff
@@ -0,0 +1,128 @@
...
+		16="Portamento on"
+		90="Algorithm"
+		91="Ratio C"
+		92="Ratio A"
+		16="Ratio B"
+		17="Harmonics"
+		18="Detune"
+		19="Feedback"
+		20="Mix"
+		="Ratio C offset"
```
seangoodvibes: Missing CC

contrib/sd_card/MIDI_DEVICES/DEFINITION/Elektron/Digitone II.xml
stosem: Hm.. The original .csv also has CC duplications: https://github.com/pencilresearch/midi/blob/main/Elektron/Digitone%20II.csv 
How we can deal with it?

contrib/sd_card/MIDI_DEVICES/DEFINITION/Elektron/Digitone.xml
stosem: Thanks, @seangoodvibes ! 
Sometimes original .csv files have also empty values, like Cutoff here: https://github.com/pencilresearch/midi/blob/main/Waldorf/Iridium.csv , sometimes it's an nrpn values.
But I've solved this issue in the converter by simply excluding empty CCs.

PR #4137, opened 2025-10-11
kit sound display name fixes

Metamere:


src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -3657,37 +3662,41 @@ bool InstrumentClipView::handleNoteRowEditorPadAction(int32_t x, int32_t y, int3
 	return true;
 }
 
-// handles editing notes on the grid
```
seangoodvibes, line 3660: This PR is undoing recent changes

Metamere: Oh, that was an accident, as I only added a little bit. I must have had an out of date file somehow, so I'll try to fix that.

PR #4139, opened 2025-10-12
kit sound display name fixes

Metamere:

### DIFF
--- a/src/deluge/gui/ui/browser/sample_browser.cpp
+++ b/src/deluge/gui/ui/browser/sample_browser.cpp
@@ -994,7 +994,7 @@ doLoadAsSample:
 				drum->name.clear();
 
 				String newName;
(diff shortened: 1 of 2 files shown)


seangoodvibes:
@Metamere shouldn't this have targeted community?

I just noticed. You'll need to open a separate PR.

Thanks

seangoodvibes:
@Metamere can you please confirm if you will cherry pick this to community?

I don't want it to get forgotten

Thanks

Metamere:
I thought we were going to turn this into the community branch, so that's why I put it here. I will cherry pick onto community and start another PR if that's what is needed.

seangoodvibes:
> I thought we were going to turn this into the community branch, so that's why I put it here. I will cherry pick onto community and start another PR if that's what is needed.

No decision was made on that so yes it is needed thanks.

PR #4140, opened 2025-10-12
fix(midi): boost midi performance when doing sysex loading

bfredl:
I haven't bisected it exactly yet but the performance of sysex firmware loading became very slow a few months ago. I get the sense that MIDI processing is polling driven now which means MIDI is being throttled by the scheduler. Is is probably a good default but makes the MIDI engine really struggle with processing multiple megabytes of data.

As an easy hack, tenfold the target speed of the MIDI engine task when an update is incoming (only when compile time enabled) I kinda expeted just lowering the `backOffPeriod` should be enough for when playback is not active and nothing really should compete with the CPU, but I had to decrease the `targetInterval` as well.

Perhaps there is a less special cased way e.g. the MIDI engine could observe that more data is being queued on the hardware level and request the next callback to happen ASAP. I could investigate more.

### DIFF
--- a/src/OSLikeStuff/scheduler_api.h
+++ b/src/OSLikeStuff/scheduler_api.h
@@ -68,6 +68,7 @@ double getAverageRunTimeforCurrentTask();
 double getSystemTime();
 void setNextRunTimeforCurrentTask(double seconds);
 void removeTask(TaskID id);
+void boostTask(TaskID id);
 void yield(RunCondition until);
 /// timeout in seconds, returns whether the condition was met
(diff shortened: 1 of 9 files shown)


Metamere:
I thought it was just taking longer because the bin size is larger now, but I guess it was a bigger change than that would account for. In the last couple of weeks I noticed that occasionally the firmware would load twice as fast as usual, and I thought something was going wrong, but it seemed that it loaded correctly. So something is funky with it. Having faster build/load/test cycles would be a nice development accelerator for when you need to futz with OLED adjustments or bisect your way through a big set of commits.

PR #4148, opened 2025-10-16
Bugfix/notes ignoring conditions

### LINKED ISSUE #3491: Row in fill mode triggering by itself on the Nightly FW

I'm wondering if anyone experienced a row on a drum kit in fill mode (a clap sound for example) triggering some of the notes itself on the nightly FW ? when I switch to Chopin its totally fine but if I go back to nightly build its triggering some notes on the fill row again and some other notes on the different synth tracks also. Is it an issue with the unfinished nightly build? Btw I composed the track on Chopin. But also if I compose a brand new track on the latest nightly build, its doing it again. 

a kit row in fill mode has to play only when I press the sync-scaling button

I'm experiencing this issue at least for over 3 months. I'm frequently switching between Chopin and Nightly but the is issue still there

no

OLED

Nightly

deluge-v1_3_0-nightly+2025_03_19-ba32521

_No response_

### LINKED ISSUE #3933: Sometimes fill doesn't work

Was jamming with 1.3 a few days ago and was using fill and noticed that sometimes it didn't activate properly.

Specifically I had set 3 note rows of FM kick drums to "not fill" and whenever I pressed sync scaling it seemed like some notes would still play. But when I muted a couple rows, it seemed to work properly.

Someone else reported a similar issue on discord.

All notes respect their fill setting.

Few days ago

_No response_

OLED

Beta

_No response_

_No response_

### LINKED ISSUE #4264: bug: Drone notes re-trigger endlessly

Create a new synth
Add 5 drone notes
Hit play

You will hear that the drone notes continue to retrigger every time the deluge loops around. It causes excessive voices to be created and leads to culling and distortion of sound.

Bisecting points to this PR as the cause: https://github.com/SynthstromAudible/DelugeFirmware/pull/3229

Video of the problem:

https://github.com/user-attachments/assets/b9b42104-b0db-4ac5-9144-c4ec26d023c4

It should sound like this:

https://github.com/user-attachments/assets/3ddcdfff-3921-460e-8e81-bbc41b830f2b

Today

3229

OLED

Beta

_No response_

_No response_

m-m-adams:
Note: Needs at least some user testing, I'm not sure if I've covered all possible combinations of arp modes properly 

Fix #3933
Fix #3491
Fix #4264


m-m-adams:
Fix #3491 as well

m-m-adams:
I'm pretty sure this is good to go - tested the kit arp, drum arp, and midi arp without issues

soymonitus:
SHouldn't noteStatus be reset to OFF when in the code we are also filling the rest of noteCode elements of the array to ARP_NOTE_NONE? Like for example here

```
for (int32_t n = 1; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
  // Clean rest of chord note slots
  active_note.noteCodeOnPostArp[n] = ARP_NOTE_NONE;
}
```

Or is it intentional that when you cleanup the array of note codes to ARP_NOTE_NONE, that you leave the noteStatus untouched? for example when a note evaluates to "play chord" all the notes in the array may get filled with values, but if the next note is "normal note", just element 0 is filled, and the rest up to ARP_MAX_INSTRUCTION_NOTES are filled with ARP_NOTE_NONE, but noteStatus from n=1 to max, could have PENDING or PLAYING values.

Saying this because there is a function in arpeggiator.h that gives isPending as true is ANY of the values has pending status:
```
	bool isPending() {
		return std::ranges::any_of(noteStatus, [](ArpNoteStatus status) { return status == ArpNoteStatus::PENDING; });
	}
```

src/deluge/model/instrument/kit.cpp
```diff
@@ -821,6 +822,7 @@ void Kit::renderNonAudioArpPostOutput(deluge::dsp::StereoBuffer<q31_t> output) {
 			}
 			if (instruction.arpNoteOn != nullptr) {
 				for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
+					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
```
soymonitus: Why setting it to PLAYING before checking if the notecode is ARP_NOTE_NONE and the corresponding "break" on the next line? the index of the array for this note would end up being PLAYING while the corresponding noteCode is ARP_NOTE_NONE.

src/deluge/model/instrument/kit.cpp
```diff
@@ -768,6 +768,7 @@ void Kit::setupAndRenderArpPreOutput(ModelStackWithTimelineCounter* modelStackWi
 					                          kitInstruction.arpNoteOn->mpeValues, 0, kitInstruction.sampleSyncLengthOn,
 					                          0, 0);
 				}
+				kitInstruction.arpNoteOn->noteStatus[0] = ArpNoteStatus::PLAYING;
```
soymonitus: I have two questions:
1. Why did you set just index 0 of the array to playing?
2.  In this same file there is another place (doTickForwardForArp) where it calls thisNoteRow->drum->noteOn, why not adding the same like to set status to playing in the other place?

src/deluge/model/instrument/non_audio_instrument.cpp
```diff
@@ -66,6 +66,7 @@ void NonAudioInstrument::renderOutput(ModelStack* modelStack, deluge::dsp::Stere
 					if (instruction.arpNoteOn->noteCodeOnPostArp[n] == ARP_NOTE_NONE) {
 						break;
 					}
+					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
```
soymonitus, line 69: Here you correctly set note status AFTER the conditional "break"

src/deluge/modulation/arpeggiator.h
```diff
@@ -197,15 +209,16 @@ class ArpReturnInstruction {
 
 class ArpeggiatorBase {
 public:
+	virtual ~ArpeggiatorBase() = default;
 	ArpeggiatorBase() {
-		noteCodeCurrentlyOnPostArp.fill(ARP_NOTE_NONE);
 		glideNoteCodeCurrentlyOnPostArp.fill(ARP_NOTE_NONE);
-		outputMIDIChannelForNoteCurrentlyOnPostArp.fill(0);
```
soymonitus, line 203: is outputMIDIChannelForNoteCurrentlyOnPostArp not used anymore? how do you track now the output channel for the current notes?|
hmmmmm is this because this all moved to be inside arpNote class? just in one place?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -1215,7 +1225,7 @@ void Arpeggiator::rearrangePatterntArpNotes(ArpeggiatorSettings* settings) {
 		arpByPatternNote->noteCode = arpNote->inputCharacteristics[util::to_underlying(MIDICharacteristic::NOTE)];
 	}
 }
-
+// todo: make sure reference taken after copy to activeNote, seems to be missed for first note on while arp active
```
soymonitus: What is this TODO about?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -1339,61 +1348,61 @@ void Arpeggiator::switchNoteOn(ArpeggiatorSettings* settings, ArpReturnInstructi
 
 		// Wipe noteOn codes
 		for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
-			noteCodeCurrentlyOnPostArp[n] = ARP_NOTE_NONE;
 			arpNote->noteCodeOnPostArp[n] = ARP_NOTE_NONE;
+			arpNote->noteStatus[n] = ArpNoteStatus::OFF;
```
soymonitus: See my previous comment about you missing this "reset" of status values in the array in some places of the code

soymonitus: Search in the code for " = ARP_NOTE_NONE;" and you will find all the places where note values are reset. Worth a look to see if there are some more places where it could be important to also reset noteStatus state.

src/deluge/model/instrument/kit.cpp
m-m-adams: Kits can't play chords can they? I don't see a way for any other notes in the ArpNote to be played

src/deluge/modulation/arpeggiator.h
m-m-adams: Yeah, this is moved to be tracked inside the active_note member now. The glide note could be refactored to also be an arp note but I decided that's out of scope for now in the interest of not breaking stuff

src/deluge/modulation/arpeggiator.cpp
m-m-adams: Forgot to remove it, I had a bug at one point where note ons with the active arp would be sent twice

src/deluge/model/instrument/kit.cpp
soymonitus: Ok to the comment about chords not available in kits.

On the other hand I would put this line inside the "if (thisNoteRow->drum != nullptr) {" (as it is already done in kit.cpp inside doTickForwardForArp method)

src/deluge/model/instrument/kit.cpp
```diff
@@ -1195,6 +1197,8 @@ int32_t Kit::doTickForwardForArp(ModelStack* modelStack, int32_t currentPos) {
 				thisNoteRow->drum->noteOn(modelStackWithThreeMainThings, kitInstruction.arpNoteOn->velocity,
 				                          kitInstruction.arpNoteOn->mpeValues, 0, kitInstruction.sampleSyncLengthOn, 0,
 				                          0);
+				// no check needed - will be held by the drum's own arp if it can't start immediately
+				kitInstruction.arpNoteOn->noteStatus[0] = ArpNoteStatus::PLAYING;
```
soymonitus, line 1201: Why is this done AFTER having called drum->noteOn while for nonAudioDrums (in this same file) it is set BEFORE calling noteOnPostArp? I feel the logic is a bit obscure here and hard to follow, maybe I am missing something

src/deluge/model/instrument/kit.cpp
```diff
@@ -824,6 +825,7 @@ void Kit::renderNonAudioArpPostOutput(deluge::dsp::StereoBuffer<q31_t> output) {
 					if (instruction.arpNoteOn->noteCodeOnPostArp[n] == ARP_NOTE_NONE) {
 						break;
 					}
+					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
```
soymonitus, line 828: if you do a search for "= ArpNoteStatus::PLAYING" you will find several places, for nonaudiodrums, nonaudioinstrument and sound. In each of them setting the status to PLAYING is done sometimes before the noteOnPostArp call and sometimes afterwards. Could we make it consistent and set it either all afterwards or either all before noteOnPostArp? in case it doesn't make any difference, just to keep consistency

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -216,15 +221,13 @@ void ArpeggiatorForDrum::noteOff(ArpeggiatorSettings* settings, int32_t noteCode
 	// If no arpeggiation...
 	if ((settings == nullptr) || settings->mode == ArpMode::OFF) {
 		instruction->noteCodeOffPostArp[0] = noteCodePreArp;
-		instruction->outputMIDIChannelOff[0] = arpNote.outputMemberChannel[0];
-		noteCodeCurrentlyOnPostArp[0] = ARP_NOTE_NONE;
-		outputMIDIChannelForNoteCurrentlyOnPostArp[0] = MIDI_CHANNEL_NONE;
+		instruction->outputMIDIChannelOff[0] = active_note.outputMemberChannel[0];
+		active_note.noteCodeOnPostArp[0] = ARP_NOTE_NONE;
+		active_note.outputMemberChannel[0] = MIDI_CHANNEL_NONE;
```
soymonitus, line 226: Do we need here to also set notestatus to OFF to keep consistency with other places where we assign ARP_NOTE_NONE to the active_note noteCodeOnPostArp?

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -1407,13 +1428,25 @@ bool Arpeggiator::hasAnyInputNotesActive() {
...
+	return active_note.velocity;
 }
 
+bool ArpeggiatorBase::checkPendingNotes(ArpeggiatorSettings* settings, ArpReturnInstruction* instruction) {
+	if (active_note.isPending()) {
+		ARP_PRINTLN("found a pending drum note");
+		instruction->arpNoteOn = &active_note;
+		return true;
+	}
+	instruction->arpNoteOn = nullptr;
```
soymonitus, line 1441: Could this be an antipattern to have a method that is called "check.." to produce a side effect of nulling the arpNoteOn from the instruction?

Not sure why this is needed and what kind of side effects it is meant to produce. Could you check the usages of checkPendingNotes and see if nulling the arpNoteOn is not an undesirable thing to happen? Maybe could it be refactored to remove the antipattern?

src/deluge/processing/sound/sound.cpp
```diff
@@ -1540,11 +1540,19 @@ void Sound::noteOn(ModelStackWithThreeMainThings* modelStack, ArpeggiatorBase* a
 			if (instruction.arpNoteOn->noteCodeOnPostArp[n] == ARP_NOTE_NONE) {
 				break;
 			}
-			atLeastOneNoteOn = true;
-			invertReversed = instruction.invertReversed;
-			noteOnPostArpeggiator(modelStackWithSoundFlags, noteCodePreArp, instruction.arpNoteOn->noteCodeOnPostArp[n],
-			                      instruction.arpNoteOn->velocity, mpeValues, instruction.sampleSyncLengthOn, ticksLate,
-			                      samplesLate, fromMIDIChannel);
+			if (AudioEngine::allowedToStartVoice()) {
```
soymonitus, line 1543: Just checking here to see if this was intentional. You added here an "if" for AudioEngine::allowedToStartVoice while you removed it in note_row.cpp. My guess is that at  note_row level it was not needed as that will be handled internally by each sound.

soymonitus (commented):
Some more comments!

src/deluge/model/instrument/kit.cpp
m-m-adams: It doesn't matter, it's just marking the note as being dealt with

src/deluge/model/instrument/kit.cpp
m-m-adams: They don't matter though and anything making it matter would probably be a bug. It's just marking the note as dealt with and not needing to be played a second time

src/deluge/modulation/arpeggiator.cpp
m-m-adams: This is the main part, it's looking for whether there's a pending note and setting it as the arp return instruction if found. That's how the arp ends up finding notes that were delayed a tick

I'll rename and add a doc comment for clarity

src/deluge/processing/sound/sound.cpp
m-m-adams: Yeah that's the change, it's now pushed down to be checked within the pending note logic of the arp instead of the sequencer. That means that the max note on limits are now applied after the note has been fully selected to play so it avoids interfering with the sum to 100 logic and any other sequencing parameters

PR #4153, opened 2025-10-18
Fix automation menu view crash

### LINKED ISSUE #3592: Bug with entering automation view editor from menu

Can't enter automation view editor from menu when automation view was the Root UI prior to entering the menu.

You can enter the automation view editor fine if automation view was not the root UI prior to entering the menu.

Pressing song (in arranger) or clip (in clip view) while in a parameter menu should open automation editor.

Today

_No response_

OLED

Nightly

_No response_

_No response_

seangoodvibes:
Fixed crash that would occur when turning select encoder after entering menu from automation view

Also fixed issue where the selected parameter in automation view was not updating when changing the horizontal menu item selection

Fix #3592

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -376,7 +376,7 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 					goUpOneLevel();
 				}
 
-				handlePotentialParamMenuChange(b, inCardRoutine, currentMenuItem, getCurrentMenuItem());
+				handlePotentialParamMenuChange(b, inCardRoutine, currentMenuItem, getCurrentMenuItem(), false);
 
 				if (currentUIMode == UI_MODE_HOLDING_AFFECT_ENTIRE_IN_SOUND_EDITOR
 				    && getCurrentMenuItem()->usesAffectEntire() && editingKit()) {
@@ -420,7 +420,7 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 					goUpOneLevel();
 				}
 
-				handlePotentialParamMenuChange(b, inCardRoutine, currentMenuItem, getCurrentMenuItem());
+				handlePotentialParamMenuChange(b, inCardRoutine, currentMenuItem, getCurrentMenuItem(), false);
 
 				if (currentUIMode == UI_MODE_HOLDING_AFFECT_ENTIRE_IN_SOUND_EDITOR
 				    && getCurrentMenuItem()->usesAffectEntire() && editingKit()) {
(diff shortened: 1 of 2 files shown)


src/deluge/gui/menu_item/submenu.cpp
```diff
@@ -182,8 +182,6 @@ void Submenu::selectEncoderAction(int32_t offset) {
 		} while (offset < 0);
 	}
 	updateDisplay();
-	updatePadLights();
```
seangoodvibes, line 185: updating pad lights and automation view parameter while scrolling select didn't make any sense so I removed these

updating automation view parameter was definitely causing a crash

src/deluge/gui/menu_item/horizontal_menu.cpp
```diff
@@ -299,7 +299,7 @@ void HorizontalMenu::switchVisiblePage(int32_t direction) {
 
 	renderUIsForOled();
 	updatePadLights();
-	(*current_item_)->updateAutomationViewParameter();
+	// (*current_item_)->updateAutomationViewParameter();
```
seangoodvibes: commenting all these calls out as I can't figure out why they cause crashes. seems to be because they're trying to update non automatable param menu items

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -582,7 +582,7 @@ void SoundEditor::handlePotentialParamMenuChange(deluge::hid::Button b, bool on,
 			if (getRootUI() == &automationView) {
 				// if on menu view, swap out root UI to previous UI
 				if (automationView.onMenuView) {
-					previousItem->buttonAction(b, on, inCardRoutine);
+					previousItem->buttonAction(b, true, inCardRoutine);
```
seangoodvibes, line 606: with the change to select encoder button action, it resulted in !on being passed. for this specific function we always want to pass true so there's no need to forward the actual "on" condition from the button press

src/deluge/gui/ui/sound_editor.cpp
```diff
@@ -959,10 +959,9 @@ void SoundEditor::selectEncoderAction(int8_t offset) {
 	MenuItem* item = getCurrentMenuItem();
 	RootUI* rootUI = getRootUI();
 
-	// Forward to automation view.
-	// - skipped for submenus, since vertical menus don't need it, and horizontal menus need extra case
-	// - TODO: this could be handled by the Automation class via regular forwarding for all cases
-	if (!item->isSubmenu() && rootUI == &automationView && isEditingAutomationViewParam()
+	// Forward to automation view if holding a note
```
seangoodvibes, line 982: updated this section to only forward to automation view if you're holding a step in automation view, otherwise forward to menu item

also removed extra condition below that I do not believe was necessary

seangoodvibes:
Draft until I incorporate latest feedback from Immers :)

PR #4227, opened 2025-12-06
Feature/DX7 Convert

bzawad:
### Converting DX7 Patch Banks **[1.4]**

If you have DX7 patch banks in .syx format that you want to convert to individual preset files, you can use the DX7 Convert feature:

1. Navigate to the DX7 SOURCE menu
2. Find the "CONVERT DX7 .SYX FILE" option
3. Select this option to open a file browser showing available .syx files
4. Select a .syx file to convert it
5. The converter will create a new folder in `SYNTHS/DX7/` with the same name as the .syx file
6. Individual patches from the bank will be saved as .XML files in this folder
7. You can then load these individual presets using the regular DX7 loading workflow

The conversion process:
- If a folder already exists for the SYX then the message ALREADY CONVERTED will show.
- Presets with empty names and duplicate names are skipped
- Existing converted presets are not overwritten (they will be skipped)
- Invalid filename characters in preset names are replaced with underscores
- A progress indicator shows while presets are being converted and a completion message shows how many presets were converted and how many were skipped

Note: When converting a SYX file, the final preset that was converted will remain loaded in a temporary slot. However, once you power the unit off and back on, that temporary preset is cleared just like any other unsaved instrument.

image


seangoodvibes:
@bzawad checks are failing fyi

bzawad:
> @bzawad checks are failing fyi

@seangoodvibes I excluded the dx7_converter from the memory tests since I'm not sure how else to handle this.  The converter has a dependency on song/song.h since it needs that to create the instrument but that brings in a chain of dependencies.  I tried including argon but caused a bunch of test failures unrelated to the converter as far as I can tell.

Let me know if this works or if you have any other ideas.

Also let me know if any of the remaining clang issues need to be fixed and I'll get those.  I didn't want to keep pushing things through CI.  Thanks!

m-m-adams:
Could you rebase onto main? We’ve rolled back a pile of problematic refactorings

bzawad:
@m-m-adams @seangoodvibes This is rebased to the new main now.

PR #4239, opened 2025-12-16
bugfix: default settings for upstream port 2 (beta)

m-m-adams:
Same as #4237 but for beta


seangoodvibes (approved):

seangoodvibes:
Can you please update the menu hierarchies documentation and add a note in the change log about this?

Thanks

PR #4240, opened 2025-12-16
Bugfix: send note off before replacing note in case of duplicate

m-m-adams:
When both sending and receiving from MPE the deluge could cause dropped notes when the same note code is sent on two different channels. Easy fix is turning off the first note (matches what the internal synth engine does).


seangoodvibes:
Can you please update the menu hierarchies documentation and add a note in the change log about this?

Thanks

Issue #4223, opened 2025-11-29
Crash when scrolling AUDIO FILES menu too quickly


bzawad:
### Please describe the problem:

I'm seeing a consistent crash when scrolling files in AUDIO FILES menu too quickly.  This seems to be caused by PR #4055 after tracking down which commit it started in.

<img width="592" height="397" alt="Image" src="https://github.com/user-attachments/assets/6ed0609a-6574-4079-8d32-1ef08356f92c" />

### What is the expected behavior?

In 1.3 and prior to the PR you can scroll through audio files quickly without crashes.

### When did you start noticing the issue?

I noticed this testing my own feature and that it happens only in 1.4 and not 1.3 or 1.2.

### Is there a relevant Pull request?

#4055

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Built from community

### What is the firmware name:

community 1.4

### If possible provide the steps to reproduce the issue and upload additional media:

git checkout ddb9e054cb21f5cf4e95450fdae16206330a9512
./dbt build release
Flash the bin
Create an audio track or try to load an audio file to a synth.
At the AUDIO FILES go to a folder full of audio files and scroll fast.  
Everything is ok.

git checkout b9da59672091c592ee352abd0a097451b59ed281
./dbt build release
Flash the bin
Create an audio track or try to load an audio file to a synth.
At the AUDIO FILES go to a folder full of audio files and scroll fast.  

You may see SD Card Error messages or a crash.  It's easier to crash then to get the SD Card Error mesage.  This happens when you scroll too fast which might not be that fast as all.  If you scroll slowly then things might be ok.

The same crash also happens in community head as of 2025-11-28.

stosem:
The same bug: https://github.com/SynthstromAudible/DelugeFirmware/issues/3809

Issue #4098, opened 2025-09-28
Simple freeze


brandonhabanero:
### Please describe the problem:

The Deluge froze, displayed an error message ("E058"), and the pad grid lit randomly after scrolling quickly through kit presets.

![Image](https://github.com/user-attachments/assets/e331d7cb-16b7-4eb1-a19b-82ba3d1f89db)

### What is the expected behavior?

The Deluge should load kit presets when scrolling through them normally, or at the very least, briefly pause scrolling through presets to allow itself enough time to load each one it scrolls through.

### When did you start noticing the issue?

9/25/2025; however, I haven't been able to reproduce this issue since then. 

### Is there a relevant Pull request?

No.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.3.0 BETA-7331FCA

### If possible provide the steps to reproduce the issue and upload additional media:

1. Turned on Deluge.
2. Change first track to a midi clip and tap in a short melody.
3. Navigate to song mode (rows).
4. Create new track.
5. Enter clip edit view.
6. Change clip to kit.
7. Scroll through presets as fast as possible.

Issue #3809, opened 2025-06-03
Crash with E078 while browsing sample-based presets


stosem:
### Please describe the problem:

Deluge freezes with error `E078` when I browse Boards of Deluge 7 presets (which are heavy FX and are mostly sample based). I basically play a few notes and select the next preset, then Deluge crashes on a random preset.
Confirmed in the latest beta 1.3.0-880AE4A, but also exists in alpha and previous versions. And not only with BoD7, but BoD6, BoD5 as well (which are also sample-based). It seems OK with BoD4 (no samples here).

![Image](https://github.com/user-attachments/assets/3ed97f83-69b7-4c3c-a646-b6a28affbbba)

### What is the expected behavior?

Deluge shouldn't crash.

### When did you start noticing the issue?

just now and before

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.3.0-BETA-880AE4A

### If possible provide the steps to reproduce the issue and upload additional media:

1. Turned on
2. Created a new synth clip
3. Load + Synth
4. Go to BoD presets
5. Choose a preset
6. Play a few notes
7. Switch to next
8. Repeat until Deluge crash with error (usually after 10-15 presets).

What other details could I provide that would be helpful?

m-m-adams:
FYI this won't crash in the final release, it's a debug feature for betas. It's indicating a gremlin in the code that's been around since official v4.0 but hopefully we'll find it soon

stosem:
BTW, if I press SELECT, it comes back, but now crashing more frequently (every second preset). Is there any way to debug this?

m-m-adams:
If you can get it to happen reliably then sharing your steps might help find a fix but otherwise just use 1.2.1 and it won't crash. There's nothing user facing to debug and we don't know the root cause

m-m-adams:
As far as I can tell this is fixed in nightly by #3617 

We aren't going to be able to fix it in beta as this has been around for years and needed a large refactoring. The freeze on this error code will be disabled automatically when the beta is changed to a release candidate

stosem:
Thanks, @m-m-adams ! I can confirm, that I've not faced this issue in c1.2.1.
Unfortunately, I can still reproduce this bug in the latest nightly build: FW: v1_3_0-nightly+2025_06_20-893bf4b
[...]

PR #4252, opened 2025-12-30
BugFix/ Release With Cluster Loading (Nightly)

seangoodvibes:
Updated routineWithClusterLoading calls to fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4223. Will cherry-pick these changes to 1.3 separately.

Also related to these issues which seem to be present in 1.3 which indicates I think that some of the yields might be problematic in 1.3 as well. I'll look at a 1.3 specific PR to address those if I can reproduce the crashes.
- https://github.com/SynthstromAudible/DelugeFirmware/issues/3809
- https://github.com/SynthstromAudible/DelugeFirmware/issues/4098

**Summary**

Depending on where RWCL is called from, the Deluge may not respond as expected if you Yield to Audio instead of a direct call to the Audio Routine. From my testing, this is a very delicate area to touch.

As part of this commit I have identified the problematic RWCL calls which:

- cause crashes when browsing for files on the SD card.
- cause the Sample Recorder in Kit's to not function as expected (gives SD Card Error).
- cause song loading to not function as expected (gives an F1 error when it shouldn't).

To fix the above, instead of yielding, we can continue to run the audio routine directly, but run it through the scheduler so that the scheduler is aware.

### DIFF
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -380,7 +380,7 @@ void routineWithClusterLoading(bool mayProcessUserActionsBetween) {
 
 void runRoutine() {
 	// check if we've setup the audio routine task
-	if (routine_task_id != -1) {
+	if (routine_task_id != -1) [[likely]] {
 		// run AudioEngine::routine() task so that scheduler is aware
 		runTask(AudioEngine::routine_task_id);
 	}


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -381,16 +381,34 @@ void routineWithClusterLoading(bool mayProcessUserActionsBetween, bool useYield)
...
+		            runTask(AudioEngine::routine_task_id);
+		        }
+		        else {
+		            AudioEngine::routine();
+		        }
+		    }
+		*/
+		//	if (!audioRoutineLocked) {
+		if (AudioEngine::routine_task_id != -1) {
+			runTask(AudioEngine::routine_task_id);
```
m-m-adams: oh nice yeah that should work and way simpler than what I was imagining. maybe just centralize this logic in a function so it's not duplicated in a few places

seangoodvibes: Sweet thanks for taking a look, will do!

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -362,16 +362,33 @@ int32_t getNumVoices() {
...
+		logAction("call runRoutine() from routineWithClusterLoading()");
+		runRoutine();
 	}
 }
 
-void routineWithClusterLoading(bool mayProcessUserActionsBetween) {
-	yieldToAudio();
+void runRoutine() {
+	// check if we've setup the audio routine task
+	if (routine_task_id != -1) {
```
m-m-adams: mark this path as likely please (tag as `[[likely]]`), otherwise I think it's good to go

m-m-adams (approved):

Issue #3578, opened 2025-04-09
Short glitchy sound when currently playing and scheduling next song load (1.3 alpha)


lburygin:
### Please describe the problem:

If you play the attached song and while playing, press LOAD and choose the same song as currently playing, you will hear a short but very noticeable glitchy stutter-ish sound

For some reason it occurs on stable basis only with this song for me (at the moment)

### What is the expected behavior?

When scheduling next song to load, there shouldn't be any audio artifacts

### When did you start noticing the issue?

Since 1.3 alpha, doesn't reproduce on the 1.2 chopin

### Is there a relevant Pull request?

No AFAIK

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.3.0-BETA-72AA06F

### If possible provide the steps to reproduce the issue and upload additional media:

1. load the attached song
2. press play
3. press load
4. choose the attached song from the list and press load (or select)

[glitchy loading.zip](https://github.com/user-attachments/files/19665284/glitchy.loading.zip)

lburygin:
Reproduced this on other CPU-heavy songs. So a more general case is:
- make a very CPU heavy song, you can use BOD 6-7 patches to quickly eat all CPU
- save the song
- play the song
- press load
- choose the song from the menu and press load

seangoodvibes:
I think you're probably hearing the initialize of the grain FX in that song. It's a known bug atm. If you change the mod fx type on "SYNT3" to something other than Grain FX I think it should load ok.

Unfortunately there's no solution for this yet. As @m-m-adams explained on discord:

"it’s the allocation time for the 4mb buffer, if it has to steal clusters it takes too long and causes the beep

one solution might be to use @stellar_aria’s object pool for the buffer and keep one pre allocated"

lburygin:
Yeah confirmed, the grain is the cause of this buzzy sound

seangoodvibes:
This PR fixed this buzz sound in nightly: https://github.com/SynthstromAudible/DelugeFirmware/pull/3962

Should be able to port the fix to beta.

Issue #3881, opened 2025-06-25
Performance drop on 1.3 beta compared to 1.2 fw


lburygin:
### Please describe the problem:

I tested some scenarios and noticed the difference between 1.2 and 1.3 beta fw.

1. Raw synth voices test: 70 long notes, with the length of the clip length, playing simultaneously.
- 1.2: stable 55 voices, no fluctuations
- 1.3: fluctuates between 51 and 55 voices
- [bonus] 1.1:  fluctuates between 63-67 voices

2. Raw samples in memory test: 40 samples with the length of 4 bars playing simultaneously 
- 1.2: stable 29 voices, no fluctuations
- 1.3: in average 28 voices but often drops to 5-10 voices for a short time
- [bonus] 1.1 : fluctuates between 32-37 voices

3. Raw streamed samples test: 14 kit rows with samples, 400mb each, playing in parallel
- 1.2: 12 voices
- 1.3: 12 voices
On par here :))

4. Raw streamed samples mixed with raw synth voices test: 14 kit rows with samples, 400mb each, plus 24 synth notes playing in parallel
- 1.2: fluctuates between 28-35 voices
- 1.3: fluctuates between 16-35 voices, with average at about 23
- [bonus] 1.1: more stable 34-36 voices

In addition (this is a positive thing but just for info), in the last test there were noticeable clicks at 1.1 fw, clicks also were there on 1.2 fw but softer, and on 1.3 fw there were no clicks at all

### What is the expected behavior?

Perfomance should be at least on par with 1.2 fw

### When did you start noticing the issue?

Today

### Is there a relevant Pull request?

No

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

866edb8be07032e3143859c5f6e03d304a4cf56a

### If possible provide the steps to reproduce the issue and upload additional media:

-

m-m-adams:
![Image](https://github.com/user-attachments/assets/0c2d9b9b-83b7-4542-9c76-f98fc490bd87)

![Image](https://github.com/user-attachments/assets/328c3a2e-a132-420f-8490-7ef2ea4bf4d3)

A main cause of the bug seems to be an issue where UI rendering and input processing takes a bizarrely long time which then messes up the schedule for a while

lburygin:
Does the time depend on the UI content complexity? (e.g. simple numeric menu vs horizontal menu with lots of graphics)

[...]

PR #4276, opened 2026-01-08
BugFix/ StereoSample initialization performance issue

seangoodvibes:
Initializing the StereoSample l and r variables to zero introduces performance issues. It is most notable with the initialization of GrainFX where a buzz sound is heard.

This may have also been the culprit of other performance issues.

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/3578


seangoodvibes:
@lburygin could you also check if this fix helps at all with your other issue?

https://github.com/SynthstromAudible/DelugeFirmware/issues/3881

Thanks!

PR #4280, opened 2026-01-11
Performance: replace dual maps with sorted vector

m-m-adams:
Don't eagerly erase to avoid thrashing - will be erased later when they reach front of queue

Use a sorted queue to manage the cluster loading instead of two maps. Reduces allocations and keeps them contiguous


m-m-adams:
putting it in draft because I want to play with some options for erase

seangoodvibes (approved):

PR #4295, opened 2026-02-01
Performance/revert afm changes

m-m-adams:
Removes the maps and returns the use of the resizeable array. This seems to improve performance but I think there's also a bug with culling. Voices get culled with render times around 85 samples which should be totally fine

### DIFF
--- a/src/deluge/model/global_effectable/global_effectable_for_clip.cpp
+++ b/src/deluge/model/global_effectable/global_effectable_for_clip.cpp
@@ -119,9 +119,10 @@ GlobalEffectableForClip::GlobalEffectableForClip() {
 
 	// Render saturation
 	if (clippingAmount != 0u) {
+		int32_t shiftAmount = getShiftAmountForSaturation();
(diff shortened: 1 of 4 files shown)


src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -332,9 +334,10 @@ void forceReleaseOneVoice(size_t num_samples) {
...
 	const Sound::ActiveVoice* best = &all_voices.front();
 	for (const auto& voice : all_voices | std::views::drop(1)) {
-		// if the voice is already releasing faste than this we'd rather release another voice
-		if (voice->envelopes[0].state >= EnvelopeStage::FAST_RELEASE
-		    && voice->envelopes[0].fastReleaseIncrement >= 4096) {
+		// if a voice is already fast releasing just speed it up
+		if (voice->envelopes[0].state == EnvelopeStage::FAST_RELEASE) {
+			voice->speedUpRelease();
+			return;
 			continue;
```
seangoodvibes: do we need the continue here?

PR #4308, opened 2026-02-08
Update website downloads

seangoodvibes:
Updated website downloads

Remove 1.3 beta only keep nightly


website/src/content/docs/downloads.mdx
m-m-adams: This should be main now

seangoodvibes:
```suggestion
    href="https://github.com/SynthstromAudible/DelugeFirmware/blob/main/CHANGELOG.md"
```

PR #4315, opened 2026-02-15
Fix kit rows sending 64 MIDI note-on messages when arpeggiator is dis…

alifeinbinary:
…abled

Bug: When a kit row with a MIDI drum had its individual arpeggiator disabled but was included in the kit arpeggiator, it would send 64 note-on messages per note during playback instead of a single message. This occurred only during playback, not during auditioning.

Root cause: The doTickForward() function called handlePendingNotes() before checking if the arpeggiator was disabled. When disabled, handlePendingNotes() would find pending notes and return true, causing doTickForward() to return 0 (immediate callback) and create an infinite loop that sent repeated MIDI messages.

Fix: Moved the handlePendingNotes() call to after the ArpMode::OFF check in doTickForward(). Now when the arpeggiator is disabled, the function returns early without processing pending notes, preventing the infinite loop while preserving normal operation when the arpeggiator is enabled.

Testing: Verified that MIDI drums now send exactly one note-on message per note during playback when their arpeggiator is disabled, and normal arpeggiator/auditioning functionality remains intact.

### DIFF
--- a/src/deluge/modulation/arpeggiator.cpp
+++ b/src/deluge/modulation/arpeggiator.cpp
@@ -1436,11 +1436,6 @@ bool ArpeggiatorForDrum::hasAnyInputNotesActive() {
 bool ArpeggiatorBase::handlePendingNotes(ArpeggiatorSettings* settings, ArpReturnInstruction* instruction) {
 	if (active_note.isPending()) {
 		instruction->arpNoteOn = &active_note;
-		for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
-			if (active_note.noteStatus[n] == ArpNoteStatus::PENDING) {
-				active_note.noteStatus[n] = ArpNoteStatus::PLAYING;
-			}
-		}
 		return true;
 	}
 	instruction->arpNoteOn = nullptr;
--- a/src/deluge/model/instrument/kit.cpp
+++ b/src/deluge/model/instrument/kit.cpp
@@ -1265,6 +1265,7 @@ int32_t Kit::doTickForwardForArp(ModelStack* modelStack, int32_t currentPos) {
 						break;
 					}
 					nonAudioDrum->noteOnPostArp(instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn, n);
+					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
 				}
 			}
 


src/deluge/modulation/arpeggiator.cpp
```diff
@@ -1517,15 +1517,14 @@ int32_t ArpeggiatorBase::doTickForward(ArpeggiatorSettings* settings, ArpReturnI
 	if (clipCurrentPos == 0) {
 		notesPlayedFromLockedRandomizer = 0;
 	}
-	if (handlePendingNotes(settings, instruction)) {
```
m-m-adams, line 1520: This will break pending notes. They need to pick up immediately

alifeinbinary:
Thanks for the feedback @m-m-adams . I reverted the check at line 1527 in arpeggiator.cpp back to its original position and followed your advice by marking the notes as PLAYING after the direct noteOnPostArp call in midi_drum.cpp. arpeggiator.cpp also marks PLAYING as a defensive backstop for any other callers that don't immediately mark the trigger status. The same issue existed in gate_drum.cpp so the solution is applied there as well. 

I tested kit and instrument arpeggiators and all work as expected.

src/deluge/modulation/arpeggiator.cpp
```diff
@@ -1435,6 +1436,11 @@ bool ArpeggiatorForDrum::hasAnyInputNotesActive() {
 bool ArpeggiatorBase::handlePendingNotes(ArpeggiatorSettings* settings, ArpReturnInstruction* instruction) {
 	if (active_note.isPending()) {
 		instruction->arpNoteOn = &active_note;
+		for (int32_t n = 0; n < ARP_MAX_INSTRUCTION_NOTES; n++) {
```
m-m-adams: They can actually end up pending for more than one render cycle if there's a ton of notes being processed and the cpu load is already high

I'll try to think of a more robust way to handle this in general but if you remove this bit I'll merge in your fix for now 

Thanks a ton!!

alifeinbinary: I hadn't tested it under heavy load but I see what you're saying. I removed that loop from arpeggiator.cpp and added one line to kit.app at 1268 inside the ARP_MAX_INSTRUCTION_NOTES for loop

`instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;`

It marks PLAYING at the point of actual playback in the per-drum tick loop. 

I tested under load with a few CPU heavy BoD synths sequenced (the play button was flickering) it and it work dependably without any stuttering or dropped beats.

m-m-adams (approved):

ogabrielluiz:
Hi,
Awesome work! Will this be merged into `community` soon?

alifeinbinary:
@ogabrielluiz community branch is deprecated I believe, main is the branch to use going forward. This has been merged into main. If you build from main it will contain the fix.

ogabrielluiz:
> @ogabrielluiz community branch is deprecated I believe, main is the branch to use going forward. This has been merged into main. If you build from main it will contain the fix.

Thank you! Good to know

PR #4332, opened 2026-03-03
Add community setting to swap tempo and select encoders

CaptPat:
Adds a community feature toggle to swap the select and tempo encoder functions (Settings > Community Features). When enabled, the physical tempo encoder navigates menus and the physical select encoder changes tempo, including correct button press detection for tempo tap. Uses the existing `RuntimeFeatureSettings` pattern.

Discussion #4206.

## Test plan
- [ ] Toggle on → physical tempo encoder navigates menus, physical select encoder changes tempo
- [ ] Toggle off → default behavior restored
- [ ] Setting persists across power cycles

### DIFF
--- a/src/deluge/hid/encoders.cpp
+++ b/src/deluge/hid/encoders.cpp
@@ -163,16 +163,14 @@ checkResult:
 				     || (getCurrentUI() == &automationView && automationView.inNoteEditor()))
 				    && runtimeFeatureSettings.get(RuntimeFeatureSettingType::Quantize)
 				           == RuntimeFeatureStateToggle::On) {
-					instrumentClipView.tempoEncoderAction(
-					    limitedDetentPos,
-					    Buttons::isButtonPressed(deluge::hid::button::TEMPO_ENC),
-					    Buttons::isShiftButtonPressed());
+					instrumentClipView.tempoEncoderAction(limitedDetentPos,
+					                                      Buttons::isButtonPressed(deluge::hid::button::TEMPO_ENC),
+					                                      Buttons::isShiftButtonPressed());
 				}
 				else {
-					playbackHandler.tempoEncoderAction(
-					    limitedDetentPos,
-					    Buttons::isButtonPressed(deluge::hid::button::TEMPO_ENC),
-					    Buttons::isShiftButtonPressed());
+					playbackHandler.tempoEncoderAction(limitedDetentPos,
+					                                   Buttons::isButtonPressed(deluge::hid::button::TEMPO_ENC),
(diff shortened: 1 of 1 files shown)


kastenbalg:
Both of these encoder knobs can be interacted with by turning, by holding shift and turning, and by pressing. This does not work with shift turning or pressing; they appear not to be swapped.

CaptPat:
Good catch! The swap was only applied to encoder **turn** events in `interpretEncoders()`. Pressing the physical encoder buttons was dispatched separately in `buttonAction()` without any swap.

Fixed in e1990677 — added an early button identity swap in `buttonAction()` (before any handler checks `b`), so pressing the physical tempo encoder now triggers select actions and vice versa. This covers all downstream `buttonAction()` handlers across all UIs automatically.

`buttonStates[]` still records the physical button state (set before the swap), so `isButtonPressed()` reflects physical reality.

### Test plan update
- [ ] Swap ON: **turn** tempo encoder → acts as select (and vice versa)
- [ ] Swap ON: **press** tempo encoder → acts as select press (and vice versa)
- [ ] Swap ON: **SHIFT+press** tempo encoder → acts as SHIFT+select (and vice versa)
- [ ] Swap OFF: all encoders behave normally

kastenbalg:
There are also shift + turn and shift + press, swap not complete unless all encoder functions swapped.

CaptPat:
Updated in 6f065a3d — the swap was incomplete because `isButtonPressed(TEMPO_ENC)` / `isButtonPressed(SELECT_ENC)` are called directly across many files (session_view, instrument_clip_view, audio_clip_view, playback_handler, load_song_ui, value.h, patch_cable_strength.cpp, etc.) for shift+turn combos and press-while-turn detection.

Fix: `isButtonPressed()` itself now swaps TEMPO_ENC↔SELECT_ENC queries when the setting is active. This means all callers across the codebase get the swapped state transparently — no need to find and patch every call site.

The swap now covers all four interactions:
- **Turn** — encoder name swap in `interpretEncoders()`
- **Press** — button identity swap in `buttonAction()`
- **Shift+turn** — `isButtonPressed()` swap returns correct encoder button state
- **Shift+press** — both `buttonAction()` swap and `isButtonPressed()` swap active

CaptPat:
The swap covers all interactions via three swap points: encoder identity in `encoders.cpp` (turn/shift+turn), button identity in `buttonAction()` (press/shift+press), and `isButtonPressed()` (press+turn combos). Is there a specific interaction you're seeing that isn't swapped?

PR #4333, opened 2026-03-03
Fix metronome playing during count-in when toggled off

### LINKED ISSUE #3250: Metronome is always on during count-in and unable to be toggled off

The recording count-in (engaged by pressing rec then play) is always accompanied by an audible metronome, regardless of whether the metronome is toggled on or not (shift + tap tempo)

Furthermore, the metronome can not be toggled off during the count-in (via shift + tap tempo) as the 1.2 changelog would suggest (even though that was fixed in #2653)

If the metronome is off in general then the count-in should be silent

Also if the metronome is on during count-in you should be able to toggle it off using shift + tap tempo (as per #2653)

I first noticed it on 2025-01-08 using the 1.2 release

_No response_

7SEG

Release 1.2/Chopin

_No response_

https://github.com/user-attachments/assets/7d951d1e-a92b-448d-b5cd-2dc1d19120d0

CaptPat:
The `goto doMetronome` at the end of the count-in block served two purposes: playing the metronome click AND skipping past playback tick processing (`considerLaunchEvent`/`doTickForward`) that shouldn't run during count-in.

Wrapping the goto with `if (metronomeOn)` fixes the metronome sound but when metronome is off, execution falls through into playback processing, which prevents recording from ever starting.

Fix: wrap goto with metronome guard + add early return for the metronome-off case so playback processing is correctly skipped during count-in regardless of metronome state.

Fixes #3250.

## Test plan
- [ ] Metronome OFF + count-in → silent count-in, recording starts normally
- [ ] Metronome ON + count-in → clicks during countdown, recording starts normally
- [ ] No count-in + metronome ON → metronome works as before
- [ ] No count-in + metronome OFF → silent, playback works as before

### DIFF
--- a/src/deluge/playback/playback_handler.cpp
+++ b/src/deluge/playback/playback_handler.cpp
@@ -910,7 +910,9 @@ void PlaybackHandler::actionSwungTick() {
 				display->displayPopup(buffer, 0, true, 255, 2);
 			}
 			swungTicksTilNextEvent = 2147483647;
-			goto doMetronome;
+			if (metronomeOn) {
+				goto doMetronome;
+			}
 		}
 	}
 


seangoodvibes:
This doesn't build

kastenbalg:
I believe I have a binary with this in it. There appears to be a logic error. If the metronome is on, it plays during the record countdown and continues playing once recording starts. If the metronome is off, the record countdown does not occur and the device never starts recording.

CaptPat:
Fixed both issues in 76efef98:

**Build failure**: Rebased onto current `main` (was 4 commits behind).

**Logic error** (@kastenbalg's report): The original `goto doMetronome` served double duty — it played the metronome click AND jumped past `considerLaunchEvent`/`doTickForward`, effectively exiting the function early during count-in. Wrapping it with `if (metronomeOn)` fixed the sound but broke the control flow: when metronome was off, execution fell through into playback tick processing, which prevented recording from ever starting.

Fix: added an early `return` (with proper flag cleanup) for the metronome-off path, so count-in correctly skips playback processing regardless of metronome state.

### Updated test plan
- [ ] Metronome OFF + count-in → silent count-in, recording starts normally after countdown
- [ ] Metronome ON + count-in → clicks during countdown, recording starts normally
- [ ] No count-in + metronome ON → metronome works as before
- [ ] No count-in + metronome OFF → silent, playback works as before

PR #4336, opened 2026-03-03
Fix long-press back in browser loading the previewed preset

### LINKED ISSUE #4038: Holding back in a load preset menu causes the selected preset to be loaded

Using the new long press feature to automatically back out of multiple folders to exit a preset browser will cause the selected preset to be loaded.

It should keep the preset that was present when the browser menu was entered, like happens when you press back multiple times to back out. The preset should only be loaded when you press the select encoder or the load button.

Today, though it has probably been an issue since that new feature was added.

_No response_

OLED

Beta

1.3.0-BETA- 6716343

Press load+synth in a synth clip or load+kit in a kit clip, turn the select knob to preview a new preset, then hold back to exit the menu. The new preset will be loaded.

CaptPat:
`Browser::exitUI()` (long-press back) called `close()` directly, bypassing the virtual `exitAction()` where subclasses revert state. For `LoadInstrumentPresetUI`, this skipped `revertToInitialPreset()`, loading the previewed preset instead of restoring the original. Changed to call `exitAction()`.

Fixes #4038.

## Test plan
- [ ] Enter preset browser → scroll to different preset → long-press back → original preset restored

### DIFF
--- a/src/deluge/gui/ui/browser/browser.h
+++ b/src/deluge/gui/ui/browser/browser.h
@@ -104,7 +104,7 @@ public:
 
 	// ui
 	bool exitUI() override {
-		Browser::close();
+		exitAction();
 		return true;
 	}
 


kastenbalg:
Tested on device appears to work.

PR #4338, opened 2026-03-04
Fix Shift+pad shortcuts opening wrong menu for DX7 synths

### LINKED ISSUE #3369: Shift + Name pulls up Clip Name instead of OP4 Tune/Level while inside DX7 editor grid

The shift + name button combo is not pulling up the correct parameter while inside the DX7 editing mode. 

The expected behavior is that shift + name pulls up the OP4 Tune/Level parameter while editing DX7 synth parameters. It seems shift + every other button selects the appropriate parameter (or doesn't select anything on purpose) except for shift + Name which pulls up the clip name menu. 

I have only just recently learned the DX7 engine, but I believe I'm on the release version of 1.2 firmware.

unknown

OLED

Release 1.2/Chopin

deluge-c1_2_0.bin

0) Have DX7 engine enabled in community features
1) Custom 1 + Synth to create a new DX7 initial synth, then press Back to get out of the menu
2) Go to Keyboard mode by pressing the keyboard button
3) The audition/section column of buttons should be lit up blue and green. Shift + any of those buttons to start editing the DX7 engine parameters. 
4) Shift + Name to try to pull up OP4 Tune/Level but instead it pulls up Clip Name.

CaptPat:
Shift+pad shortcuts on DX7 instruments fell through to standard menus (e.g., Clip Name) because the DX7 interception only worked when the sound editor was already open. Now detects DX7 instruments (via `oscType == OscType::DX7`) and opens the DX editor before processing the shortcut pad.

Fixes #3369.

## Test plan
- [ ] DX7 synth → Shift+Name pad → opens DX7 parameter (not Clip Name)
- [ ] Non-DX7 synths → standard shortcuts still work

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -1116,27 +1116,10 @@ ActionResult SoundEditor::potentialShortcutPadAction(int32_t x, int32_t y, bool 
 		}
 
 		else {
-			if (runtimeFeatureSettings.get(RuntimeFeatureSettingType::EnableDX7Engine)
-			    == RuntimeFeatureStateToggle::On) {
-				bool inDxEditor = getCurrentUI() == &soundEditor && getCurrentMenuItem() == &dxParam;
-
-				// If the sound editor isn't open yet and the instrument is DX7,
-				// open the DX editor so shortcut pads route to DX parameters
-				// instead of standard shortcuts like Clip Name (#3369)
-				if (!inDxEditor && getCurrentUI() != &soundEditor && getCurrentClip()->type == ClipType::INSTRUMENT) {
-					Output* output = getCurrentClip()->output;
-					if (output->type == OutputType::SYNTH) {
-						auto* sound = static_cast<SoundInstrument*>(output);
-						if (sound->sources[0].oscType == OscType::DX7) {
-							if (setup(getCurrentClip(), &dxParam, 0)) {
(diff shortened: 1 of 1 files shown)


seangoodvibes:
This doesn't build

kastenbalg:
Was able to reproduce bug on deluge-v1_3_0-nightly+2026_03_02-5bdf316. Accessing performance view and pressing Shift+Name pulls up name of patch instead of OP4 Fine Tuning parameter.

CaptPat51's provided build appears to fix this, but breaks the UI in another way. In performance view/keyboard mode, pressing Shift+Name does allow editing of the correct DX7 operator parameter. BUT... outside of performance view, shift+press of any grid pad edits DX7 operator parameters, not the default parameters such as Filter Cutoff, effects settings, etc.

CaptPat:
Closing — the original bug (#3369) was already fixed upstream by PR #2720 (`b2fd7849` "Disable non dx shortcuts"), which made `potentialShortcutPadAction()` always return `true` when in the DX editor, blocking fallthrough to normal shortcuts like Clip Name.

Our fix incorrectly forced entry into the DX editor from any view, which prevented access to normal synth parameters (filter, effects, etc.) on DX7 instruments outside keyboard/performance mode. Thanks to Will Parks for catching this.

kastenbalg:
Not convinced this fixed in earlier PR, PR significantly predates bug report.

PR #4340, opened 2026-03-04
Fix notes sticking when ENV1 sustain moved to zero

### LINKED ISSUE #4029: Played notes sticks when ENV1 sustain is moved to zero.

Under certain ENV1 conditions note(s) held or or played at the same time ENV1 sustain is turned to zero the note(s) can get stuck at full sustain.

When sustain is moved to zero, no sound should be heard after the decay time is over.

just now

_No response_

OLED

Beta

1.3.0 20250813

1: Any synth (not DX7) or sample.
2: Set ENV1  A = Any, D <=5, S=1, R >= 5
3: Select S in ENV1 in Horizontal.
4: Audition one or more note(s)
5: Turn select a few clicks CCW, S = 0 and the note drones.
6: Turn of by changing S > 0 (or quickly pressing some notes or switching synth type)

CaptPat:
Envelope sustain smoothing could overshoot past zero, wrapping `smoothedSustain` negative. This caused the envelope output to jump to near-max amplitude and get stuck permanently. Clamps `smoothedSustain >= 0` after each smoothing update in the DECAY and SUSTAIN stages.

Fixes #4029.

## Test plan
- [ ] Play notes, sweep ENV1 sustain to zero → notes go silent (not stuck at full volume)

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -72,6 +72,7 @@ add_executable(UnitTests
         sync_tests.cpp
         chord_tests.cpp
         time_tests.cpp
+        envelope_tests.cpp
 )
 add_test(NAME UnitTests
         COMMAND UnitTests)
--- a/tests/unit/envelope_tests.cpp
+++ b/tests/unit/envelope_tests.cpp
@@ -1,0 +1,56 @@
+#include "CppUTest/TestHarness.h"
+#include "util/fixedpoint.h"
+#include <algorithm>
+#include <cstdint>
+
+namespace {
+
+// Standalone reimplementation of the sustain smoothing from Envelope::render().
+// Tests the fix for #4355: smoothedSustain going negative when sustain=0.
+struct EnvelopeSustainSim {
+	int32_t smoothedSustain{0};
+
+	// Mirrors the smoothing calculation in Envelope::render() DECAY/SUSTAIN stages
+	void step(uint32_t numSamples, uint32_t sustain) {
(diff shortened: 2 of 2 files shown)


kastenbalg:
Reproduced on deluge-v1_3_0-nightly+2026_03_02-5bdf316 using method in bug report.

Env1: A=5, D=5, S=1, R=5. Play note from audition pad, hold audition pad, turn sustain to zero. Sticks note at higher volume. If audition pad still held and sustain turned to back 1, note returns to low sustain level. If audition pad released while note is stuck after turning sustain to zero, note appears to enter release phase only if sustain turned back to 1.

Same method did not reproduce on earlier 1.2.1.

Same method did not reproduce on build sent to me by CaptPat with patch present. No apparent bad effect from changing sustain to/from 0 while playing notes from audition pad.

src/deluge/gui/views/session_view.cpp
```diff
@@ -4494,6 +4494,7 @@ void SessionView::gridTransitionToSessionView() {
 }
 
 void SessionView::gridTransitionToViewForClip(Clip* clip) {
+	PadLEDs::skipGreyoutFade();
```
seangoodvibes: I don't think this is related to this PR

kastenbalg: Agree. Looks like part of PR #4339. I'll follow up. If there have been problems building this kind of thing may be why.

CaptPat:
Rebased onto main. The session_view.cpp diff was a GitHub artifact from the base being behind upstream — the only change in this PR is the `smoothedSustain` clamp in `envelope.cpp`.

PR #4342, opened 2026-03-04
Fix reverb filter encoder value jumping backwards

### LINKED ISSUE #4234: Select encoder value jumps backwards when adjusting reverb filters

I have had odd behaviour in _some_ menus when using the select encoder and finally paid enough attention to properly describe the issue.

On a new song with nothing playing, if I open the reverb settings, go to the second tab and select the HPF value, I have unexpected behavior when using the select encoder. If I increase the value one click at a time, I will always get stuck at 5. Increasing the value by one click does nothing. Decreasing the value by one click will go from 5 to 3, skipping 4.

The only way I can increase the filter value to over 5 is by turning the select encoder really fast, presumably skipping over multiple values.

It looks like sometimes the encoder will decrease the selected value by one regardless of my input: either it looks like nothing happened (+1 and -1  = 0) or it deducts two (-1 and -1 = -2). This seems to reliably happen twice every 10 clicks, so it takes me 8 clicks to go from 30 to 20 and another 8 clicks from 20 to 10.

For reference, this issue does not happen in any other reverb setting, only on (both) filters.

Select encoder should advance by one value on every click.

Just now

_No response_

OLED

Nightly

deluge-v1_3_0-nightly+2025_12_13-714256d

Open HPF settings in mutable reverb and use the select encoder to increase the value one by one until it gets stuck.

CaptPat:
Reverb HPF/LPF `readCurrentValue()` truncated instead of rounding the float→int conversion, causing the encoder to fight user input (e.g., `0.12 * 50 = 5.999` → 5 instead of 6). Adds `std::round()` to match the other reverb params.

Fixes #4234.

## Test plan
- [ ] Reverb HPF/LPF → turn encoder one click at a time → value increments smoothly

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -72,6 +72,7 @@ add_executable(UnitTests
         sync_tests.cpp
         chord_tests.cpp
         time_tests.cpp
+        reverb_filter_roundtrip_tests.cpp
 )
 add_test(NAME UnitTests
         COMMAND UnitTests)
--- a/tests/unit/reverb_filter_roundtrip_tests.cpp
+++ b/tests/unit/reverb_filter_roundtrip_tests.cpp
@@ -1,0 +1,42 @@
+// Reverb filter encoder round-trip tests.
+// Verifies that std::round() prevents float-to-int truncation drift
+// when reading back HPF/LPF values in the reverb menu.
+
+#include "CppUTest/TestHarness.h"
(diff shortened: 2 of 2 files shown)


seangoodvibes:
This doesn't build

CaptPat:
The build failure was a transient network issue (doxygen-awesome-css download). Rebased onto main to re-trigger CI.

PR #4346, opened 2026-03-04
Fix song search with numbers not working in browser

### LINKED ISSUE #4105: Load Song searches that start with numbers do not work

Many of my song projects are by date - for example, 250929 GoaFolk- but when going to load an saved project by typing the date in the keyboard, either nothing happens or the wrong numbers appear. 

When in Load Song mode, typing numbers causes a the correct numbers to appear in the display.

Searching Songs that begin with letters appear to work as expected. 

A couple months ago on a previous 1.3 release. Still happening on current 1.3 beta as of 30 Sep

_No response_

OLED

Beta

1.3.0-BETA-795C48B

_No response_

CaptPat:
Numeric prefix search in the browser failed because `strcmpspecial`'s natural number ordering caused the "~" upper-bound trick to sort before higher-numbered items (e.g., "1~" < "15.XML"). Adds a forward prefix check at position `i` in addition to the existing backward check at `i-1`.

Depends on #4341. Fixes #4105.

## Test plan
- [ ] Type "1" in song browser → finds first song starting with "1"
- [ ] Alphabetic prefix search still works

### DIFF
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -1277,7 +1277,8 @@ gotErrorAfterAllocating:
 	else {
 		i--;
 		// If it didn't match exactly, try some other stuff before accepting that result.
-		if (memcasecmp(((FileItem*)fileItems.getElementAddress(i))->displayName, enteredText.get(), enteredTextEditPos)) {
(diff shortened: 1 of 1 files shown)


CaptPat:
Note: this PR depends on #4341 (text search last item fix) — both modify `browser.cpp` in the same area. Merge #4341 first.

PR #4347, opened 2026-03-04
Fix velocity view blocking multi-note deletion

### LINKED ISSUE #3985: cannot delete multiple notes at once in velocity view

While in normal clip view, it is possible to delete multiple notes at once by pressing them at the same time.

In velocity view, only one of the notes gets deleted. (Velocity view does support entering multiple notes at once.)

All simultaneously pressed notes should be deleted in velocity view as well.

Just now.

_No response_

OLED

Beta

1.3.0-BETA-F46E61F

_No response_

CaptPat:
In velocity/note editor view, pressing multiple velocity pads for deletion was blocked by the multi-pad ramp logic. Pads at the velocity head position (where the note's current velocity sits) now bypass ramp logic and register independently for deletion on release. Also clears `deleteOnDepress` in `setVelocityRamp` to prevent accidental deletion after applying a ramp.

Fixes #3985.

## Test plan
- [ ] Hold velocity-head pad on one note, press velocity-head pad on another → both delete on release
- [ ] Velocity ramp (hold one column, press different Y on another) still works
- [ ] After applying a ramp, releasing pads does not delete notes

### DIFF
--- a/src/deluge/gui/views/automation_view.cpp
+++ b/src/deluge/gui/views/automation_view.cpp
@@ -2747,18 +2747,15 @@ void AutomationView::velocityEditPadAction(ModelStackWithNoteRow* modelStackWith
 	bool refreshVelocityEditor = false;
 	bool showNewVelocity = true;
 
-	// For simultaneous presses at the velocity head (within 80ms), skip multi-pad logic
-	// so each press is registered independently for deletion on short release.
-	// The 80ms threshold matches editPadAction's length-edit timing guard.
-	bool isSimultaneousVelocityHeadPress =
-	    velocity && squareInfo.numNotes != 0
-	    && (int32_t)(instrumentClipView.timeLastEditPadPress + 80 * 44 - AudioEngine::audioSampleTimer) >= 0
-	    && nonPatchCableMinPadDisplayValues[y] <= squareInfo.averageVelocity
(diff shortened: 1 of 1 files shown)


CaptPat:
Thanks for the feedback — you're right, the previous approach was a regression. Bypassing multi-pad logic entirely for velocity-head presses broke ramp functionality.

**Revised approach:** Instead of checking *what* the second press is (velocity head or not), I now check *when* it arrives relative to the first press:

- **Within 80ms** of the first press AND at the velocity head → treat as simultaneous independent presses. Both get registered in `editPadPresses[]` and can delete on short release via the existing `deleteOnDepress` + `kShortPressTime` mechanism.
- **After 80ms** → enter multi-pad logic as normal (ramp/average fully preserved).

The 80ms threshold matches `editPadAction`'s own length-edit timing guard (line 2296 in `instrument_clip_view.cpp`), which also prevents the second `editPadAction` call from accidentally triggering note-length editing.

Additionally, `setVelocityRamp` now clears `deleteOnDepress` for all active presses as a safety measure, preventing accidental note deletion after a ramp gesture.

kastenbalg:
Tested in UI. Appeared to only allow deletion of up to two notes at a time - should be able to delete as many notes as you can hold. Basing selection of notes to delete on timing probably not the right approach.

seangoodvibes:
Going to close this Pr as I'll take care of this fix

Thanks

PR #4352, opened 2026-03-05
fix(clock): gate and MIDI clock output dies under external MIDI clock

ogabrielluiz:
This PR fixes a bug where the gate clock output (gate 4) and MIDI clock output stop after an initial burst when the Deluge is slaved to an external MIDI clock via USB.

## Root cause

The clock output scheduling depends on internal timer tick state (`lastTimerTickActioned`, `nextTimerTickScheduled`) which freezes when external clock takes over. After a few ticks the scheduling threshold is exceeded and no more clock output gets scheduled.

## Fix

- Adds external-clock-aware scheduling paths in `scheduleTriggerClockOutTick()` and `scheduleMIDIClockOutTick()` that compute timing from the input tick time base instead
- Emit-at-most-one strategy: if behind, emit one tick immediately then either schedule the next (if close) or resync (if far behind) -- avoids blocking rendering or flooding gear with back-to-back ticks
- Fixed a `clockState` parity bug in `resyncAnalogOutTicksToInternalTicks()` that caused phase inversion (halved clock rate) after large tempo changes
- Pure scheduling logic extracted into `computeExternalClockSchedule()` with a type-safe `ClockScheduleAction` enum (None, Schedule, EmitAndSchedule, EmitAndResync) for the result
- 18 unit tests covering all scheduling paths, edge cases, ratio math, and overflow

## Test plan

1. Set gate 4 to clock output with PPQN 24
2. Connect a USB MIDI clock source (e.g. Ableton) to the Deluge
3. Start playback on the clock source and verify continuous clock pulses on an external module connected to gate 4 (tested with Turing Machine)
4. Verify MIDI clock output (F8) continues to flow under external clock
5. Tempo jump test: change external clock tempo significantly and verify clean tracking without doubled/halved rate
6. Disconnect USB and verify clock output still works normally under internal clock

### DIFF
--- a/src/deluge/playback/clock_output_scheduler.cpp
+++ b/src/deluge/playback/clock_output_scheduler.cpp
@@ -11,18 +11,15 @@ ClockSchedulingResult computeExternalClockSchedule(int64_t lastTickDone, int64_t
 	int64_t nextOutTick = lastTickDone + 1;
 	int64_t internalTickForNextOut = nextOutTick * internalTicksPer / outTicksPer;
 
-	// If behind (or exactly on boundary), emit one immediate tick.
-	if (internalTickForNextOut <= currentInternalTick) {
-		result.shouldEmitTick = true;
+	bool shouldEmit = (internalTickForNextOut <= currentInternalTick);
 
-		// Simulate post-emit state: the caller will increment lastTickDone by 1.
+	if (shouldEmit) {
+		// Look one tick past the one we're about to emit (see header comment for + 2 rationale).
 		int64_t tickAfterEmit = lastTickDone + 2;
 		int64_t internalTickForTickAfterEmit = tickAfterEmit * internalTicksPer / outTicksPer;
 
-		// If strictly behind after emitting one, resync. Exactly on the boundary means the
-		// tick is due now — schedule it for the ISR to fire with proper temporal spacing.
 		if (internalTickForTickAfterEmit < currentInternalTick) {
-			result.shouldResync = true;
+			result.action = ClockScheduleAction::EmitAndResync;
 			return result;
 		}
 
@@ -36,8 +33,12 @@ ClockSchedulingResult computeExternalClockSchedule(int64_t lastTickDone, int64_t
 	if (internalTicksUntilNext >= 0 && timePerInputTickMovingAverage > 0) {
 		uint32_t timeUntilNext = (uint32_t)((uint64_t)internalTicksUntilNext * timePerInputTickMovingAverage
 		                                    * inputTicksPer / internalTicksPerInput);
-		result.shouldSchedule = true;
+		result.action = shouldEmit ? ClockScheduleAction::EmitAndSchedule : ClockScheduleAction::Schedule;
 		result.scheduledTime = inputTickTime + timeUntilNext;
+	}
+	else if (shouldEmit) {
+		// Emitted a tick but can't compute a schedule (e.g. zero tempo) -- treat as resync.
+		result.action = ClockScheduleAction::EmitAndResync;
 	}
 
 	return result;
--- a/src/deluge/playback/clock_output_scheduler.h
+++ b/src/deluge/playback/clock_output_scheduler.h
@@ -1,21 +1,27 @@
 #pragma once
 #include <cstdint>
 
-/// Result of computing the next clock output scheduling decision.
-/// Invariant: shouldResync and shouldSchedule are mutually exclusive.
+/// The four possible outcomes of the external clock scheduling decision.
+enum class ClockScheduleAction : uint8_t {
+	None,            ///< No action needed (e.g. tempo not yet established)
+	Schedule,        ///< Schedule next tick at scheduledTime; nothing to emit now
+	EmitAndSchedule, ///< Emit one immediate tick, then schedule the following tick at scheduledTime
+	EmitAndResync,   ///< Emit one immediate tick, then resync (too far behind to schedule)
+};
+
 struct ClockSchedulingResult {
-	bool shouldEmitTick{false};
-	bool shouldResync{false};
-	bool shouldSchedule{false};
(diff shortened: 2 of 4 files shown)


src/deluge/playback/playback_handler.cpp
```diff
@@ -786,9 +786,51 @@ void PlaybackHandler::doTriggerClockOutTick() {
 	}
 }
 
+// Maximum catch-up ticks per input tick to prevent ISR overrun (2 bars at 24ppqn)
+static constexpr int32_t kMaxClockCatchUpTicks = 48;
+
 // Check these are enabled before calling this!
 void PlaybackHandler::scheduleTriggerClockOutTick() {
 
+	if (isExternalClockActive()) {
```
m-m-adams, line 793: could you drop all the code in this block into a new function?

src/deluge/playback/playback_handler.cpp
```diff
@@ -786,9 +786,51 @@ void PlaybackHandler::doTriggerClockOutTick() {
...
+		uint32_t internalTicksPerInput;
+		getInternalTicksToInputTicksRatio(&inputTicksPer, &internalTicksPerInput);
+
+		int64_t currentInternalTick = lastInputTickReceived * internalTicksPerInput / inputTicksPer;
+
+		int64_t nextAnalogOutTick = lastTriggerClockOutTickDone + 1;
+		int64_t internalTickForNextAnalogOut = nextAnalogOutTick * internalTicksPer / analogOutTicksPer;
+
+		int32_t catchUpCount = 0;
+		while (internalTickForNextAnalogOut <= currentInternalTick && catchUpCount < kMaxClockCatchUpTicks) {
```
m-m-adams: have you tested this somehow? There's a hardware issue where two gates in quick succession won't actually go low, they need to be seperated by a tiny bit of time

Plus this looks like it would block rendering while it's catching up

src/deluge/playback/playback_handler.cpp
```diff
@@ -813,6 +855,45 @@ void PlaybackHandler::scheduleTriggerClockOutTickParamsKnown(uint32_t analogOutT
...
+		getInternalTicksToInputTicksRatio(&inputTicksPer, &internalTicksPerInput);
+
+		int64_t currentInternalTick = lastInputTickReceived * internalTicksPerInput / inputTicksPer;
+
+		int64_t nextMIDIClockOutTick = lastMIDIClockOutTickDone + 1;
+		int64_t internalTickForNextMIDIOut = nextMIDIClockOutTick * internalTicksPer / midiClockOutTicksPer;
+
+		int32_t catchUpCount = 0;
+		while (internalTickForNextMIDIOut <= currentInternalTick && catchUpCount < kMaxClockCatchUpTicks) {
+			doMIDIClockOutTick();
```
m-m-adams: same question here, if multiple clock ticks end up in one usb midi packet it crashes a lot of gear

src/deluge/playback/playback_handler.cpp
ogabrielluiz: I did experience that exact thing a few times in the past few days. I'll get back to you on that. Thanks!

src/deluge/playback/playback_handler.cpp
ogabrielluiz: Done. Extracted the external clock blocks into scheduleTriggerClockOutTickFromExternalClock() and scheduleMIDIClockOutTickFromExternalClock(). Went further and pulled the pure scheduling logic into a testable free function (computeExternalClockSchedule) with 18 unit tests.

src/deluge/playback/playback_handler.cpp
ogabrielluiz: You were right on both counts. Replaced the catch-up loop with emit-at-most-one then resync if strictly behind. The ISR handles subsequent ticks with proper temporal spacing. Also fixed a clockState parity bug in resyncAnalogOutTicksToInternalTicks() that was causing phase inversion (halved clock rate) after large tempo changes. Tested on hardware with Pam's Pro Workout and tracks tempo jumps cleanly now.

src/deluge/playback/playback_handler.cpp
ogabrielluiz: Same approach as the gate fix. Emit at most one F8, resync if further behind. The scheduling logic is shared via computeExternalClockSchedule() so both paths have identical behavior.

src/deluge/playback/clock_output_scheduler.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+#include <cstdint>
+
+/// Result of computing the next clock output scheduling decision.
+/// Invariant: shouldResync and shouldSchedule are mutually exclusive.
```
m-m-adams: If these are exclusive can they be a three option enum?

m-m-adams: Can you make them an enum so that's enforced by the type system?

src/deluge/playback/clock_output_scheduler.cpp
```diff
@@ -0,0 +1,44 @@
...
+	int64_t currentInternalTick = lastInputTickReceived * internalTicksPerInput / inputTicksPer;
+
+	int64_t nextOutTick = lastTickDone + 1;
+	int64_t internalTickForNextOut = nextOutTick * internalTicksPer / outTicksPer;
+
+	// If behind (or exactly on boundary), emit one immediate tick.
+	if (internalTickForNextOut <= currentInternalTick) {
+		result.shouldEmitTick = true;
+
+		// Simulate post-emit state: the caller will increment lastTickDone by 1.
```
m-m-adams: Can you make this comment clearer? I'm not sure I understand the purpose of adding 2. Is it to compute the next next output tick and schedule it?

ogabrielluiz:
Hey @m-m-adams, comments addressed. Let me know if this needs any more changes.

PR #4361, opened 2026-03-07
Fix grainBufferStolen() leaving processor in stale state

### LINKED ISSUE #4357: grainBufferStolen() leaves wrapsToShutdown in stale state, allowing processing with uninitialized buffer

When the Stealable system steals the grain buffer, it calls GranularProcessor::grainBufferStolen() (GranularProcessor.h:62), which only sets grainBuffer = nullptr. It does not reset wrapsToShutdown, grainInitialized, bufferFull, or the grain array.

On the next call to processGrainFX() (GranularProcessor.cpp:47), the entry condition anySoundComingIn || wrapsToShutdown >= 0 is true because wrapsToShutdown retains its pre-steal value. The code then:

                    Calls getBuffer() at line 55, which allocates a fresh, uninitialized buffer
                    Proceeds to processOneGrainSample() at line 63, where active grains (with stale length > 0 from before the steal) read from the uninitialized buffer at arbitrary offsets

This appears to be undefined behavior (reading uninitialized memory).

Expected behavior: grainBufferStolen() should reset the processor to a clean state, similar to clearGrainFXBuffer():

This ensures that after a steal, re-entry to processGrainFX either exits early (if wrapsToShutdown is 0 and no sound coming in) or starts fresh (grains won't read stale positions).

While investigating expanded tests for edge cases / stability.

None Found

OLED

Nightly

nightly (commit 5bdf3166)

Steps to reproduce:

Create a GranularProcessor and process several blocks of audio to fill the buffer and activate grains
Call grainBufferStolen() (simulating Stealable system reclaiming the 4MB buffer under memory pressure)
Call processGrainFX() again with anySoundComingIn = true
Grains read from freshly allocated, uninitialized buffer memory

Files: 
src/deluge/dsp/granular/GranularProcessor.h line 62, 
src/deluge/dsp/granular/GranularProcessor.cpp lines 47-87

CaptPat:
## Summary

- Reset `wrapsToShutdown`, `grainInitialized`, `bufferFull`, and all grain lengths when the grain buffer is stolen
- Previously only `grainBuffer` was set to nullptr, leaving stale state that caused `processGrainFX()` to read from a freshly allocated uninitialized buffer at stale offsets

Fixes #4357

## Test plan

- [ ] Trigger memory pressure to force grain buffer steal (e.g., load many samples)
- [ ] Verify granular FX recovers cleanly after buffer is stolen and re-allocated
- [ ] No audio glitches or crashes from stale grain state

### DIFF
--- a/src/deluge/dsp/granular/GranularProcessor.h
+++ b/src/deluge/dsp/granular/GranularProcessor.h
@@ -59,7 +59,15 @@ public:
 	                    q31_t reverbAmount);
 
 	void clearGrainFXBuffer();
-	void grainBufferStolen() { grainBuffer = nullptr; }
+	void grainBufferStolen() {
+		grainBuffer = nullptr;
+		wrapsToShutdown = 0;
+		grainInitialized = false;
+		bufferFull = false;
+		for (auto& grain : grains) {
+			grain.length = 0;
+		}
+	}
 
 private:
 	void setupGrainFX(int32_t grainRate, int32_t grainMix, int32_t grainDensity, int32_t pitchRandomness,


m-m-adams:
Thanks - the grain buffer used to be fully initialized, this and the other grain issue come from the change to lazy initialization

PR #4366, opened 2026-03-07
feat: add per-clip time signature support

CaptPat:
Adds per-clip time signature (numerator 1-32, denominator 2/4/8/16) as musical notation metadata. Does NOT change playback speed — purely affects beat/bar display, metronome accent pattern, count-in length, and launch quantization grid.

- `TimeSignature` struct with `barLengthInBaseTicks()` / `beatLengthInBaseTicks()` (96 ticks = whole note)
- Song stores a `defaultTimeSignature` that new clips inherit
- Timeline view uses clip's time signature for bar:beat:sub display
- Metronome accents on beat 1 of the clip's bar, clicks on each beat
- Launch quantization uses clip's bar/beat length
- XML serialized only when non-default; denominator validated to 2/4/8/16
- Menu: Clip Settings → Time Signature → Beats Per Bar / Beat Unit

Orthogonal to per-clip tempo ratio (which changes tick speed). Can be combined: e.g., 7/8 time sig + 7:8 tempo ratio = 7 eighth notes at 87.5% speed with correct metronome.

## Test plan
- [ ] Set clips to 3/4, 7/8, 5/4 → beat/bar display and metronome follow correctly
- [ ] New clips inherit song's default time signature
- [ ] Save/load round-trip preserves time signature
- [ ] Existing 4/4 songs are unaffected


m-m-adams:
This doesn't work. It changes the playback speed so that bars are the same length, it does not visually group notes together. 

It's the opposite of the description

CaptPat:
Thanks for the feedback @m-m-adams — you're right, the previous approach was fundamentally wrong. It was changing playback speed by deriving clip loop length from the time signature, rather than just changing the visual/audio feedback.

Revised in 6b7c4fbf — the per-clip time signature now **only** affects:
- **Metronome**: accent pattern follows the clip's time sig (e.g., 3/4 gives 3 clicks per bar)
- **Display**: bar/beat position shown per the clip's time sig

It does **not** change:
- Clip loop length
- Playback speed or tempo
- Launch quantization
- Count-in timing

Also reverted the denominator back to standard beat units (2, 4, 8, 16) and the numerator to 1-32.

### Test plan
- [ ] Set clip to 3/4 → metronome clicks 3 beats per bar with accent on beat 1
- [ ] Set clip to 7/8 → metronome clicks 7 eighth-note beats per bar
- [ ] Bar/beat display shows correct position for non-4/4 time sigs
- [ ] Clip loop length unchanged when time signature changes
- [ ] Playback speed/tempo unchanged
- [ ] Saves/loads correctly in song XML
- [ ] Old songs without time sig tags load as 4/4 (default)

PR #4367, opened 2026-03-07
feat(clip): per-clip tempo ratio for polyrhythmic playback

CaptPat:
Clips can run at fractional speeds relative to global tempo using integer Bresenham tick scaling. A 7:8 ratio clip plays 7 clip-ticks per 8 global ticks (87.5% speed), enabling polyrhythmic arrangements where clips of different lengths align over shared time spans.

Engine: `scaleGlobalToClipTicks()` in `Clip::incrementPos()` with ceiling-division feedback for the scheduler. 4 touch points: tick scaling, scheduler event conversion, UI position interpolation, audio clip sample timing. 1:1 ratio clips have zero overhead (early-exit branch).

Menu: Clip Settings → Tempo Ratio — 35 presets displayed as percentage with underlying ratio (e.g., "88% (7:8)"). All unique reduced fractions with N,D ∈ [1,8], sorted by percentage, ranging from 25% (1:4) to 400% (4:1). XML serialized only when non-default.

### Bresenham precision

The Bresenham accumulator guarantees **exact** long-term tracking for any integer ratio — no drift, no rounding. A 7:8 ratio produces exactly 7 clip ticks per 8 global ticks, with the accumulator returning to zero at each cycle boundary. All 43 reduced fractions with N,D ∈ [1,8] are verified to cycle exactly in unit tests.

**Performance:** ~0.002% CPU (5 clips × 48 ticks/sec × 40 cycles on 400MHz Cortex-A9). Tick-rate only, never per-sample.
**Memory:** 12 bytes/clip + 4 bytes/audio guide + ~500 bytes static (menu). 23 unit tests included (13 original + 10 edge cases).

## Test plan
- [ ] Two clips at 1:1 and 7:8 — verify different playback speeds, no stuck notes
- [ ] Audio clip at non-1:1 ratio — sample timing adjusts correctly
- [ ] Save/load round-trip preserves ratio
- [ ] 1:1 clips behave identically to before (zero overhead)
- [ ] Scroll through percentage menu — values change smoothly, ratio applies immediately


m-m-adams:
Closing - far below contribution standards and I have no hope that you'll fix that

I'm especially annoyed that even this is obviously wrong, so you're not even bothering to read the posts you make

image

PR #4368, opened 2026-03-07
feat: per-clip time signature (display and metronome only)

CaptPat:
Add a per-clip time signature that affects metronome accent pattern and bar/beat position display without changing playback timing.

- **Metronome**: accents beat 1 of each bar per the clip's time sig (e.g., 3/4 gives 3 clicks per bar)
- **Display**: bar/beat position shown per the clip's time sig
- New clips inherit the song's default time signature
- Serialized per-clip and per-song in XML (backward compatible — old songs load as 4/4)
- Menu: Clip Settings → Time Signature → Beats Per Bar (1-32) / Beat Unit (2, 4, 8, 16)

Does **not** change clip loop length, playback speed, launch quantization, or count-in timing.

**Memory:** 2 bytes/clip + 2 bytes on Song.

## Test plan
- [ ] Set clip to 3/4 → metronome clicks 3 beats per bar with accent on beat 1
- [ ] Set clip to 7/8 → metronome clicks 7 eighth-note beats per bar
- [ ] Bar/beat display shows correct position for non-4/4 time sigs
- [ ] Clip loop length unchanged when time signature changes
- [ ] Save/load round-trip preserves time signature
- [ ] Old songs without time sig tags load as 4/4


CaptPat:
Revised based on feedback from @m-m-adams on #4366 — the previous approach incorrectly changed playback speed by deriving clip loop length from the time signature.

This revision scopes per-clip time signature to **display and metronome only**:
- Metronome accent pattern follows the clip's time sig
- Bar/beat position display uses the clip's time sig
- No changes to clip loop length, playback speed, launch quantization, or count-in timing

Also reverted denominator to standard beat units (2, 4, 8, 16) and numerator to 1-32.

m-m-adams:
Not remotely what people are requesting

PR #4370, opened 2026-03-08
Revert "Fix grainBufferStolen() leaving processor in stale state"

m-m-adams:
Reverts SynthstromAudible/DelugeFirmware#4361

Took a closer look and this is duplicate code. It's reset when getBuffer is called in the next render

### DIFF
--- a/src/deluge/dsp/granular/GranularProcessor.h
+++ b/src/deluge/dsp/granular/GranularProcessor.h
@@ -59,15 +59,7 @@ public:
 	                    q31_t reverbAmount);
 
 	void clearGrainFXBuffer();
-	void grainBufferStolen() {
-		grainBuffer = nullptr;
-		wrapsToShutdown = 0;
-		grainInitialized = false;
-		bufferFull = false;
-		for (auto& grain : grains) {
-			grain.length = 0;
-		}
-	}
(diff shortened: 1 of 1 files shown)


CaptPat:
getBuffer() only resets bufferFull and bufferWriteIndex, not reset:  wrapsToShutdown, stale countdown value; grainInitialized, remains true, so the init path is skipped; grains[].length, stale grain lengths cause reads from freshly allocated, uninitialized buffer memory.  Sequence after buffer theft remains:  grainBufferStolen() → grainBuffer = nullptr, nothing else reset; 
Next render → getBuffer() allocates fresh buffer, resets only bufferFull/bufferWriteIndex; grainInitialized is still true → skips re-initialization; Stale grains read from uninitialized buffer.  this was issue #4357.  PR 4371 addresses (part) of the issue - might want to add bool grainInitialized{};      // false on construction  and  int32_t wrapsToShutdown{};     // 0 on construction; but that only fixes construction not the runtime grainBufferStolen() path. For that, grainBufferStolen() needs to reset those fields when the buffer is taken.

m-m-adams:
The buffer can't be stolen if grains are active. Marking the buffer as not full stops new grains being made until it is full

Issue #3898, opened 2025-07-03
Crash when exiting Settings (when tapping quickly Back several times)


soymonitus:
### Please describe the problem:

If you press Back quickly while you are exiting Settings menu, it crashes

bash-3.2$ arm-none-eabi-addr2line -Capife build/Release/deluge.elf 0x2013507d 0x20114be5 0x20107f7f 0x201085f5
0x2013507d: disk_read.isra.0 at ??:?
0x20114be5: Userdef_INTC_HandlerExe at /Users/admin/Development/Projects/Deluge/DelugeFirmware/src/RZA1/intc/intc_userdef.c:734
 (inlined by) INTC_Handler_Interrupt at /Users/admin/Development/Projects/Deluge/DelugeFirmware/src/RZA1/intc/intc_handler.c:135
0x20107f7f: move_window at /Users/admin/Development/Projects/Deluge/DelugeFirmware/src/fatfs/ff.c:1070
 (inlined by) move_window at /Users/admin/Development/Projects/Deluge/DelugeFirmware/src/fatfs/ff.c:1056
0x201085f5: dir_find at /Users/admin/Development/Projects/Deluge/DelugeFirmware/src/fatfs/ff.c:2433

### What is the expected behavior?

No crash

### When did you start noticing the issue?

Latest beta

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

ok-reza:
possibly related: #2759

seangoodvibes:
Same issue reported on discord today:

![image](https://github.com/user-attachments/assets/04fbdec3-a0ad-4df3-ba8b-66688af573af)

kastenbalg:
This one is intermittent. 

Seem to be able to reproduce it reliably on deluge-v1_3_0-nightly+2026_03_12-af12fb9 with following steps:

1) Start deluge.
2) Shift+press select into settings.
3) Navigate to firmware version, click into it to see version.
4) Press back a few times to trigger crash.

kastenbalg:
Crash I reproduce is similar location in source as original report.

arm-none-eabi-addr2line -Capife build/Release/deluge.elf 0x20140599 0x20146a44 0x2011530f 0x201159b5
0x20140599: disk_read.isra.0 at ??:?
0x20146a44: icciar_read_complete at C:/Users/wilbo/source/repos/DelugeFirmware/src/RZA1/compiler/asm/irqfiq_handler.S:161
[...]

Issue #2759, opened 2024-10-05
Hitting BACK a couple times from the main menu while 'SAVING SETTING' is displayed will cause a crash.


ok-reza:
### Please describe the problem:

Entering the main menu and then exiting with repeated BACK presses will cause a crash. Sometimes it only takes one BACK while 'SAVING SETTING' appears to trigger the crash, but it's more consistent if you press multiple times.

Additionally, once I reboot the Deluge, I get a Start Up Song error, despite having no start up song previously enabled. it displays "START UP OFF,  REASON:'

### What is the expected behavior?

No crash on exit.
No issue with start up song displays.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Beta

### What is the firmware name:

1.2 beta

### If possible provide the steps to reproduce the issue and upload additional media:

![IMG_7637](https://github.com/user-attachments/assets/31fc41be-14f1-47ce-a371-c7bff9d4e94b)

m-m-adams:
I can't reproduce this - I'm slightly suspicious that your SD card might have a problem based on the symptoms, maybe try repairing it on a computer?

ok-reza:
im suspecting it might be an issue of the slower read/write speed then? im using the sd card that the deluge comes with. but i'll check w repairing too!

m-m-adams:
That message should only be present if you have a song named SONGS/__STARTUP_OFF_CHECK_ on your card - do you see one? 

It would have been created by the startup song feature failing at some point in the past

pmatila:
I'm seeing this kind of thing a lot on 1.2, so much that it's been keeping me on 1.1. For me a fairly reliable reproducer is settings -> midi -> devices -> upstream port 2 -> mpe -> in -> lower zone -> change value (and back, if you like), and then hit back quickly until it it has exited the menu and saved (as in, just keep hitting back without waiting for feedback):
![IMG_20241206_164100_HDR](https://github.com/user-attachments/assets/42fe7d82-c536-4043-bc49-f2184221f68d)

I'm also using the SD card that came with Deluge. There *was* the _STARTUP_OFF_CHECK file present on today's first crash, but after removing that file on PC and running fsck on the card filesystem I can still reproduce the crash pretty much at will, tried several times, rebooting each time.  ... except that after testing it on some random CV related setting from which I slowly backed out and didn't crash, I seem to have lost the ability to reproduce it anymore.

[...]

PR #4384, opened 2026-03-21
Back menu exit, don't remind, unset timer

kastenbalg:
Two crash paths can reach exitCompletely() and write to the card re-entrantly.

1. A short press of back button from navigationDepth == 0 gets deferred as REMIND_ME_OUTSIDE_CARD_ROUTINE. This causes exitCompletely() to get called again when the back button press is replayed. This happens outside of a card routine, but that is not enough, because the replay is between two of them. exitCompletely() calls _two_ functions that both write to the card; presumably the replay calls exitCompletely() a second time once the first call's first write finished, during the first call's second write. Fix returns DEALT_WITH for a back button press at navigationDepth == 0 if inCardRoutine is true. If you press back at that depth of the menu while you are already using the card, you don't want to use the card again, just exit the menu. That press should not get replayed. There may still be the possibility of a rare unlucky press, in the brief moment between those two functions that write (inCardRoutine = false), but the existing code guarantees a crash by replaying the press for you as soon as it thinks it can.
2. If BACK_EXIT_MENU timer is set and you press the back button enough times to call exitCompletely(), you can have another call to exitCompletely() during the button press triggered card write, when the timer fires and calls exitCompletely() again. Fix unsets the timer before calling the functions that write to the card. Once you reach the point of writing, you do not want the timer to write again.

I see the first one as the primary cause, perhaps in practice the sole one, but found the second one along the way.

Tried other approaches that deferred the extra write. These fixes should mean no extra write at all.

Fixes: https://github.com/SynthstromAudible/DelugeFirmware/issues/3898, https://github.com/SynthstromAudible/DelugeFirmware/issues/2759

### DIFF
--- a/src/deluge/gui/ui/sound_editor.cpp
+++ b/src/deluge/gui/ui/sound_editor.cpp
@@ -409,6 +409,12 @@ ActionResult SoundEditor::buttonAction(deluge::hid::Button b, bool on, bool inCa
 		    || currentUIMode == UI_MODE_HOLDING_AFFECT_ENTIRE_IN_SOUND_EDITOR) {
 			if (on) {
 				if (inCardRoutine) {
+					// At navigationDepth 0, exitCompletely() is already in progress (it does
+					// card writes that yield via routineForSD). Consuming the press avoids a
+					// deferred replay that would re-enter the card writes and crash.
+					if (navigationDepth == 0) {
+						return ActionResult::DEALT_WITH;
+					}
 					return ActionResult::REMIND_ME_OUTSIDE_CARD_ROUTINE;
 				}
 


m-m-adams:
I'm sorry but these words don't make sense in the order claude is using them. I've read it repeatedly and can't tell what it's trying to talk about and I'm not convinced either claude or you know what re entrancy means

kastenbalg:
Claude didn't write that but as long as it's fixed I don't care. I'm not
going to submit anything written by claude. Claude will bumble around a lot
unless you guide it very carefully.

The crash path is inside routineforSD(), I believe; it calls a function to
keep the UI going that deals with deferred button presses. The two writes
are the problem, they both go into routineforSD() and yield (maybe wrong
word) out of it. Best of my understanding, the second write function of the
two that exitCompletely() calls goes into routineforSD() and then
routineforSD() replays the back button and calls exitCompletely() again,
which cheerfully calls the first write function while the second one is
still writing.

Maybe this is incoherent too. I probably should have written something like
function1() -> function2() -> function3().

What I don't know for sure is whether the other crash path is legit. Ie,
whether the timer BACK_MENU_EXIT actually gets set in such a way that it
can fire and call exitCompletely() again in a similar way. I know that
unsetting the timer does not fix the crash on its own.

PR #4388, opened 2026-03-23
Compressor hpf stereo rms

kastenbalg:
All the other filters do it like this - existing code calculates RMS subtly incorrectly. May be hard to hear but not right.


kastenbalg:
My mistake sorry, this is only supposed to change one file.

Issue #4199, opened 2025-11-11
Creating multiple clip instruments introduces incrementally longer wait times during each instrument creation


ok-reza:
### Please describe the problem:

Every time you add a new clip instrument, the time it takes to create the clip (to populate the designated pad/row) gets longer and longer.

The weird part about this is that ~the lag seems to be directly tied to how far down the list of available synth presets the Deluge has to parse through.~ The lag seems to be based on 1) amount of instruments already created and 2) significant more time added if the deluge has to find the next available preset from a new folder.

In my video attached, synth presets populate until factory preset #27. For whatever reason, SYNT400 comes after 27 despite there being all 170+ factory presets available. In any case, the creation of synth presets 400+ takes **significantly** longer to populate then say the 20th or 21st synth preset in the browser (however this video demonstrates the lag present between say the first 5 instruments created and once you get to ~20).

Also points to something funky going on with the browser: ignores presets after 27, enters a folder to populate new instruments, dialog messages appear saying no more presets are available after around 35 instruments.

https://github.com/user-attachments/assets/2245302a-1f51-42fe-9d56-85a065e826b9

### What is the expected behavior?

I'm wondering if this lag introduced with every instrument creation can be optimized. Obviously 25+ unique instruments is pushing the Deluge in a way it likely wasn't 'intended' for, but if there are optimizations possible, then users would experience some gains for more normal 8-16 instrument clip creations as delay seems to be noticeable around 8-10 clips in.

Creating the 27th instrument from the 27th preset on the list in the root synth folder takes about 2-3 seconds, compared to almost 9 seconds for my 28th instrument which actually happens to be the first preset nested in a folder (titled '400').

### When did you start noticing the issue?

Slow clip creation I've noticed during playback/creation of a song, but always figured it was CPU related due to all other things going on in song. This is something different, so unable to determine.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Release 1.2/Chopin

### What is the firmware name:

1.2.1 also 1.3 beta

seangoodvibes:
Does this happen in 1.1.1?

ok-reza:
i've replicated this issue in 1.1.1 and 1.0.1 (only ones i've tried). trying OLED 4.1.4 this delay is not an issue, however it seems after ~30 instruments it is also unable to 'find any presets'

seangoodvibes:
[...]

PR #4392, opened 2026-03-24
fix(browser): cache FilePointer on Instrument to avoid repeated SD reads

CaptPat:
Addresses #4199.

`setupWithInstrument()` calls `f_open()` for every in-memory instrument each time
the browser builds its file list. With N instruments, creating a new clip triggers
N SD reads, making each successive creation slower.

Cache the FAT `FilePointer` on `Instrument`. Invalidate on save and clone.
Copy from directory-read FileItems during dedup when available.

Tested with ~60 loaded instruments — clip creation near-instant regardless of count.


CaptPat:
This fix has been tested on device - I used a song file with about 60 instruments and added 5 or 6 clips - seemed to be fast.

m-m-adams:
The files are reloaded because not reloading them is unsound. Maybe try caching enough metadata to remake the file list instead? We don't actually need to read the files we just know where they exist in the directory

CaptPat:
Closing this PR as #4393  addresses the issue and is cleaner.

PR #4393, opened 2026-03-25
defer checking if file exists

### LINKED ISSUE #4199: Creating multiple clip instruments introduces incrementally longer wait times during each instrument creation

m-m-adams:
Fix #4199

### DIFF
--- a/src/deluge/gui/context_menu/delete_file.cpp
+++ b/src/deluge/gui/context_menu/delete_file.cpp
@@ -99,6 +99,7 @@ bool DeleteFile::acceptCurrentOption() {
 			display->displayPopup(l10n::get(STRING_FOR_ERROR_PRESET_IN_USE));
 		}
 		else if (toDelete->instrument) {
+			display->displayPopup(l10n::get(STRING_FOR_FILE_DELETED));
 			// it has an instrument, it's not on the card, it's not in use, let's remove it
 			browser->currentFileDeleted();
 		}
--- a/src/deluge/gui/ui/browser/browser.cpp
+++ b/src/deluge/gui/ui/browser/browser.cpp
@@ -419,6 +419,9 @@ void Browser::deleteFolderAndDuplicateItems(Availability instrumentAvailabilityR
 				if (!nextItem->instrument && !nextItem->isFolder) {
 					if (!strcasecmp(readItem->displayName, nextItem->displayName)) {
 						// if (readItem->filename.equalsCaseIrrespective(&nextItem->filename)) {
+						if (readItem->maybeExistsOnCard && readItem->filePointer.sclust == 0) {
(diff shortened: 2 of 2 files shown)


CaptPat:
Mark - I'll close PR #4392.  Is there a cardReinserted() invalidation needed?   In audio_file_manager.cpp, at the end of cardReinserted(), iterate all instruments in the song (both active and hibernating) and set mightExistOnCard = false. ??

This handles the scenario where the user:
  Loads a song with 30 instruments (all get mightExistOnCard = true from the song file)
  Ejects the card, deletes some preset XML files on their PC  
  Reinserts the card

CaptPat:
Also in browser.cpp - predictExtendedText(), sclust is 0 with deferred resolution - so preview triggers may not occur:
if (fileItem->filePointer.sclust != oldClust) {
    currentFileChanged(0);  // trigger preview
}
perhaps compare fileIndexSelected instead?  this is robust regardless of whether FilePointers are resolved:
if (fileIndexSelected != oldFileIndex) {
    currentFileChanged(0);
}

m-m-adams:
Thanks @CaptPat, I'll take a look at how to deal with those

PR #4404, opened 2026-03-27
Feature/transpose screen

m-m-adams:
Hold <> and turn ^v to transpose the notes on the current screen only. If you also hold ^v it locks the notes into their original octave so you get pseudo voice leading


CaptPat:
Tested it... works good.  merged on top of my nightly - clean merge -  running it now.   nice.

seangoodvibes (approved):

PR #4441, opened 2026-04-05
Fix/ arranger automation bug incorrectly getting sequence direction by casting song timeline counter as clip

### LINKED ISSUE #4020: Deluge crashes/freeze when using performancemode effects while recording in arranger

when i use Performance effects while recording in arranger the deluge crashes. 

no crashing

"just now

_No response_

7SEG

Beta, Built from community

_No response_

record into arranger over the shortcut "record + song" and then use performance effects for example the bitcrusher
after that the deluge crashes

seangoodvibes:
**This seems to be a long standing bug in arranger automation view, dating back to official firmware.**

Official: 
- https://github.com/SynthstromAudible/DelugeFirmware/blob/695b22af1e9fcc060e59868c83cf8e504bd57ff1/src/AutoParam.cpp#L913
- https://github.com/SynthstromAudible/DelugeFirmware/blob/695b22af1e9fcc060e59868c83cf8e504bd57ff1/src/AutoParam.cpp#L146
- https://github.com/SynthstromAudible/DelugeFirmware/blob/695b22af1e9fcc060e59868c83cf8e504bd57ff1/src/ModelStack.cpp#L63

Main:
- https://github.com/SynthstromAudible/DelugeFirmware/blob/4ab93c5308319e018ccee84caae4756d94732162/src/deluge/modulation/automation/auto_param.cpp#L1118
- https://github.com/SynthstromAudible/DelugeFirmware/blob/4ab93c5308319e018ccee84caae4756d94732162/src/deluge/modulation/automation/auto_param.cpp#L156
- https://github.com/SynthstromAudible/DelugeFirmware/blob/4ab93c5308319e018ccee84caae4756d94732162/src/deluge/model/model_stack.cpp#L67

**Issue:**

- Error "E445" is triggered in arranger view by trying to record delay amount automation is found in auto_param.cpp.
- When you look at the code, it falls under a branch that says "// Or, playing reversed..."
- Well this is a red flag ... arranger view playback is never reversed.

**Cause:**
- Following the code back up, we find where the "reversed" flag is set:
- In void AutoParam::setCurrentValueInResponseToUserInput:
- bool reversed = modelStack->isCurrentlyPlayingReversed();
- So then I went and checked isCurrentlyPlayingReversed, and it doesn't check the song's timeline counter, which is what is used when recording song automation.
- Instead it was always casting the timeline counter as a clip and returnin the reversed flag of the clip.
- Well that's wrong..

**Solution:**
- To avoid casting the timeline counter to the wrong type, I moved sequence direction mode and reversing status to the timeline counter class which both clips and the song inherit from.
- This fixed the E445 crash and maybe some other gremlins resulting from landing in that "reversed" branch.

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4020

### DIFF
--- a/src/deluge/model/model_stack.cpp
+++ b/src/deluge/model/model_stack.cpp
@@ -63,13 +63,12 @@ ModelStackWithNoteRow* ModelStackWithNoteRowId::automaticallyAddNoteRowFromId() 
 }
 
 bool ModelStackWithNoteRow::isCurrentlyPlayingReversed() const {
-
-	// Under a few different conditions, we just use the parent Clip's reversing status.
+	// Under a few different conditions, we just use the Song's or the parent Clip's reversing status.
 	if (!noteRow
 	    || (noteRow->sequenceDirectionMode == SequenceDirection::OBEY_PARENT
 	        && (!noteRow->loopLengthIfIndependent
-	            || ((Clip*)getTimelineCounter())->sequenceDirectionMode != SequenceDirection::PINGPONG))) {
-		return ((Clip*)getTimelineCounter())->currentlyPlayingReversed;
+	            || getTimelineCounter()->sequenceDirectionMode != SequenceDirection::PINGPONG))) {
+		return getTimelineCounter()->currentlyPlayingReversed;
 	}
 
 	// Otherwise, we use the NoteRow's local one.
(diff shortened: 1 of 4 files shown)


src/deluge/model/model_stack.cpp
m-m-adams: I think maybe currentlyPlayingReversed and sequenceDirectionmode should move from Clip to TimelineCounter? This is still a bit of a hack and it could lead to bugs if we ever do add more modes to song sequencing

seangoodvibes: Good idea, I can make that change

PR #4464, opened 2026-04-20
Interrupt-driven encoders

stellar-aria:
Switch the encoder hardware backend from polling to interrupt driven via IRQs. Note that this does not decrease the (perceived) user latency yet, since interpretEncoders is still called at the same rates. It _should_ prevent "missed" detents, though.

### DIFF
--- a/src/deluge/hid/encoders.cpp
+++ b/src/deluge/hid/encoders.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "hid/encoders.h"
+#include "OSLikeStuff/timers_interrupts/timers_interrupts.h"
 #include "definitions_cxx.hpp"
 #include "extern.h"
 #include "gui/ui/ui.h"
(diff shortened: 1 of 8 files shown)


src/deluge/hid/encoders.cpp
```diff
@@ -59,11 +129,16 @@ void init() {
...
 	getEncoder(EncoderName::MOD_1).setNonDetentMode();
+
+	initInterrupts();
 }
 
 void readEncoders() {
-	for (auto& encoder : encoders) {
-		encoder.read();
+	for (size_t i = 0; i < util::to_underlying(EncoderName::MAX_ENCODER); i++) {
+		int8_t edges = encoderEdgeDeltas[i].exchange(0, std::memory_order_acquire);
```
m-m-adams: I think this can be relaxed - its ordering wrt any following operations doesn't matter

m-m-adams (approved):
I'm not 100% on that comment so feel free to merge if you disagree, it's just a difference of one `dmb ish` in each use

src/deluge/hid/encoders.cpp
stellar-aria: Done

m-m-adams (approved):

PR #4470, opened 2026-04-25
Feature/kit midi out at kit level

AaronRito:
This feature allows for a midi channel to be assigned to an entire kit clip. The feature will auto assign rows to standard GM kit notes (row 0 = note 32, row 1 = 33, row 2 = 34, and so on). This will align the deluge with most other commercial drum machines that output standard GM starting at note 32. Individual row channel and note assignment feature is unchanged and will override the auto assign if enabled.


seangoodvibes:
Hi,

Can you please separate the individual bug fixes into separate PR's?

Thank you

AaronRito:
> Hi,
> 
> Can you please separate the individual bug fixes into separate PR's?
> 
> Thank you

Posted

src/deluge/model/instrument/kit.cpp
```diff
@@ -1294,6 +1294,18 @@ void Kit::noteOnPreKitArp(ModelStackWithThreeMainThings* modelStack, Drum* drum,
 	}
 	NoteRow* thisNoteRow = ((InstrumentClip*)activeClip)->getNoteRowForDrum(drum, &drumIndex);
 	if (drumIndex != -1 && thisNoteRow->drum != nullptr) {
+		// Kit-level MIDI output: channel lives on the InstrumentClip so it survives kit preset swaps
```
m-m-adams: I'm not sure about this - it might make more sense to keep it in the preset since that would let it be used for controlling specific drum machines?

m-m-adams (commented):
Would it make sense to have the output base note get set when the midi learn kit base note is set too?

PR #4472, opened 2026-04-25
allow naming MIDI & gate drums

nikodemus:
- name storage moves from SoundDrum to Drum, changes to std::string, which causes much fallout - but hopefully less bugs. Member storing it renamed to drumName and getter follows suit: this can be undone, but it added some clarity when dealing with the fallout - easier to know we're actually dealing with drum names...

- MIDI drum description squeezed to one line, so that it still fits nicely when we add then name

### DIFF
--- a/src/deluge/processing/sound/sound_drum.h
+++ b/src/deluge/processing/sound/sound_drum.h
@@ -27,7 +27,6 @@ class ModelStackWithTimelineCounter;
 
 class SoundDrum final : public Sound, public Drum {
 public:
-	String name;
 	String path;
 	bool nameIsDiscardable = false;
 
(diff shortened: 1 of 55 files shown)


src/deluge/processing/sound/sound_drum.h
```diff
@@ -50,7 +49,7 @@ class SoundDrum final : public Sound, public Drum {
 	Error loadAllSamples(bool mayActuallyReadFiles) override;
 	void writeToFile(Serializer& writer, bool savingSong, ParamManager* paramManager) override;
 	void writeToFileAsInstrument(bool savingSong, ParamManager* paramManager);
-	void getName(char* buffer) override {}
+	std::string getDrumName() override;
```
m-m-adams, line 52: should this be a string view? I guess maybe not since it's replacing a copy API?

nikodemus: That, and in case of MIDI and CV drums the name is going to be at least partially constructed on request - so there's no backing store to return a pointer to.

m-m-adams (approved):

PR #4477, opened 2026-04-26
Fix ghost notes at start of linear MIDI recording (#4225)

### LINKED ISSUE #4225: First note of first beat over usb midi adds extra notes

When recording any loop using the REC+New Row Looping feature, the first note seems to create a kind of cluster chord around the single note. I'm not sure of the velocity of each additional note, but it does have a noticeable effect on the audio. Deluge is in usb host mode connected to a midi keyboard (minilab) using the midi follow mode. I have found this occurs with almost all of the synth kits. This seems to be a newish problem as it didn't occur in the nightly build from a few weeks ago. 

When recording a new loop variation of another synth row, the notes should not add additional notes over midi. 

It happened to me with yesterdays nightly build. 

_No response_

OLED

Beta

deluge-v1_3_0-beta+2025_11_27-549828d.bin

1. Turn on
2. choose synth kit
3. press play
4. hold rec+ press row underneath desired new loop orientation
5. play midi note from midi keyboard using midi follow usb mode
6. go into clip mode and see cluster chord

AaronRito:
When beginLinearRecording() inserts early-arrived notes at position 0, ignoreNoteOnsBefore_ was only set when the note had already been released (still_active == false). For still-held notes, the live MIDI stream (or the clip-length-doubling replay path) could fire another note-on before tick 1 and ghost-duplicate the note with zero or corrupted velocity.

Fix: always set ignoreNoteOnsBefore_ = 1 after inserting an early note at position 0, regardless of still_active / earlyNoteStillActive. The flag only suppresses note-ons before tick 1; note-offs and subsequent presses are unaffected, so held notes still release and re-trigger correctly.

Fix #4225

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -461,13 +461,11 @@ Error InstrumentClip::beginLinearRecording(ModelStackWithTimelineCounter* modelS
 					Iterance iterance = noteRow->getDefaultIterance();
 					int32_t fill = noteRow->getDefaultFill(modelStackWithNoteRow);
 					noteRow->attemptNoteAdd(0, 1, velocity, probability, iterance, fill, modelStackWithNoteRow, action);
-					if (!thisDrum->earlyNoteStillActive) {
-						D_PRINTLN("skipping next note");
-
-						// We just inserted a note-on for an "early" note that is still sounding at time 0, so ignore
-						// note-ons until at least tick 1 to avoid double-playing that note
-						noteRow->ignoreNoteOnsBefore_ = 1;
-					}
+					// Always suppress re-trigger from the live MIDI stream: we just inserted this note at position 0
+					// via earlyNotes, so any note-on arriving before tick 1 (including clip-length-doubling replays)
+					// would produce a ghost note with zero/corrupted velocity. This applies whether or not the note
+					// is still physically held — note-offs and presses at tick >= 1 are unaffected.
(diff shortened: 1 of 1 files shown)


src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -461,13 +461,11 @@ Error InstrumentClip::beginLinearRecording(ModelStackWithTimelineCounter* modelS
...
-
-						// We just inserted a note-on for an "early" note that is still sounding at time 0, so ignore
-						// note-ons until at least tick 1 to avoid double-playing that note
-						noteRow->ignoreNoteOnsBefore_ = 1;
-					}
+					// Always suppress re-trigger from the live MIDI stream: we just inserted this note at position 0
+					// via earlyNotes, so any note-on arriving before tick 1 (including clip-length-doubling replays)
+					// would produce a ghost note with zero/corrupted velocity. This applies whether or not the note
+					// is still physically held — note-offs and presses at tick >= 1 are unaffected.
+					noteRow->ignoreNoteOnsBefore_ = 1;
```
m-m-adams, line 468: what happens if a midi note arrives during tick 0? e.g. not an early one but before tick 1

m-m-adams (approved):
eh seems fine in practice

seangoodvibes:
Going to revert this because this does not solve the issue

This issue did not always exist (it's fine in 1.2) - so changing this code (which has existed since official) is not the answer

PR #4478, opened 2026-04-28
Fix MIDI program change off-by-one and missed clip activations

AaronRito:
Two related fixes for MIDI program/bank/sub-bank handling:

1. Display off-by-one (preset.h): drawValue, drawInteger, and renderInHorizontalMenu showed the raw 0-indexed stored value, so users saw 0 when their synth was on Program 1. Display getValue()+1 for the program number range while keeping storage and the wire value at the spec-compliant 0-127 range. NONE (128) unchanged.

2. PC not firing on first activation / grid view (midi_instrument.cpp): shouldSendPGMs required activeClip != nullptr AND that the new clip had a different PGM/Sub/Bank than the previous clip. The first activation of a MIDI clip never met the first condition, and re-launching a clip with the same program in grid view never met the second. Replace with: send whenever switching to a different clip that has any of PGM/Sub/Bank configured.

### DIFF
--- a/src/deluge/gui/menu_item/midi/preset.h
+++ b/src/deluge/gui/menu_item/midi/preset.h
@@ -38,7 +38,7 @@ public:
 			text = l10n::get(l10n::String::STRING_FOR_NONE);
 		}
 		else {
-			intToString(this->getValue(), buffer, 1);
+			intToString(this->getValue() + 1, buffer, 1); // show 1-based
 			text = buffer;
 		}
 		canvas.drawStringCentred(text, yPixel + OLED_MAIN_TOPMOST_PIXEL, textWidth, textHeight);
@@ -49,7 +49,7 @@ public:
 			display->setText(l10n::get(l10n::String::STRING_FOR_NONE));
 		}
 		else {
-			display->setTextAsNumber(this->getValue());
+			display->setTextAsNumber(this->getValue() + 1); // MIDI programs are 0-indexed internally; show 1-based
 		}
 	}
 
@@ -79,7 +79,7 @@ public:
 			size_y = kTextSpacingY;
 		}
 		else {
-			paramValue.appendInt(getValue());
+			paramValue.appendInt(getValue() + 1); // show 1-based
 			size_x = kTextTitleSpacingX;
 			size_y = kTextTitleSizeY;
 		}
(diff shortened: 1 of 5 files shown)


src/deluge/gui/menu_item/midi/preset.h
```diff
@@ -38,7 +38,7 @@ class Preset : public Integer {
 			text = l10n::get(l10n::String::STRING_FOR_NONE);
 		}
 		else {
-			intToString(this->getValue(), buffer, 1);
+			intToString(this->getValue() + 1, buffer, 1); // show 1-based
```
m-m-adams: Is this better? Most of my gear seems to use 0 based PCs

m-m-adams: maybe it should be a per channel option?

m-m-adams (commented):
Not sure about changing the indexing but the other one lgtm

src/deluge/gui/menu_item/midi/preset.h
AaronRito: Its better, it works correctly now.....None of my gear start at preset 0. Arturia, Roland, Moog. You can still navigate to 0.

m-m-adams: Quad cortex, morningstar, strymon and airstep are all 0 based (arguably elektron too, they're explicit that PC 0 selects preset 1 and you see it that way on the screen). On the sending side I think it's more confusing to send 0 when you select 1 as a default.

I'm happy to have it as an option but I'm not gonna merge it as a global change

AaronRito: I really dont care if it starts on 0 or 1, it just needs to send the right value, which it doesnt right now. It just failed some more robust manual testing I did, so I'll put it in draft until I can get it working right.

AaronRito: @m-m-adams One thing I think is weird about this feature, is that it sends the program change while youre adjusting the setting. So as youre turning the encoder, its pushing PCs that cause devices to PC one by one, which could be considered "undesirable" especially if you havent saved some existing change on the client side. How do you feel about disabling that, and sending the PC only after the selection is complete, or not at all until it's actually invoked by a clip activation or otherwise?

m-m-adams: Yeah that makes sense to me, maybe send it a second after the last change?

PR #4508, opened 2026-05-15
WIP: make flash via vendored OpenOCD with renesas_spibsc driver

heeen:
WIP. Adds `make flash` (and `./dbt flash`) wired to a custom `renesas_spibsc` OpenOCD flash driver for the Deluge's SPIBSC0 controller. The driver's `smart_program` command CRC-sweeps the flash and only reprograms 4 KiB sub-sectors that differ — typically ~10x faster than upstream `program` for incremental builds.

OpenOCD source: https://github.com/heeen/openocd (branch `deluge-spibsc-v0.12.0`, v0.12.0 + driver patch). Build-openocd.sh clones it, configures with `--prefix` pointing at the DBT-toolchain openocd dir, and `make install`s over the stock xPack binary — single openocd on $PATH afterwards. Linux + macOS supported; Windows guard with a clear error message (needs a one-off CI build to produce `openocd.exe` with the patch applied, separate task).

Also includes an abandoned-spike doc (`scripts/debug/flash/deluge-flash-tcl.py.experimental`) explaining why driving the mailbox over stock openocd's TCL RPC doesn't work — three independent walls verified on hardware, kept as documentation so future contributors don't re-investigate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ else
 		export DBT := ./dbt
 endif
 
+# Build configuration for `make flash` (release | debug | relwithdebinfo | CMake name).
+CONFIG ?= release
+
 all: debug release
 
 debug:
@@ -14,3 +17,7 @@ release:
 
 clean:
 	$(DBT) clean
+
+.PHONY: flash
+flash:
+	$(DBT) flash $(CONFIG)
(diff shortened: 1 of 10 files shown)


seangoodvibes:
Hi, just letting you know that we are going closing this PR because it conflicts with / will be superseded by ongoing efforts for the deluge 2.0 firmwares. Thanks

PR #4516, opened 2026-05-25
Add column-wide affect-entire combo to grid session view

AaronRito:
## Summary
Adds a new gesture in the grid session view: hold any clip pad in a column together with AFFECT_ENTIRE to broadcast gold-encoder parameter changes to every clip in that column at once.

## Details
- Routes changes through each clip's clip-level paramManager only, so per-drum overrides on kit note rows are preserved.
- Absolute snap, momentary behavior (changes apply while the combo is held).
- Held column tints toward white as a visual cue.
- AFFECT_ENTIRE LED blinks while the combo is engaged.

## Files changed
- src/deluge/gui/views/session_view.cpp (+98 / -2)
- src/deluge/gui/views/session_view.h  (+8)

## Test plan
- [ ] Hold a clip pad in a column + AFFECT_ENTIRE, turn gold encoder, verify all clips in that column receive the change.
- [ ] Verify per-drum overrides on kit note rows are preserved.
- [ ] Verify column tints toward white while held.
- [ ] Verify AFFECT_ENTIRE LED blinks while engaged.
- [ ] Release combo and confirm normal AFFECT_ENTIRE behavior resumes

### DIFF
--- a/src/deluge/gui/views/session_view.cpp
+++ b/src/deluge/gui/views/session_view.cpp
@@ -387,9 +387,16 @@ moveAfterClipInstance:
 
 	// Affect-entire button
 	else if (b == AFFECT_ENTIRE) {
+		// Track held state independently of the toggle below — the toggle is gated on UI_MODE_NONE,
+		// but the column-affect-entire combo needs to detect the button while a clip pad is held.
+		affectEntireButtonHeld = on;
 		if (on && currentUIMode == UI_MODE_NONE) {
 			currentSong->affectEntire = !currentSong->affectEntire;
 			view.setActiveModControllableTimelineCounter(currentSong);
+		}
+		updateAffectEntireLED();
+		if (currentSong->sessionLayout == SessionLayoutType::SessionLayoutTypeGrid) {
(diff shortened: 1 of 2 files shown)


soymonitus:
I hope this gets added into 1.3 release. It makes so much sense to include it. Remember I added Affect-Entire holding to affect all rows in a kit. This is the same but for clips in a track. It is one of those utility features that makes life easier. The code changes are so small I think we could include it, please!! @seangoodvibes 
I did a quick test and works nice

seangoodvibes:
> I hope this gets added into 1.3 release. It makes so much sense to include it. Remember I added Affect-Entire holding to affect all rows in a kit. This is the same but for clips in a track. It is one of those utility features that makes life easier. The code changes are so small I think we could include it, please!! @seangoodvibes 
> 
> I did a quick test and works nice

Sorry we can't accept brand new features anymore.

The code may be small but we would have to review and look at it from the perspective of: "is it the right way to do this, is it accessible", etc etc

And accepting means we have to maintain it for the release if anything comes up.

PR #4522, opened 2026-06-02
Create EMX-1 profile

IainD92:
Profile for korg emx-1.

All parameters that are controllable from CC WITHOUT using nrpn.

I've included value tables for the stuff that's a pain to remember (osc type, fx type, mod destination, etc). So you can reference it without scouring the internet every time you forget.
Taken from the official korg midi implementation for the EMX-1, and converted to decimal from hex for ease of use.


IainD92:
Will test this properly in the morning.

Are comments in XMLs allowed?

seangoodvibes:
> Will test this properly in the morning.
> 
> 
> 
> Are comments in XMLs allowed?

Thanks

Comments in the XML is fine if it works :)

IainD92:
Tested, works like a charm. Thanks for adding midi profiles!

seangoodvibes (approved):

PR #4526, opened 2026-06-10
Feature Extend/ Add midi follow control for specific tracks

seangoodvibes:
Added midi follow control for up to 16 specific midi follow track channels for deluge tracks 1-16.

- forward's cc's, notes, pitch bend and after touch to specific tracks
- add's default cc's 89 and 90 to forward mute and solo commands respectively to specific tracks when cc value received is greater than 0

To do:
- [x] Add track mute / solo commands
- [x] Add shift control toggle
- [x] Move all midi follow settings to MIDIFollow.xml


src/deluge/gui/l10n/english.json
```diff
@@ -889,6 +889,23 @@
     "STRING_FOR_FOLLOW_CHANNEL_A": "Channel A",
     "STRING_FOR_FOLLOW_CHANNEL_B": "Channel B",
     "STRING_FOR_FOLLOW_CHANNEL_C": "Channel C",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK": "Channel Track",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK1": "Channel Track 1",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK2": "Channel Track 2",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK3": "Channel Track 3",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK4": "Channel Track 4",
+    "STRING_FOR_FOLLOW_CHANNEL_TRACK5": "Channel Track 5",
```
stellar-aria: these should all be a single string of "channel track" and then just get the number appended at runtime/compile time

src/deluge/gui/l10n/seven_segment.json
```diff
@@ -497,6 +497,23 @@
         "STRING_FOR_FOLLOW_CHANNEL_A": "A",
         "STRING_FOR_FOLLOW_CHANNEL_B": "B",
         "STRING_FOR_FOLLOW_CHANNEL_C": "C",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK": "TRK",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK1": "T1",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK2": "T2",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK3": "T3",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK4": "T4",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK5": "T5",
+        "STRING_FOR_FOLLOW_CHANNEL_TRACK6": "T6",
```
stellar-aria: same here, just use "T" and the number at runtime

stellar-aria (changes requested):

soymonitus:
Right now CC 122 makes Shift engaged from ANY channel (intentianal?)
Anyways, I think the Shift command should be included as a global MIDI COMMAND, learnable, so anyone can decide which CC they want. This feature could be useful outside the midi follow context, because some people could need an external button to act as Shift (maybe it is more comfortable for their setup, or maybe they have a faulty Deluge shift)

seangoodvibes:
> Right now CC 122 makes Shift engaged from ANY channel (intentianal?)
> 
> Anyways, I think the Shift command should be included as a global MIDI COMMAND, learnable, so anyone can decide which CC they want. This feature could be useful outside the midi follow context, because some people could need an external button to act as Shift (maybe it is more comfortable for their setup, or maybe they have a faulty Deluge shift)

It's already usable outside midi follow context

It's just a default cc

Anyway I can do global midi command but I'll have to disable cc 122 again as it wasn't intended for anything else.

src/definitions_cxx.hpp
```diff
@@ -362,6 +362,7 @@ enum class MIDIFollowFeedbackAutomationMode : uint8_t {
 	LOW,
 	MEDIUM,
 	HIGH,
+	LAST,
```
stellar-aria: Don't use LAST inside of an enum class, it's intended to be a closed set of possible values. instead, define a constant like kNumMIDIFollowFeedbackAutomationModes

stellar-aria (changes requested):

stellar-aria (approved):

PR #4535, opened 2026-06-12
Fix DIN device disappearing when scrolling MIDI devices menu

### LINKED ISSUE #4520: Midi Device list glitches out when scrolling

See video, the Din-device disappears/becomes invisible when scrolling it out of the viewport

Steps to reproduce If you have an Arturia minilab mk2 or similar usb-only device, connect through the usb b/a adapter, use Deluge as midi host. And use Nightly Firmware.

https://github.com/user-attachments/assets/b53dbd33-a73a-45cf-80b6-18a4534ec8bd

Device list not glitching out.

An hour ago

_No response_

OLED

Nightly

_No response_

_No response_

kazemio:
**Issue:** In Settings → MIDI → Devices on OLED, scrolling down past DIN and then scrolling back up never brought DIN (or the upstream USB ports) back into view, and the selection highlight vanished.

  **Cause:** `currentScroll` was declared `size_t`, but the menu uses negative indices for built-in cables (DIN = -3, upstream USB = -2/-1). The signed/unsigned comparison in `selectEncoderAction` (`getValue() < currentScroll`) converted -3 to a huge unsigned value, so the viewport never scrolled back up once it had advanced to a hosted device (index ≥ 0) — the selection moved to DIN off-screen.
  
  **Fix:** Declare `currentScroll` as `int32_t` so all scroll comparisons are signed.

  **Testing:** Scroll down in the MIDI devices list until DIN leaves the frame, then scroll back up — DIN and upstream USB ports scroll back into view with the highlight intact. See attached videos.
  
- Before:
https://github.com/user-attachments/assets/0dad464b-a8dd-4d0c-8276-cc1f788ce7e9

- After: 
https://github.com/user-attachments/assets/5294b4f4-caa1-46b6-b246-6d7e913ae742

Fixes #4520.

### DIFF
--- a/src/deluge/gui/menu_item/midi/devices.cpp
+++ b/src/deluge/gui/menu_item/midi/devices.cpp
@@ -52,7 +52,7 @@ void Devices::beginSession(MenuItem* navigatedBackwardFrom) {
 
 	soundEditor.currentMIDICable = getCable(this->getValue());
 	if (display->haveOLED()) {
-		currentScroll = this->getValue();
+		currentScroll_ = this->getValue();
 	}
 	else {
 		drawValue();
@@ -86,8 +86,8 @@ void Devices::selectEncoderAction(int32_t offset) {
 	// Don't show devices which aren't connected. Sometimes we won't even have a name to display for them.
 
 	if (display->haveOLED()) {
-		if (this->getValue() < currentScroll) {
-			currentScroll = this->getValue();
+		if (this->getValue() < currentScroll_) {
+			currentScroll_ = this->getValue();
 		}
 		//
 		if (offset >= 0) {
@@ -95,7 +95,7 @@ void Devices::selectEncoderAction(int32_t offset) {
 			int32_t numSeen = 1;
 			while (d > lowestDeviceNum) {
 				d--;
-				if (d == currentScroll) {
+				if (d == currentScroll_) {
 					break;
 				}
 				auto device = getCable(d);
@@ -104,7 +104,7 @@ void Devices::selectEncoderAction(int32_t offset) {
 				}
 				numSeen++;
 				if (numSeen >= kOLEDMenuNumOptionsVisible) {
-					currentScroll = d;
+					currentScroll_ = d;
 					break;
 				}
 			}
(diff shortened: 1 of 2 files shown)


src/deluge/gui/menu_item/midi/devices.cpp
```diff
@@ -86,16 +87,14 @@ void Devices::selectEncoderAction(int32_t offset) {
 	// Don't show devices which aren't connected. Sometimes we won't even have a name to display for them.
 
 	if (display->haveOLED()) {
-		if (this->getValue() < currentScroll) {
```
kazemio, line 89: Changed this to use `std::min` to resolve clang error.

src/deluge/gui/menu_item/midi/devices.cpp
```diff
@@ -25,6 +25,7 @@
 #include "io/midi/cable_types/usb_hosted.h"
 #include "io/midi/midi_device.h"
 #include "io/midi/midi_device_manager.h"
+#include <algorithm>
```
kazemio, line 28: Required for `std::min`.

src/deluge/gui/menu_item/midi/devices.h
```diff
@@ -32,7 +32,7 @@ class Devices final : public Value<int32_t> {
 	void drawPixelsForOled();
 
 private:
-	size_t currentScroll;
+	int32_t currentScroll_ = 0;
```
seangoodvibes: since you're touching currentScroll, can you please change it to current_scroll instead and without the trailing underscore?

we're moving away from camelCase to snake_case.

thanks!

seangoodvibes (changes requested):

seangoodvibes:
I tested it and it looks like it fixes it, plus it also fixes issue I noticed without any midi devices connected where if you go into the din menu, scroll down and press back that the devices menu is blank.

Just one small formatting change request and then we're good to merge

Thanks for fixing this

src/deluge/gui/menu_item/midi/devices.h
stellar-aria: the underscore is correct here, indicates private member. it's in the clangd stuff

seangoodvibes: > the underscore is correct here, indicates private member. it's in the clangd stuff

woopsies

ok ya current_scroll_

kazemio: Done!

kazemio:
> I tested it and it looks like it fixes it, plus it also fixes issue I noticed without any midi devices connected where if you go into the din menu, scroll down and press back that the devices menu is blank.
> 
> Just one small formatting change request and then we're good to merge
> 
> Thanks for fixing this

Woohoo! No worries, trying to help where I can!

seangoodvibes (approved):

PR #4539, opened 2026-06-13
Fix tempo automation stepping instead of ramping when recorded in Arranger

### LINKED ISSUE #3221: Tempo automation in Arranger view not played back correctly

After recording tempo automation in Arranger view with an arrangement of 16 bars and with 1 or 2 tempo changes of 1 beat per bar, the playback skips a lot of tempo change automation, resulting in sudden changes of tempo in stead of smooth tempo change.
Extra info: 2 of 4 tracks are Midi tracks.

The tempo should be adjusted according to automation.

Just now

No

OLED

Release 1.2/Chopin

C1.2.0

Create 4 clips (1 kits, 1 synth, 2 midi) in Song view and arrange 16 bars in arranger view on all 4 clips. 
The kit clip is 2 bars long, the other clips just 1 bar.
In arranger view press Record and start playing. Turn tempo knop 1 beat up 1 or 2 times every bar. At end of arrangement stop playing, exit record mode, and press play to hear result.
The tempo now is updated very irregularly because a lot of tempo changes are not played back.

kazemio:
**Problem**
  Recording tempo automation in Arranger view (e.g. a few tempo changes across a 16-bar arrangement) produced sudden tempo jumps instead of smooth ramps. It only ramped between points that happened to be recorded in the same continuous knob movement.

  **Cause**
  Tempo is stored as automation nodes on the `UNPATCHED_TEMPO` param. Playback only ramps between two nodes when the later node is flagged `interpolated`; otherwise it jumps straight to the value. That flag is decided at record time by a generic 150 ms proximity heuristic (`TIME_TO_INTERPOLATE_WITHIN`) — only nodes recorded <150 ms apart get interpolated. That heuristic suits dense synth-param gestures (filter sweeps, etc.) but is wrong for tempo, which is inherently sparse (even at 300 BPM a beat is 200 ms), so virtually every recorded tempo point ended up stepped.

  **Fix**
  In the live-record path (`AutoParam::setCurrentValueInResponseToUserInput`), force interpolation for the tempo param only (`UNPATCHED_TEMPO` + `Kind::UNPATCHED_GLOBAL`). Recorded tempo nodes now ramp to the next by default.

Fixes #3221.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -5972,12 +5972,24 @@ ModelStackWithAutoParam* Song::getModelStackWithParam(ModelStackWithThreeMainThi
 }
 
 void Song::updateBPMFromAutomation() {
+	// Evaluate the tempo from the automation at the current playhead position, rather than reading the param's
+	// interpolated currentValue. The generic per-sample interpolation (AutoParam::tickSamples) rounds its increment
+	// to zero for the tempo param - tempo values are tiny relative to the int32 param range - so currentValue never
+	// advances between interpolated nodes and the tempo freezes. getValueAtPos() interpolates exactly between the
+	// surrounding nodes (and still steps at non-interpolated nodes, preserving deliberate "snap" tempo changes).
+	char model_stack_memory[MODEL_STACK_MAX_SIZE];
+	ModelStackWithThreeMainThings* model_stack_with_three_main_things =
+	    setupModelStackWithSongAsTimelineCounter(model_stack_memory);
+	ModelStackWithAutoParam* tempo_param =
+	    getModelStackWithParam(model_stack_with_three_main_things, params::UnpatchedGlobal::UNPATCHED_TEMPO);
+
+	int32_t current_tempo = tempo_param->autoParam->getValueAtPos(arrangement.lastProcessedPos, tempo_param);
+
 	// There seems to be a param manager bug where it occasionally reports unautomated params as 0 so just ignore
 	// that
-	int32_t currentTempo = currentSong->paramManager.getUnpatchedParamSet()->getValue(params::UNPATCHED_TEMPO);
-	if (currentTempo && currentTempo != intBPM) {
-		setBPMInner((float)currentTempo / 100, false);
-		intBPM = currentTempo;
+	if (current_tempo && current_tempo != intBPM) {
+		setBPMInner((float)current_tempo / 100, false);
+		intBPM = current_tempo;
 	}
 }
 


stellar-aria:
Ramp-by-default isn't really a great solution, since sharp tempo changes are musically valid. We need to respect the automation interpolation setting instead

stellar-aria (changes requested):
Ramp-by-default isn't a good solution here, we need to allow sharp changes as well

kazemio:
> Ramp-by-default isn't really a great solution, since sharp tempo changes are musically valid. We need to respect the automation interpolation setting instead

Good catch! Fixed it to honor the setting instead of forcing true. Thanks!

src/deluge/modulation/automation/auto_param.cpp
```diff
@@ -192,6 +192,15 @@ void AutoParam::setCurrentValueInResponseToUserInput(int32_t value, ModelStackWi
 				}
 			}
 
+			// When recording tempo automation, honor the Automation View interpolation toggle instead of the generic
+			// 150ms proximity heuristic above (which suits dense synth gestures but leaves sparse tempo points
+			// stepped). This lets the user choose smooth ramps (interpolation on) or sharp tempo changes (interpolation
+			// off).
+			if (modelStack->paramId == deluge::modulation::params::UNPATCHED_TEMPO
+			    && modelStack->paramCollection->getParamKind() == deluge::modulation::params::Kind::UNPATCHED_GLOBAL) {
+				shouldInterpolateRegionStart = automationView.interpolation;
```
seangoodvibes: this is not a good fix either because the automation view interpolation setting is only used when entering automation on the grid in automation view (which for the tempo param, doesn't exist yet in arranger view)

it should not be used if you're not in automation view and recording tempo automation with the tempo knob.

seangoodvibes (changes requested):

seangoodvibes:
> > Ramp-by-default isn't really a great solution, since sharp tempo changes are musically valid. We need to respect the automation interpolation setting instead
> 
> Good catch! Fixed it to honor the setting instead of forcing true. Thanks!



> > Ramp-by-default isn't really a great solution, since sharp tempo changes are musically valid. We need to respect the automation interpolation setting instead
> 
> Good catch! Fixed it to honor the setting instead of forcing true. Thanks!

Unfortunately that isn't the correct fix either

src/deluge/modulation/automation/auto_param.cpp
kazemio: Yeah, that makes sense. What *should* decide whether nodes interpolate? Is that why that 150 ms heuristic was in there to begin with?

stellar-aria: That's what it looks like, yes. There was no automation view or manual automation node placement when the firmware was open-sourced, and that code has been there since the initial release on GitHub.

seangoodvibes: > Yeah, that makes sense. What _should_ decide whether nodes interpolate? Is that why that 150 ms heuristic was in there to begin with?

I think it was indeed there to determine if it should interpolate the nodes before the recorded automation node.

This function is really only called when you record automation with gold knobs and maybe external midi cc's

Tempo automation is recording automation using tempo encoder which is not as smooth as gold knobs, so I don't know if the timing affects things.

seangoodvibes: Update I confirmed it's not the encoder that's the problem. As I tried recording tempo automation with gold knobs and had same issue.

seangoodvibes: Last point: I wouldn't say the tempo automation is being recorded stepped either. It looks like tempo is not even getting recorded properly.

If it was stepped I would still see all the values (you can record stepped in automation view and you'll see the values you recorded on the screen as it plays back the automation, even if stepped).

kazemio: It's definitely a little weird. Can I ask how you were able to learn the tempo to the gold knobs? And how you were able to record stepped tempo in automation view? I don't see tempo as an available automation lane.

seangoodvibes: > It's definitely a little weird. Can I ask how you were able to learn the tempo to the gold knobs? And how you were able to record stepped tempo in automation view? I don't see tempo as an available automation lane.

I just changed the encoder code to get the gold knobs to call playbackHandler.tempoEncoderAction instead of currentUI->modEncoderAction

As for automation view, I didn't actually use automation view for tempo step recording. I just mean if you step record other regular parameters in automation view, the values don't get skipped like they do with tempo which is shown on the display.

kazemio: Got it. I think this latest change fixed something!

Bug:
https://github.com/user-attachments/assets/2bb9551d-4c53-4cbe-9829-72a7550cbbd6

Fix:
https://github.com/user-attachments/assets/f75b57e8-652a-4008-9920-3edd2385b994

kazemio: Here's Claude's summary of the above fix:

### Problem
  Recording tempo automation in Arranger and playing it back would get "stuck" — the tempo would jump at the start of a change, then freeze at an intermediate value for the rest of the playback (e.g. a ramp recorded to 140 BPM would stick at 121). It was intermittent: sometimes the same gesture played back fine.

  ### Cause
  Between two **interpolated** automation nodes, the song's tempo is advanced only by the generic per-sample interpolation,
  `AutoParam::tickSamples`, which computes:

  round(valueIncrementPerHalfTick × timePerInternalTickInverse / 2³²) × 6 × numSamples

  It rounds to an integer **before** scaling by `numSamples`. Tempo values are tiny relative to the int32 param range (BPM×100 ≈ 12000), so the increment rounds to **0** and `currentValue` never advances — the tempo freezes. It only ever moved at **non‑interpolated** nodes, where playback jumps the value directly. Whether a recording contains non‑interpolated nodes depends on the recorder's 150 ms proximity heuristic, which is why the bug looked intermittent. This has existed since tempo automation was added (#2367).
  
  ### Fix
  `updateBPMFromAutomation` now reads the tempo straight from the automation at the current playhead position via `AutoParam::getValueAtPos()`, instead of relying on the broken interpolated `currentValue`. This interpolates exactly between nodes and still steps at non‑interpolated nodes (so deliberate "snap" tempo changes are preserved). It's tempo‑only and doesn't touch the shared per‑sample interpolation path.

  ### How we narrowed it down
  1. **Decoded the saved song's `tempo=` automation data** → the nodes were always recorded correctly and reached the target value, ruling out a recording bug.
  2. **Stripped the song to tempo automation only** (no MIDI, no tracks) → still froze, ruling out MIDI clock-out and clips.
  3. **Added on-device logging** (per-tick: playhead pos, tempo value, interpolation increment) → showed the increment was set and the playhead advanced, but the tempo value never moved → isolated it to the per‑sample interpolation.
  4. **Added counters** for whether that interpolation block ran → it *was* running every block, so the increment itself was computing to zero → pinpointed the `round(...) × numSamples` precision loss.

seangoodvibes: hey thanks for the investigative work. Unfortunately I think the solution is more complex than this and I need to revisit how tempo automation is actually done. 

Ultimately we need to be able to rely on the current value from the autoParam class. We shouldn't work around it.

Something is fundamentally wrong with how tempo automation is stored.

I know the solution I just need to work it out.

I also have to integrate that with the action logger system because there's another related issue to all this pertaining to that - so they go hand in hand.

You've pointed me in the right direction so that's awesome :)!

seangoodvibes:
Just changing this to draft for now because I need to look at some stuff to figure out the best solution.

stellar-aria:
Closing in favor of #4540

PR #4568, opened 2026-06-22
Add a priority slot for CV when OLED is being sent

### LINKED ISSUE #3520: CV tracks: gate timing problems when pressing UI grid buttons

Timing of gate impulses in CV clips is massively impacted by pressing UI buttons in grid mode. When unmuting a clip the displayed countdown in the first row also seems to impact CV gate timing.

UI interaction does not impact gate timing

Observed in deluge-v1_3_0-nightly+2025_03_25-541d2dc and nightlies before

No

OLED

Nightly

deluge-v1_3_0-nightly+2025_03_25-541d2dc

Setup: Create a song in grid view and populate some columns with clips. Add a CV clip on channel 1and evenly fill it with notes (e.g. eights). Connect an external synth which is triggered by the gate 1 output (in my case Wirehead instruments Freaq FM)
Start sequencer and unmute CV 1 clip. Go back to grid view and randomly mute/unmute other clips --> you will hear that the external synth plays unevenly until the muted/unmuted clip has switchted its state.

stellar-aria:
Basically makes it so that if there's any OLED sending in-flight and a CV message needs to be sent, the next message out is _always_ the CV message.

### DIFF
--- a/src/RZA1/oled/oled_low_level.c
+++ b/src/RZA1/oled/oled_low_level.c
@@ -37,12 +37,29 @@ void setupSPIInterrupts()
     setupAndEnableInterrupt(cvSPITransferComplete, INTC_ID_SPRI0 + SPI_CHANNEL_CV * 3, 5);
 }
 
+// Priority CV channel: CV words bypass the OLED frame queue so gate timing isn't held
+// hostage by OLED refreshes on the shared SPI bus. A single latest-wins slot matches the queue's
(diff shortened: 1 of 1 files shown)


seangoodvibes (approved):

therealstaubi:
I can confirm that CV timing is not affected anymore. I understood that the root cause was the OLED-modification which caused a race condition with the CV timing.
Great work. Thanks for finding out and solving!

PR #4573, opened 2026-06-23
Recency clock: date saved files in save order (no RTC needed)

CommmandrCody:
## What this does

The Deluge has no real-time clock, so `get_fattime()` returned `0` and every saved file got a zero/1980 date. That makes it impossible to sort the song browser by how recently you actually worked on something - all the dates are the same.

This adds a small monotonic "recency clock": each save gets a packed FAT datetime later than the last one, so sorting files by date is the same as sorting by save order. At card mount the counter is seeded from the newest date already in `SONGS/` (one bounded, read-only directory scan), so new saves always land after what is already on the card and the order never goes backwards across power cycles.

The synthetic dates look a little odd if you pop the card into a computer, but on the device they give true "most recently worked on" ordering, which is the thing you actually want when you sit down to keep working on a track.

It also clamps the seed: a corrupt directory entry sitting up near the FAT year ceiling (~2106) can't get adopted as "newest" and pin the clock up there forever. Without the clamp one bad entry poisons every later save, and a computer renders ~2106 as a 1970 garbage date, so it looks like every save lost its date.

## Why

This is the groundwork for sorting the song browser by recency. On its own it just makes saved files carry a sensible, ordered date instead of all sharing date 0.

## Files

- `src/RZA1/diskio.c` - the recency clock (`get_fattime`, `fatPack`, `fatClockSeedFromPacked`) plus the seed clamp
- `src/deluge/storage/storage_manager.cpp` - seed the clock at card mount from the newest date in `SONGS/`
- `src/fatfs/ff.h` - declare `fatClockSeedFromPacked`

## Testing

- Built Release for 7SEG.
- On hardware: saved several songs in sequence, confirmed each lands newest; power-cycled and confirmed new saves still sort after the existing ones (clock re-seeds from the card).
- Tested on a 7-segment Deluge.

## Notes

No menus, no settings, no behaviour change anywhere except the date stamped on saved files. Does not touch playback or the audio path.

### DIFF
--- a/src/RZA1/diskio.c
+++ b/src/RZA1/diskio.c
@@ -294,9 +294,111 @@ DRESULT disk_ioctl(BYTE pdrv, /* Physical drive nmuber (0..) */
     }
 }
 
+/* ---- Recency clock -------------------------------------------------------
+ * The Deluge has no real-time clock, so it used to stamp every saved file with
+ * time 0. Instead we COUNT: each save gets a packed FAT datetime later than the
+ * last, so sorting by date == sorting by save order. The counter is seeded at
+ * card mount from the newest date already on the card (fatClockSeedFromPacked,
+ * called from storage_manager) so it never goes backwards across power cycles.
+ * The resulting "dates" are synthetic (they look odd on a computer) but give
+ * true recency on the device. */
+static uint32_t fatClockSecs2               = 0;  /* 2-second units since 1980-01-01 */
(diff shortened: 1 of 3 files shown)


seangoodvibes:

Thanks

CommmandrCody:

m-m-adams:
What? Yes they do. The split is at the 1.2 release

PR #4574, opened 2026-06-23
Autosave + crash recovery: don't lose an unsaved jam

CommmandrCody:
## What this does

The Deluge has no autosave, so a power loss or a stray button press can wipe a jam you never got around to naming. This adds a small safety net, with no impact on live playing.

It comes in three parts that share one invariant.

**Autosave.** Any undoable edit or structural change flags the song dirty. A low-priority, card-safe task (`RESOURCE_SD_ROUTINE`) writes `SYSTEM/RECOVER.XML` when transport is stopped and the card is free, via a temp file plus atomic rename so a power loss mid-write can't corrupt the recovery copy. It never runs during playback, so it can't glitch live audio. RECOVER lives outside `SONGS/`, so it never shows up in the song browser and can't collide with a song you named.

**Recover on boot.** If `SYSTEM/RECOVER.XML` is present at startup, the recovered song is loaded instead of the normal startup song, through the same proven load path, and the device says `Recovered unsaved song`. Its name is cleared and it's pointed at `SONGS/`, so saving works like any normal song with no "RECOVER" duplicates.

**Load-guard.** A first LOAD over an unsaved song just arms and warns (`Unsaved - press LOAD again to load over`); a second LOAD actually loads. Navigating away or BACK resets the arm. This stops an accidental double-LOAD from throwing away unsaved work.

The invariant: **RECOVER exists if and only if there is unsaved work.** A manual Save (or loading a different song) clears it, since there's no clean-shutdown signal to key off.

## Why

A no-RTC, card-based instrument has no equivalent of a DAW's autosave. Losing an unnamed jam to a power blip is a real and painful failure mode. This recovers it without changing how saving or playing works.

## Files

- `storage_manager.{cpp,h}` - `writeRecoveryFile()` / `clearRecoveryFile()` + the dirty flag
- `deluge.cpp` - the card-safe autosave task, and recover-on-boot in `setupStartupSong()`
- `action_logger.cpp`, `session_view.cpp` - set the dirty flag on edits / structural changes
- `save_song_ui.cpp` - clear recovery on a manual Save
- `load_song_ui.{cpp,h}` - the load-guard

## Testing

- Built Release for 7SEG.
- On hardware (7-segment): edited an unnamed jam, pulled power, confirmed it came back on the next boot with the "Recovered unsaved song" message; confirmed a manual Save clears the recovery file; confirmed a double-LOAD warns before loading over unsaved work.

## Notes

Autosave only writes while stopped and only when the card is otherwise free, so it stays off the live-audio path entirely. No new menus or settings.

### DIFF
--- a/src/deluge/deluge.cpp
+++ b/src/deluge/deluge.cpp
@@ -402,6 +402,16 @@ void setupStartupSong() {
 	auto defaultSongFullPath = "SONGS/DEFAULT.XML";
 	auto filename =
 	    startupSongMode == StartupSongMode::TEMPLATE ? defaultSongFullPath : runtimeFeatureSettings.getStartupSong();
+
+	// Recovery: if an unsaved song was left behind (power loss), load it INSTEAD of the normal startup
+	// song, through this exact same proven path so it inherits the crash-protection canary below.
+	// Presence of the file means unsaved work existed (a Save deletes it).
+	bool recoveringUnsavedSong = StorageManager::fileExists("SYSTEM/RECOVER.XML");
+	if (recoveringUnsavedSong) {
+		filename = "SYSTEM/RECOVER.XML";
+		startupSongMode = StartupSongMode::LASTOPENED; // force the load path (not TEMPLATE/BLANK)
+	}
+
 	String failSafePath;
 	failSafePath.concatenate("SONGS/__STARTUP_OFF_CHECK_");
 	auto size = strlen(filename) + 1;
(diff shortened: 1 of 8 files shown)


seangoodvibes:

Thanks

CommmandrCody:

One conflict on the way over, in `load_song_ui.cpp`: main removed the `LoadUI::enterKeyPress()` call before `performLoad()`. I took main's version and kept the load-guard around it, so this now inherits that change.

CommmandrCody:
> Hi @CommmandrCody just wanted to let you know that your PR is targeting the community branch which is deprecated. PR's should target the main branch.
> 
> Thanks

Hi @seangoodvibes 

Thanks for the nudge. I havent looked the PR comments in a bit. All four of those PRs  are rebased onto main now and retargeted, they build clean.

One thing you might want to flag for others: main and community have no common ancestor, so git rebase  on main just errors out and you have to cherrypick onto a fresh branch. Then the PR base needs retargeting separately or GitHub shows somehting like 1200 changed files. Took me a minute to work out what was going on but I added a cmooment as to not confuse anyone. 

Cheers

PR #4575, opened 2026-06-23
Retrospective MIDI capture: dump the noodle you just played into the clip

CommmandrCody:
## What this does

Ableton and Push 3 have a "capture" feature that grabs the notes you just played while you were noodling around, even though you weren't recording. This brings the same idea to keyboard view on the Deluge.

While the feature is on, every note you play in a keyboard layout gets stashed in a small background buffer. Timing is kept in audio samples, so it works whether or not the transport is running. When you play something you like, **SHIFT + RECORD** dumps the buffer into the current clip's piano roll. The buffer clears when you change track, change layout, or leave keyboard view, so it only ever holds the take you're actually working on.

The dump measures how long you played and grows the clip to fit, rounded to whole bars (multiples of the clip's current length), so two bars played over a one-bar loop become a real two-bar clip instead of folding back onto bar one. Notes go in through the same path live recording uses (`getOrCreateNoteRowForYNote` + `attemptNoteAdd`), so it's a single undoable action and it respects the scale.

Melodic only for now; kit rows are skipped with a `MELODIC ONLY` popup. The popup after a dump shows how many notes landed.

## Why

The most painful little loss on the Deluge is playing a great phrase while just messing about, not recording, and having no way to get it back. A DAW would have caught it. This catches it, on the device, without changing how recording or playback work.

This is a sibling to the autosave/recovery idea: never lose an idea you just had.

## How it's gated

A Community Features toggle, **Retrospective MIDI Capture** (7-seg: `RCAP`), on by default. When off, nothing is buffered and the dump trigger does nothing.

## Files

- `keyboard_screen.cpp` - the capture buffer + dump, and the note-on / note-off / SHIFT+RECORD / clear-on-context-change hooks
- `runtime_feature_settings.{cpp,h}` - the toggle
- `menu_item/runtime_feature/settings.cpp` - the menu entry
- `l10n/strings.h`, `g_english.cpp`, `g_seven_segment.cpp` - the feature name

## Testing

- Built Release for 7SEG.
- On hardware (7-segment): with the feature on, noodled along to a drum loop without recording, hit SHIFT+RECORD, and the phrase landed in the clip at the right length (including the case of playing two bars over a one-bar loop, which grew the clip to two bars). Confirmed the buffer clears on layout change / leaving keyboard view, and that kit clips show MELODIC ONLY.

## Notes

The buffer holds up to 128 notes and keeps the earliest take if it fills (rather than overwriting), so a long noodle won't silently lose its start. No effect on live audio: capture is just bookkeeping on note on/off, and the dump only happens on an explicit SHIFT+RECORD.


seangoodvibes:

Thanks

owlet-labs:
to wit, if there's appetite I can re-open the retro sampler pr now based on main. Just checking on key combos--seems you're using shift record. IIRC there was some collision on that? FWIW, I am trying to use record+sync to trigger the retro sampler. cc @seangoodvibes

CommmandrCody:

CommmandrCody:
@owlet-labs sorry for the slow reply on this.

You were right about the collision, it just wasn't the one you thought. RECORD+SYNC is clear of me. The problem was my end. SHIFT+RECORD is already output recording in stock firmware and I was stepping on it. A UI handler that returns DEALT_WITH means `buttons.cpp` never sees the press at all, so with my feature on in a keyboard layout you'd just quietly lose resampling. Good catch.

Fixed. Capture only takes SHIFT+RECORD now if there's actually something in the buffer. Nothing buffered and it falls through, resampling works like it always did. So I only borrow the combo right after you played something, which is the only time you'd have meant it anyway.

Separate thing, might save you some time since I was already in there. RECORD+SYNC is free but SYNC on its own isn't. In `view.cpp` it's either sync-scaling or Fill Mode depending on the `syncScalingAction` setting, and the Fill path fires `changeFillMode(on)` on press AND release. So for anyone running that, holding SYNC as a modifier flips fill mode on them while your sampler goes. Might want to eat the SYNC press while RECORD is held. There's already a pattern for it, `recordButtonPressUsedUp`, that's what I use so the dump doesn't also arm recording.

Nothing blocking you from my side. Would be good to see the retro sampler back... yours grabs the audio, mine grabs the notes.

PR #4601, opened 2026-06-28
Fix/ sample osc transpose + horizontal menu note range editing

### LINKED ISSUE #4599: Transposition parameter of OSC's does not work in multisampled synths.

When using multisampled synths the transposition of oscillators is not possible (SHIFT + TRANSPOSE), neither coarse nor fine. While the horizontal menus don't show the "pitch ranges" screen when attempting to transpose vs. showing in regular menus, the inability to transpose is not dependent on it nor whether horizontal menus are on or off. 

The Transpose setting (SHIFT + TRANSPOSE for given oscillator) changes pitch of musltisamped synths by coarse and fine amount, it works as with any other synth type. 

Long ago while using FW 1.3,  to the point I'm not even sure this setting is supposed to work here. 

_No response_

OLED

Beta, Nightly

deluge-v1_3_0-nightly+2026_06_28-b79a2e3

Load any multisampled synth. Attempt to transpose the sound by SHIFT + TRANSPOSE for the given oscillator. When community feature "horizontal menus" is OFF a pitch ranges screen will show. Select desired range. "OSC 1 Transpose" screen will follow. Attempting to change the pitch either by coarse or fine amount does not work. In case of community feature "horizontal menus" being ON, the pitch ranges screen will be skipped and transpose setting available in the horizontal menu. Attempting to change the pitch either by coarse or fine amount does not work.

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4599

Fixed several issues:

1) refactoring of osc menu item code for horizontal menus broke oscillator multiple sample transpose (you could no longer edit the transpose for specific multi sample note ranges. It would always edit multi range index 0.

2) note range menu was inaccessible from horizontal menus. This meant that you could not select a specific range and edit the transpose value for that range.

With the above fixes, you can now:
- enter note range menu with horizontal menu enabled using shift/audition + osc transpose
- when you press select on a specific note range, it will open the horizotnal menu with the transpose menu item updated to show the transpose value of that note range. changing the transpose value now affects the transpose of the note range selected.
- when in the osc horizontal menu, if you press the button to select the osc transpose menu item, it will open the note range menu again.
- when you press select on transpose in horizontal menu it enters the "select patch cable source" menu. If you press back to go back to horizontal menu it will keep the previously selected note range and display that note range's transpose value (until you select another horizontal menu item)

### DIFF
--- a/src/deluge/gui/menu_item/horizontal_menu.cpp
+++ b/src/deluge/gui/menu_item/horizontal_menu.cpp
@@ -38,7 +38,19 @@ namespace deluge::gui::menu_item {
 using namespace hid::display;
 
 void HorizontalMenu::beginSession(MenuItem* navigatedBackwardFrom) {
+	::MultiRange* selected_range = soundEditor.currentMultiRange;
+	int16_t selected_range_index = soundEditor.currentMultiRangeIndex;
+	const bool current_child_is_range_dependent = current_item_ != items.end() && (*current_item_)->isRangeDependent();
+	const bool should_preserve_range =
+	    current_child_is_range_dependent
+	    || (navigatedBackwardFrom != nullptr && navigatedBackwardFrom->isRangeDependent());
+
 	Submenu::beginSession(navigatedBackwardFrom);
+
+	if (should_preserve_range) {
+		soundEditor.currentMultiRange = selected_range;
+		soundEditor.currentMultiRangeIndex = selected_range_index;
+	}
 
 	for (const auto it : items) {
 		it->parent = this;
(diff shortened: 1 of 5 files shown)


src/deluge/gui/menu_item/horizontal_menu.cpp
```diff
@@ -409,7 +421,14 @@ void HorizontalMenu::handleInstrumentButtonPress(std::span<MenuItem*> visible_pa
 			}
 
 			// Update the currently selected item
-			initializeItem(*current_item_);
+			const MenuPermission permission = initializeItem(*current_item_);
```
lburygin, line 424: maybe now we can get rid of "initializeItem" wrapper and just do straight call to "checkPermissionToBeginSession"? since now the intention changed a bit

PR #4604, opened 2026-06-29
[DRAFT] [Feature] Add option to only display defined CCs from device definition file in MIDI clip automation view

kazemio:
When loading a device definition file in a MIDI clip, I found it slightly annoying to scroll through (potentially many) undefined CC automation lanes to get at the parameters that were defined in the file.

Added an option to only display defined CCs from the device definition file in the MIDI clip automation view. This is a setting that can be toggled on/off (default off to preserve existing behavior).

This setting is persisted in the device definition file (saved using SAVE + GOLD KNOB).

Demo:
https://github.com/user-attachments/assets/d44b075b-2187-4100-9b64-14c5adfff559


seangoodvibes:
Thanks for proposing this addition. As the MIDI Device Definition feature is my feature I have gone ahead and implemented this the way I would have envisioned it, so there is no need for this PR anymore. See: https://github.com/SynthstromAudible/DelugeFirmware/pull/4787

PR #4627, opened 2026-07-05
Fix/ Resume notes after transposing

### LINKED ISSUE #2939: Master Transpose causes drone notes to cut off until note can be retriggered

If you have a drone note and use master transpose or midi transpose to transpose that drone note, the drone note will cut off until the sequence loops back around for the drone note to retrigger

Drone notes immediately retrigger after being transposed

_No response_

OLED, 7SEG

Beta

_No response_

_No response_

seangoodvibes:
Fix transpose cutting off notes by resuming them after transposing

Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/2939

To do: 
- [ ] adjust approach - instead of stopping and restarting notes, don't stop and set a flag to tell the voice to re-calculate pitch on next render

## Before Video

https://github.com/user-attachments/assets/1388de05-8af0-418c-8218-d1c681b8539c

## After Video

https://github.com/user-attachments/assets/233e441b-38a0-4efe-aa20-9e536a802875

### DIFF
--- a/src/deluge/model/clip/instrument_clip.cpp
+++ b/src/deluge/model/clip/instrument_clip.cpp
@@ -1302,10 +1302,47 @@ void InstrumentClip::seeWhatNotesWithinOctaveArePresent(NoteSet& notesWithinOcta
 	}
 }
 
+bool InstrumentClip::stopAllNotesPlayingForTranspose(ModelStackWithTimelineCounter* modelStack) {
+	bool shouldResumePlayingNotes = playbackHandler.isEitherClockActive() && modelStack->song->isClipActive(this);
+
+	// Stop every old pitch first. For active clips, keep sequenced as a temporary marker so only rows that were
+	// actually sounding before the transpose are restarted at their new pitch.
+	for (int32_t i = 0; i < noteRows.getNumElements(); i++) {
+		NoteRow* thisNoteRow = noteRows.getElement(i);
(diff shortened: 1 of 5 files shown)


src/deluge/model/clip/instrument_clip.cpp
```diff
@@ -1302,10 +1302,47 @@ void InstrumentClip::seeWhatNotesWithinOctaveArePresent(NoteSet& notesWithinOcta
...
+	bool shouldResumePlayingNotes = playbackHandler.isEitherClockActive() && modelStack->song->isClipActive(this);
+
+	// Stop every old pitch first. For active clips, keep sequenced as a temporary marker so only rows that were
+	// actually sounding before the transpose are restarted at their new pitch.
+	for (int32_t i = 0; i < noteRows.getNumElements(); i++) {
+		NoteRow* thisNoteRow = noteRows.getElement(i);
+		bool wasSequenced = thisNoteRow->sequenced;
+		ModelStackWithNoteRow* modelStackWithNoteRow =
+		    modelStack->addNoteRow(getNoteRowId(thisNoteRow, i), thisNoteRow);
+		thisNoteRow->stopCurrentlyPlayingNote(modelStackWithNoteRow);
```
m-m-adams, line 1315: I don't think we should actually stop and restart the notes - that's pretty heavy and will probably lead to culling. Can the notes be moved to their new row while playing?

PR #4633, opened 2026-07-07
Fix USB MIDI receive packet loss + chainload corruption

### LINKED ISSUE #4762: smSysex: inbound SysEx larger than 48 bytes is silently dropped

Any inbound SysEx message longer than **48 data bytes** is silently dropped a large fraction of the time. 48 bytes or fewer is 100% reliable. The cliff is exactly one byte wide.

48 SysEx bytes = 16 USB-MIDI packets = 64 bytes on the wire = the bulk endpoint's `wMaxPacketSize`. Anything needing a second USB transfer is exposed.

Measured over USB MIDI Port 3, padding a `dir` request's `path` value so only the *inbound* size varies (the reply stays small):

| request bytes | USB-MIDI packets | USB wire bytes | replies |
|---|---|---|---|
| 28–47 | ≤16 | ≤64 | 100% (20 samples each size) |
| **48** | 16 | **64** | **40/40** |
| **49** | 17 | **68** | **17/40** |
| 50 | 17 | 68 | 24/40 |
| 97 | 33 | 132 | 0/15 |

It is purely a function of inbound size, not of the command. The same `ping` command padded across the threshold behaves the same way:

| request | replies |
|---|---|
| `{"ping":{}}` — 19 B | 20/20 |
| `{"ping":{}}` + padding — 47 B | 20/20 |
| `{"ping":{}}` + padding — 53 B | 5/20 |

A dropped request is never answered — no error, no late reply, no log. Replies that do arrive are always complete and correct, in 21–29ms, even at 2KB, so the outbound direction is fine. Reproduced identically on two different USB cables.

This makes the smSysex file API unreliable in practice, since command size depends on arguments:

- `{"read":{"fid":1,"addr":0,"size":1024}}` is 47 bytes and works, but `"addr":4096` makes it 50 and it starts failing — so reads get flaky deeper into a file.
- `open` with a real path like `/SYNTHS/FACT/SYNT000.XML` is ~60 bytes and needs several retries.
- `write` carries a 128-byte binary payload, ~1170 bytes total, and fails most of the time.

Inbound SysEx of any length up to the documented maximum should be received reliably, and a request the firmware cannot parse should get an error reply rather than being silently discarded.

Noticed while writing a Web MIDI client against smSysex. Not a regression I can date — the code paths involved look long-standing.

No.

7SEG

Beta

1.3 beta. Source analysis below against `beta` @ `ea4e69eb`.

Connect the Deluge over USB and open Chrome to any `https://` or `localhost` page. Paste into the console and click Allow when prompted:

```js
const acc = await navigator.requestMIDIAccess({ sysex: true });
const out = [...acc.outputs.values()].find(o => o.name === 'Deluge Port 3');
const inp = [...acc.inputs.values()].find(o => o.name === 'Deluge Port 3');
let waiter = null;
inp.onmidimessage = e => { if (waiter) waiter(new Uint8Array(e.data)); };
let mid = 9;

// {"dir":{"path":"/AAA..."}} padded to hit a target total SysEx length
function dirMsg(pad) {
  const j = new TextEncoder().encode(JSON.stringify({ dir: { path: '/' + 'A'.repeat(pad) } }));
  mid = mid >= 0x7f ? 9 : mid + 1;
  return new Uint8Array([0xF0, 0x00, 0x21, 0x7B, 0x01, 0x04, mid, ...j, 0xF7]);
}

for (const pad of [20, 21, 22]) {          // -> 48, 49, 50 byte requests
  let ok = 0;
  for (let i = 0; i < 40; i++) {
    const m = dirMsg(pad);
    if (await new Promise(r => { waiter = d => r(d); setTimeout(() => r(null), 500); out.send(m); })) ok++;
    waiter = null;
  }
  console.log(`${dirMsg(pad).length} bytes: ${ok}/40`);
}
```

Observed:

```
48 bytes: 40/40
49 bytes: 17/40
50 bytes: 24/40
```

The path is deliberately bogus so the reply is a small `err` response — this keeps the outbound size constant and isolates the inbound direction. Any `ping`/`dir` pair padded across 48 bytes shows the same thing.

---

### Analysis

Two separate defects combine here.

**1. The receive pipe is armed for exactly one max-packet, and the re-arm is blocked by SD access.**

`receiveData` is 64 bytes ([`midi_device_manager.h:81`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/io/midi/midi_device_manager.h#L81)) and every arm sets `tranlen = 64` ([`midi_engine.cpp:926`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/io/midi/midi_engine.cpp#L926)), matching `wMaxPacketSize` in the endpoint descriptor. So one armed transfer absorbs at most 16 USB-MIDI packets = 48 SysEx bytes. A longer message *must* have the pipe re-armed mid-message.

That re-arm only happens inside `checkIncomingUsbMidi`, which is on the `midi routine` task registered with `RESOURCE_SD | RESOURCE_USB` ([`deluge.cpp:547-548`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/deluge.cpp#L547-L548)), and which additionally early-returns on SD access ([`midi_engine.cpp:855-861`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/io/midi/midi_engine.cpp#L855-L861)):

```cpp
if (!usbCurrentlyInitialized
    || currentlyAccessingCard != 0) { // hack to avoid SysEx handlers clashing with other sd-card activity.
```

Meanwhile the FIFO-read task `check usb midi` holds only `RESOURCE_USB` ([`deluge.cpp:540`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/deluge.cpp#L540)) and keeps draining `g_usb_data_cnt` toward zero. When the re-arm loses that race the driver hits:

```cpp
if (g_usb_data_cnt[pipe] < numBytesReceived)
    return USB_READOVER;
```

[`r_usb_plibusbip.c:612-613`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/RZA1/usb/r_usb_basic/src/driver/r_usb_plibusbip.c#L612-L613) → `usb_pstd_forced_termination(pipe, USB_DATA_ERR)` → `usb_cstd_do_aclrm`, which clears the hardware FIFO. The tail of the message — including the packet carrying `F7` — is discarded with no error, no counter, no callback.

`ping` at 8 bytes is 3 packets = 12 wire bytes, always inside one armed window, so it is structurally immune. That is why it never drops.

**2. A request that fails to parse is popped without any reply.**

In `handleNextSysEx` ([`smsysex.cpp:791-848`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/storage/smsysex.cpp#L791-L848)), `done:` is also the natural fall-through of the tag-matching loop:

```cpp
parser.match('{');                                            // return value discarded
while (*(tagName = parser.readNextTagOrAttributeName())) {
    if (!strcmp(tagName, "open")) { openFile(...); goto done; }
    ...
    parser.exitTag();
}
done:
    SysExQ.pop_front();
```

If the payload is truncated or corrupted, `readNextTagOrAttributeName()` returns `""`, the loop exits normally, and the entry is popped with **no reply generated**. `msgSeqNum` and `de.cable` are both in hand at that point, so an error reply is possible without a successful parse.

Defect 1 loses the bytes; defect 2 converts what should be a recoverable parse error into a silent unanswerable hang. Fixing 2 alone would at least let clients detect and retry instead of waiting on a reply that will never come.

Some other silent-failure paths noticed while tracing, not required for this repro:

- `copyFile`/`moveFile` return without replying when `parseFileOpParams` fails; `deleteFile`/`createDirectory`/`rename` skip the reply on an empty path.
- Inbound SysEx over 1024 bytes cannot be received at all — `incomingSysexBuffer` is `uint8_t[1024]` ([`midi_device.h:160`](https://github.com/SynthstromAudible/DelugeFirmware/blob/beta/src/deluge/io/midi/midi_device.h#L160)) and overflow resets `incomingSysexPos` to 0 mid-message, so the `[0] == 0xf0` guard then fails and the message is dropped. smSysex sizes its own buffer at 1280 (`blockBufferMax + 256`), so it expects messages the reassembly buffer cannot hold.
- `checkIncomingUsbSysex` treats CIN `0x5` unconditionally as SysEx-end, but the USB-MIDI spec also uses it for single-byte System Common.

### Possible fixes

1. Decouple the re-arm from `RESOURCE_SD` — re-arming the receive pipe touches no SD state. Alternatively move the drain and re-arm into the `check usb midi` task so reading and re-arming stay adjacent.
2. Make the receive buffer larger than one max-packet (the pipe itself is configured `USB_BUF_SIZE(512u)`), with a matching `tranlen`.
3. Send an explicit error reply on the `handleNextSysEx` fall-through instead of dropping, and stop discarding `parser.match('{')`'s result.
4. Don't silently `aclrm` on `USB_READOVER` — NAK the host or at least bump a diagnostic counter.

owlet-labs:
# Fix USB MIDI receive packet loss + chainload corruption (makes SysEx firmware loading reliable)

## Summary

Two independent bugs conspired to make USB SysEx firmware loading (`ENABLE_SYSEX_LOAD`) unreliable, and the first one also silently degrades any dense inbound USB MIDI stream in peripheral mode. Both are fixed here; together they make unpaced firmware upload from a modern Linux host work reliably at full speed.

## Commit 1: `fix(usb): stop dropping USB MIDI packets received as peripheral`

As a USB peripheral, receive transfers are only re-armed from `checkIncomingUsbMidi()`, polled at ~1ms by the task scheduler. The BRDY completion handler leaves the pipe at PID=BUF (the stock `usb_pstd_data_end()` NAK/disable logic is `#if 0`'d out as "not needed"), so between a transfer completing and the next poll, the hardware keeps ACKing incoming bulk packets. Such a packet reaches the handler with no transfer armed, takes the `USB_READOVER` branch, and is destroyed by `usb_pstd_forced_termination()` (its ACLRM clears the FIFO) — after the host already saw it ACKed on the wire. Silent loss, no retry, because bulk flow control was bypassed.

Any host sending bulk packets spaced closer than ~1ms loses most packets after the first. This is why SysEx firmware loading breaks on hosts that don't pace their MIDI writes (reported on Linux 6.16+), and it is likely the underlying cause of the slowness that #4140 worked around: only one 64-byte packet survived per poll.

The fix makes the BRDY handler: (1) NAK the pipe before touching the FIFO, restoring bulk flow control — the host now retries instead of losing data; (2) wait out PBUSY like the stock `usb_cstd_set_nak()` did, since a packet mid-transaction still completes into the FIFO; (3) drain every resident packet (the pipe is double-buffered) into `receiveData`, which grows 64 → 192 bytes (+768 bytes BSS); `checkIncomingUsbMidi()` already parses any whole number of 4-byte events; (4) never call `usb_pstd_forced_termination()` for a benign empty/mismatched FIFO — besides discarding data, its completion callback fires with `USB_DATA_ERR`, which `usbReceiveCompletePeripheralOrA1()` ignores as a detach, and that could leave the armed transfer permanently stuck.

Host-mode receive has a similar structure and may deserve the same treatment in a follow-up; this PR deliberately touches only the peripheral path.

## Commit 2: `fix(chainload): flush data cache before running the MMU-off chainloader`

With packet loss fixed, uploads still intermittently failed to boot: upload completes, image CRC passes, `chainload_from_buf()` jumps — and the unit hangs (blank OLED, progress LEDs frozen). Reliably reproducible after rendering audio; reliably absent from a fresh boot.

`deluge_chainload` runs with the MMU and caches disabled and stores straight to physical RAM, but its per-word DCCMVAU maintenance writes back any line still dirty in the disabled-but-not-emptied data cache, clobbering the word it just stored. All RAM is mapped write-back (`TTB_PARA_NORMAL_CACHE`), and the stage-2 chainloader is copied into `spareRenderingBuffer` — an audio rendering buffer, full of dirty lines whenever any voice has rendered since boot. Stale audio data written back over the copied stub means jumping into garbage. The same mechanism can corrupt the copied image itself via dirty destination lines. This is very likely the stub's longstanding "works (mostly) reliably" caveat.

Fix: clean & invalidate (`v7_dma_flush_range`) everything the chainloader touches — the received image, the copy destination, and the stub area — after interrupts are disabled, while the MMU is still on.

## Testing

- Hardware: Deluge (OLED), host: Linux kernel 7.0 (issue first observed on 6.16+).
- Before: `dbt loadfw` failed with truncated segments unless writes were paced to ~48-byte chunks with multi-ms delays; after commit 1, all 3,895 segments of a full image upload succeed unpaced.
- Before: chainload after audio had been rendered hung the unit; after commit 2, upload + boot succeeds repeatedly mid-session.
- Normal peripheral MIDI (notes/clock from host) regression-tested alongside firmware-upload traffic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


seangoodvibes:
hey @owlet-labs thanks for opening the PR, would it be possible to split the Pr’s commits into two separate PRs (one for USB, one for chain loading).

owlet-labs:
Per the feedback: split the chainload fix into its own PR — #4634 (12 lines, `chainload.cpp` only, fully independent). This PR is now the USB receive fix alone; the branch has been trimmed to just that commit. The two fixes compose but neither depends on the other: this one makes dense inbound USB MIDI (and unpaced SysEx upload) reliable, #4634 makes the subsequent boot reliable. Happy for them to be considered separately.

owlet-labs:
Rebased onto current `main` (was 7 commits behind); the diff is byte-for-byte unchanged by the rebase. Verified on our fork's CI at the rebased head `ca07d6f79`: [tests run — all green](https://github.com/owlet-labs/DelugeFirmware/actions/runs/28997518667) (x86 + QEMU). The workflow run on this repo is pending approval.

m-m-adams:
@owlet-labs I'm currently reviewing this. I'm having a hard time figuring out what some of the changes do but the gist seems to be a bunch of things rohan commented out as "not needed" turn out to actually be needed?

m-m-adams (approved):
I'm pretty sure this is good. I've tested host and peripheral with a few different scenarios and don't see a regression

There's a couple spots where I'm not sure what's going on and just left the original llm comment, otherwise I tried to add human readable comments where I could

stellar-aria (approved):

PR #4634, opened 2026-07-07
Fix chainload corruption: flush data cache before the MMU-off chainloader

owlet-labs:
# Fix chainload corruption: flush data cache before running the MMU-off chainloader

Split out of #4633 per maintainer feedback — this is the chainload fix alone (12 lines, `chainload.cpp` only). It is fully independent of the USB receive changes and fixes a real bug for any image source, including SD-adjacent workflows and paced/slow SysEx uploads.

## The bug

Chainloading a firmware image over USB SysEx hung the unit (blank OLED, progress LEDs frozen) whenever audio had been rendered since boot — while the same image booted fine from a fresh boot. Reliably reproducible, reliably absent on fresh boots.

## Root cause

`deluge_chainload` runs with the MMU and caches disabled and stores straight to physical RAM, but its per-word DCCMVAU maintenance writes back any line still **dirty** in the disabled-but-not-emptied data cache, clobbering the word it just stored (plus the rest of the line). All RAM is mapped write-back (`TTB_PARA_NORMAL_CACHE`), and the stage-2 chainloader is copied into `spareRenderingBuffer` — an audio rendering buffer, full of dirty lines whenever any voice has rendered since boot. Stale audio data written back over the copied stub means jumping into garbage. The same mechanism can corrupt the copied image itself via dirty lines overlapping the copy destination (mostly masked because the pre-chainload CRC pass happens to stream-evict the source lines).

This is very likely the stub's longstanding "works (mostly) reliably" caveat.

## The fix

Clean & invalidate (`v7_dma_flush_range`, already present in `drivers/cache/invalidate.S`) everything the chainloader touches — the received image, the copy destination range, and the stage-2 stub area — after interrupts are disabled, while the MMU is still on.

## Testing

- Hardware: Deluge (OLED), host: Linux kernel 7.0.
- Before: chainload after audio had been rendered hung the unit every time; after: upload + boot succeeds repeatedly mid-session.
- Fresh-boot chainloads (the previously-working case) regression-tested unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/deluge/util/chainload.S
+++ b/src/deluge/util/chainload.S
@@ -33,15 +33,15 @@
 // is implicitly position-independent because all jumps (except `bx/blx` to
 // registers) in armv7a are relative.
 //
-// All of this work needs to be done with the MMU and caches disabled because
-// the RZ/A1L cache is extremely broken, and doesn't respect most of the cache
-// maintenance operations. This includes CP15 MCRs, the usual DMB/IMB
-// instructions, and the ARM-recommended L1$ flush logic.
-//
-// It's possible all of this actually works and it's the L2$ misbehaving -- We
-// haven't tried flushing that cache instead yet. But disabling the MMU is
-// easy, doesn't cost meaningful performance here, and seems to work (mostly)
-// reliably.
+// All of this work is done with the MMU and caches disabled so the copies go
+// straight to physical RAM. (This comment used to claim the RZ/A1 cache
+// "doesn't respect" maintenance operations; it does. The historical
+// unreliability had a different cause: disabling the D-cache doesn't empty
+// it, so lines left dirty by the old firmware were written back over freshly
+// copied words by the DCCMVAU housekeeping below. chainload_from_buf() now
+// cleans & invalidates every range this code touches - while the MMU is still
+// on, before jumping here - which is what the "works (mostly) reliably"
+// caveat that used to live in this comment was missing.)
   .align 4
   .text
   .arm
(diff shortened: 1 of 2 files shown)


stellar-aria (approved):

stellar-aria:
@owlet-labs the tests need fixing before merge

owlet-labs:
Fixed — the x86 test build links `chainload.cpp`, and the flush calls referenced ARM-only symbols (`v7_dma_flush_range` is the ARM assembly cache routine). The cache-maintenance block is now under the same `#if defined(__arm__)` guard the chainloader jump already uses, so the host test build is back to the exact pre-PR shape. Verified locally that the host-compiled object has no undefined references.

owlet-labs:
Verified on our fork's CI (same workflow, same commit `0170004bd`): [tests run — all green](https://github.com/owlet-labs/DelugeFirmware/actions/runs/28992303594), including the previously-failing x86 job. The upstream run here is pending workflow approval.

owlet-labs:
Pushed one more amendment (`b374deb30`): corrected the stale header comment in `chainload.S` that claimed the RZ/A1 cache "doesn't respect" maintenance operations — the root cause turned out to be dirty lines surviving cache-disable, now documented where the next reader will look. Comment-only change: the assembled object is byte-identical (verified with the project toolchain), so prior hardware testing still applies. Fork CI re-run on the amended head: [all green](https://github.com/owlet-labs/DelugeFirmware/actions/runs/28993385724) (x86 + QEMU).

stellar-aria (approved):

PR #4659, opened 2026-07-11
Fix zoom animation for out-of-bounds clip area on empty rows

### LINKED ISSUE #3363: Zooming out beyond the clip's length will not animate the clip rows properly.

Now that we can zoom out beyond the clip length, by doing so the area that isn't part of the clip length doesn't properly animate during zoom in/out transitions UNLESS the row has note data in it.

Maintain zoom-in/out animated transitions for the out-of-bounds clip areas.

a couple weeks at least

#1962

OLED

Nightly

DD80936

https://github.com/user-attachments/assets/3a1e0d34-ee00-42f3-85e6-843b816c7cfa

stellar-aria:
Since zoom can now go beyond a clip's length (#1962), every row renders a grey "undefined area" whose boundary moves during a zoom transition. But InstrumentClipView::tellMatrixDriverWhichRowsContainSomethingZoomable only flagged rows with notes for animation, so on note-less rows that grey boundary snapped instead of sliding.

Animating a row that renders identically before and after is invisible, so skipping rows is only ever a micro-optimization. Make "animate every row" the TimelineView base default and drop the now-identical InstrumentClipView and AudioClipView overrides. ArrangerView keeps its override: its empty output rows are solid black with no moving grey boundary, so skipping them is still valid.

Fixes #3363

### DIFF
--- a/src/deluge/gui/views/audio_clip_view.cpp
+++ b/src/deluge/gui/views/audio_clip_view.cpp
@@ -854,10 +854,6 @@ bool AudioClipView::setupScroll(uint32_t oldScroll) {
 	return ClipView::setupScroll(oldScroll);
 }
 
-void AudioClipView::tellMatrixDriverWhichRowsContainSomethingZoomable() {
-	memset(PadLEDs::transitionTakingPlaceOnRow, 1, sizeof(PadLEDs::transitionTakingPlaceOnRow));
-}
-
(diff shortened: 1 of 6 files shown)


seangoodvibes (approved):

seangoodvibes:
Tested confirmed fixed

PR #4673, opened 2026-07-13
Feature: round robin

bomtelecom:
## Summary

Adds MPC-style multi-sample playback to sample-based oscillators: each sample zone can hold
up to 4 variants — the primary sample plus 3 alternates — that are played in rotation
(Cycle / Random / No Repeat) each time the zone is triggered. This is the classic
round-robin workflow for reducing the "machine gun" effect on repeated drum hits, and the
first step toward general multi-sample-per-pad support.

## What's new

- **New `VARIANTS` submenu** under `OSC1`/`OSC2` (sample mode only): a MODE selector
  (Cycle / Random / No Repeat) plus four slot entries. Each slot opens its own horizontal,
  icon-based menu — the same paging style `OSC1`/`OSC2`s own menu already uses one level up —
  with FILE, STRT, END, and TRANSPOSE (a single combined semitones+cents value, matching
  how OSC-level pitch is already one entry rather than two separate ones). Slot 1 is the
  primary sample; alternates fill in order, and SHIFT + SAVE/DELETE clears one. Once a
  second slot is loaded, `OSC1`/`OSC2`s own FILE/START POINT/END POINT/TRANSPOSE items step
  aside from the horizontal page automatically, since they'd otherwise be a second path to
  editing slot 1's exact same data. 
- **VARIANTS is scoped to single-zone (non-key-split) instruments — on a multi-zone instrument
  it doesn't appear at all, since every other zone-specific item is a flat, single-hop param
  while VARIANTS is a real nested submenu, and composing that with multi-zone navigation
  isn't supported yet (see Future work).
- **Menu-driven audition**: while a slot's submenu (or its marker editor) is open,
  auditioning the drum/note plays that exact slot without advancing the round-robin cycle —
  so you can always hear what you're editing.
- **Kit pad feedback**: auditioning a pad with variants loaded pops up the file name and
  slot position of the variant that just played, e.g. `kick_2.wav (3/4)`.
- **Community feature toggle**: `SETTINGS > COMMUNITY FEATURES > ROUND-ROBIN SAMPLE
  VARIANTS` (on by default). When off, the menu is hidden and playback always uses the
  primary sample; existing variant data in songs is preserved either way.
- Works identically on OLED and 7SEG.

## Save format

Alternates are stored per-zone as a `roundRobinAlternates` array (file name, transpose,
cents, start/end/loop points) plus an `rrMode` tag when not Cycle, nested inside the
existing sample-range scope. Songs that don't use variants are byte-for-byte unaffected;
songs that do still load correctly on firmware without this feature, since unknown tags are
skipped and the primary sample loads as normal.

## Implementation notes

- Alternates are lazily allocated: a zone that never uses variants costs 8 bytes over the
  previous `MultisampleRange` layout (a pointer plus 4 small fields); the alternate-holder
  table itself is only allocated once a second sample is actually loaded into a zone.
  Layout is pinned with `static_assert`s.
- Slot resolution (`Voice::noteOn` → `MultisampleRange::resolveVariant`) is allocation- and
  display-free, safe on the audio path. A `noteMightBeConstant` flag threaded from
  `NoteRow` through `Kit`/`Drum` ensures drone-note continuations reuse the last-played slot
  rather than advancing the cycle.
- The XML parser is shared between the single-range and multi-range save-file paths via a
  templated helper (`round_robin_serialization.h`) and is unit-tested on the host with a
  mock deserializer, independent of hardware.
- Menu definitions are generated from the existing Python DSL
  (`dmenus/oscillator/round_robin.py`) rather than hand-edited into `g_menus.inc`.

## Bug fixes found and fixed along the way

- `rrMode` was being reset to Cycle on every song load regardless of tag order in the file
  (now covered by a regression test).
- A use-after-free in the sample browser when converting a WAVETABLE source to SAMPLE while
  targeting an alternate slot (the target holder was fetched before the range array was
  reallocated).
- Loading a wavetable-marked file into an alternate slot no longer converts the whole source
  to WAVETABLE mode, which would have destroyed the other variants.
- A failed load into a newly-created alternate slot no longer leaves a silent empty slot
  (which would mute every Nth trigger).
- `MultisampleRange::loadAllAudioFiles` now reports the first error from any slot instead of
  only the primary's.

## Size impact

Release binary: 1,698,524 bytes vs. current main's 1,685,904 bytes (+12,620 bytes / +0.75%).

## Future work

This PR deliberately keeps scope to the Cycle / Random / No Repeat round-robin core. A few
natural extensions were considered and left for follow-up PRs:

- **Velocity-based layer selection** — an MPC-style mode where each slot owns a velocity
  range instead of (or alongside) rotation. `Voice::noteOn` already receives velocity, so
  this would compose as another `RRMode` option on the existing resolver rather than a
  parallel system. Deluge's `MultiRange` zones are key-ranges, so velocity-based selection
  is an orthogonal axis, not a conflict — this is the biggest remaining gap vs. typical
  hardware-sampler layer modes.
- **Variants on multi-zone (key-split) instruments — today VARIANTS only appears on a
  single-zone oscillator. Supporting a key-split zone's own alternates would need either
  teaching the shared note-range-picker navigation about nested submenus, or a separate
  entry point that bypasses it entirely — real work for a narrow case. If per-sample
  variation beyond a single zone turns out to be wanted, it likely makes more sense as an
  extension of the velocity-based layer selection described above, which is already an
  orthogonal axis to key-split zones and wouldn't need to go through the note-range picker
  at all, rather than as dedicated support for key-split zones specifically.
- **Raising the 4-slot cap** — the lazily-allocated `RoundRobinAlternates` pointer table
  means more slots stay cheap for zones that actually use them (e.g. 8 slots instead of 4
  costs 16 extra bytes, only where populated). The cap today is really a menu-design
  constraint, not a data-model one: the generator DSL builds a fixed number of slot
  submenus, so a genuinely dynamic count would need per-instrument menu generation or a
  scrollable slot list rather than four fixed rows.
- **Per-variant level and pan** — would live alongside the alternate holders in
  `RoundRobinAlternates` (not on `SampleHolderForVoice`, which is embedded by value in
  every zone and would grow all of them, used or not) as small offsets relative to the
  primary's existing level/pan. Slot 1 would need no extra storage, and zones without
  alternates would pay nothing, consistent with the rest of this feature's memory design.
- **Multiple choke groups per kit** — today `PolyphonyMode::CHOKE` is a single global
  bucket: `Kit::choke()` unconditionally silences every `CHOKE`-flagged drum in the kit
  whenever any one of them triggers, so hi-hats, snares, and anything else on Choke all
  share one group. Would be awesome if we could support several independent choke
  groups per kit (closed/open hi-hat in one, snare rolls in another, etc.). This would need
  a per-drum group id in place of today's boolean-like flag, and `Kit::choke()` restricted
  to drums sharing the triggering drum's group. Unrelated to round-robin's data model, but
  a natural adjacent kit-editing improvement.

## Test plan
- [x] Test everything in both OLED mode and 7SEG
- [x] Load 2-4 variants into a Synth zone and a Kit row; confirm Cycle/Random/No Repeat all
      behave as expected on both OLED and 7SEG
- [x] Load 2-4 variants into a Synth zone and a Kit row; confirm Cycle/Random/No Repeat all
      behave as expected on both OLED and 7SEG
- [x] Edit each slot's Start/End/Transpose/Cents independently
- [x] Audition from inside a slot's submenu/marker editor and confirm it plays that exact
      slot without disturbing the cycle
- [x] Clear an alternate slot mid-playback (SHIFT+SAVE/DELETE)
- [x] Clear all slots except the first to confirm menus reapear under OSC
- [x] Save the song, reload it, confirm variants and mode persist
- [x] Load a song saved before this feature — confirm it's unaffected
- [x] Toggle the community feature off/on and confirm menu visibility and playback fallback
- [x] `./dbt test` — all unit test suites pass, including the new round-robin resolver and
      XML parser tests


m-m-adams:
Have you already followed the test plan and confirmed it's ok?

bomtelecom:
@m-m-adams I tested again, there was bug that froze the device by my toggle for this feature while navigating the settings/community-features. Fixing it now and I will remake the commits. If I test on an OLED with 7seq turned on, is it enough, or should someone with 7seq test?

m-m-adams:
I don't feel like that answered the question. How did you conduct your tests and what was the result?

bomtelecom:
@m-m-adams I conducted the testing according to the described test plan in the summary, and switched to 7 seg mode to see that everything works as expected there, and it does. I'll want to do one more commit with some menu improvements for round robins and notify you when that's also done and retested

bomtelecom:
@m-m-adams I added a new commit on the PR with a couple of menu changes that improve the overall logic when multiple variants are assigned to a pad. 
Now we have horizontal menus under each SLOT inside VARIANTS, and some options which were previously duplicated on the OSCILATOR menu layer will now show only under slot 1, but only when multiple variants are assigned to a pad.
I redid the provided test scenario, everything looks good 👍

When loading a multi keyed sampled instruments from a folder (the original multisample code on which I've expanded in this PR), the VARIANTS section will hide so that the flow remains as it was. I don't think cycling samples made that much sense in that context, but maybe in the future, if we add a velocity mode beside cycling and random, it would make more sense to show that under note range.

PR #4695, opened 2026-07-18
fix: serialize filter mode OFF correctly in XML

### LINKED ISSUE #4684: Filter mode "off" not saving in XML files

When customizing a kit settings I like to start with the filters off at all levels. I will deliberately set song, kit, and kit sound filters to off. When I save the file and come back to it later I have to set the filters off again everytime. This only seems to happen when setting the mode to "off' where as other filter types survive a restart after saving. Upon inspecting the xml files It appears the modes are being saved as "flanger" 

Image
Image

Filter modes should persist in kits, synths, and song files however only when setting the filters to "off" they don't.

when firmware moved from nightly to beta status 

_No response_

OLED

Beta

1.3.0 - beta- EBB059B

1. power on
2. select kit track
3. select a kit
4. set a kit sound and kit fx filters to "off"
5. save
6. power off
7. power on
8. select same kit
9. see the standard filters in LP24 or HP12 mode.

douglas-carmichael:
Fixes #4684

## Problem

Setting an LPF/HPF filter mode to **off** doesn't survive a save/load cycle. The XML ends up with garbage in the mode attribute (users observed `lpfMode="flanger"`), and on reload the filters come back as LP24 / HP12.

## Root cause

`FilterMode::OFF` has underlying value 6, but the serialization map is `EnumStringMap<FilterMode, kNumFilterModes>` where `kNumFilterModes == 6` deliberately excludes OFF — so the map's string table has entries for values 0–5 only:

- **Writing** an OFF filter (`lpfTypeToString` in `ModControllableAudio::writeAttributesToFile`) indexes `stringList_[6]` on a 6-element array — an out-of-bounds read whose garbage pointer happens to print as `"flanger"` on current builds.
- **Reading** any unrecognized string falls back to the last mapped variant, `HPLADDER` — so OFF could never round-trip.

## Fix

- Add `{FilterMode::OFF, "off"}` to the map and size it `kNumFilterModes + 1`. OFF now round-trips as `"off"`.
- Because the unknown-string fallback returns the last variant, it now resolves to OFF instead of HPLADDER — which also **repairs files already saved with a garbage mode string** by affected builds: they load as OFF, matching what the user had actually set.
- Bounds-check `EnumStringMap::operator()(Enum)` so a future enum/map size mismatch can't read out of bounds again.

## Testing

- **Not yet tested on hardware** — repro steps from the issue double as the test: set kit sound + kit FX (and synth/song) filters to off, save, power cycle, reload → modes must still be off, and the saved XML must show `lpfMode="off"` / `hpfMode="off"`. Also worth loading a song saved by a current beta with an OFF filter (garbage mode string) → should now load as off.

### DIFF
--- a/src/deluge/model/mod_controllable/filters/filter_config.cpp
+++ b/src/deluge/model/mod_controllable/filters/filter_config.cpp
@@ -5,12 +5,16 @@
 // converts lpf/hpf mode to string for saving
 namespace deluge::dsp::filter {
 
-EnumStringMap<FilterMode, kNumFilterModes> filterMap({{{FilterMode::TRANSISTOR_12DB, "12dB"},
-                                                       {FilterMode::TRANSISTOR_24DB, "24dB"},
-                                                       {FilterMode::TRANSISTOR_24DB_DRIVE, "24dBDrive"},
-                                                       {FilterMode::SVF_BAND, "SVF_Band"},
-                                                       {FilterMode::SVF_NOTCH, "SVF_Notch"},
-                                                       {FilterMode::HPLADDER, "HPLadder"}}});
+// OFF must be in the map so it round-trips through XML (issue #4684). It also makes the
+// string->enum failure fallback (last variant) resolve to OFF, so files saved by builds
+// which wrote garbage for OFF load back as OFF rather than HPLadder.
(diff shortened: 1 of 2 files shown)


seangoodvibes:

PR #4696, opened 2026-07-18
fix: load MIDI CC learns for song sections from song files

### LINKED ISSUE #4680: MIDI CCs learned to launch song sections do not save with song

Song sections learned to MIDI CCs do not save with the song in the way that MIDI note-on messages do.  Both types of midi messages work the same when initially assigning them to song sections (triggering their correspondent section when sent), but, after saving a song and restarting the deluge, the note-on assignments persist and the CCs do not.

The expected behavior is that once MIDI CCs are learned to song sections and then saved, those same assignments will still be linked after the Deluge is shut down and then turned on again.

March 11 2024 is when I first noticed the issue (or at least the first time I can find myself posting about it on the discord).  

In a conversation on the deluge discord on 1/12/2026, Sean.Good.Vibes dug up this fix from 1.1 that seems to be where the functionality of learning CCs to song sections was added (with the lack of saving presumably not being noticed at that time):  https://github.com/SynthstromAudible/DelugeFirmware/pull/961

 https://github.com/SynthstromAudible/DelugeFirmware/pull/961

OLED

other

I reproduced it on 1.0.1, 1.1.1, a number of nightly builds at various times, and the current (07/15/2026) 1.3 beta. I also tested 4.1.4 and confirmed that the functionality is not present at all there.

learn a midi cc value to a song section and save the song. The midi signal should effectively start that song section. Restart the deluge. The same midi cc should now not trigger that song section.

douglas-carmichael:
Fixes #4680

## Problem

MIDI CCs learned to launch song sections work at runtime, but after saving and restarting only note-on learns survive — CC learns are gone.

## Root cause

The save side is fine: sections write the raw `launchMIDICommand.channelOrZone`, and `LearnedMIDI` encodes CC learns as `channel + IS_A_CC` (18–35) and MPE zones as 16–17, so the CC data has been present in song files all along.

The bug is in the hand-rolled section loader in `Song::readFromFile`:

```cpp
if (channel < 16 && note < 128) {
```

This guard predates CC learning (added in #961) and silently drops any `channelOrZone` ≥ 16 — i.e. every CC learn (and MPE-zone note learns too). The generic `LearnedMIDI::readFromFile` used by every other CC-learnable feature imposes no such restriction, which is why MIDI-knob/global-command CC learns persist and section learns don't.

## Fix

Accept the full learned range: `channel < IS_A_CC + NUM_CHANNELS` (0–35: plain channels, MPE zones, CC-encoded channels).

A nice side effect: songs saved before this fix already contain the CC assignment, so they start working on load without re-learning.

## Testing

- **Not yet tested on hardware** — test procedure: learn a CC to a section (it launches), save, power cycle, reload, send the same CC → the section must launch. Note-on learns should behave exactly as before.

### DIFF
--- a/src/deluge/model/song/song.cpp
+++ b/src/deluge/model/song/song.cpp
@@ -1952,7 +1952,9 @@ unknownTag:
 						}
 
 						if (id < kMaxNumSections) {
-							if (channel < 16 && note < 128) {
+							// channelOrZone can also be an MPE zone (16-17) or a CC learn
+							// (18-35, offset by IS_A_CC) - see LearnedMIDI (issue #4680)
+							if (channel < IS_A_CC + NUM_CHANNELS && note < 128) {
 								sections[id].launchMIDICommand.cable = cable;
 								sections[id].launchMIDICommand.channelOrZone = channel;
 								sections[id].launchMIDICommand.noteOrCC = note;


seangoodvibes:

PR #4698, opened 2026-07-18
fix: audio track specific-source selection oddities

### LINKED ISSUE #4677: Audio track specific track - strange behaviour

See videos attached below

Essentially couple behaviours that seem expected:

1) audio track can select itself as the input source
2) when you select a synth as the input source, it shows synth - synth name, but if you scroll select it says audio - synth name as well
3) if you set a audio track's source to be a synth and then delete the synth, the specific track will show audio - synth name (synth name gets stuck), until you switch input source and then go back to specific track
4) sometimes it just shows synth and not the synth name as the input source until you scroll select

Not sure if intentional but:

- audio track can't select itself as input source?
- that would probably inherently solve showing audio - synth name issue
- if synth is the input source, it should show the synth name always

today

_No response_

OLED

Nightly

_No response_

Couple videos showing the behaviour in song view and arranger view

https://github.com/user-attachments/assets/c5a72528-821a-432d-b1fd-f766c2ce862b

https://github.com/user-attachments/assets/02732880-f2e9-44ae-9200-6073824b639f

douglas-carmichael:
Fixes #4677

Four related defects around picking another track as an audio track's input source ("specific track"):

## 1. An audio track could select itself (pad shortcut)

`AudioInputSelector::padAction` filtered MIDI and CV pads but not the audio track's own pad. `setOutputRecordingFrom(self)` silently rejects the assignment, so the UI claimed a selection that never happened.
**Fix:** filter it out like MIDI/CV, with a "Can't record from itself!" popup.

## 2. The TRACK scroll selector offered every output

`SpecificSourceOutputSelector::selectEncoderAction` clamped over *all* outputs — including the track itself, MIDI and CV. Landing on a rejected candidate desynced the scroll index from the stored source, which is where displays like "AUDIO - \<synth name\>" came from.
**Fix:** scrolling now steps to the next output that can actually be recorded from (skips self/MIDI/CV; stays put at the ends).

## 3. OLED drew the type and the name from different places

The type line came from the scroll index, the name line from the stored source pointer — they diverge (see 2) — and the MIDI channel variable was left uninitialized for non-MIDI outputs, garbling the "Audio …" label nondeterministically. The name line also dereferenced the stored source without a null check (7SEG path too).
**Fix:** render both lines from the stored source; zero-init the channel; pass the real `AudioOutputMode` for audio sources (matching `View::displayOutputName`); show "None"/"NONE" when no source is set.

## 4. Deleting an edited synth left a stale reference

Deleting an *edited* instrument hibernates it (`moveInstrumentToHibernationList`) rather than truly deleting it, and the hibernation path never called `clearRecordingFromReferencesTo()` — only `Song::deleteOutput` did. Audio tracks kept displaying (and pointing at) a source that was no longer in the song until the user cycled the input.
**Fix:** clear recording-from references when an instrument moves to the hibernation list, same as on deletion.

## Testing

- **Not yet tested on hardware** — test procedure: pad-select the track itself → popup, selection unchanged; scroll the TRACK selector both directions → never lands on self/MIDI/CV, type prefix always matches the name; delete an edited synth used as a source → display clears to "None" immediately; song with no valid source candidates → "None", no crash; and normal specific-track recording/monitoring still works.

### DIFF
--- a/src/deluge/gui/context_menu/audio_input_selector.cpp
+++ b/src/deluge/gui/context_menu/audio_input_selector.cpp
@@ -171,7 +171,7 @@ void AudioInputSelector::selectEncoderAction(int8_t offset) {
 ActionResult AudioInputSelector::padAction(int32_t x, int32_t y, int32_t on) {
 	if (on && getUIUpOneLevel() == &sessionView) {
 		auto track = (&sessionView)->getOutputFromPad(x, y);
-		if (track && track->type != OutputType::MIDI_OUT && track->type != OutputType::CV) {
+		if (track && track != audioOutput && track->type != OutputType::MIDI_OUT && track->type != OutputType::CV) {
 			audioOutput->inputChannel = AudioInputChannel::SPECIFIC_OUTPUT;
 			audioOutput->setOutputRecordingFrom(track);
 			display->popupTextTemporary(track->name.get());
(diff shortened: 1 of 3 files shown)


seangoodvibes:

Issue #4721, opened 2026-07-23
Severe Voice Limiting in Multisampled Instruments (pops, crashes)


AndrewLevinMusic:
### Please describe the problem:

I have been getting bad clicks and pops with all of my multisampled instruments even when setting voice limiting as low as 2-3. It has even crashed my Deluge project a few times. Reverted my SD card back to 1.2 to confirm it did not happen on the older Firmware. I absolutely love 1.3 so far and this is my only issue, but it makes it somewhat unusable unfortantely as I used multisampled instruments frequently.

### What is the expected behavior?

It should be able to handle many voices without clicks and pops

### When did you start noticing the issue?

7/21/2026

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

1.3 Nightly

### If possible provide the steps to reproduce the issue and upload additional media:

I have been using a Rephazzer sample pack, but any somewhat completely multi-sampled instrument will likely do the trick.

seangoodvibes:
Hi @AndrewLevinMusic can you share a collect all save of the problem song that works on 1.2 but not 1.3 please?

You can zip it and attach here

Thanks

m-m-adams:
looks like it could be fixed

seangoodvibes:
> looks like it could be fixed

Re-opening cause this is related to that song with the arrangement where only the first half was fixed. second half of the arrangement still culls significantly compared to 1.2.

AndrewLevinMusic:
just commenting here to bump this as I've had the issue on other songs now as well. @seangoodvibes are there any specific tests I could run that might help you solve this issue? for example, recording audio tracks until they start dropping out and posting a video, etc.

seangoodvibes:
> just commenting here to bump this as I've had the issue on other songs now as well. @seangoodvibes are there any specific tests I could run that might help you solve this issue? for example, recording audio tracks until they start dropping out and posting a video, etc. 

Nothing for now but I'm working on a cpu usage menu that should help with gaining more insight into what's going in people's projects. (Eg how many notes are on, are there any heavy fx, etc)

I'll reach out when I'm ready if you want to help me test it by checking what it says for your problem songs.

PR #4722, opened 2026-07-24
fix: voice-cull victim selection never filtered the first voice (#4721)

stellar-aria:
All four cull selectors initialized `best` with the first voice and only applied the already-fast-releasing skip filter to the rest. A just-fast-released voice has the highest priority rating, so when it sat at the front it won the comparison and absorbed every subsequent cull in the same render window.

Not a fun bug. :(

### DIFF
--- a/src/deluge/processing/engines/audio_engine.cpp
+++ b/src/deluge/processing/engines/audio_engine.cpp
@@ -387,9 +387,7 @@ void routineWithClusterLoading(bool mayProcessUserActionsBetween) {
 	audioFileManager.loadAnyEnqueuedClusters(128, mayProcessUserActionsBetween);
 
 	if (!routineBeenCalled) {
-		// This catch-up routine is already running after SD / cluster-loading work delayed audio. Release 1.2 skipped
-		// culling here so a one-off storage stall didn't get treated as a synth load problem.
-		bypassCulling = true;
+		// bypassCulling = true; // yolo? Sean: not sure if this is necessary
 
 		logAction("call runRoutine() from routineWithClusterLoading()");
 		runRoutine();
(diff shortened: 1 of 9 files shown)


src/deluge/processing/sound/sound.cpp
```diff
@@ -4941,22 +4941,31 @@ void Sound::terminateOneActiveVoice() {
...
 
 	if (best == nullptr) {
 		return;
 	}
 
 	const ActiveVoice& voice = *best;
-	bool still_rendering = voice->doFastRelease(4 * SOFT_CULL_INCREMENT);
+	// 2x SOFT_CULL_INCREMENT = a 64-sample (~1.5 ms) fade. Deliberately faster than 1.2's voice-limit steal (1x,
+	// which was felt to reclaim voices too slowly) but not the 4x (32-sample, sub-millisecond) rate that read as a
+	// pop on sustained material when a low maxVoices makes every note-on steal (issue #4721).
```
m-m-adams: Can we confirm that's still a problem? We want the voice to stop as quickly as possible, otherwise it's still contributing to our dsp load when we go to cull again next cycle 

Or possibly we should pass the speed as a parameter so that voice limits are slower than culling

m-m-adams: Terminate's place in the cull system was originally a literal immediate voice unassign so I don't want to drift too far from that. Ideally we shouldn't ever get to this stage

m-m-adams: actually if we're culling the window is probably longer than 64 anyway so meh

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -439,19 +445,16 @@ void cullVoices(size_t numSamples, int32_t numAudio, int32_t numVoice) {
 			culled = true;
 		}
 
-		// Or if it's just a little bit dire, do a soft cull with fade-out, but only cull for sure if numSamples
-		// is increasing
+		// Or if it's just a little bit dire, do a soft cull with fade-out. Release 1.2 did this immediately when over
+		// the limit; waiting for a rising trend lets max-release voices stack up until the hard-cull path fires.
```
m-m-adams: This is no longer true with the fix to cap the release rate

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -406,27 +421,18 @@ uint8_t numRoutines = 0;
...
 			D_PRINTLN("Culling %d voices", num_to_cull);
 
-			for (int32_t i = num_to_cull / 2; i < num_to_cull; i++) {
-				// cull with fast release
-				terminateOneVoice(numSamples);
-			}
-			for (int32_t i = 1; i < num_to_cull / 2; i++) {
+			for (int32_t i = 0; i < num_to_cull; i++) {
 				// cull with immediate release
 				killOneVoice(numSamples);
```
m-m-adams, line 444: This pops, we might want to keep the split to terminate first and only kill if we're stopping multiple voices

src/deluge/processing/engines/audio_engine.cpp
```diff
@@ -475,8 +478,7 @@ void cullVoices(size_t numSamples, int32_t numAudio, int32_t numVoice) {
 /// set the direness level and cull any voices
 inline void setDireness(size_t numSamples) { // Consider direness and culling - before increasing the number of samples
 	// number of samples it took to do the last render
-	auto dspTime = (int32_t)(getAverageRunTimeForTask(routine_task_id) * 44100.);
-	size_t nonDSP = numSamples - dspTime;
+	auto dspTime = (int32_t)(getLastRunTimeForTask(routine_task_id) * 44100.);
```
m-m-adams: I'm not sure about this change. We've been using the smoothed average since 1.0, last is very sensitive to transient jumps from starting voices or forcibly loading a cluster

m-m-adams (approved):

PR #4753, opened 2026-07-30
Fix/ send automated midi cc changes on section launch

### LINKED ISSUE #4749: Automation: CC value of zero is not getting sent

When automating MIDI, if the entire clip has a value of 0 nothing seems to be sent even though it is showing as automated on the Deluge.

CC values of zero should be sent when automated, even if the entire clip has a value of zero for that parameter.

This is important as many device use a value of zero to switch toggles (e.g. arp on/off)

07.30.26

No

OLED

Nightly

deluge-v1_3_0-nightly+2026_07_07-4a8bead

Set up a midi clip and automate a midi cc to zero for the entirety of that clip. Set corresponding control on a hardware device to something other than zero and note that it doesn't respond.

I've been using arp on/of on my Sub37 to test.

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4749

This fixes the following scenario:

You have a song with two midi clips from the same midi output.

The first midi clip automated midi cc value to 0 across the entire clip timeline The second midi clip automated midi cc value to 8 across the entire clip timeline

When you switch between clips either by manually launching or through section launching, the deluge would never send out the automated midi cc value of the clip you're changing to (so first clip plays and you set value to 0 which got send out to your midi device, then you switch to second clip and your midi device still has its value set to 0 even though the second clip automated cc value to 8).

The reason this happens is because the deluge only sends out midi cc value's if the current value changed, but that current value is local to the clip so when you switch to another clip its just comparing automated value to the current value of that clip and thus doesn't send anything out (0 = 0 or 8 = 8 -> don't send anything).

This fixes that by using the same logic as midi program changes which is to send the current automated midi cc values once when the active clip is changed.

--

Due to the potential to overwhelm the midi protocol without a midi queuing system in place, this change is gated behind a new community features menu toggle "SendMIDICCOnClipSectionLaunch"


m-m-adams:
good to go when we merge the priority queues

PR #4755, opened 2026-07-30
Oled screensaver 

### LINKED ISSUE #4736: [Feature Request] OLED Display Auto Off / Timeout after user inactivity
Hi,

To prevent **OLED screen burn-in** when the Deluge is left unused, it would be useful to add a feature that **automatically turns off the display after a period of inactivity.**

Here is the proposed behavior:
- **Configurable timer**: Set the display to turn off completely after 15s, 30s, 1 min, 5 mins, 15 mins, or OFF.
- Wake-up triggers: The display turns back on via physical user interaction (buttons, grid, encoders).

MIDI handling: Include an option to choose whether incoming **MIDI events wake up the display or not**. This allows users synced to external clock/gear to keep the display turned off without clock messages constantly triggering it.

This would protect OLED displays without interrupting workflow.

Thanks for your work on the community firmware.

Regards,

stellar-aria:
Adds three screensaver modes: Blank, Starscape, and Deluge (default), for preventing OLED burn-in. Default timeout is 5 minutes, but can go down to 1 minute.


alifeinbinary:
I've been wanting a screen saver on the Deluge for so long and you've just gone ahead and done it. Great work @stellar-aria, both animation modes are really cool! Thank you!

seangoodvibes (approved):

PR #4758, opened 2026-07-31
Fix/ Enable midi follow feedback to send feedback to both track + reg channel

seangoodvibes:
This fixes issue reported to me by @busa-projects. I worked with them to test that it works as expected now.

With the addition of midi follow control of specific tracks, a "track" feedback option was added.

This option sends feedback to the selected track if it was configured for track control.

The downside however is that now regular midi follow channels don't receive feedback.

To address this, add 3 more midi feedback channels to send feedback to the track AND either channel A, B, or C

--

Complementary change: made the MidiEngine::midiFollowChannelType[] array smaller by removing NONE and TRACK from the array.


m-m-adams:
Is this desireable? It feels like it's just duplicating data

seangoodvibes:
> Is this desireable? It feels like it's just duplicating data

Yes it’s desirable for users with controllers like Electra one or Neuzit

It’s because right now it only sends feedback for either the specific track, or regular midi follow channel, not both

So if you’re using one controller for both (which is common), you would want both channels to remain in sync with current deluge values

Here’s the example from @busa-projects:

“Lets say you have LPF on track 1 (CH1) AND track one is the selected clip for midi follow (CH16), than both would need the update if you change the value on the deluge side. Otherwise one would drift out of sync.”

m-m-adams (approved):

PR #4759, opened 2026-08-01
Fix midi follow not reading midi device

### LINKED ISSUE #4757: MIDI-FOLLOW device assignment and per-device MPE zone settings do not persist across power cycle
Firmware: Community 1.3.0-BETA-EA4E69E

Environment: SD card: brand new, freshly FAT32-formatted, clean install (factory content + firmware only, no migrated files)
Setup: Deluge connected via a CME H2MIDI Pro (USB host interface) hosting one or two class-compliant USB MIDI controllers (tested with a Novation Launch Control XL mk2 alone, and together with an Exquis MPE controller — same result in both cases)

Steps to reproduce:

Fresh clean-installed SD card, first boot.
Connect a USB MIDI controller through the host interface.
SETTINGS > MIDI > MIDI-FOLLOW > CHANNEL >A or B — Learn the controller to a channel.
SETTINGS > MIDI > DEVICES > [controller] > MPE > IN — set Lower/Upper Zone to 0 (disable MPE for this device).
Confirm in-session: CC's correctly control synth parameters via MIDI Follow.
Full power cycle the Deluge.
Return to MIDI-FOLLOW > CHANNEL > A or B: the channel number is retained, but the device shows as "Device Unassigned" and CC's no longer have any effect.
Return to DEVICES > [controller] > MPE: the zone value set in step 4 has reverted.
Re-doing Learn (step 3) and the MPE reset (step 4) restores full functionality for the current session only — the same loss repeats on every subsequent power cycle.

Expected behavior: Once a MIDI-FOLLOW channel is learned to a device and a device's MPE zone is configured, both should persist across power cycles.

Actual behavior: The numeric channel value persists; the device association and the MPE zone configuration do not.

Additional notes:

Reproduced identically with a single connected USB controller (rules out multi-device interaction as the sole cause) and with a completely fresh SD card (rules out storage corruption).
The current beta's regenerated MIDIFollow.XML now stores Channel A/B/C and per-track channel assignments directly in a <settings> block within the file itself (previously, per the changelog, these lived in flash memory). Manually hand-editing these values in the XML and reloading does cause the Deluge to read them correctly (verified for Channel A/B) — suggesting the read path is fine, and the issue may be specific to how certain settings are committed to storage via the on-device Learn/config UI.

seangoodvibes:
Fix #4757

Fixed issue where midi follow would save midi device to MIDIFollow.xml and would read it but not store it in the midi follow channel type variable.

### DIFF
--- a/src/deluge/io/midi/midi_follow.cpp
+++ b/src/deluge/io/midi/midi_follow.cpp
@@ -1808,7 +1808,8 @@ void MidiFollow::readSpecificChannelSettingsFromFile(Deserializer& reader, MIDIF
 		}
 		// step into <device> tag
 		else if (!strcmp(tag_name, MIDI_DEFAULTS_SETTINGS_DEVICE_TAG)) {
-			MIDICable* cable = MIDIDeviceManager::readDeviceReferenceFromFile(reader);
+			midiEngine.midiFollowChannelType[util::to_underlying(type)].cable =
+			    MIDIDeviceManager::readDeviceReferenceFromFile(reader);
 		}
 		// exit so we can step into next tag
 		reader.exitTag();


PaoloQuaranta:
Thank you! Hope you don't mind if I have a few questions, as I am new here..
You fixed this very fast! Now I see you need approval to get it to the next beta release: how often do you release a new beta?
Do you think this will also fix the commands MIDI learn problem, or should I open another issue?

seangoodvibes:
> Thank you! Hope you don't mind if I have a few questions, as I am new here..
> 
> You fixed this very fast! Now I see you need approval to get it to the next beta release: how often do you release a new beta?
> 
> Do you think this will also fix the commands MIDI learn problem, or should I open another issue?

It will get approved when someone has a chance to review it.

I don't know what the global midi commands issue is so I'd open another PR.

Thanks

m-m-adams (approved):

PR #4765, opened 2026-08-02
MIDI/ Add midi queuing system

### LINKED ISSUE #4687: deluge clock random drop outs (both, midi and gate/clock)

I'm having clocking issues.

both, gate outs and midi keep skipping ticks randomly.

this is only happening with the clock. 

when sending out notes/gates from a clip the delly is very stable.

consistent midi clock

since community firmware. not sure if this is also with the og firmware

_No response_

7SEG

other

with all community firmwares, also with 1.3. (not sure about synthstorm firmware...)

Image

seangoodvibes:
Adds a MIDI Queue Manager which prioritizes sending out MIDI by type, only sending MIDI when queue MIDI of a higher priority type has been fully sent.

Order of priority:

1. CLOCK
2. NOTES,
3. EXPRESSION,
4. CC,
5. SYSEX

Maybe Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4687

### DIFF
--- a/src/deluge/io/midi/midi_device_manager.cpp
+++ b/src/deluge/io/midi/midi_device_manager.cpp
@@ -649,22 +649,22 @@ void ConnectedUSBMIDIDevice::bufferMessage(uint32_t fullMessage, QueuePriority p
 	// If backlog grows, opportunistically kick a flush to keep latency bounded.
 	if (queued > 16) {
 		// Only trigger a new flush when a send transaction is not already active.
-		if (!anyUSBSendingStillHappening[0]) {
+		if (anyUSBSendingStillHappening[0] == 0) {
 			midiEngine.flushUSBMIDIOutput();
 		}
 	}
 
 	// Occupancy of just the selected priority lane we are about to enqueue into.
-	uint16_t queueSize = MidiQueueManager::usbQueueCount(this, priority);
+	uint16_t queue_size = MidiQueueManager::usbQueueCount(this, priority);
 	// Keep one slot free in a ring buffer so full/empty states stay distinguishable.
-	if (queueSize >= (MIDI_SEND_BUFFER_LEN_RING - 1)) {
+	if (queue_size >= (MIDI_SEND_BUFFER_LEN_RING - 1)) {
 		// If nothing is currently transmitting, try flushing now to free space.
-		if (!anyUSBSendingStillHappening[0]) {
+		if (anyUSBSendingStillHappening[0] == 0) {
 			midiEngine.flushUSBMIDIOutput();
 		}
 		// Re-check after opportunistic flush.
-		queueSize = MidiQueueManager::usbQueueCount(this, priority);
-		if (queueSize >= (MIDI_SEND_BUFFER_LEN_RING - 1)) {
+		queue_size = MidiQueueManager::usbQueueCount(this, priority);
+		if (queue_size >= (MIDI_SEND_BUFFER_LEN_RING - 1)) {
 			// Still full: drop this message rather than overwrite unread queued data.
 			// TODO: show some error message
 			return;
--- a/src/deluge/io/midi/midi_device_manager.cpp
+++ b/src/deluge/io/midi/midi_device_manager.cpp
@@ -25,6 +25,7 @@
 #include "io/midi/device_specific/specific_midi_device.h"
 #include "io/midi/midi_device.h"
 #include "io/midi/midi_engine.h"
+#include "io/midi/midi_queue_manager.h"
 #include "mem_functions.h"
 #include "memory/general_memory_allocator.h"
 #include "storage/storage_manager.h"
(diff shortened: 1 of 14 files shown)


src/deluge/io/midi/midi_engine.cpp
```diff
@@ -580,22 +591,20 @@ void MidiEngine::sendUsbMidi(MIDIMessage message, int32_t filter) {
...
-
-	if (messageLength >= 2) {
-		bufferMIDIUart(message.data1);
-
-		if (messageLength == 3) {
-			bufferMIDIUart(message.data2);
-		}
-	}
+void MidiEngine::sendSerialSysex(uint8_t const* data, int32_t len) {
+	// Route SysEx through queue manager so long transfers yield to higher-priority
```
m-m-adams: You can't do this, it ends the sysex transmission

src/deluge/io/midi/midi_queue_manager.cpp
```diff
@@ -0,0 +1,418 @@
...
+	// Realtime lane can preempt and send a single byte immediately.
+	if (!serialPriorityQueues_[kClockIdx].empty()) {
+		if (budgetBytes < 1 || uartSpace < 1 || maxLen < 1) {
+			return 0;
+		}
+		serialPriorityQueues_[kClockIdx].pop(outBytes[0]);
+		return 1;
+	}
+
+	// For channel messages, prefer popping whole messages to avoid fragmentation.
```
m-m-adams: You cannot ever interrupt messages. Always needs to send the full message

src/deluge/io/midi/midi_queue_manager.cpp
```diff
@@ -0,0 +1,418 @@
...
+		return 0;
+	}
+
+	constexpr size_t kClockIdx = static_cast<size_t>(QueuePriority::CLOCK);
+	constexpr size_t kNotesIdx = static_cast<size_t>(QueuePriority::NOTES);
+	constexpr size_t kExpressionIdx = static_cast<size_t>(QueuePriority::EXPRESSION);
+	constexpr size_t kCcIdx = static_cast<size_t>(QueuePriority::CC);
+	constexpr size_t kSysexIdx = static_cast<size_t>(QueuePriority::SYSEX);
+
+	// Realtime lane can preempt and send a single byte immediately.
```
m-m-adams: Why? What?

src/deluge/io/midi/midi_queue_manager.cpp
```diff
@@ -0,0 +1,418 @@
...
+	if (!serialPriorityQueues_[kClockIdx].empty()) {
+		if (budgetBytes < 1 || uartSpace < 1 || maxLen < 1) {
+			return 0;
+		}
+		serialPriorityQueues_[kClockIdx].pop(outBytes[0]);
+		return 1;
+	}
+
+	// For channel messages, prefer popping whole messages to avoid fragmentation.
+	auto tryPopWholeMessage = [&](size_t queueIdx) -> int32_t {
```
m-m-adams: Why is this a lambda?

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
```diff
@@ -0,0 +1,287 @@
...
+bool MIDIQueueManagerUSBUpstream::pop_priority_message(ConnectedUSBMIDIDevice* device, uint32_t& message_out,
+                                                       int32_t& cc_budget_packets_remaining) {
+	for (uint8_t p = QUEUE_PRIORITY_CLOCK; p < QUEUE_PRIORITY_COUNT; p++) {
+		QueuePriority priority = static_cast<QueuePriority>(p);
+		if (!queue_count(device, priority)) {
+			continue;
+		}
+
+		if (priority == QUEUE_PRIORITY_CC) {
+			// Inspect the CC-lane head packet to decide whether CC fairness rules apply.
```
m-m-adams, line 72: what does this mean?

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
```diff
@@ -0,0 +1,287 @@
...
+			uint16_t head_idx = device->ringBufReadIdx[p] & MIDI_SEND_RING_MASK;
+			uint32_t head_message = device->sendDataRingBuf[p][head_idx];
+
+			if (is_channel_cc(head_message)) {
+				// Per-transfer CC cap prevents low-priority bursts from dominating the batch.
+				if (cc_budget_packets_remaining <= 0) {
+					// Skip CC for now and keep scanning lower lanes (e.g. SysEx) this pass.
```
m-m-adams, line 79: why?

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
```diff
@@ -0,0 +1,287 @@
...
+					continue;
+				}
+				// Pop one fair-selected CC packet (RR baseline + debt preference).
```
m-m-adams, line 82: expand on this comment please

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	return true;
+}
+
+/// Pops one queued USB channel-CC packet using controller fairness.
+///
+/// Selection flow:
+/// 1. Capture each controller's first queued CC offset.
+/// 2. Establish RR baseline from rotating controller cursor.
```
m-m-adams, line 221: what is "RR"?

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -636,44 +638,70 @@ void readAHostedDeviceFromFile(Deserializer& reader) {
 
 } // namespace MIDIDeviceManager
 
-void ConnectedUSBMIDIDevice::bufferMessage(uint32_t fullMessage) {
-	uint32_t queued = ringBufWriteIdx - ringBufReadIdx;
+/// Queues one USB-MIDI packet into the selected priority lane with backpressure handling.
+void ConnectedUSBMIDIDevice::bufferMessage(uint32_t fullMessage, QueuePriority priority) {
```
m-m-adams, line 642: why is the priority passed in instead of set by the get priority for message function?

src/deluge/io/midi/midi_device_manager.cpp
```diff
@@ -636,44 +638,70 @@ void readAHostedDeviceFromFile(Deserializer& reader) {
...
+	// Defensive fallback: unknown priority values are treated as regular CC lane.
+	if (priority >= QUEUE_PRIORITY_COUNT) {
+		priority = QUEUE_PRIORITY_CC;
+	}
+
+	// Total packets currently queued across all priority lanes for this device.
+	uint32_t queued = MIDIQueueManagerUSBUpstream::total_queued_messages(this);
```
m-m-adams, line 649: why is the upstream queue manager involved in downstream hosted devices?

src/deluge/io/midi/midi_device_manager.h
```diff
@@ -90,9 +88,25 @@ struct ConnectedUSBMIDIDevice {
 	// Any code which wants to send midi data would use the writing side and append more messages.
 	// When we are ready to send data on this device, we consume data on the reading side and move it into the
 	// smaller dataSendingNow buffer above.
-	uint32_t sendDataRingBuf[MIDI_SEND_BUFFER_LEN_RING];
-	uint32_t ringBufWriteIdx;
-	uint32_t ringBufReadIdx;
+	// Messages are queued in priority-specific rings and consumed in priority order.
+#ifdef __cplusplus
```
m-m-adams, line 92: uhhhhh what is going on here the C++ and C definitions are not equivalent

m-m-adams: and also why???

m-m-adams (commented):
This is still really hard to follow and seems overly complicated. Why are there so many different classes for queue management? Just make one and instantiate it multiple times. There's a ton of code duplication and it's hard to spot differences between them or understand their purpose

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
seangoodvibes: This was added because sending a lot of cc automation messages without a cap causes a cumulative impact on clock drifting and ultimately notes skipping.

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
seangoodvibes: This relates to the midi cc cap mentioned above. When the cap is reached, cc's are skipped from being sent. Next time you come around you need to send out the cc's that were skipped.

Essentially it's trying to fairly send out all cc's so that it all averages out and no cc's are missed in the end.

E.g. round 1:

send cc 0, 2-8, skip 9

round 2:

send 9 then send the rest up to the cap

src/deluge/io/midi/queue_manager_types/usb_device_cable.cpp
seangoodvibes: round robin

src/deluge/io/midi/midi_device_manager.cpp
seangoodvibes: That's my bad with the naming. I was aligning the class structure with the cable types folder which had broken things down into din, usb_device_cable, usb_hosted and got things mixed up.

src/deluge/io/midi/midi_device_manager.cpp
seangoodvibes: > why is the priority passed in instead of set by the get priority for message function?

It's cause if you look at MidiEngine::sendUsbMidi it gets the priority from the MIDIMessage

```
void MidiEngine::sendUsbMidi(MIDIMessage message, int32_t filter) {
QueuePriority usb_priority = MIDIQueueManager::classify_message(message);
```

The message that then gets sent to bufferMessage is not the same:

```
uint32_t channeled_message = fullMessage | (p << 4);
connectedDevice->bufferMessage(channeled_message, usb_priority);
```

So you would have derive the MIDIMessage from channeled_message within bufferMessage and then get the priority from that.

Also in another spot the priority for sysex is just hardcoded as you're not getting the priority from the message in that case. So you couldn't derive a MIDIMessage to get the priority from that.

```
uint32_t packed = ((uint32_t)data[5] << 24) | 0x007DF004 | (portNumber << 4);
connectedDevice->bufferMessage(packed, QUEUE_PRIORITY_SYSEX);
```

It's done similarly for DIN also in that the priority is determined before enqueuing the message:

```
void MIDIQueueManagerDINPorts::enqueue_serial_message(MIDIMessage message) {
	// Convert message to wire bytes and queue by shared priority policy.
	uint8_t status_byte = message.channel | (message.statusType << 4);
	uint8_t raw_bytes[3] = {status_byte, message.data1, message.data2};
	int32_t message_length = bytesPerStatusMessage(status_byte);
	QueuePriority priority = MIDIQueueManager::classify_message(message);
	if (priority == QUEUE_PRIORITY_CC && coalesce_queued_cc(message)) {
		// For dense CC streams, replace pending controller value instead of appending another packet.
		return;
	}

	// Atomic lane enqueue; false means lane full and message is intentionally dropped.
	bool queued_ok = enqueue_serial_bytes(priority, raw_bytes, message_length);
```

src/deluge/io/midi/midi_device_manager.h
seangoodvibes: That's a def hallucination. LLM was trying to "future proof" the solution by adding C compatible definitions that are not currently needed. Removing that.

the cplusplus def is needed cause the header is used by the c usb drivers

seangoodvibes:
Closing, I'll open another PR when I'm ready

PR #4794, opened 2026-08-07
fix: bound screen transpose to the playable note range

### LINKED ISSUE #4778: You can transpose notes outside usable range

1. Max note is G8, Min note is C-2
2. Add note to G8 note row
3. Holding <> + turning vertical encoder lets you transpose that note outside the usable range (e.g. G8 becomes A8)
4. Holding <> + press and turning vertical encoder also lets you transpose note outside usable range (e.g. A8 becomes B8)
5. After doing the above, the note range becomes permanently changed (eg. max B8, min C-2)

You can't transpose notes outside the usable range (respect max and min of note range). Usable note range doesn't change.

today

_No response_

OLED

Beta

_No response_

_No response_

stellar-aria:
Fixes #4778

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -6242,6 +6242,40 @@ void InstrumentClipView::commandTransposeScreen(int32_t offset, bool inOctave) {
 	int32_t screenStartPos = xScroll;
 	int32_t screenEndPos = xScroll + (xZoom * kDisplayWidth);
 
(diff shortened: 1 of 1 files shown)


seangoodvibes:
Is this a bug with the screen transpose?

1. Go to any root note, add a note
2. Hold horizontal encoder down
3. Press and turn vertical encoder right (I would expect it to move forward a root note, but it seems to move backwards to a non-root note)


https://github.com/user-attachments/assets/6f4d6b44-11fd-44ae-92bb-26e3ebb2bb39

seangoodvibes (approved):

PR #4795, opened 2026-08-07
Fix text highlight alignment in browser / rename UI

### LINKED ISSUE #4781: Text highlighting in browser / rename UI doesn't look good

When in the browser or rename UI where the text entered is highlighted, if you scroll the horizontal encoder it moves the highlighting to show cover letters. Currently the highlighting is not properly aligned so it covers parts of the letters.

Highlighting doesn't cover letters and keeps things readable

Today

_No response_

OLED

Beta

_No response_

_No response_

stellar-aria:
Fixed the inverted highlight bar in the browser and rename UIs clipping the character to the left of the cursor when moving through the text with the horizontal encoder.

Regression from #4233, which reinstated this logic with a hardcoded -2 in place of the original adjustment (0 inside the string, 1 at its start).

### DIFF
--- a/src/deluge/gui/ui/qwerty_ui.cpp
+++ b/src/deluge/gui/ui/qwerty_ui.cpp
@@ -160,7 +160,10 @@ void QwertyUI::drawTextForOLEDEditing(int32_t xPixel, int32_t xPixelMax, int32_t
 			highlightStartX = 0;
 		}
 		else {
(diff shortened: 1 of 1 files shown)


seangoodvibes (approved):

seangoodvibes:
Tested, works, lgtm

PR #4803, opened 2026-08-11
Oops this shoulda been a max

### LINKED ISSUE #4802: Threshold recording to audio clips requires loop margins to be turned off

For threshold recording to work with audio clips, you need to turn off the loop margins settings in the settings > recording menu. 

Otherwise when loop margins is on the first sound recorded at threshold gets cut off

Threshold recording works with loop margins turned on

Today

_No response_

OLED

Beta

_No response_

_No response_

m-m-adams:
Fix #4802

### DIFF
--- a/src/deluge/model/sample/sample_recorder.cpp
+++ b/src/deluge/model/sample/sample_recorder.cpp
@@ -336,7 +336,7 @@ void SampleRecorder::setRecordingThreshold(RecorderConfig config) {
 		}
 
 		thresholdRecording = true;
-		minThresholdMargin = std::min<uint32_t>(sample->fileLoopStartSamples, 256);
+		minThresholdMargin = std::max<uint32_t>(sample->fileLoopStartSamples, 256);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


seangoodvibes (approved):

src/deluge/model/sample/sample_recorder.cpp
```diff
@@ -336,7 +336,7 @@ void SampleRecorder::setRecordingThreshold(RecorderConfig config) {
 		}
 
 		thresholdRecording = true;
-		minThresholdMargin = std::min<uint32_t>(sample->fileLoopStartSamples, 256);
+		minThresholdMargin = std::max<uint32_t>(sample->fileLoopStartSamples, 256);
```
seangoodvibes, line 339: This has a side effect in that the audio recorder in a kit displays "recording" right away instead of "waiting" until the threshold is passed.

Works fine on main currently.

seangoodvibes: Fixed

PR #4810, opened 2026-08-12
Fix/save name numbering window

stellar-aria:
Basically a replace the save-numbering-suffix stuff, for both OLED and 7SEG. Needs serious testing and potentially changes before merging, but _should_ fix the class of bugs.

Includes a bunch of unit tests so this doesn't happen again.


walkerab:
I've done some limited testing on OLED hardware.

On a song called "SONG 9", when I hit SAVE I now see "SONG 10" as the suggested name instead of what I see on the Beta build - "SONG 2" (which already exists).

PR #4833, opened 2026-08-20
Remove a memory leak on every boot with a startup song

cyface:
## Summary

Every boot that loads a startup song (TEMPLATE / LASTOPENED / LASTSAVED mode) unnecessarily allocates a Song-sized block of RAM for the whole session, reducing headroom for voices and sample clusters.

The allocation in `setupStartupSong()` is simply dead code: the pointer is never used, null-checked, or freed — `performLoad()` does its own Song allocation when it actually creates the song. The fix is deleting the line.

## Testing

- Validated that saved songs still load fine on boot with this line removed on OLED Deluge and DelugEmu.
- `./dbt build release` compiles and links cleanly with the change.

### DIFF
--- a/src/deluge/deluge.cpp
+++ b/src/deluge/deluge.cpp
@@ -476,7 +476,6 @@ void setupStartupSong() {
 			}
 		}
 		// Load song, if we got this far!
-		void* songMemory = GeneralMemoryAllocator::get().allocMaxSpeed(sizeof(Song));
 		currentSong->setSongFullPath(filename);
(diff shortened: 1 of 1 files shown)


m-m-adams:
Please keep our contribution policy in mind wrt PR descriptions. It's not a leak (just a static 2kb waste of space) so I'm gonna close this and just incorporate it into some other work on the binary size I'm doing

cyface:
@m-m-adams thanks for looking! I appreciate all your work here. 🙏🏻

PR #4852, opened 2026-08-23
Clarify community_features.mdx

jpursey:
- Added explicit references to the toggle menu options within the community features in docs. Some were obvious others were not. This change makes all of them explicit for consistency.

- Removed reference to changing option for Patch Cable Modulation Resolution option, as it no longer seems to be configurable.

- Removed redundancy in "the point the point".

- Corrected "4.3.5 - Automation View" section to refer to the AUTOMATION menu in DEFAULTS instead of COMMUNITY FEATURES, as that is where it actually is now.

- Updated the default for "TRIM FROM START OF AUDIO CLIPS" to be "ON" instead of "OFF", since that is what it was (and I just installed the firmware for the first time).


website/src/content/docs/features/community_features.mdx
seangoodvibes, line 117: Related to this, can you remove Mod. Depth Decimals (MOD.) from the Community Features menu section at the bottom of this document

jpursey: Done!

seangoodvibes (approved):

PR #4855, opened 2026-08-24
Add per device midi thru toggle

seangoodvibes:
Requested by Jules_Etro on discord to be able to selectively disable specific devices (e.g. their MPE controller) from sending MIDI thru:

> Hi dear Delugers, question about MPE. Is there a way to exclude MPE from being passed through Midi Through to prevent trigerring random other Synths when using the MPE controller to play Deluge synths?

I asked if they tried turning off MIDI Thru but they confirmed that they need it for some devices.

> Yes but I need to use midi through to pass midi data to other devices

Also I think addresses legacy comment left by Rohan:

https://github.com/SynthstromAudible/DelugeFirmware/blob/7ba8a467338de46d6c17b6cc9a7511caf4b54d70/src/deluge/io/midi/midi_engine.cpp#L1082

- Added new `MIDI Thru` toggle within the configuration of each MIDI device in the `Settings > MIDI > Devices` menu.
  - If Devices have not already been set up, or if you Reset Device Settings, then this will inherit by default the Global MIDI Thru setting.
  - After the per Device MIDI Thru setting has been setup, the Global MIDI Thru setting will no longer be used to determine whether MIDI from a specific Device should be sent through
  
Confirmed working by Jules on discord


src/deluge/gui/menu_item/midi/device_is_relative.h
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2014-2023 Synthstrom Audible Limited
- *
+ * Copyright (c) 2026 Sean Ditny
```
stellar-aria, line 2:
```suggestion
 * Copyright (c) 2026 Synthstrom Audible Limited
```

stellar-aria:
clang-tidy has a number of valid nits over the implicit bool conversions, could we fix those first?

stellar-aria (changes requested):

src/deluge/gui/menu_item/midi/device_is_relative.h
seangoodvibes: Oh I was updating this one cause I added that toggle :p

seangoodvibes:
> clang-tidy has a number of valid nits over the implicit bool conversions, could we fix those first?

Will do!

seangoodvibes:
> clang-tidy has a number of valid nits over the implicit bool conversions, could we fix those first?

tackled the clang tidy's but then it flagged a couple more. I think if I touch these it will create a whole bunch more clang tidy's. let me know if you want me to fix the remaining two.

Thx

stellar-aria (approved):

PR #4856, opened 2026-08-24
Fix for digital reverb right channel filtering

cyface:
## Summary

Hello! While working on some reverb stuff, I came across this little bug where the right-side reverb was being sent through the left-side filter setting:

```cpp
right_sum = right_sum - dsp::OnePole(hp_l_, right_sum, hp_cutoff_);
right_sum = dsp::OnePole(lp_l_, right_sum, lp_cutoff_);
```

I see it correctly routed to the right side in other places, so I am pretty sure this is a typo.

## Change
Just  changing `l` to `r` on the `right_sum` lines.

## Testing

- Verified on OLED Deluge using headphone output.


seangoodvibes:
Looks like a valid typo, good catch :) thanks! Just getting another team member to check.

stellar-aria (approved):

Issue #4828, opened 2026-08-17
Audio click issue in C1.3.0 (Aug 17 build or earlier) when switching scenes on one particular song


Mark-Mosher:
### Please describe the problem:

In C1.3 (Aug. 17 nightly and prior), I hear a reproducible audible click in one specific song when advance to another section. It's always the same set of sections if I trim the song down to only include those two sections.

The issue does not occur in C1.2.1.

This is isolated to one of the 30+ songs I use live, but it is consistently reproducible. The song is in Row mode.

### What is the expected behavior?

C1.2.1 Behavior: The transition to the next section for this song always plays without an audible click.

### When did you start noticing the issue?

When I started using any version of C1.3.0 on or prior to Aug 17, 2026.

### What hardware did you reproduce it with?

OLED

### What firmware did you reproduce it with?

Nightly

### What is the firmware name:

deluge-v1_3_0-beta+2026_08_17-685b4fb.bin

### If possible provide the steps to reproduce the issue and upload additional media:

I tested on two OLED machines and get the same behavior.

I included a song (with all samples collected in a folder) and a video. It's in a dropbox [here](https://www.dropbox.com/scl/fo/26w9h4plde1p7l6z2uf12/AHEBC7lvsRUvAdX4enDLY2g?rlkey=a9qcrb24woyjnxjdg7rnmg26t&st=lr3k3zm7&dl=0). 

For some reason, the click is less prominent in the video example, but I think you’ll hear it clearly if you load the demo song.

1. Load Cosmic Dreams 6.
2. Start the transport.
3. Section 1 starts and automatically advances to Section 2. (The issue also occurs when Section 1 is set to repeat indefinitely and I manually advance to the next section.)
4. When Section 2 begins, an audible click occurs.

Additional observations
- If I stop the transport and repeat the procedure, the click does not occur.
- If I leave the transport running and return to Section 1 and advance again, the click does not occur.
- Reloading the song sometimes reproduces the click.
- If this is the first song played after starting the machine, the click occurs every time.

seangoodvibes:
Fixed by #4858

PR #4858, opened 2026-08-24
Fix click on an audio track's first note after loading a song

cyface:
Fixes Issue SynthstromAudible/DelugeFirmware#4828 (audible click when audio clip launches the first time after loading a song).

## Cause

* `Envelope::smoothedSustain` starts at 0 and is only populated by the `Voice` overload of `noteOn()`. 
* `AudioOutput::resetEnvelope()` used the bare `noteOn(bool)` and didn't set a sustain value like Synths are doing, resulting in a 0 `smoothedSustain` initial value only for audio clips.
* As a result, the clip's first play dipped from max down toward that zero initial `smoothedSustain` before ramping up rapidly, causing a pop.
* Once smoothedSustain reaches the target for a clip, it doesn't get reset, which is why the pop doesn't happen when the clip is played again. The default zero value has been erased.

Synth/kit voices were unaffected, but audio clips weren't setting a smoothedSustain initial value, causing it to default to zero.

## Fix

* pass `AudioOutput::resetEnvelope()` a sustain starting value through a new `sustain` param for `Envelope::noteOn`.
* set the audio sustain value to max since they do not have a sustain envelope param value like synths do - so `smoothedSustain` starts at its target and has nothing to ramp toward
* synths now pass their starting sustain setting through that new `sustain` param instead of setting smoothedSustain directly

## Verification

* Validated with reporter's song and samples on OLED Deluge.
* Confirmed no additional memory impact.

### DIFF
--- a/src/deluge/modulation/envelope.cpp
+++ b/src/deluge/modulation/envelope.cpp
@@ -118,7 +118,8 @@ considerEnvelopeStage:
 	return (lastValue - 1073741824) << 1; // Centre the range of the envelope around 0
 }
 
-int32_t Envelope::noteOn(bool directlyToDecay) {
+int32_t Envelope::noteOn(bool directlyToDecay, int32_t sustain) {
+	smoothedSustain = sustain;
 	ignoredNoteOff = false;
 	pos = 0;
 	if (!directlyToDecay) {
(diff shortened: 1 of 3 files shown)


seangoodvibes:
Can you explain why this did not occur on c1.2.1 but occurs in c1.3? We need to make sure we are addressing the real cause and not covering up a symptom which could introduce other bugs. 
- e.g. If we accidentally changed something compared to c1.2.1 we should restore what we changed as opposed to introducing new stuff.,

cyface:
> Can you explain why this did not occur on c1.2.1 but occurs in c1.3? We need to make sure we are addressing the real cause and not covering up a symptom which could introduce other bugs.

For sure! MarkA introduced `smoothedSustain` in https://github.com/SynthstromAudible/DelugeFirmware/pull/2805 - which is great, and a useful feature. It just missed that audio clips don't have a sustain setting, so they default to zero. That PR was merged after 1.2.1 release was cut, so `smoothedSustain` doesn't exist in 1.2.1.

seangoodvibes:
I don't think this is the right fix.

Audio outputs shouldn't be using smooth sustain - so calling a function that changes smooth sustain to make audio outputs sound correct is wrong. It's papering over the issue. Smooth sustain was introduced for modulating sustain with expression which isn't relevant for audio outputs.

Previously audio outputs would pass max sustain value via envelope.render

The real issue is that the sustain value audio outputs are passing to render is not being used the same way anymore.

A better fix would probably be to pass a parameter that says whether smooth sustain should be used. If not, use the sustain value provided.

seangoodvibes:
I honestly would leave this fix to @ilovesteak since they know this subsystem the best and going back and forth isn't the most efficient.

m-m-adams:
Yeah I'll just fix this, this is gonna cause the opposite problem in synth voices

m-m-adams:
Misread in my first review, this is fine

m-m-adams (approved):

PR #4863, opened 2026-08-25
Track muted note in note editor explicitly

### LINKED ISSUE #4862: Turning off note in Note Editor will cause infinite notes when leaving note editor

Steps:
1. Enter note editor by pressing note on grid + select
2. Note is sustaining
3. Press note again to stop note from sounding. It is still selected / blinking
4. Exit note editor by pressing Back / Undo
5. Press any note in the audition column

Result: Note is triggered and held forever at max velocity and/or volume. Changing modes, playing the song, etc. has no effect. Every new note pressed via the audition buttons is added and held forever.

Normal playback occurs and sound stops when audition key is released.

After I built and uploaded firmware from github.

_No response_

OLED

Built from community

4f4a17676c261a032152ffbf91f4a3c387c7f58c

See above

jpursey:
Velocity 255 is used in the note editor to be a sentinel to indicate it is muted. However, when exiting the note editor, this leaks into "indendedVelocity" which results in every new note auditioned getting a max velocity. One approach would be to just check for that (making sure velocities are between 1 and 127 before starting), but it seemed cleaner to me to explicitly track the state and then not promote the 255 (which is invalid anyway) beyond the note editor.

Fix #4862

### DIFF
--- a/src/deluge/gui/views/instrument_clip_view.cpp
+++ b/src/deluge/gui/views/instrument_clip_view.cpp
@@ -3559,6 +3559,7 @@ bool InstrumentClipView::enterNoteEditor() {
 }
 
 void InstrumentClipView::exitNoteEditor() {
+	noteEditorAuditionMuted = false;
 	if (lastSelectedNoteXDisplay != kNoSelection && lastSelectedNoteYDisplay != kNoSelection) {
 		if (isUIModeActive(UI_MODE_NOTES_PRESSED)) {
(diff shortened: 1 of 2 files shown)


jpursey:
I'm not sure how to tag or correlate this to the bug I logged for this issue: https://github.com/SynthstromAudible/DelugeFirmware/issues/4862

seangoodvibes (approved):

PR #4870, opened 2026-08-26
Increase the usb descriptor size limit to 512 and add some usb messages

sirex99:
My keyboard, the medeli mk 49 was failing to work with the deluge. It seems this was because the usb descriptors for the device are slightly longer than the 256byte limit. This patch increases the limit to 512, which resolves that problem.

It also adds a message window to the oled screen for a brief second at boot that gives an overview of the usb setup process, to provide a little more clarity on what's happening.


seangoodvibes:
Hi now im a bit confused cause we recently merged this:

https://github.com/SynthstromAudible/DelugeFirmware/pull/4836

Can you take a look?

Is that other PR wrong when it says descriptions can't exceed 255 chars?

I don't know how this is fixing it  when the other one caps it to 255.

Thanks

m-m-adams:
They're not the same field

PR #4873, opened 2026-08-26
Use the stored root note when entering the keyboard

### LINKED ISSUE #4754: Keyboard View: New root note does not stick

To reproduce problem:

1. In a new song, press 'keyboard' button, entering kb-view,
2. hold 'Scale' button and press keyboard-pad to pick a new root note, it gets highlighted as expected.
3. press 'Scale' button so only root is lighted up, root sticks to new selection
4. press 'Scale' again, root falls back to prior root note.

I would expect to be able to switch back and forth between steps 3 and 4 (chromatic and scale) without it affecting the root note

noticed behavior on c1.2.1, upgraded to c1.3 to check if still there, which it is

_No response_

OLED

Beta

1.3.0-BETA-E40A9F0

_No response_

stellar-aria:


seangoodvibes:
Sorry it took me so long to check this @stellar-aria  I just tested and confirmed its good.

seangoodvibes (approved):

PR #4881, opened 2026-08-28
Feature/sound editor colours

m-m-adams:
Show possible params to select when shift is held. Supports everything except modulating patch cables just because that's a mess of special cases I can't untangle just yet.

It dynamically updates the overlay with what's possible to select so it'll only light pads that are usable 

In clips it'll render while shift is held, in the song it only renders once you've entered the edit menu

60E2DE79-05CD-4320-8B67-6869CF91D415_1_201_a


src/deluge/gui/l10n/seven_segment.json
```diff
@@ -453,6 +453,7 @@
         "STRING_FOR_COMMUNITY_FEATURE_ALTERNATIVE_TAP_TEMPO_BEHAVIOUR": "TAPT",
         "STRING_FOR_COMMUNITY_FEATURE_TRIM_FROM_START_OF_AUDIO_CLIP": "TRIM",
         "STRING_FOR_COMMUNITY_FEATURE_SHOW_BATTERY_LEVEL": "BATT",
+        "STRING_FOR_COMMUNITY_FEATURE_SHORTCUT_OVERLAY": "OVER",
```
seangoodvibes, line 456: would OLAY be easier to identify on 7seg?

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -7324,6 +7343,27 @@ bool InstrumentClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplay
 		return true;
 	}
 
+	if (shouldRenderShortcutsOverview()) {
+		// what even is this view???
```
seangoodvibes: Which view? I am confused lol

src/deluge/gui/views/instrument_clip_view.cpp
```diff
@@ -7324,6 +7343,27 @@ bool InstrumentClipView::renderMainPads(uint32_t whichRows, RGB image[][kDisplay
...
+		renderedShortcutPads = true;
+		SoundEditor::renderMainShortcutsOnly(
+		    (ModControllableAudio*)view.activeModControllableModelStack.modControllable, image, occupancyMask,
+		    SoundEditor::shouldEditKitAffectEntire());
+		return true;
+	}
+	else if (renderedShortcutPads) {
+		for (int32_t yDisplay = 0; yDisplay < kDisplayHeight; yDisplay++) {
```
seangoodvibes: What is this doing?

seangoodvibes: Oh is it clearing the shortcut pads before rendering the notes on the grid again?

src/deluge/gui/views/instrument_clip_view.cpp
m-m-adams: The global one called view! I decided not to use it though

src/deluge/gui/views/instrument_clip_view.cpp
m-m-adams: Yeah the row render only touches squares that have contents so this makes sure everything is cleared

src/deluge/gui/l10n/seven_segment.json
m-m-adams: Yeah that might be clearer... we should probably make these scroll lol

seangoodvibes (approved):

PR #4868, opened 2026-08-26
Midi Queue Manager

stellar-aria:
Supersedes #4840 

Basically makes it so that a bunch of CC automation messages can't preempt and delay clock and notes. Coalescing is an opt-in feature based on the class of message. See the previous PR for an overview of the general design. Relevant documentation is in docs/dev/systems.

Includes unit tests to prevent regressions and validate behavior.

PR #4886, opened 2026-08-31
USB MIDI: fix send ring for pipelined requests

cyface:
Hello!

I am working on a computer-based editor/librarian (starting from the existing smsysex work), and I encountered very slow performance when saving .WAV files to the Deluge over USB MIDI.

Working with the agent to debug over the course of many hours, we traced one cause of the slowness to an issue where the USB MIDI send path on the Deluge was corrupting replies to "pipelined" sends. This meant that packets had to be sent one at a time, which slowed things down significantly. 

Pipelining enables sending multiple requests at a time, rather than serial request/response.

This PR addresses the pipelining corruption with three fixes and one advertised value:

1) Don't accept another event if the send ring is already full. Current firmware accepts 1 more event after the ring is full, which ends up corrupting the reply.

2) Check to make sure we have room reserved for the whole message, so that we don't truncate a message mid-send.

3) Make sure we clear the whole send ring when a device connects, not just a subset of it. Otherwise, the uncleared parts could have previous message data.

4) Set up a "pipe" value in the session grant to tell clients whether the firmware supports pipelining and if so what depth. That way clients can fall back to slower serial sends for older firmware.

5. (Added after testing against #4868) Hold each incoming request in the receive queue until the send buffer has room for its whole reply. This makes pipelined clients slow down automatically to whatever the buffer can handle, instead of relying on dropped replies and client retries. With this change I measured pipeline widths of 3 and 4 running clean at ~163 KB/s with zero retries, so pipelining is now safe at any width.

I stress tested this with a very large number of scripted uploads to the card of big WAV files, XML files, and .bin files.

These fixes speed up uploads by about 30% - a 5MB sample that took about 135 seconds now takes about 100.

I can provide my python and/or WebMIDI test harnesses upon request.

Thank you!

Tim

### DIFF
--- a/src/deluge/io/midi/cable_types/usb_common.cpp
+++ b/src/deluge/io/midi/cable_types/usb_common.cpp
@@ -16,7 +16,11 @@
  */
 
 #include "usb_common.h"
+#include "io/debug/log.h"
 #include "io/midi/midi_engine.h"
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, readability-identifier-naming) - matches the USB driver's definition
+extern uint8_t anyUSBSendingStillHappening[];
 
 void MIDICableUSB::connectedNow(int32_t midiDeviceNum) {
 	connectionFlags |= (1 << midiDeviceNum);
@@ -93,6 +97,19 @@ void MIDICableUSB::sendSysex(const uint8_t* data, int32_t len) {
 
 	if (!connectedDevice) {
 		return;
+	}
+
+	// Reserve room for the whole message (one event per 3 bytes, plus one for the 0x7D header below):
+	// dropping individual events mid-message would leave a well-framed but corrupt SysEx.
+	int32_t eventsNeeded = ((len + 2) / 3) + 1;
+	if (connectedDevice->sendBufferSpace() < eventsNeeded * 3) {
+		if (anyUSBSendingStillHappening[0] == 0) {
+			midiEngine.flushUSBMIDIOutput();
+		}
+		if (connectedDevice->sendBufferSpace() < eventsNeeded * 3) {
+			D_PRINTLN("sendSysex: send ring full, dropping %d byte message", len);
+			return;
+		}
 	}
 
 	int32_t pos = 0;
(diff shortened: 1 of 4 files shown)


seangoodvibes:
Hi, fyi you might want to check the midi queue manager PR cause this seems to conflict with / will be superseded by that work. I think that PR will resolve your issue also.

If you want to help test that PR that would also be appreciated :) thx

PR here: https://github.com/SynthstromAudible/DelugeFirmware/pull/4868

cyface:
> Hi, fyi you might want to check the midi queue manager PR cause this seems to conflict with / will be superseded by that work. I think that PR will resolve your issue also.
> 
> If you want to help test that PR that would also be appreciated :) thx

Excellent! Looks like it probably gets me very close. Will test it and report back on that PR!

src/deluge/io/midi/cable_types/usb_common.cpp
```diff
@@ -16,8 +16,12 @@
  */
 
 #include "usb_common.h"
+#include "io/debug/log.h"
 #include "io/midi/midi_engine.h"
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, readability-identifier-naming) - matches the USB driver's definition
+extern uint8_t anyUSBSendingStillHappening[];
```
m-m-adams: Can you fill in the size here please

m-m-adams (commented):
Hey I think this looks good, would you mind sharing your test code in the contrib folder?

src/deluge/io/midi/cable_types/usb_common.cpp
cyface: Done!

cyface:
> Hey I think this looks good, would you mind sharing your test code in the contrib folder?

You bet - it's here now: https://github.com/cyface/DelugeFirmware/blob/fix/usb-send-ring-overflow/contrib/usb_midi_stress/smsysex_pipeline_stress.py

m-m-adams (approved):

cyface:
@m-m-adams - Yep, my bad, I was able to reproduce. Fix is in [this PR now](https://github.com/SynthstromAudible/DelugeFirmware/pull/4901). My script only tested against release builds and so didn't trip this. Updated script tests both, and fix was to remove debug logging from drop paths.

PR #4887, opened 2026-09-01
Decreased SPI clock frequency to work with 2026 batch of OLEDs

rohanhill:
Our newest batch of OLEDs for Deluge production, despite being the same OEL9M0087-W-E model as always, glitch if their clock frequency is above about 8.5MHz.

Previously, the code sets the SPI bus frequency at 10MHz and this had worked fine and been within the OLED datasheet's spec. This branch reduces it to 4MHz to be safe with the new OLEDs.

Hopefully this doesn't have too detrimental an effect in the instance where SPI has to be sent to the OLED and to the CV DAC at the same time (well, quick succession). There appears to be some provision in the code for CV SPI messages to jump ahead of OLED ones? Of course we want those to take priority.


stellar-aria (approved):

m-m-adams:
Would 8MHz work? That seems like it would alleviate any latency issue

PR #4891, opened 2026-09-02
Fix iterance migration: account for FIRST/LAST shift in legacy song files

### LINKED ISSUE #4890: Notes with iteration dependence load as FIRST/LAST after upgrading from C1.2.1 to 1.3 beta

Songs saved in C1.2.1 (or any firmware before FIRST/LAST iterance presets were added in commit 3ce3d41d) with notes set to "iteration dependence" (e.g. "1 of 2") load incorrectly in the 1.3 beta. The notes display as "FIRST" or "LAST" instead of their original value, and behave accordingly — playing only on the very first loop rather than every other loop.

Root cause: commit 3ce3d41d prepended two new presets (FIRST at index 1, LAST at index 2) to the beginning of iterancePresets[], shifting the existing presets (1of2…8of8) from indices 1–35 to indices 3–37. The migration path for old-format song files (noteHexLength == 22 in note_row.cpp) was not updated — it still maps raw probability byte values directly to iterancePresets[n] without compensating for the 2-position shift. A song with "1 of 2" stored as probability byte 21 (kNumProbabilityValues + 1) now resolves to iterancePresets[0] = FIRST instead of iterancePresets[2] = 1of2.

Notes with iteration dependence set in C1.2.1 should load with the same setting in 1.3. A note set to "1 of 2" should display and play as "1 of 2".

Workaround: Hold the affected note pad and turn SELECT 3 clicks clockwise to step past FIRST (index 1) and LAST (index 2) to reach "1 of 2" (index 3). Repeat for each affected note.

After upgrading from C1.2.1 to the 1.3 beta (git 3f898e953e).

no

OLED

Beta

1.3.0-BETA-3F898E9

1. Create a song in C1.2.1 with a synth or kit clip
2. Hold a note pad and turn SELECT clockwise to set iteration dependence to "1 of 2" (or any X of Y value)
3. Save the song
4. Upgrade to 1.3 beta and load the song
5. Hold the same note pad — it now displays "FIRST" instead of "1 of 2"
6. Play the song — the note only triggers on the very first loop, then never again

walkerab:
Fixes #4890

## Problem

Commit 3ce3d41d (`save, load, playback, menu working #4844`) prepended FIRST (index 0) and LAST (index 1) to `iterancePresets[]`, shifting the existing 1of2…8of8 presets from array indices 0–34 to indices 2–36.

Three migration paths in `note_row.cpp` that read older song formats were not updated. As a result, any song saved with iteration-dependence set in C1.2.1 (or the early 1.3 nightlies that wrote `noteHexLength==26`) loads with the wrong preset:

| Legacy format | Value stored | Old resolution | After 3ce3d41d (bug) |
|---|---|---|---|
| `noteHexLength==22` (C1.2.1) | probability byte 21 = "1 of 2" | `iterancePresets[0]` = 1of2 ✓ | `iterancePresets[0]` = FIRST ✗ |
| `noteHexLength==26` (early 1.3 nightly) | preset index 1 = "1 of 2" | `fromPresetIndex(1)` = 1of2 ✓ | `fromPresetIndex(1)` = FIRST ✗ |
| Pre-lift `else` path | same as noteHexLength==22 | same | same ✗ |

## Fix

Two new helpers in `iterance_migration.cpp`:

- **`iteranceFromLegacyProbabilityByte(probability)`** — applies the +2 array offset for the `noteHexLength==22` and pre-lift-else paths
- **`iteranceFromLegacyPresetIndex(presetIndex)`** — shifts the stored index by +2 for the `noteHexLength==26` path

`note_row.cpp` now calls these instead of the raw array accesses and `fromPresetIndex()`.

## Tests

`tests/unit/iterance_migration_tests.cpp` covers:
- Boundary values: 1of2, 2of2, 1of3, 8of8 for both helper functions
- OFF (index 0) and custom (index 36) for `fromLegacyPresetIndex`
- Out-of-range probability bytes (below/above the legacy preset range) fall back to OFF
- Sweep asserts that no legacy probability byte or preset index maps to FIRST or LAST

Red → green verified locally (12 failures before fix, 0 after).

## Testing

Verified in the [DelugEmu](https://github.com/gramster/delugemu) QEMU-based emulator against a real C1.2.1-era song containing notes with legacy iterance-encoded probability bytes (21 = "1 of 2", 22 = "2 of 2"), loaded through the actual firmware LOAD SONG UI:

- **Before this fix** (current `main`): the notes decode to **FIRST** and **LAST**, reproducing the bug exactly as described above.
- **After this fix** (this branch): the same notes decode correctly to **1 of 2** and **2 of 2**.

Also confirmed the fixed firmware boots cleanly under emulation with no crash/fault, and that the real ARM-compiled `iterancePresets[]` data matches the values the migration helpers expect.

### DIFF
--- a/src/deluge/model/iterance/iterance_migration.cpp
+++ b/src/deluge/model/iterance/iterance_migration.cpp
@@ -1,0 +1,51 @@
+/*
+ * Copyright © 2024 Synthstrom Audible Limited
+ *
+ * This file is part of The Synthstrom Audible Deluge Firmware.
+ *
+ * The Synthstrom Audible Deluge Firmware is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with this program.
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "model/iterance/iterance_migration.h"
+#include "definitions_cxx.hpp"
+#include <array>
+#include <cstdint>
+
+// iterancePresets is defined in lookuptables.cpp.
+extern const std::array<Iterance, kNumIterancePresets> iterancePresets;
+
+Iterance iteranceFromLegacyProbabilityByte(int32_t probability) {
+    // The old preset index (1-based) is encoded as probability - kNumProbabilityValues.
+    // FIRST and LAST were prepended to iterancePresets[], so the old index N now sits at
+    // array position N+1 (i.e. +2 for the prepend, -1 for 0-based indexing = net +1).
+    int32_t idx = probability - kNumProbabilityValues + 1;
+    // Defensive guard: idx should be 2..36 for any valid C1.2 file (old kNumIterancePresets=35).
+    if (idx < 2 || idx >= kNumIterancePresets) {
+        return kDefaultIteranceValue;
+    }
+    return iterancePresets[idx];
+}
+
+Iterance iteranceFromLegacyPresetIndex(int32_t presetIndex) {
+    // Shift all non-zero indices by +2 to skip the newly prepended FIRST and LAST.
+    if (presetIndex == 0) {
+        return kDefaultIteranceValue;
+    }
+    int32_t newIndex = presetIndex + 2;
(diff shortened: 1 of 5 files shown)


seangoodvibes:
Hi, did you test these changes on a hardware deluge? Your description mentions "DelugeEmu" but not the actual hardware.

Thanks

m-m-adams (dismissed):

m-m-adams (commented):
Whoops meant to approve the workflow run. I think the code looks fine but needs to be tested with some old/early 1.3/current nightly files, preferably on hardware

walkerab:
> Hi, did you test these changes on a hardware deluge? Your description mentions "DelugeEmu" but not the actual hardware.

I haven't run on real hardware yet. Just the emulator.

I should have some time later this week to test on real hardware.

walkerab:
Tested fix on deluge hardware. Works great.

PR #4893, opened 2026-09-03
Fixed wrong key combination in Pattern doc page

boblemarin:
Loading a Melodic Pattern without overwriting existing Notes is done with the LOAD button, not the SAVE one.


seangoodvibes (approved):

seangoodvibes:
Thank you for catching this and helping improve our docs. Really appreciate it :)

PR #4896, opened 2026-09-04
Cleanup dx7 memory allocation

seangoodvibes:
Cleaned up dx7 memory allocation

- switched voice allocation to object pool
- switched dx patch allocation to low speed
- add dx patch deallocation to source destructor
- removed dx prealloc references

### DIFF
--- a/src/deluge/dsp/dx/engine.cpp
+++ b/src/deluge/dsp/dx/engine.cpp
@@ -56,9 +56,9 @@ DxEngine::DxEngine() {
 
 gsl::owner<DxVoice*> DxEngine::solicitDxVoice() {
 #ifdef DX_PREALLOC
-	if (firstUnassignedDxVoice) {
-		DxVoice* toReturn = firstUnassignedDxVoice;
-		firstUnassignedDxVoice = firstUnassignedDxVoice->nextUnassigned;
+	if (dxEngine && dxEngine->firstUnassignedDxVoice) {
+		DxVoice* toReturn = dxEngine->firstUnassignedDxVoice;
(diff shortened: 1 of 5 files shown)


src/deluge/dsp/dx/engine.cpp
```diff
@@ -50,35 +54,35 @@ DxEngine::DxEngine() {
 #endif
 }
 
-DxVoice* DxEngine::solicitDxVoice() {
+gsl::owner<DxVoice*> DxEngine::solicitDxVoice() {
 #ifdef DX_PREALLOC
```
m-m-adams: this is now stale so just remove it (voice pool is pre alloc but consistent with the rest of deluge)

src/deluge/processing/source.cpp
```diff
@@ -47,6 +48,12 @@ Source::Source() {
 }
 
 Source::~Source() {
+	// destruct dxPatch if it was allocated
+	if (dxPatch != nullptr) {
+		dxPatch->~DxPatch();
+		delugeDealloc(dxPatch);
```
m-m-adams, line 54: were you going to move the patch to external memory? I don't see a change for that

seangoodvibes: Up top I changed allocMaxSpeed to allocLowSpeed

DxPatch* DxEngine::newPatch(void) {
	void* memory = allocLowSpeed(sizeof(DxPatch));
	return new (memory) DxPatch;
}

src/deluge/dsp/dx/engine.cpp
seangoodvibes: Will do, this part was confusing me

m-m-adams (approved):

PR #4900, opened 2026-09-04
Revert "USB MIDI: fix send ring for pipelined requests"

m-m-adams:
Reverts SynthstromAudible/DelugeFirmware#4886

Causes crashes when many logs are written at the same time. @cyface flushing in the send is probably not safe since it's done in an isr as well


cyface:
I will review!

PR #4919, opened 2026-09-13
Fix/ param lookup safety and prevent null ptr deref

seangoodvibes:
Add some additional safety checks to auto param lookups to prevent invalid auto param setup / nullptr dereference


stellar-aria (approved):

src/deluge/model/model_stack.cpp
```diff
@@ -269,11 +269,14 @@ ModelStackWithAutoParam* ModelStackWithThreeMainThings::getPatchCableAutoParamFr
 }
 
 ModelStackWithAutoParam* ModelStackWithThreeMainThings::getExpressionAutoParamFromID(int32_t newParamId) {
-	if (newParamId >= kNumExpressionDimensions) {
+	if (!paramManager || newParamId < 0 || newParamId >= kNumExpressionDimensions) {
```
m-m-adams, line 272: Why is this guarded? Is it possible to have a modelstackwiththreemainthings without a param manager?

src/deluge/modulation/params/param_set.cpp
```diff
@@ -365,6 +365,9 @@ void ParamSet::backUpParamToAction(int32_t p, Action* action, ModelStackWithPara
 }
 
 ModelStackWithAutoParam* ParamSet::getAutoParamFromId(ModelStackWithParamId* modelStack, bool allowCreation) {
+	if (modelStack->paramId < 0 || modelStack->paramId >= numParams_) {
```
m-m-adams, line 368: If the other one needs a check for null param manager than so does this

src/deluge/model/model_stack.cpp
```diff
@@ -269,11 +269,14 @@ ModelStackWithAutoParam* ModelStackWithThreeMainThings::getPatchCableAutoParamFr
...
 		return addParamCollectionAndId(nullptr, nullptr, 0)->addAutoParam(nullptr); // "No param"
 	}
 
-	paramManager->ensureExpressionParamSetExists(); // Allowed to fail
+	// if it fails to allocate then we would be accessing a null param collection below
+	if (!paramManager->ensureExpressionParamSetExists()) {
+		return addParamCollectionAndId(nullptr, nullptr, 0)->addAutoParam(nullptr);
+	}
```
m-m-adams, line 279: check for the null param collection, not the param set

m-m-adams (commented):
I think this is guarding the wrong pointer. Summary will always exist and only the param collection can be null

PR #4920, opened 2026-09-13
Fix param manager leaks

seangoodvibes:
First fix ensures that a backed up param manager that is about to be deleted is destructed, it's done elsewhere in the codebase (e.g. in Song::deleteBackedUpParamManagersForModControllable), but not in Song::deleteBackedUpParamManagersForClip.

Second fix. ensures that if you are going to steal param collections that you destruct the current param collections held as they are going to be replaced, otherwise you're just leaving previously allocated memory out there.


src/deluge/model/song/song.cpp
```diff
@@ -3935,6 +3935,9 @@ void Song::deleteBackedUpParamManagersForClip(Clip* clip) {
 				ModControllableAudio* modControllable = backedUp->modControllable;
 
 				// We have to delete that element...
+				// The main-only steal above retains any expression collection in
```
m-m-adams, line 3938: What is a "main only steal"

src/deluge/modulation/params/param_manager.cpp
```diff
@@ -116,11 +116,25 @@ void ParamManager::stealParamCollectionsFrom(ParamManager* other, bool stealExpr
...
+	// don't steal from ourselves
+	if (other == this) {
+		return;
+	}
 
 	int32_t mpeParamsOffsetOther = other->getExpressionParamSetOffset();
 	int32_t mpeParamsOffsetHere = getExpressionParamSetOffset();
 	int32_t stopAtOther = mpeParamsOffsetOther;
 
+	// Main collections already owned by the destination are about to be
```
m-m-adams, line 128: no they're not? If the destination has expression params then the source collection is deallocated.

m-m-adams (changes requested):

src/deluge/modulation/params/param_manager.cpp
m-m-adams: Oh main params as in not expression?

src/deluge/model/song/song.cpp
seangoodvibes: Steal only main param collections

src/deluge/model/song/song.cpp
```diff
@@ -3935,6 +3935,9 @@ void Song::deleteBackedUpParamManagersForClip(Clip* clip) {
...
+				// the removed element, and this raw vector does not run destructors.
```
m-m-adams, line 3939: No it doesn't - expression in the removed element is either deallocated or moved to the new one

src/deluge/modulation/params/param_manager.cpp
seangoodvibes: > Oh main params as in not expression? 

Correct not expression

src/deluge/model/song/song.cpp
m-m-adams: Ok - please update that comment for clarity. It's not terminology used anywhere else in the codebase

m-m-adams (approved):
Just updated the comment myself

m-m-adams:
@seangoodvibes I can't request your review on your own pull request but review please :)

seangoodvibes:
> @seangoodvibes I can't request your review on your own pull request but review please :)

The changes look good thanks. Will request approval from Katherine.

Issue #4911, opened 2026-09-09
Weird kit culling issue


seangoodvibes:
I experienced a strange issue

1. i load the attached song and i hit play. the song plays all 3 kits fine.
2. i then mute some drums in the third kit and the kit starts culling voices

> In kit DDD-1, I'm muting drums TOMH, TOMM, TOML, and RIM
> If i mute TOMH and TOMM, no issue. As soon as i mute one more (e.g. TOML or RIM) it starts culling. Play button blinks. Audible audio dropouts.
> From play off with those drums muted, if i start play it starts culling.
> if i unmute all of those drums it stops culling.

Why would this happen when fewer drums are playing?

[3Kits.XML](https://github.com/user-attachments/files/32018712/3Kits.XML)

PR #4922, opened 2026-09-13
Fix kit culling issue

seangoodvibes:
Fix Partially #4911 by improving rendering, but I don't think its addressing the root cause

I experienced a strange issue

1. i load the attached song and i hit play. the song plays all 3 kits fine.
2. i then mute some drums in the third kit and the kit starts culling voices

> In kit DDD-1, I'm muting drums TOMH, TOMM, TOML, and RIM
> If i mute TOMH and TOMM, no issue. As soon as i mute one more (e.g. TOML or RIM) it starts culling. Play button blinks. Audible audio dropouts.
> From play off with those drums muted, if i start play it starts culling.
> if i unmute all of those drums it stops culling.

Why would this happen when fewer drums are playing?

[3Kits.XML](https://github.com/user-attachments/files/32018712/3Kits.XML)


stellar-aria (approved):

src/deluge/processing/sound/sound.cpp
```diff
@@ -2350,12 +2350,23 @@ void Sound::stopParamLPF(ModelStackWithSoundFlags* modelStack) {
 void Sound::process_postarp_notes(ModelStackWithSoundFlags* modelStackWithSoundFlags, ArpeggiatorSettings* arpSettings,
                                   ArpReturnInstruction instruction) {
 	if (instruction.arpNoteOn) {
-		instruction.arpNoteOn->noteStatus[0] = ArpNoteStatus::PENDING;
+		// The melodic arp's arp-off path marks its returned note PLAYING before
```
m-m-adams: I'm not sure what this is saying. If it's not actually playing yet why not change that to pending instead of hacking a workaround into the code here?

src/deluge/model/instrument/kit.cpp
```diff
@@ -831,8 +831,8 @@ void Kit::renderNonAudioArpPostOutput(std::span<StereoSample> output) {
 					if (instruction.arpNoteOn->noteCodeOnPostArp[n] == ARP_NOTE_NONE) {
 						break;
 					}
-					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
 					nonAudioDrum->noteOnPostArp(instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn, n);
+					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;
```
m-m-adams, line 835: Why is the start/mark order being changed here?

seangoodvibes: Cause it's wrong as far as I can tell.

No where else is the status marked playing before the note on is sent.

Previously we'd mark it playing here, then it would get overridden with pending in process_postarp_notes (instruction.arpNoteOn->noteStatus[0] = ArpNoteStatus::PENDING;)

So playing was doing nothing and we were covering it up with setting the pending status.

Now there's a guard in process_postarp_notes which checks if the status is pending.

Examples of how playing is set elsewhere:

### GateDrum::noteOn

> 			noteOnPostArp(instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn, n);
> 			instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;

### MIDIDrum::noteOn

> 			noteOnPostArp(instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn, n);
> 			instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;

### Kit::setupAndRenderArpPreOutput

> 					// Do row note on
> 					ModelStackWithThreeMainThings* modelStackWithThreeMainThings =
> 					    modelStackWithTimelineCounter
> 					        ->addNoteRow(kitInstruction.arpNoteOn->noteCodeOnPostArp[0], thisNoteRow)
> 					        ->addOtherTwoThings(thisNoteRow->drum->toModControllable(), &thisNoteRow->paramManager);
> 					thisNoteRow->drum->noteOn(modelStackWithThreeMainThings, kitInstruction.arpNoteOn->velocity,
> 					                          kitInstruction.arpNoteOn->mpeValues, 0, kitInstruction.sampleSyncLengthOn,
> 					                          0, 0);
> 					kitInstruction.arpNoteOn->noteStatus[0] = ArpNoteStatus::PLAYING;

### NonAudioInstrument::renderOutput

> 					noteOnPostArp(instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn, n);
> 					instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;

### Sound::NoteOn

> 				noteOnPostArpeggiator(modelStackWithSoundFlags, noteCodePreArp,
> 				                      instruction.arpNoteOn->noteCodeOnPostArp[n], instruction.arpNoteOn->velocity,
> 				                      mpeValues, instruction.sampleSyncLengthOn, ticksLate, samplesLate,
> 				                      fromMIDIChannel);
> 				instruction.arpNoteOn->noteStatus[n] = ArpNoteStatus::PLAYING;

m-m-adams: > Cause it's wrong as far as I can tell.
> 
> No where else is the status marked playing before the note on is sent.
> 

It's just different, it doesn't matter. The note status isn't used in noteonpostarp and there's no way for it to fail

seangoodvibes: If it doesn't matter then we can change it for consistency with elsewhere?

m-m-adams: I'd rather not since it implies there's a causal relationship

seangoodvibes: So then change everywhere else to before?

I just don't like the same code being written differently in different spots.

There's multiple note post arp calls and all except this one set playing after.

It's confusing

I pasted all the other ones above for reference

seangoodvibes: Maybe I'm not comparing apples to apples I'm not sure.

It's been a while since I looked at this and don't remember anymore.

m-m-adams: It literally does not matter whether it's before or after. I don't know why you're changing it in this PR or why you're calling it "wrong" 

Your copy paste comment above is referring to completely different functions that are called at different times. It's implying this is a bug when noteOnPostArp is never checking or setting the status in the first place. I don't know if this is an LLM PR or what but the actual bug fix is the change in sound.cpp

m-m-adams (approved):
it doesn't matter so just gonna leave the ordering change

PR #4929, opened 2026-09-14
Fix droning notes don't respect non default prob iter fill

### LINKED ISSUE #4926: Drone notes don't respect probability / iterance / fill

Noticed it cause of the new first iterance value

if a note is entered as a drone note, the first iterance value doesn't do anything - note will drone endlessly

If a note is entered as a drone note with fill set, fill has no effect

If a note is entered as a drone note with probability set, probability works up until the note is auditioned and then the note drones endlessly

I think notes entered as drone notes should only work like drone notes when probability, iterance or fill are the default values, otherwise they should respect the probability, iterance and fill settings.

Today

_No response_

OLED

Nightly

_No response_

_No response_

seangoodvibes:
Fix https://github.com/SynthstromAudible/DelugeFirmware/issues/4926

Drone notes were not respecting probability, iteration or fill settings.

Example #1:
- Add a drone note
- Set iterance to first
- Press play
- Note drones indefinitely when it should only play through until the end of the loop and stop

Example #2:
- Add a drone note
- Set fill
- Press play
- Note drones indefinitely irrespective of fill mode being active when it should only drone while fill is active

Example #3:
- Add a drone note
- Set probability < 100%
- Press play
- Note drones indefinitely irrespective of probability setting when it should only drone when probability condition passes

--

Fixed it by adding drone notes to pending note on list with a flag to identify if the pending note on relates to a currently sounding drone note

If conditions pass, note continues to drone and note row is kept active
if conditions fail, drone note is stopped and note row is made inactive

Other than that, I did some refactoring clean up as well to remove some go to's.

-- 

As an aside, I noticed a FIXME comment that says:

"// FIXME: this is almost certainly a bad idea, we can't handle more than 8-10 note ons per render without culling Otherwise, just send it now."

Might be something worth looking into separately


darkelement1987:
This is what you are describing correct? 

(Bottom.bar = iter 1:2, top = iter 2:2)

https://github.com/user-attachments/assets/5aa47c5e-8750-4ea4-bc39-c0af7c084cd7

seangoodvibes:
> This is what you are describing correct? 
> 
> 
> 
> (Bottom.bar = iter 1:2, top = iter 2:2)
> 
> 
> 
> https://github.com/user-attachments/assets/5aa47c5e-8750-4ea4-bc39-c0af7c084cd7
> 
> 
> 
> 

Yes

m-m-adams:
I think the original behaviour is clearer - a drone note is a note that never ends, once it starts it should continue playing imo

darkelement1987:
@m-m-adams Not sure if you are saying this is intended behavior to you, i dont really agree 

I used the same approach for arps, not necessarily for a "droning note" and when i add iteration or probability to a note (whether its a screen filling note of not) the deluge should respect that. 

The note should sound/be heard (whatever the English term is) for as long as it is drawn imo. 

The probability/iteration settings affect that duration so is say this is an issue, to me that is

seangoodvibes:
> I think the original behaviour is clearer - a drone note is a note that never ends, once it starts it should continue playing imo  

Your "last" iteration functionality will never work for drone notes then

So now we have to tell people using drone notes that they can't use any probability or iteration or fill settings for those notes even though you can change those settings

Drone notes will always play

I don't like this

m-m-adams:
Last and fill not working seems like a bug, I meant first and probability. Once the note starts it should just drone on imo

m-m-adams:
> @m-m-adams Not sure if you are saying this is intended behavior to you, i dont really agree 
> 
> 
> 
> I used the same approach for arps, not necessarily for a "droning note" and when i add iteration or probability to a note (whether its a screen filling note of not) the deluge should respect that. 
> 
> 
> 
> The note should sound/be heard (whatever the English term is) for as long as it is drawn imo. 
> 
> 
> 
> The probability/iteration settings affect that duration so is say this is an issue, to me that is 

Yeah but if that's the behaviour you want then you can just not use a drone note and achieve the same effect

